PR #11, opened 2020-08-30
Add builtin for searching vim.v.oldfiles

smolck:


tjdevries:
I think you can try to send the results with`Job:send` rather than the writer strategy... but I'm not 100% sure. I'd have to test it out. You should not have to shell out to `echo` to make it work.

Also, you have considered instead trying to use ` sorter` to fulfill the same idea you're doing here? I can show an example later if you want. I don't think we need to require fzf to make this happen (which is kind of the point of telescope ;) )

smolck:
> I think you can try to send the results with`Job:send` rather than the writer strategy... but I'm not 100% sure. I'd have to test it out. You should not have to shell out to `echo` to make it work.

Okay, I'll try that out and see if it works. Thanks! The echo definitely feels clunky.

> Also, you have considered instead trying to use ` sorter` to fulfill the same idea you're doing here? I can show an example later if you want. I don't think we need to require fzf to make this happen (which is kind of the point of telescope ;) )

I have not; an example would be great! I guess I didn't really think about the fzf requirement; my bad!

smolck:
@tjdevries PR no longer uses FZF ;)

lua/telescope/builtin.lua
```diff
@@ -236,6 +236,20 @@ builtin.grep_string = function(opts)
   }
 end
 
+builtin.oldfiles = function()
+  local oldfiles_finder = finders.new {
+    results = vim.tbl_map(
+      function(x) return (x:gsub('\n', '')) end,
```
smolck, line 242: Could also use `require('plenary.functional').first(x:gsub('\n', ''))` here, which (from the tests I did when working on https://github.com/nvim-lua/plenary.nvim/pull/10) is just a bit faster.

tjdevries: You can just use `vim.split` maybe?

smolck: How so? I got an error there regarding newlines being in the vim.v.oldfiles string, so this line is to remove those newlines; is that not necessary/is there a better way?

tjdevries: I was thinking `results = vim.split(vim.v.oldfiles)` or something.

smolck: Hmm, that won't work since vim.v.oldfiles is a table, but I am not having the issue with there being a newline in the vim.v.oldfiles strings, so I just switched it to `results = vim.v.oldfiles`.

tjdevries:
Had to rebase to fix some conflicts. I'm going to change this a bit, since I'm changing the interface, but I'll merge for now. TY!

PR #12, opened 2020-08-30
View the planets

smolck:
Currently I get a stack smashing error if I try to run `:lua require'telescope.builtin'.planets()`, but I'm not sure why. If you remove mars (or I presume any planet) it makes the issue less frequent, but it still occurs from time to time (and I think it happens less if at all with only a one or two planets).


tjdevries:
Save planets as files under Memes directory.

tjdevries:
Then just cat those files. You can search the files the normal way.

tjdevries:
@smolck I pushed an example of how you can do this without shelling out to `ls` every time. I think it needs some fixing, cause `cat` doesn't seem to work for me every time... i don't know why.

smolck:
@tjdevries Thank you, your solution is an instructive one (I didn’t even know `require('plenary.debug_utils').sourced_filepath()` existed).

> cause `cat` doesn't seem to work for me every time... i don't know why.

Would you mind elaborating?

tjdevries:
Sometimes I just get "Process exited 0" (like cat has completed).

I'm not sure why that is.

tjdevries:
@smolck do you want me to do the rebase and merge this? I think if you use the `cat` previewer and stuff now this might work.

If not, we can merge and I'l ltry and find a way to get cat to show correctly (perhaps just with a job or something dumb).

PR #14, opened 2020-08-31
Renames the quickfix-related builtin variables from lsp references

megalithic:
Prior to this commit, there looks to have been some missed renames and cleanup for variables for the `quickfix` builtin function. Was referencing the `LSP Reference` builtin variables and labels.


tjdevries:
Thanks!

PR #16, opened 2020-09-01
Add a method to show loclist

himansingh241:


lua/telescope/builtin.lua
```diff
@@ -112,6 +112,13 @@ end
 
 builtin.quickfix = function(opts)
   local locations = vim.fn.getqflist()
+  if next(locations) == nil then
+    locations = vim.fn.getloclist(0)
+    for _, value in pairs(locations) do
+      value.filename = vim.fn.bufname("%")
```
tjdevries, line 118: vim.api.nvim_buf_get_name(0) would be better

lua/telescope/builtin.lua
```diff
@@ -112,6 +112,13 @@ end
 
 builtin.quickfix = function(opts)
```
tjdevries, line 113: Can you add an opt for `loclist_fallback`?

tjdevries: Or.... hmm, maybe what we want is just two different functions.

tjdevries: They mean very different things.

himansingh241: Done!

himansingh241:
I can add the info about this in README.md if you think the implementation is okay.

lua/telescope/builtin.lua
```diff
@@ -126,6 +126,28 @@ builtin.quickfix = function(opts)
   }):find()
 end
 
+builtin.loclist = function(opts)
+  local locations = vim.fn.getloclist(0)
+
+  for _, value in pairs(locations) do
+    value.filename = vim.api.nvim_buf_get_name(0)
```
tjdevries, line 133: probably only need to get this once. Just for perf that doesn't matter.

PR #22, opened 2020-09-03
docs: Add docs for mappings.

rockerBOO:


README.md
```diff
@@ -66,53 +70,53 @@ wrappers over common tasks).
...
+  prompt = 'LSP References'
 }
 ```
 
 ```lua
 require'telescope.builtin'.quickfix{
-    -- See Picker for additional options
-    prompt = 'Quickfix'
+  -- See Picker for additional options
+  prompt = 'Quickfix'
```
tjdevries, line 97: I don't think we need to put `prompt` in these -- it is auto supplied by the builtin. I think it makes people think they have to include it.

rockerBOO: This is fair but I wanted to show the API for each builtin and each one only uses prompt right now. 

Showing these, or any options, helps those who don't know lua syntax. Maybe this is outside of the scope of this project but making shown in syntax can help show others how lua will allow options. 

So ultimately do we want to show no options for these? The API will change some so maybe we can re-approach it after the global level is setup.

tjdevries:
hahaha, that's ok :) that's why we make PRs :laughing:

PR #28, opened 2020-09-04
Normalize mappings

runiq:
Modes now use lowercase keys, mappings use the internal representation via `nvim_replace_termcodes()`.

### DIFF
--- a/lua/telescope/mappings.lua
+++ b/lua/telescope/mappings.lua
@@ -98,7 +98,9 @@ mappings.apply_keymap = function(prompt_bufnr, attach_mappings, buffer_keymap)
   local applied_mappings = { n = {}, i = {} }
 
   local map = function(mode, key_bind, key_func, opts)
-    applied_mappings[mode][key_bind] = true
+    local mode = string.lower(mode)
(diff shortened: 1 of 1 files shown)


tjdevries:
So I thought about this... but doesn't this make it impossible to map `J`?

(I do need to fix it for the "<CR>" vs. "<cr>" business though)

tjdevries:
I was thinking maybe we should use this: `nvim_replace_termcodes({str}, {from_part}, {do_lt}, {special})` ?

tjdevries:
since that might change `<CR>` and `<cr>` to the same item.

runiq:
Oh *rha

runiq:
@tjdevries like this?

runiq:
So the `applied_mappings` table uses the internal representation now since those are normalized, but `telescope_map` still gets the original, non-normalized mapping. The reason for this is that `nvim_buf_set_keymap()` unconditionally appears to run its input through `nvim_replace_termcodes()`, and running that twice isn't necessarily idempotent.

tjdevries:
ty ty

PR #30, opened 2020-09-04
feat: add buffers

tjdevries:
@Conni2461 Hey, can you check out this PR?

I actually was playing around w/ buffers this morning.

Can you tell me what you think of this?

### DIFF
--- a/lua/telescope/actions.lua
+++ b/lua/telescope/actions.lua
@@ -34,6 +34,7 @@ function actions.get_selected_entry(prompt_bufnr)
   return actions.get_current_picker(prompt_bufnr):get_selection()
 end
 
+-- TODO: It seems sometimes we get bad styling.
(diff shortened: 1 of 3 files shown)


Conni2461:
It works great.

Just two little remarks.
Do you want only visible buffers to be listed because you are filtering for loaded buffers? Because neither fzf nor vim-clap do that.

If you have single and multi-digit buffer numbers, the aligning does not fit exactly.

tjdevries:
If I merge this, would you be interested in adding the two features that you've mentioned? We can add an opt for not filtering out the buffers that aren't visible and we can do some text aligning as well.

I'd be happy to help you w/ those PRs if you need anything with it at all.

Conni2461:
Yeah i would be happy to help with that.

tjdevries:
@Conni2461 feel free to submit PR w/ improvements.

PR #32, opened 2020-09-05
Adds shorten_path, show_all_buffers opts and fixes alignment (buffers builtin)

Conni2461:
Fixes the remarks i made in #30.

Adds shorten_path opt.
Add show all buffers opt, this will also show unloaded buffers.
Fixes alignment with a dynamic solution. This means that the bufnr field adapts to the largest possible bufnr.
Previously I had a field size of 3 hardcoded but this makes it ugly when only single digit bufnr are displayed.

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -180,6 +180,8 @@ function make_entry.gen_from_quickfix(opts)
 end
 
 function make_entry.gen_from_buffer(opts)
+  opts = opts or {}
+
   local get_position = function(entry)
     local tabpage_wins = vim.api.nvim_tabpage_list_wins(0)
     for k, v in ipairs(tabpage_wins) do
@@ -189,6 +191,18 @@ function make_entry.gen_from_buffer(opts)
     end
 
     return {}
+  end
+
+  local make_display = function(entry)
+    local display_bufname
+    if opts.shorten_path then
+      display_bufname = utils.path_shorten(entry.filename)
+    else
+      display_bufname = entry.filename
+    end
+
+    return string.format("%" .. opts.bufnr_width .. "d : %s",
+                         entry.bufnr, display_bufname)
   end
 
   return function(entry)
@@ -206,7 +220,7 @@ function make_entry.gen_from_buffer(opts)
 
       value = bufname,
       ordinal = bufnr_str .. " : " .. bufname,
-      display = bufnr_str .. " : " .. bufname,
+      display = make_display,
 
       bufnr = entry,
       filename = bufname,
--- a/README.md
+++ b/README.md
@@ -4,14 +4,14 @@ Gaze deeply into unknown regions using the power of the moon.
 
 ## What is Telescope?
 
-Telescope is a highly extendable fuzzy finder over lists. Items are shown in a popup with a prompt to search over.  
+Telescope is a highly extendable fuzzy finder over lists. Items are shown in a popup with a prompt to search over.
 
 Support for:
 
 * LSP (references, document symbols, workspace symbols)
-* Treesitter 
-* Grep 
-* Files (git, fd) 
+* Treesitter
+* Grep
+* Files (git, fd)
 * Vim (command history, quickfix, loclist)
 
 [What is Telescope?](https://www.twitch.tv/teej_dv/clip/RichDistinctPlumberPastaThat)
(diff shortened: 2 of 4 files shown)


lua/telescope/make_entry.lua
```diff
@@ -178,12 +178,18 @@ function make_entry.gen_from_buffer(opts)
...
       value = bufname,
-      ordinal = bufnr_str .. " : " .. bufname,
-      display = bufnr_str .. " : " .. bufname,
+      ordinal = string.format("%" .. opts.bufnr_width ..
+                              "s : %s", bufnr_str, bufname),
+      display = string.format("%" .. opts.bufnr_width ..
+                              "s : %s", bufnr_str, bufname),
 
       bufnr = entry,
       filename = bufname,
```
tjdevries, line 195: With `shorten_path`, you change the filename, which makes the previewer unable to find the buffer.

After you fix that, I think I'll merge. Very cool! TY

tjdevries (changes requested):

lua/telescope/make_entry.lua
Conni2461: Ah makes sense. Should be fixed.

Conni2461:
After testing it again, i noticed that Treesitter highlighting does not work, when i open an unloaded buffer. e.g. if i open a session and then navigate to another file, i have to run `:e` to get highlighting.

```diff
diff --git a/lua/telescope/actions.lua b/lua/telescope/actions.lua
index 3114679..7097552 100644
--- a/lua/telescope/actions.lua
+++ b/lua/telescope/actions.lua
@@ -84,6 +84,7 @@ local function goto_file_selection(prompt_bufnr, command)
     a.nvim_set_current_win(original_win_id)
     if entry_bufnr then
       a.nvim_win_set_buf(original_win_id, entry_bufnr)
+      vim.cmd(":e")
     else
       vim.cmd(string.format(":%s %s", command, filename))
```

tjdevries:
Hmmm perhaps instead of "e" we should do something like `doautocmd FileType` or similar

tjdevries:
I have been struggling with this as well to figure out what the best way is to do this -- I don't have a 100% idea yet.

Conni2461:
I picked `vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))` [previewers.lua:116](https://github.com/nvim-lua/telescope.nvim/blob/master/lua/telescope/previewers.lua#L116) and used it instead of `vim.cmd(":e")` and that fixes the problem as well. 

But if I activate this line in previewers, it does not fix the highlighting in the preview window.

lua/telescope/make_entry.lua
```diff
@@ -178,12 +178,25 @@ function make_entry.gen_from_buffer(opts)
...
+    local dsp = make_display(bufname)
+
     return {
       valid = true,
 
       value = bufname,
-      ordinal = bufnr_str .. " : " .. bufname,
-      display = bufnr_str .. " : " .. bufname,
+      ordinal = dsp,
+      display = dsp,
```
tjdevries, line 199: display can actually just take the function itself, it allows functions to be passed to it, so as to not call display if it's expensive for every line (since sometimes they will be hidden).

So you can actually just say `display = make_display` here.

tjdevries:
We can probably merge w/out the fix to the previewer -- I will have to investigate how that is done in some other plugins maybe or something. No worries on that. TY for PR!

lua/telescope/make_entry.lua
Conni2461: Ah okay. Thanks. I am slowly getting the hang of it.

I'm not sure what to set `ordinal` to because you always set `ordinal` to `line` when you use `shorten_path`.

Conni2461:
I rebased the branch and added the missing documentation for buffers. My autocmd has removed some trailing whitespaces.

Conni2461:
@tjdevries Just did another rebase. Highlighting works now in preview for unloaded buffers. Thank you for fixing this.

I'm still not sure what `ordinal` should be (like i said [here](https://github.com/nvim-lua/telescope.nvim/pull/32#discussion_r484309946), not sure if you saw it) but otherwise it is ready to merge.

Thank you for creating such an amazing plugin.

tjdevries:
Thanks for such an awesome PR! Everything looks good on my end. Thanks for the docs updates as well.

Nice work!

PR #35, opened 2020-09-05
added preview for files which spaces in names.

SubeetKD:


tjdevries:
Ah, this probably fixes the problem, but what we might want to do is probably switch over to the way of sending a list to `termopen` or something. That will handle auto-escaping for us (like if the file has a `"` in it, i think the current way will break, for example).

What do you think about that?

SubeetKD:
Yes you are right it will break. I don't know much of lua and vimscript so I don't think I can solve this.

tjdevries:
That's ok, this is still a good improvement! We can try and fix the other thing later. Thanks for the PR!!

PR #43, opened 2020-09-08
Tweak vertical layout

clason:
This is a slight tweak to the vertical layout strategy (useful for "thin" windows) with the following changes:
1. Instead of a percentage of neovim window width, make the telescope windows full width with a two column border.
2. Instead of fixing the size of the results window, fix the size of the preview window to half neovim window height and adjust the remaining windows to fit.
3. Leave two rows of space after the prompt window (to mirror the two windows of space before the preview window).

I feel this "50% vertical split" strategy is more visually pleasant and fits better with the default "50% horizontal split" strategy.
Screenshot 2020-09-08 at 18 28 59

### DIFF
--- a/lua/telescope/pickers/layout_strategies.lua
+++ b/lua/telescope/pickers/layout_strategies.lua
@@ -97,7 +97,7 @@ layout_strategies.vertical = function(self, max_columns, max_lines, prompt_title
   local results = initial_options.results
   local prompt = initial_options.prompt
 
-  local width_padding = math.ceil((1 - self.window.width) * 0.5 * max_columns)
+  local width_padding = 3
   local width = max_columns - width_padding * 2
   if not self.previewer then
     preview.width = 0
@@ -110,12 +110,12 @@ layout_strategies.vertical = function(self, max_columns, max_lines, prompt_title
   -- Height
   local height_padding = 3
 
-  results.height = 10
   prompt.height = 1
 
-  -- The last 2 * 2 is for the extra borders
+  -- Plus 2 * 2 for the extra borders
   if self.previewer then
-    preview.height = max_lines - results.height - prompt.height - 2 * 2 - height_padding * 2
+    preview.height = math.ceil(0.5 * max_lines - 4)
+    results.height = max_lines - preview.height - prompt.height - 2 * 3 - height_padding * 2
   else
     results.height = max_lines - prompt.height - 2 - height_padding * 2
   end


rockerBOO:
Making it take up more horizontal space is good but making the popup be almost fullscreen feels a little weird. 

My issue is it is wayy to wide since I leave it on all the time. Seems the max width is not being used? (Should be at like 120 columns max)

clason:
Interesting. My terminals are ~100 cols wide (that's why I want a vertical layout -- I thought widescreeners used the default horizontal layout).

Good point about the max width, though, I'll see about fixing that.

clason:
Actually, I don't see any max width in the vertical layout?

rockerBOO:
mockup

How I want it to work, without a preview. I would probably want a way to toggle on previews but that's basically what I'm used to. This works fine for small or larger screens and doesn't take a flash of blank so I don't lose context too long.

rockerBOO:
> Actually, I don't see any max width in the vertical layout?

I'm not sure how it's implemented but somewhere around here. https://github.com/nvim-lua/telescope.nvim/blob/d4459a6b2af669cde7573499f032ff174a272108/lua/telescope/pickers/layout_strategies.lua#L36

Basically it should get no larger than 120 IMO by default.

clason:
Ah, I never tested without previewers, so that code-path is taken unchanged (besides the width) from @tjdevries old layout.

Sounds to me like you want a "minimal" layout", really?

clason:
> > Actually, I don't see any max width in the vertical layout?
> 
> I'm not sure how it's implemented but somewhere around here.
> 
> https://github.com/nvim-lua/telescope.nvim/blob/d4459a6b2af669cde7573499f032ff174a272108/lua/telescope/pickers/layout_strategies.lua#L36
> 
> Basically it should get no larger than 120 IMO by default.

I'm basing all calculations on `max_columns`, so not sure what the problem is for you?

rockerBOO:
> Ah, I never tested without previewers, so that code-path is taken unchanged (besides the width) from @tjdevries old layout.
> 
> Sounds to me like you want a "minimal" layout", really?

For my files searching I would love this. I don't need to see the files usually, and having a toggle to preview would make it very workable.

rockerBOO:
> > > Actually, I don't see any max width in the vertical layout?
> > 
> > 
> > I'm not sure how it's implemented but somewhere around here.
> > https://github.com/nvim-lua/telescope.nvim/blob/d4459a6b2af669cde7573499f032ff174a272108/lua/telescope/pickers/layout_strategies.lua#L36
> > 
> > Basically it should get no larger than 120 IMO by default.
> 
> I'm basing all calculations on `max_columns`, so not sure what the problem is for you?

As my screenshows, it goes full width when it shouldn't be full width. 

2020-09-08-155953_4480x1440_scrot

This is how master looks now. Maybe its just not calculating max width properly and just limiting it's size through percents before.

I do think max width is appropriate for this to have some sane default.

clason:
That must be a problem with the upstream code, then. The old code didn't have a max width, either -- you just didn't notice (as much) since it took a percentage of max width, so it looked (relatively) fine on your wide screen (but too narrow on my thin screen).

rockerBOO:
Ok that's fair, I'll get @tjdevries to fix it ;)

tjdevries:
<will work on this in the next coupole days>

tjdevries:
Now works with `layout_config` and similar options.

PR #47, opened 2020-09-09
feat: add scrolling in preview window

haorenW1025:
This PR tries to add a member function `scroll_fn` to Previewer. For previewer like `bat` you can use `chansend` to send termcodes of `<c-u>` or `<c-d>` to scrolling. For now this only works for `bat` but similar things should be able to port to vim buffer or other previewer.

Todo:
- [x] add docs for mappings.
- [ ] add `scroll_fn` for vim buffer(maybe in another PR?)

### DIFF
--- a/lua/telescope/actions.lua
+++ b/lua/telescope/actions.lua
@@ -34,13 +34,6 @@ function actions.get_selected_entry(prompt_bufnr)
   return actions.get_current_picker(prompt_bufnr):get_selection()
 end
 
-local function preview_scrolling(prompt_bufnr, termcode)
-  local preview_win = state.get_status(prompt_bufnr).preview_win
-  local bufnr = vim.api.nvim_win_get_buf(preview_win)
-  local channel = vim.api.nvim_buf_get_option(bufnr, "channel")
-  vim.fn.chansend(channel, termcode)
-end
-
 function actions.preview_scrolling_up(prompt_bufnr)
   actions.get_current_picker(prompt_bufnr).previewer:scroll_fn(-1)
 end
--- a/lua/telescope/actions.lua
+++ b/lua/telescope/actions.lua
@@ -34,6 +34,21 @@ function actions.get_selected_entry(prompt_bufnr)
   return actions.get_current_picker(prompt_bufnr):get_selection()
 end
 
+local function preview_scrolling(prompt_bufnr, termcode)
+  local preview_win = state.get_status(prompt_bufnr).preview_win
+  local bufnr = vim.api.nvim_win_get_buf(preview_win)
+  local channel = vim.api.nvim_buf_get_option(bufnr, "channel")
+  vim.fn.chansend(channel, termcode)
+end
+
+function actions.preview_scrolling_up(prompt_bufnr)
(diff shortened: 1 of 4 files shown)


haorenW1025:
Yeah it does make sense. I'll refactor this PR a little bit later, will ping you when it's ready.

haorenW1025:
I've refactored it into two layers and add documentation for mappings:)

lua/telescope/actions.lua
```diff
@@ -34,6 +34,21 @@ function actions.get_selected_entry(prompt_bufnr)
   return actions.get_current_picker(prompt_bufnr):get_selection()
 end
 
+local function preview_scrolling(prompt_bufnr, termcode)
```
tjdevries, line 37: Is this function used anywhere?

lua/telescope/previewers.lua
```diff
@@ -55,6 +57,18 @@ function Previewer:teardown()
   end
 end
 
+function Previewer:send_input(input)
+  if self._send_input then
```
tjdevries, line 61: if we don't have `_send_input`, we should display some error to the user (so that they know why it isn't working)

lua/telescope/previewers.lua
```diff
@@ -55,6 +57,18 @@ function Previewer:teardown()
...
+    self:_send_input(input)
+  end
+end
+
+function Previewer:scroll_fn(direction)
+  if self._scroll_fn then
```
tjdevries, line 67: (same issue here w/ alerting user)

Probably can use `vim.api.nvim_err_writeln` or similar.

lua/telescope/previewers.lua
```diff
@@ -186,6 +200,21 @@ previewers.cat = defaulter(function(opts)
...
     end,
 
+    send_input = function(self, input)
+      termcode = vim.api.nvim_replace_termcodes(input, true, false, true)
+      if self.state.termopen_id then
+        pcall(vim.fn.chansend, self.state.termopen_id, termcode)
+      end
+    end,
+
+    scroll_fn = function(self, direction)
```
tjdevries, line 210: Do you think it'd be possible to add something like "count", similar to `v:count`? so someone can do `5<c-d>` or `scroll_fn(5)` to scroll five lines?

tjdevries (changes requested):

lua/telescope/actions.lua
haorenW1025: Ahh that's some redundant function left out. Delete it:)

lua/telescope/previewers.lua
haorenW1025: Done in the latest commit:)

lua/telescope/previewers.lua
haorenW1025: I've changed that to using direction as count. So right now the default scrolling up is `scroll_fn(30)` and `scroll_fn(-30)`.

tjdevries:
I may play around a bit with the scroll and input functions, but I like what we have here and it's working for me.

Thanks @haorenW1025 :) always nice to work together.

Issue #36, opened 2020-09-06
Request: Commands


TimUntersberger:
It would be cool if you could list all of the commands that are currently defined and then execute the command when selected.

`:command` provides you with the list of defined commands along with how many arguments they take. This could be used to conditionally also provide a way to enter these arguments.

TimUntersberger:
I have written one myself. This uses the `vim.api.nvim_get_commands` function to retrieve the list of commands. When a command gets selected it either gets executed immediately if `nargs=0` else it just prefills the cmdline with `:<NAME> `.

```lua
pickers.new({}, {
  prompt = 'Commands',
  finder = finders.new_table {
      results = (function()
        local command_iter = vim.api.nvim_get_commands({})
        local commands = {}

        for _, cmd in pairs(command_iter) do
          table.insert(commands, cmd)
        end

        return commands
      end)(),
      entry_maker = function(line)
        return {
          valid = line ~= "",
          entry_type = make_entry.types.GENERIC,
          value = line,
          ordinal = line.name,
          display = line.name
        }
      end
  },
  sorter = sorters.get_norcalli_sorter(),
  attach_mappings = function(prompt_bufnr, map)
      local run_command = function()
          local selection = actions.get_selected_entry(prompt_bufnr)
          actions.close(prompt_bufnr)
          local val = selection.value
          local cmd = string.format([[:%s ]], val.name)

          if val.nargs == "0" then
            cmd = cmd .. "<CR>"
          end

          vim.cmd [[stopinsert]]
          vim.fn.feedkeys(cmd)
      end

      map('i', '<CR>', run_command)
      map('n', '<CR>', run_command)

      return true
  end
})
```

tjdevries:
Ah, this is cool. Do you want to submit PR? I would merge this (with a few improvements)

TimUntersberger:
@tjdevries  #54

PR #54, opened 2020-09-10
builtin.commands

TimUntersberger:
This PR contains the picker mentioned in #36 and also fixes a bug on windows regarding file separators.

### DIFF
--- a/lua/telescope/previewers.lua
+++ b/lua/telescope/previewers.lua
@@ -185,7 +185,7 @@ previewers.cat = defaulter(function(opts)
     end,
 
     teardown = function(self)
-      if self.state.termopen_id then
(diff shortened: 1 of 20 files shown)


lua/telescope/previewers.lua
```diff
@@ -185,7 +185,7 @@ previewers.cat = defaulter(function(opts)
     end,
 
     teardown = function(self)
-      if self.state.termopen_id then
+      if self.state and self.state.termopen_id then
```
TimUntersberger, line 188: This line actually always crashed on my system, because self.state is nil.

tjdevries:
Awesome, seems good to me. May change slightly later, but I like what it's doing right now. Will need to test a few more things for commands I think, but it seems good for now!

PR #61, opened 2020-09-11
Add sorter to live_grep

clason:
Even if the matching is exact, the sorting is still helpful when multiple results are found.


tjdevries:
Let's try it out for a bit.

PR #70, opened 2020-09-13
feat: Add support for ripgrep in "find_files" builtin

### LINKED ISSUE #69: Feature: support ripgrep for file finder
The only advantage of `fd` that I can find is that you can directly give it a string to search, which might be useful in a "live find files" situation.
This is purely personal, but I don't use `fd` since a simple `rg --files` does the trick. And if I ever want to filter over that, I'll just use `rg` again.
Doing `rg --files | wc -l` and `fd --type f | wc -l` gives me the same result.

I'll be making a PR for this as it's quite easy to do even with my limited knowledge of lua.
What I'm wondering is which should be verified first? They all seem to be as fast as one another (`fd --color never --type f` (in the case of `fd` colors double the execution time from 50ms to 100ms), `find` and `rg --files`).

Luxed:
Fixes #69

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Support for:
 * LSP (references, document symbols, workspace symbols)
 * Treesitter
 * Grep
-* Files (git, fd)
+* Files (git, fd, rg)
 * Vim (command history, quickfix, loclist)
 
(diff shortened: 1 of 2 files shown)


tjdevries:
Makes sense to me. Thanks!

PR #77, opened 2020-09-14
[RDY] feat: add builtin.lsp_code_actions

TimUntersberger:
The new picker `lsp_code_actions` displays the available code actions for the current line and executes the selected action.

image


tjdevries:
ooohhhhh cool idea :)

When I use this one, I will probably use @rockerBOO 's `dropdown` view and make it smaller on the screen.

Let me know when you want me to review it.

brandoncc:
This is awesome! I do like :CocActions (when it works right). I'm slowly moving over to lsp and telescope, so I'm happy to see this!

Luxed:
That's a cool idea!
I also found CocActions to be rather hit or miss, so I ended up using the regular action selection instead.
If I could make it behave in the same way it would be awesome: input a number and it directly selects what I want from the list. Do you think this would be possible (in the future)?

TimUntersberger:
>  input a number and it directly selects what I want from the list

I mean this should be possible pretty easily. Would you want to see the number or just use the implicit index?

TimUntersberger:
The `get_generic_fuzzy_sorter` requires at least `2` characters as input. Is there a better sorter to use for this case?

TimUntersberger:
@tjdevries RDY

Luxed:
> I mean this should be possible pretty easily. Would you want to see the number or just use the implicit index?

I would want to see the index for sure. But for now, there's nothing to worry about, what has been done is more than enough!

> Also the layout seems way to big for this use case. Is there an existing layout I can use that would not take up all of the screen (Maybe something like the omni complete dropdown) ?

I'm guessing what @tjdevries was referring to by saying `dropdown` is the PR #79.
So keeping the default layout/theme might be a good idea so that the user have the ability to use whatever they'd like.

TimUntersberger:
> I would want to see the index for sure. But for now, there's nothing to worry about, what has been done is more than enough!

The current version has the index included in the name, but the problem is that you have to type at least 2 characters for the fuzzy sort to work.

tjdevries:
So the way I would do this is in my config I would just write a function that calls this and then uses something like `themes.get_dropdown()` or similar to make it small when we call it.

I think this seems fine for now (I didn't test a lot) but we can improve in subsequent PRs.

PR #65, opened 2020-09-12
feat: Add dropdown layout strategy

rockerBOO:
Adds a dropdown strategy.

Adds results_title and preview_title options.

```vim
nnoremap <Leader>f <cmd>:lua require'telescope.builtin'.find_files({ sorting_strategy = "ascending", preview_cutoff = 200,border = false, layout_strategy = "dropdown", prompt = "",
width = 50,winblend = 3, results_title = "", borderchars = { '', '', '', '', '', '', '', ''} })<CR>
```

https://clips.twitch.tv/ExpensiveExquisitePigeonVoHiYo

PR #79, opened 2020-09-15
feat: Themes

rockerBOO:
Add a starting point with themes. 
Add center layout strategy.

Would like to finalize down the theme API but this is a good start I think.

I think you'd normally just set your theme in your opts table.

`require'telescope.builtin'.find_files{ theme = "dropdown" }`

Which doesn't exist currently.

--

This is how I am using it right now.

```
local opts = GetTheme("dropdown", { winblend = 3 })
require'telescope.builtin.find_files(opts)
```

Clip of it in action:
https://clips.twitch.tv/VictoriousYawningAnacondaWOOP

See #65

### DIFF
--- a/lua/telescope/themes.lua
+++ b/lua/telescope/themes.lua
@@ -1,0 +1,57 @@
+-- Prototype Theme System
+-- Currently certain designs need a number of parameters.
+--
+-- local opts = GetTheme("dropdown", { winblend = 3 })
+-- 
+
+local Theme = {}
+Theme.__index = Theme
+
+local merge = function(table1, table2)
+  for k, v in pairs(table2) do table1[k] = v end
+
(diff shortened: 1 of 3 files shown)


lua/telescope/themes.lua
```diff
@@ -0,0 +1,57 @@
...
+
+  return setmetatable({
+    layout_strategy = opts.layout_strategy,
+    border = opts.border,
+    sorting_strategy = opts.sorting_strategy,
+    prompt = opts.prompt,
+    results_title = opts.results_title,
+    preview_title = opts.preview_title,
+    borderchars = opts.borderchars,
+  }, Theme)
```
tjdevries, line 27: just use opts. it's fine.

PR #85, opened 2020-09-16
WIP: Fuzzy Files Find

ThePrimeagen:


ThePrimeagen:
@rockerBOO Heythis should work now.  Change your git_files impl to use get_prime_fuzzy_file

I _think_ it should all work  :)

rockerBOO:
Looks good, feels good :). Yes the code is working too :)

ThePrimeagen:
Hey @rockerBOO this definitely needs some perf, but I think some of the perf wins can be done on the outside of the sorter (meaning a proper caching mechanism).

Here is an image, do you think that this sorting mechanism is correct?
Selection_139

rockerBOO:
What changes if you do something like `widget helpers`?

I think this order is correct due to near the end, continuous characters and in the filename matching. If I wanted to search on directory, I would need to either be more accurate or be more fuzzy.

I might also try `widg helpers`.

tjdevries:
prime, a good way to test that I was doing before is you can try running `score` on two items and then just asserting that they compare correctly.

```lua
local scorer = require(...)

local prompt = "widgethelpers"
local good_score = scorer:score(prompt, "packages..")
local ok_score = scorer:score(prompt, "somepackages..")

assert(good_score < ok_score)
```

And then I just had a bunch of those and then i `luafile %` so I can make sure that I'm staying stable over my stuff. At some point I will fix test harness.

ThePrimeagen:
That previous post was more along the lines, "is this correct?"

I have some ideas to make this all better, and hopefully the code much better.

Also just found a bug.  `/` apparently causes issues :)

ThePrimeagen:
It is still too slow to be used.  I'll be improving this today ten i'll merge it.

I want it to be "usable." before releasing it.

Also @tjdevries I am going to pursue that file caching mechanism :)

tjdevries:
:+1: cool sounds good.

We could try and save some of the file info a `trie` or something to save memory space if we're going to leave it floating around forever.

tjdevries:
(Example impl https://github.com/norcalli/nvim-colorizer.lua/blob/master/lua/colorizer/trie.lua that would we use / borrow )

ThePrimeagen:
hey @rockerBOO @tjdevries  there is now an option g:telescope_cache_results that will use the line cacher that I wrote last night.  I believe I have all the bugs worked out as of now.  for my 17K files project gets reduced way down and reduces the time per keypress by a decent amount of time.

Give it a try if you want too.

rockerBOO:
Does it detect if its a static or non-static job before running the caching? I think I'm getting caching of duplicate lines when I use `get_prime_fuzzy_file` on `find_files` using `fd`.

2020-09-18-132753_4480x1440_scrot
So here I think the duplicate file is from the cache. I could be wrong.

I think it would be nice for it to not cache when using non-static results, even if it's set globally. Or maybe some other option there.   

Because get_prime_fuzzy_file could take options, maybe we pass `get_prime_fuzzy_file({ cache = false })` and that would make it never use cache.

ThePrimeagen:
@rockerBOO could you attach the list of original files and your prompt that caused this? (just do `git ls-files -o -exclude-standard -c`)

So cached tracking only happens on `static` finders.  just because you cannot track dynamic non-sense.

rockerBOO:
It happens in the telescope.nvim repo if you search for `N` right now with duplicate

```
lua/telescope/_private/NGRam.lua
lua/telescope/_private/NGRam.lua
```

rockerBOO:
using `git_files` filters it properly, using the static `git ls-files `

ThePrimeagen:
weird.  With or without let g:telescope_cache_results = 1 I do not get duplicates!

rockerBOO:
with/without g:telescope_cache_results it has a duplicate. I am assuming there must be a bug with how `fd` works with telescope right now.

ThePrimeagen:
Does it happen with / without prime_fuzzy?

rockerBOO:
with the default `get_fuzzy_file()`, `N` doesn't do any sorting, `NG` puts NGRam first but shows a bunch of other items too. Even doing exact match on the filename `NGRam.lua` shows a bunch of other options too.

ThePrimeagen:
@rockerBOO Just finished line caching.  Its much faster, still not fast af, but fast.  And I think it matches better than FZF.

lua/telescope/builtin.lua
```diff
@@ -45,14 +46,18 @@ builtin.git_files = function(opts)
     opts.cwd = vim.fn.expand(opts.cwd)
   end
 
+  opts.track_results = vim.g["telescope_cache_results"]
+  local use_prime = vim.g["telescope_prime_fuzzy_find"]
```
tjdevries, line 50: If you wanted to do this for yourself, you should set a new value in `config.lua`

Probably something like `file_matching_strategy` or something, it can be set to `prime` to use your matcher.

lua/telescope/builtin.lua
```diff
@@ -45,14 +46,18 @@ builtin.git_files = function(opts)
...
+
   pickers.new(opts, {
     prompt    = 'Git File',
     finder    = finders.new_oneshot_job(
       { "git", "ls-files", "-o", "--exclude-standard", "-c" },
       opts
     ),
     previewer = previewers.cat.new(opts),
-    sorter    = sorters.get_fuzzy_file(),
+    sorter    = use_prime and sorters.get_prime_fuzzy_file(opts) or
```
tjdevries, line 59: If we add the configuration, it should apply to all file based telescope builtins.

I'm not sure I want to 100%. People can just call the builtins with `sorter = sorters.get_prime_fuzzy_file()` or similar.

lua/telescope/cache.lua
```diff
@@ -0,0 +1,137 @@
+local log = require('telescope.log')
+
+local PromptCache = {}
+
+function getFirstByteDiffIdx(a, b)
```
tjdevries, line 5: snake_case

lua/telescope/cache.lua
```diff
@@ -0,0 +1,137 @@
...
+  local max_idx = #a
+  while a:byte(idx) == b:byte(idx) and idx <= max_idx do
+    idx = idx + 1
+  end
+
+  return idx
+end
+
+function PromptCache:new(opts)
+  self.__index = self
```
tjdevries, line 16: You should set `PromptCache.__index = PromptCache` on line 4 instead of this.

lua/telescope/finders.lua
```diff
@@ -46,12 +49,17 @@ function JobFinder:new(opts)
   --        vim.loop.new_pipe (stdin / stdout). stdout => filter pipe
   --        rg huge_search | fzf --filter prompt_is > buffer. buffer could do stuff do w/ preview callback
 
+  self.__index = self
```
tjdevries, line 52: This should not be necessary.

lua/telescope/finders.lua
```diff
@@ -64,23 +72,46 @@ function JobFinder:new(opts)
...
+    local lines = self._cached_lines
+    if self.track_results then
+      local cached_results = self.cache:get_cache(prompt)
+
+      if cached_results and type(cached_results) == "table" then
+        lines = cached_results
+      end
     end
 
+    log.info("cached lines", #lines)
```
tjdevries, line 101: info is probably too high for this message.

lua/telescope/pickers.lua
```diff
@@ -422,9 +423,12 @@ function Picker:find()
       PERF("Displayed FN Amount", displayed_fn_amount)
     end)
 
+
+    local start = vim.fn.reltimefloat(vim.fn.reltime())
```
tjdevries, line 427: let's remove these.

This can be done either by profiling later or by adding it in the finder directly.

lua/telescope/sorters.lua
```diff
@@ -70,13 +71,229 @@ sorters.get_levenshtein_sorter = function()
...
+  end
+
+  local line_cache = {}
+  local function get_cache_for(line, prompt)
+    if not line_cache[line] then
+      line_cache[line] = Cache.PromptCache:new()
+    end
+
+    return line_cache[line]:get_cache(prompt)
+  end
```
tjdevries, line 223: this strikes me as though we're going to blow up on memory for some of the larger searches. I'll have to ask you about this in our call.

lua/telescope/sorters.lua
```diff
@@ -70,13 +71,229 @@ sorters.get_levenshtein_sorter = function()
...
+
+  local function cache_line_results(line, prompt, results)
+    local cache = line_cache[line]
+    cache:set_cache(prompt, results)
+  end
+
+  local function cache_remove_line(line)
+    line_cache[line] = nil
+  end
```
tjdevries, line 232: Ah, maybe my previous comment is unwarranted if we remove them as we go.

lua/telescope/sorters.lua
```diff
@@ -70,13 +71,229 @@ sorters.get_levenshtein_sorter = function()
...
+    scoring_function = function(_, prompt, line)
+      if prompt == "" then
+        return 1
+      end
+
+      local cache = get_cache_for(line, prompt) or {}
+      local areas = cache.areas or create_areas(line)
+      local new_cache = {
+        areas = areas
+      }
```
tjdevries, line 244: I'm not sure why you can't use a default_dict strategy here. It seems like it would provide you the same idea (at least for areas) with less overhead.

lua/telescope/sorters.lua
```diff
@@ -70,13 +71,229 @@ sorters.get_levenshtein_sorter = function()
...
+
+      local exact_match = false
```
tjdevries, line 246: don't know if you want to expose this or not.

lua/telescope/sorters.lua
```diff
@@ -70,13 +71,229 @@ sorters.get_levenshtein_sorter = function()
...
+      local idx = 1
+
+      while idx <= #prompt and not exact_match do
+        exact_match = is_upper_case(prompt:byte(idx))
+        idx = idx + 1
+      end
+
+      local points = cache.points or {
```
tjdevries, line 254: I've got to read a bit more about what you're doing w/ points.

lua/telescope/sorters.lua
```diff
@@ -70,13 +71,229 @@ sorters.get_levenshtein_sorter = function()
...
+          bonus_caps_matcher = 0
+        end
+
+        score = score +
+        (cc + bonus_caps_matcher) * -- Using caps should be worth more in score
+        (1 +
+          curr.area_idx / area_count + -- More specific is worth more
+          cc / #areas[curr.area_idx].word + -- More percent of an area matched
+          (curr.exact_match and 0.1 or 0) + -- case matching
+          (curr.area_idx == area_count and 0.5 or 0)) -- A bonus for final word
```
tjdevries, line 147: nit: probably indent the rest of this by one.

lua/telescope/sorters.lua
```diff
@@ -70,13 +71,229 @@ sorters.get_levenshtein_sorter = function()
...
+-- script/foo.js
+-- scriptf
+-- blah
+-- [ b, l, a
+
+sorters.get_prime_fuzzy_file = function(opts)
+  opts = opts or {}
+
+  local function create_areas(line)
+    local split_line = vim.split(line, os_sep)
```
tjdevries, line 84: You should use the `fast_split` I sent you.

lua/telescope/pickers.lua
```diff
@@ -369,7 +369,7 @@ function Picker:find()
     local process_result = function(entry)
       -- TODO: Should we even have valid?
       if entry.valid == false then
-        return
+        return -1
```
tjdevries, line 372: I'm not sure why these have to return things (It's OK if they do, but I didn't read far enough to understand that)

lua/telescope/finders.lua
```diff
@@ -46,12 +49,17 @@ function JobFinder:new(opts)
...
   --        rg huge_search | fzf --filter prompt_is > buffer. buffer could do stuff do w/ preview callback
 
+  self.__index = self
   local obj = setmetatable({
     entry_maker = opts.entry_maker or make_entry.from_string,
     fn_command = opts.fn_command,
     static = opts.static,
     state = {},
 
+    track_results = opts.track_results,
```
tjdevries, line 59: So it's not exactly about tracking results it seems, it's more that progressively narrowing results over time.

I'm not sure if this is 100% true for you, but can something not match the first parts of your prompt and then match later? I know for some of my sorters it is possible.

lua/telescope/builtin.lua
```diff
@@ -45,14 +46,18 @@ builtin.git_files = function(opts)
     opts.cwd = vim.fn.expand(opts.cwd)
   end
 
+  opts.track_results = vim.g["telescope_cache_results"]
```
tjdevries, line 49: It seems `track_results` is related to both the sorter and the finder, it should probably not be passed in by the user.

I'd say it should be determined by some interaction of the finder and the sorter.

tjdevries (changes requested):

lua/telescope/sorters.lua
ThePrimeagen: Well, points.  They will be redone... soonish.  When I make the speed improvement I am planning.  I effectively reduce the need for a tree and turn it into a list of leaves.

tami5:
@ThePrimeagen any future plans on working on this pr?

tami5:
Closed for being inactive or abandoned.

ThePrimeagen:
good close, i'll update later

Issue #71, opened 2020-09-13
Cannot search for '--' as bat assumes it to be a flag.


the-codingguy:
While searching for -- (in live grep or any other that uses bat to show preview) 

![issue](https://user-images.githubusercontent.com/27463862/93010463-7eded680-f5aa-11ea-8d25-78572f1e744e.png)

tjdevries:
Ok, think this works now. Please let me know if it isn't working anymore.

PR #87, opened 2020-09-16
use double dash to fix special filenames

errx:
fix issue #71 

Add double dash when calling underlying programs to prevent mixing of options and arguments.

This should fix all problems with filenames like '--my-file-with-strange-name'


tjdevries:
Ah, very smart! Thanks for the fix :)

PR #92, opened 2020-09-17
fix: Add nil checks for state when preview hasn't been initialized yet

### LINKED ISSUE #89: Bug: Exiting an empty search throws an error
When trying to quit an empty search, the following error happens:
image

This makes telescope completely unusable in conjunction with #68.
When this bug happens, a restart is required.

Edit: tested on Windows.

Luxed:
Fixes #89.

I'm not sure about this fix. Please tell me if something else should be done instead.


tjdevries:
Yeah, good calll, noticed that this morning. TY for beating me to the fix :grin:

PR #99, opened 2020-09-18
Respect sp, vsp, tabe mappings for buffers builtin

Conni2461:
Respect `sp`, `vsp` and `tabe` mappings.

Maybe just a hack. I have to think about this later again.


ruifm:
I can confirm this fixes https://github.com/nvim-lua/telescope.nvim/issues/98

tjdevries:
Thanks, this is a good fix for now. Will have to consider more extensible ways to open in the future, but it looks good to me.

Thanks @Conni2461 !!

PR #104, opened 2020-09-20
buffers(): Trim cwd from bufname if possible

ruifm:
Before this change, when using the `buffers()` builtin, the full path of the buffer filename was always shown because the `vim.api.nvim_buf_get_name()` API function is used (and that function always returns the full path).

I don't know lua at all so this may be a very naive solution, but it worked for me :laughing: .

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -181,6 +181,8 @@ end
 function make_entry.gen_from_buffer(opts)
   opts = opts or {}
 
+  local cwd = vim.fn.expand(opts.cwd or vim.fn.getcwd())
+
   local get_position = function(entry)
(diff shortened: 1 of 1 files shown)


tjdevries:
Thanks! nice change. May add configuration option for it in the future (in case someone always wants full path displayed) but seems good for now.

ruifm:
>Thanks! nice change.

You're welcome, looking forward to contributing more in the future as I get more comfortable with lua. Btw, I am `MrSommerfeld` from the beginbot stream :laughing: in case you remember.

> May add configuration option for it in the future (in case someone always wants full path displayed) but seems good for now.

You can always do that by passing `opts = { cwd = "/"}`  (or something illegal/gibirish that can never be part of a path) but I guess that can be considered a hack.

PR #106, opened 2020-09-22
Fix issue with trimming path separator '/' for buffers()

ruifm:
Follow up of https://github.com/nvim-lua/telescope.nvim/pull/104 that introduced a bug in which if `vim.fn.getcwd()` returns a string with a trailing path separator `/`, it causes the bufname path to be wrongly trimmed (e.g. `$(pwd)/foo/bar -> /foo/bar` instead of `$(pwd)/foo/bar -> foo/bar`).

This patch just checks the cwd trailing char and applies an offset.

@tjdevries I'm noticing that so far working with paths in lua has been quite hacky which in turn makes the appearence of these bugs easier. Is there a proper way to work with paths?

I have looked at [plenary.nvim](https://github.com/nvim-lua/plenary.nvim) , and more concretely at  [path.lua](https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/path.lua) which according to the projects README it's suppose to be the equivalent of python's pathlib. But from looking at that file, it's still far behind which leaves me for now with manual string handling.

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -212,7 +212,12 @@ function make_entry.gen_from_buffer(opts)
 
     -- if bufname is inside the cwd, trim that part of the string
     if bufname:sub(1, #cwd) == cwd  then
-      bufname = bufname:sub(#cwd + 1, #bufname)
+      local offset =  0
+      -- if  cwd does ends in '/', we need to take it off
+      if cwd:sub(#cwd, #cwd) ~= '/' then
+        offset = 1
+      end
+      bufname = bufname:sub(#cwd + 1 + offset, #bufname)
     end
 
     local position = get_position(entry)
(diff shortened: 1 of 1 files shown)


tjdevries:
should probably use os_separator, which we've got elsewhere in telescope.

Yes, we should probably be using the `path.lua` from plenary. It is in need of some more features :laughing: 

I also maintain plenary, so if you want to add any features to make path handling easier here, let me know.

tjdevries:
( I just added os sep fix. If you're interested, we can work on improving plenary's `path` module, and then use it here )

ruifm:
> ( I just added os sep fix. If you're interested, we can work on improving plenary's `path` module, and then use it here )

Thanks, it was in my TODO but it was just a busy week and it got buried in my GH notifications. I am interested, will look into it and start making some PR's that implement pathlib equivalents.

tjdevries:
Also, we should use libuvs stuff for access file system, not  viml wrappers like I had before. Can explain more later.

PR #107, opened 2020-09-22
small fixes.

sunjon:
When using the Tree-sitter finder, the target is always in the current buffer.
The `goto_file_selection()` function ran the `command` parameter every time, causing "unsaved notification" messages as it tried to reload the modified buffer.

The second change merely changes the order of  'Kind  Text' in the ordinal string, and adds parenthesis for better reading. Perhaps some dialogs could be improved with formatted columns for each field.


tjdevries:
Thanks, LGTM

Conni2461:
Hey, the changes in `actions.lua` broke something.
Easy way to reproduce the error is, open nvim (with startify installed) and open file with find_files.
Error is
```
E5108: Error executing lua ...ig/nvim/plugged/telescope.nvim/lua/telescope/actions.lua:104 Invalid buffer id: 1
```
The reason for this error is because the startify window had id: 1 and on opening a new buffer this buffer gets deleted and id: 1 does no longer exist. So we have to run `bufnr = vim.api.nvim_get_current_buf()` after `vim.cmd` again in order to get the new bufnr.

PR #108, opened 2020-09-23
New builtin: Lua packages

sunjon:
A picker to select/reload loaded Lua packages.
Requires Plenary for reload action.


lua/telescope/builtin.lua
```diff
@@ -401,6 +401,46 @@ builtin.command_history = function(opts)
...
+builtin.reloader = function(opts)
+  opts = opts or {}
+  local package_list = vim.tbl_keys(package.loaded)
+
+  -- filter out packages we don't want and track the longest package name
+  opts.column_len = 0
+  for index, module_name in pairs(package_list) do
+    if type(require(module_name)) ~= 'table' or module_name:sub(1,1) == "_" or package.searchpath(module_name, package.path) == nil then
+      table.remove(package_list, index)
+    elseif module_name:len() > opts.column_len then
```
tjdevries, line 413: I generally use `#` instead of `:len()`. Just a code style comment.

lua/telescope/make_entry.lua
```diff
@@ -277,4 +277,19 @@ function make_entry.gen_from_treesitter(opts)
   end
 end
 
+function make_entry.gen_from_packages(opts)
+  return function(module_name)
+    local path = package.searchpath(module_name, package.path) or ""
+    display = string.format("%-" .. opts.column_len .. "s : %s", module_name, vim.fn.fnamemodify(path, ":~:."))
```
tjdevries, line 283: I think it's better to use the `make_display` style functions that are used elsewhere in the file.

This way calling vim functions is deferred until we actually need to display something.

tjdevries:
Can you fix the conflicts and then squash your history to just be one commit? So I can see the changes for this PR?

Then we can probably merge.

Later I will add table support for display.

PR #110, opened 2020-09-23
Nil check for no preview window

fahadbr:
When the window is too narrow for a preview window - state is nil and causes the below error and doesn't allow the pop up window to close.

screenshot-20200922-1600831275

I added a simple nil check and this seems to work fine but please advise if something more involved is required


tjdevries:
Thanks!

Issue #115, opened 2020-09-23
 Telescope file openers preventing folds 


mhartington:
Hey there. Not sure what's happening here, but when opening files with the `find_files` or `git_files`, it seems it's preventing me from setting folds with autocmds.

Given the following autocmd:

```viml
  autocmd FileType javascript,typescript,typescript.tsx,typescriptreact,json,lua setl foldmethod=syntax
```

When I open a JS/TS/JSON/Lua file, the fold method is set, but  it cannot register any sections to fold.

As opposed to vim-clap, which works as expected.

![telescope](https://user-images.githubusercontent.com/2835826/94013681-6758d800-fd78-11ea-97b5-ff0efb595e37.gif)

Not sure why this is, as the general functionality that is used between clap and telescope are the same. Could my Autocmds be off?

tjdevries:
(Was your problem fixed by #116 ? )

mhartington:
Yes! Sorry, forgot to tag this issue in the PR. Closing

PR #116, opened 2020-09-23
run filetypedetect on edit

mhartington:
Possible fix for #115


tjdevries:
I will probably rewrite a bit of this code later, but this seems like a strict improvement.

PR #117, opened 2020-09-23
feature: Vim help-tags picker

sunjon:
A finder dialog for Vim help-tags.
Until there is a way to retrieve help-tags natively, I've used the nightly auto-generated tags from [tweekmonster/helpful.vim](https://github.com/tweekmonster/helpful.vim).

One nice thing is that the version at which a feature was introduced is in the tagfile, and available in the finder display if called with `require'telescope.builtin'.help_tags({show_version = true})`.

I've put no thought yet into which is the appropriate sort algorithm for this, and the current one can produce confusing results.

I found a function for help-tags labelled WIP in the code, that wasn't hooked into anything, and it "just worked"™  🥳


sunjon:
The help.previewer's use of search pollutes the search history.
Maybe best not to merge this until that's fixed.

sunjon:
The latest changes are an attempt to convert the output of  `taglist()[1].cmd`  from `/foo` to `search('foo')`.
There are issues with escaping characters, which I attempted to solve with `fnameescape()`, but it's not working pefectly: Queries containing quotes and some special characters break things.

sunjon:
search() string is now more correct.
still breaks on entries with single quotes in, and their appears to be a bug outside of this PR where strings with colon in aren't passed through to preview correctly. (I've left a couple of `print([debug]..)` statements in previewers.lua that illustrate it.

For the quotes issue, try query `'path'`

For the colon issue, try query `[:tab`

tjdevries:
Fixed it up on stream last night, working great for me now. Awesome work :) 

I may tinker w/ a few things later, but wanted to merge it cause it's so awesome.

PR #119, opened 2020-09-23
fix: refresh `bufnr` after executing edit command

sunjon:
Fixes problem mentioned in https://github.com/nvim-lua/telescope.nvim/pull/107#issuecomment-697346324


Conni2461:
Thanks, that is exactly what i meant.

tjdevries:
<3 Thanks

PR #120, opened 2020-09-24
Strips CR from vimgrep output on Windows

RMichelsen:
Vimgrep output from `rg` on Windows contains CRLF endings which result in lines ending in `^M` in the picker window. This should strip carriage returns from all output lines, with the precondition that vim is running on Windows.

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -102,11 +102,18 @@ function make_entry.gen_from_vimgrep(opts)
     return display
   end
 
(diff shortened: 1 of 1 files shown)


tjdevries:
So I just pushed a fix to plenary that I think will do this universally, it just filters out `\r` all the time, because no one wants to see it i think :laughing: 

If that fixes your problem, we can close this.

RMichelsen:
Just checked, that fixes it :)

PR #124, opened 2020-09-25
Fix help preview and add highlighting to search result

sunjon:
This fixes an issue with no preview content because `vim.o.tags` did not include correct path.
Also adds highlighting  in the preview window to the help_topic/query

### DIFF
--- a/lua/telescope/previewers.lua
+++ b/lua/telescope/previewers.lua
@@ -368,13 +368,25 @@ end, {})
 
 previewers.help = defaulter(function(_)
   return previewers.new {
-    preview_fn = function(_, entry, status)
+    setup = function()
+      return {}
+    end,
+
+    teardown = function(self)
+      if self.state and self.state.hl_id then
+        pcall(vim.fn.matchdelete, self.state.hl_id, self.state.hl_win)
+        self.state.hl_id = nil
+      end
+    end,
+
+    preview_fn = function(self, entry, status)
       with_preview_window(status, nil, function()
         local special_chars = ":~^.?/%[%]%*"
 
         local escaped = vim.fn.escape(entry.value, special_chars)
         local tagfile = vim.fn.expand("$VIMRUNTIME") .. '/doc/tags'
         local old_tags = vim.o.tags
+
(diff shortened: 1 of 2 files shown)


lua/telescope/previewers.lua
```diff
@@ -401,6 +405,12 @@ previewers.help = defaulter(function(_)
         vim.cmd "norm! gg"
         vim.fn.search(search_query, "W")
         vim.cmd "norm  zt"
+
+        if status.preview_search_hl_id then
+          vim.fn.matchdelete(status.preview_search_hl_id)
+          status.preview_search_hl_id = nil
```
tjdevries, line 411: we don't want to use `status` here, the `previewer` has a `state` field, that you can `setup` and `teardown` exactly for this kind of thing (so you can clear the highlight when we're done w/ the previewer as well.

lua/telescope/previewers.lua
```diff
@@ -373,7 +373,11 @@ previewers.help = defaulter(function(_)
 
         local escaped = vim.fn.escape(entry.value, special_chars)
         local tagfile = vim.fn.expand("$VIMRUNTIME") .. '/doc/tags'
+        local old_tags = vim.o.tags
+        vim.o.tags = tagfile
         local taglist = vim.fn.taglist('^' .. escaped .. '$', tagfile)
+        vim.o.tags = old_tags
```
tjdevries, line 379: I didn't think we need to set old_tags if we pass the tag file to `taglist`. When I was testing this locally, I didn't seem to need to do this.

sunjon: Perhaps you have something else dynamically seeing your tags var?

sunjon:
adding a screenshot, for proposed gallery

help

lua/telescope/previewers.lua
tjdevries: Ping on this, basically only thing stopping from merging.

tjdevries:
I added a wiki, can you make a gallery page there @sunjon? You can post as many cool pics as you want there. 😁

PR #129, opened 2020-09-28
WIP: Changed git command to show all files inside project

Norlock:
The current git command only shows the git files inside the working directory, not for the complete project. This seems to be the same functionality as the find_files command, this change will show all the files inside the project.


tjdevries:
Yeah, this makes sense I think. If some people do not like it, we can make it an option instead. Thanks!

PR #130, opened 2020-09-29
Fix if-clause after latest changes

Conni2461:
After the lastest changes in 3592b1f, this if branch is no longer reachable.

3592b1f still works but now we run `:edit #bufnr`, rather than `nvim_win_set_buf`.
This would be a valid choice but then we should remove the if clause. 
Regardless I would still vote for `nvim_win_set_buf` because the `edit` is not necessary here.


tjdevries:
Thanks, good call. Bad programming by me :)

PR #133, opened 2020-09-30
WIP: options menu

sunjon:
WIP finder using options.lua

### DIFF
--- a/lua/telescope/builtin.lua
+++ b/lua/telescope/builtin.lua
@@ -443,7 +443,7 @@ builtin.vim_options = function(opts)
 
       ok, option.current_value = pcall(vim.api.nvim_get_option, o.full_name)
       if not ok then
-        print("WHAT'S GOING ON WITH " .. o.full_name)
+        -- print("WHAT'S GOING ON WITH " .. o.full_name)
         goto continue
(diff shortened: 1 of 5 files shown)


sunjon:
wip

teto:
thanks for making a vimmer dream come true

teto:
Does this hide window-scoped options ? I could not find `relativenumber` when typing `relative` in `:Telescope vim_options`

tjdevries:
Yeah, it does curently. It's just options with global scope.

I need to implement some other stuff for this to make sense for setting window options. 

For example... You are not in the window you expect to be in when you are editing. So I will need to track and apply and it can no longer be set from command line. So it is complicated haha

tjdevries:
You can make an issue with info regarding what you want, that'd be helpful for tracking purposes.

teto:
there was this one https://github.com/nvim-lua/telescope.nvim/issues/201. As a first step it could just prefill the command line with 
`:set relativenumber` and let the user press enter so that the correct window gets modified. I understand you want something even more polished so I'll wait, so happy already I will sunset nvim-palette and advise people to use this shiny new telescope.

PR #134, opened 2020-10-01
feat: Manpages finder (output of apropos)

sunjon:
man

### DIFF
--- a/lua/telescope/builtin.lua
+++ b/lua/telescope/builtin.lua
@@ -740,13 +740,13 @@ end
 builtin.man_pages = function(opts)
   opts = opts or {}
 
-  -- TODO: make man sections configurable?
-  local cmd = "apropos --sections=1 ''"
+  local cmd = opts.man_cmd or "apropos --sections=1 ''"
+
   local f = assert(io.popen(cmd, 'r'))
   local pages = assert(f:read('*a'))
   f:close()
 
-  lines = {}
+  local lines = {}
   for s in pages:gmatch("[^\r\n]+") do
     table.insert(lines, s)
   end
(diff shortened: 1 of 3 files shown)


sunjon:
manpages

lua/telescope/builtin.lua
```diff
@@ -695,4 +695,43 @@ builtin.current_buffer_fuzzy_find = function(opts)
   }):find()
 end
 
+builtin.man_pages = function(opts)
+  opts = opts or {}
+
+  -- TODO: make man sections configurable?
+  local cmd = "apropos --sections=1 ''"
```
kamalmarhubi, line 702: drive-by: I'd definitely like to configure this. For my uses, a better default would be to search all sections.

kamalmarhubi (commented):
this is pretty cool :-)

lua/telescope/builtin.lua
tjdevries: Can be `opts.man_cmd` or similar.

lua/telescope/builtin.lua
```diff
@@ -695,4 +695,43 @@ builtin.current_buffer_fuzzy_find = function(opts)
...
+    table.insert(lines, s)
+  end
+
+  pickers.new(opts, {
+    prompt    = 'Man',
+    finder    = finders.new_table {
+      results = lines,
+      entry_maker = make_entry.gen_from_apropos(opts),
+    },
+    -- previewer = previewers.qflist.new(opts),
```
tjdevries, line 718: Do you think we can use the Man command to be the previewed?

sunjon: yeah, I think so. I can't remember why I stopped. I think it was because the rendering of the Man page itself in nvim is still slow imo.
The whole sequence of generating an uncached list from apropos(0.5s), the slower-than-other-filetypes rendering of Man already seems slow in comparison to other finders(0.5s).

I will implement it but I think something needs to be improved to make the experience tolerable.

tjdevries: I thought it was just the first loading of a man page that takes a long time (I don't know for sure though)

tami5:
This is extremely useful  and smart, thanks @sunjon . Hope it gets merged soon.

PR #135, opened 2020-10-01
Add highlight for selection caret

pwntester:


lua/telescope/pickers.lua
```diff
@@ -723,6 +723,14 @@ function Picker:set_selection(row)
       0,
```
tjdevries, line 723: Let's add this highlight not starting from 0 anymore (since we're just going to overwrite immediately after.

plugin/telescope.vim
```diff
@@ -1,6 +1,7 @@
 
 " Sets the highlight for selected items within the picker.
 highlight default link TelescopeSelection Visual
+highlight default link TelescopeSelectionCaret Visual
```
tjdevries, line 4: Let's make the default equal to TelescopeSelection, so that it's easier for people to keep their same setup.

pwntester: makes sense, thanks!

tjdevries:
Thanks!

PR #142, opened 2020-10-04
fix: Pass --paging=always to bat

kamalmarhubi:
This gets rid of a slightly annoying difference between files that fit
in the preview pane and ones that don't: hitting the scroll-preview
mappings would complain about a closed stream. Always using the pager
ensures there's always a process at the other end, even if asking it
to scroll doesn't actually do anything.


tjdevries:
Yeah, that makes sense, thanks! I think we toggled that back and forth at one point, but I like it this way. Thanks!

PR #153, opened 2020-10-07
[WIP] Global options used by pickers

parmort:
Use a theme by default by doing something like the following:

```lua
require'telescope'.setup{
    defaults = require'telescope.themes'.get_dropdown{ other config here }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,6 +308,14 @@ nnoremap <Leader>f :lua require'telescope.builtin'.find_files(require('telescope
 
 Themes should work with every `telescope.builtin` function.  
 
+To use a theme by default, use:
+
+```lua
+require('telescope').setup{
+  default = require'telescope.themes'.get_dropdown{}
+}
+```
+
 If you wish to make theme, check out `lua/telescope/themes.lua`. If you need more features, make an issue :).
 
 ## Goals
--- a/README.md
+++ b/README.md
@@ -308,11 +308,13 @@ nnoremap <Leader>f :lua require'telescope.builtin'.find_files(require('telescope
 
 Themes should work with every `telescope.builtin` function.  
 
-To use a theme by default, use:
+To use a theme globally, put it in setup
 
 ```lua
 require('telescope').setup{
-  default = require'telescope.themes'.get_dropdown{}
+  default = require('telescope.themes').get_dropdown({
+    -- Other config here
+  })
 }
 ```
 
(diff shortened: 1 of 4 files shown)


parmort:
I ran into a snag where the `select` call in line 103 of `lua/telescope/config/resolve.lua` returns `nil` because the number of variadic arguments is less than the selector. Still need to investigate further when I have some time.

parmort:
I think if we decide to use the `{ theme = ... }` in the global settings, we should use it in the finders' settings.

Once this PR was finished with the predefined method, I was planning on doing something similar in a different PR to get it done in one fell swoop.

parmort:
Are there any more options that could be set in the global settings? A few aren't, hence my previous commit.

tjdevries:
Hmmm I don't think I like this exact way of doing it.

The main reason is:

1. We always have to remember to do this every time going forward (which is hard to remember)
2. This can be accomplished by users quite easily (by passing the same options dictionary to their own builtin ones
3. The `options` that a `Picker` expects is not exactly the same as the `config` ones...

Hmm, can you give me a few days to think some more about this? It may just be that I should change how the opts are set.

parmort:
Sure. One other option to perhaps consider: have another key in the setup table like so:

```lua
require'telescope'.setup{
    defaults = { ... },
    picker_defaults = { ... },
    theme = '...'
}
```

These names are arbitrary. This way config options in `defaults` are separate from global picker options.

tjdevries:
Let's try tackling one issue at a time. I want to recap what your current problem is:

You want to be able to specify, in your setup, a default theme that you would use (unless overridden) for all of the builtins.

Is that correct?

parmort:
Yeah. Maybe for a different PR, I could look at having global options for pickers (like turning the previewer off, for instance).

Conni2461:
The pickers defaults can be set inside setup. https://github.com/nvim-telescope/telescope.nvim#customize-default-builtin-behavior

I might have stolen your PR. I am sorry about that :laughing: I have implemented it in #883

`theme = ""` is still not supported in the current setup as well as some other values. I just mentioned it here: https://github.com/nvim-telescope/telescope.nvim/issues/1003#issuecomment-917646441

I would close this PR. I am sorry about stealing your work / idea. If you wanna work on the `theme =` or other things. Feel free to. We have also a lot of other mentored projects :) https://github.com/nvim-telescope/telescope.nvim/issues/1190

PR #154, opened 2020-10-07
enable prompt_prefix option

pwntester:


tjdevries:
When you used this, did you try starting up a picker and then typing immediately after? I was getting weird race conditions when doing that before.

pwntester:
yep, works flawlessly for me

tjdevries:
Could you add a default in the config for this so people can pick one that they like by default?

lua/telescope/pickers.lua
```diff
@@ -59,6 +60,7 @@ function Picker:new(opts)
 
   return setmetatable({
     prompt = opts.prompt,
+    prompt_prefix = opts.prompt_prefix or '>',
```
pwntester, line 63: You mean something like this for the default value? or in a different place?

tjdevries:
I meant how I do it in `config.lua` and then how I get those values from `config` in the `Picker:new` section.

tjdevries:
Also, we should add a default highlight in `plugin/telescope.vim` for `TelescopePromptPrefix`

pwntester:
updated 👍

pwntester:
Im also considering adding an option to override the prompt. Currently, I need to duplicate all builtin picker code just to set the prompt to an empty string. It would be nice to be able to do that from the pickers options. Thoughts?

Speaking about that, I think that `prompt` and `prompt_prefix` are confusing. They could be refactored to `prompt_title` and `prompt` but that would be a breaking change

tjdevries:
hmmm, I think I'm fine with that. Although I'd rather not re-use `prompt` then. Let's just do something like `prompt_title`, and `prompt_text` or something like that?

tjdevries:
The breaking change is fine I think (I'm not sure how many people are setting custom prompt titles anyway)

pwntester:
Nice, will implement that in a different PR

PR #157, opened 2020-10-08
fix picking height as width

pwntester:
The problem Im trying to fix is the following: I set up a with of 120, but end up with a width of 43 (number of lines in my nvim window). By reading the code on `resolve.lua`, `resolve_width` is going to pick the minimum between `val` and the values passed to the function returned by `resolve_width`. In this case, it is going to select the min value between 120 (my desired width), 158 (number of cols) and 43 (number of lines), and therefore the width is set to 43.

Im not sure if the fix is correct though. The comments on `resolve.lua` seem to imply that its correct to pass `(col, lines)` so, if thats the case, then the fix should be changing the `_resolve_map` functions so that it picks the minimum between `val` and `lines` or `cols` (depending on what is resolving), but not the 3 of them

### DIFF
--- a/lua/telescope/pickers/layout_strategies.lua
+++ b/lua/telescope/pickers/layout_strategies.lua
@@ -140,9 +140,8 @@ layout_strategies.center = function(self, columns, lines)
   local prompt = initial_options.prompt
 
   -- This sets the height/width for the whole layout
-  local height = resolve.resolve_height(self.window.results_height)(self, columns, lines)
-  local width = resolve.resolve_width(self.window.width)(self, columns, lines)
-
+  local height = resolve.resolve_height(self.window.results_height)(self, lines)
+  local width = resolve.resolve_width(self.window.width)(self, columns)
   local max_results = (height > lines and lines or height)
   local max_width = (width > columns and columns or width)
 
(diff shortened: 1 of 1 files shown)


pwntester:
Please note that if this is the correct fix, other places in `layout_strategies` should be changed as well

rockerBOO:
Yes this looks good to me. My bad as I just sort of didn't think about the parameters.

pwntester:
line 219 (and maybe others?) also need to be adjusted then

rockerBOO:
Ok I think the current way it was doing it was accurate. I was told by @tjdevries to implement it this way but I don't fully understand how to fix it in a way to get the current correct expected value.

rockerBOO:
Well lets just say I don't know enough and maybe this will still be accurate but I felt like it wasn't correct.

tjdevries:
I know the problem, I'll fix tonight. It's because `select` doesn't just select only the first value, but all the rest of them.

We need to only pass the selected value to the things in resolve.lua

Thats the short version of the problem.

pwntester:
Yep, I had that impression but was not sure, thanks!

tjdevries:
OK, I think it's fixed now w/ latest commit :)

pwntester:
Confirm it works for me, thanks!

PR #165, opened 2020-10-09
gallery images

pwntester:


tjdevries:
I think you can just paste the image into the wiki and it will upload them, right? Or there's some way to upload images? I don't think I want to include them in the repo (as it greatly inflates the download size)

pwntester:
For some reason I cannot upload them directly to a wiki page, but I uploaded them to the original issue and then reference them in the wiki. Thanks!

PR #167, opened 2020-10-09
Add Colorscheme and Marks Support and Telescope Commands

glepnir:
@tjdevries cc

### DIFF
--- a/lua/telescope/builtin.lua
+++ b/lua/telescope/builtin.lua
@@ -774,4 +774,33 @@ builtin.man_pages = function(opts)
   }):find()
 end
 
+builtin.colorscheme = function(opts)
+  opts = opts or {}
+
+  local colors = vim.fn.getcompletion('', 'color')
+
+  pickers.new(opts,{
+    prompt = 'Change Colorscheme',
+    finder = finders.new_table {
+      results = colors
+    },
+    previewer = previewers.man.new(opts),
+    sorter = sorters.get_generic_fuzzy_sorter(),
+    attach_mappings = function(prompt_bufnr, map)
+      local change_colorscheme = function()
+        local selection = actions.get_selected_entry(prompt_bufnr)
+
+        actions.close(prompt_bufnr)
+        print(vim.inspect(selection.value))
+        vim.cmd("colorscheme " .. selection.value)
+      end
+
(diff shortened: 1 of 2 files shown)


lua/telescope/builtin.lua
```diff
@@ -774,4 +774,33 @@ builtin.man_pages = function(opts)
   }):find()
 end
 
+builtin.colorscheme = function(opts)
```
tami5, line 777: @glepnir That's very useful!!!! I wanted to create this function so bad! thank you ( also for Thinkvim). 

I wonder is it possible to include nvim-base16.lua as an option of its installed or loaded? and query from `base16.themes` array from selection, then on selection do `require'base16'(require'base16'.themes["selection"], true)` [reference](https://github.com/tami5/nvim/blob/main/lua/core/theme.lua#L75)

tjdevries: Perhaps we can do `local colors = vim.list_extend(opts.colors or {}, vim.fn.getcompletion('', 'color'))` on line 780.

This would allow users to pass their own lists of colorschemes as well (if they are not currently available).

glepnir: yep.

lua/telescope/builtin.lua
```diff
@@ -774,4 +774,33 @@ builtin.man_pages = function(opts)
...
+  opts = opts or {}
+
+  local colors = vim.fn.getcompletion('', 'color')
+
+  pickers.new(opts,{
+    prompt = 'Change Colorscheme',
+    finder = finders.new_table {
+      results = colors
+    },
+    previewer = previewers.man.new(opts),
```
tjdevries, line 787: I don't think this should set a `previewer`.

I have an idea for how to preview the colorscheme -- but it will take a quite bit of time on my side.

glepnir:
@tjdevries  Done. and add marks support

tjdevries:
I will push some examples of how to do the marks in a bit, have some other stuff to get done.

Thanks for the PRs :)

glepnir:
Ok 👌

glepnir:
check `popup.nvim`, I found it has  little code, is it necessary to use it as a plugin?

Issue #40, opened 2020-09-07
Shortcut from the popup to dynamically use other telescope functionality. 


rockerBOO:
So right now if I want to swap to another search:

1. Close out of the popup.
2. Use one of up to 14 keybinds. New functions will need to be bound first before using.

I propose an option in the Popup that allows you to swap to another format.  Sort of like the current `telescope.builtin.builtin{}` but allowing us to go into those finder modes.

`map <Leader>j :lua require'telescope...'.builtin{}` 

### Options 

These options would work either with 1 file, or for a working directory.

- LSP (workspace symbols, document symbols)
- Grep
- Git_Files
- Files
- Treesitter

rockerBOO:
You can now access other telescope functions using the `require'telescope.builtin'.builtin{}` thanks to @s-zeng on #168

PR #168, opened 2020-10-09
Shortcut from the popup to dynamically use other telescope functionality. #40

s-zeng:
My attempt at providing something akin to :Clap providers, a.k.a. a telescope prompt that lets you summon any other available telescope prompts. This would resolve Issue #40

### DIFF
--- a/lua/telescope/actions.lua
+++ b/lua/telescope/actions.lua
@@ -165,6 +165,14 @@ actions.set_command_line = function(prompt_bufnr)
   vim.cmd(entry.value)
 end
 
+actions.run_builtin = function(prompt_bufnr)
+  local entry = actions.get_selected_entry(prompt_bufnr)
+
+  actions.close(prompt_bufnr)
+
+  require('telescope.builtin')[entry.value]()
+end
+
 -- TODO: Think about how to do this.
 actions.insert_value = function(prompt_bufnr)
   local entry = actions.get_selected_entry(prompt_bufnr)
--- a/lua/telescope/builtin.lua
+++ b/lua/telescope/builtin.lua
@@ -495,31 +495,20 @@ builtin.builtin = function(opts)
 
   local objs = {}
 
-  for k, v in pairs(builtin) do
-    local debug_info = debug.getinfo(v)
-
-    table.insert(objs, {
-      filename = string.sub(debug_info.source, 2),
-      lnum = debug_info.linedefined,
-      col = 0,
-      text = k,
-
-      start = debug_info.linedefined,
-      finish = debug_info.lastlinedefined,
-    })
+  for k, _ in pairs(builtin) do
+    table.insert(objs, k)
   end
 
   pickers.new(opts, {
     prompt    = 'Telescope Builtin',
-    finder    = finders.new_table {
-      results     = objs,
-      entry_maker = make_entry.gen_from_quickfix(opts),
-    },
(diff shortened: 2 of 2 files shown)


tjdevries:
So we already have `builtin.builtin` -- I think the main thing would be just changing that existing one to use the action that you made here.

Does that make sense?

s-zeng:
Makes sense to me. I'll make the change

lua/telescope/builtin.lua
```diff
@@ -501,32 +501,21 @@ builtin.builtin = function(opts)
...
-    table.insert(objs, {
-      filename = string.sub(debug_info.source, 2),
-      lnum = debug_info.linedefined,
-      col = 0,
-      text = k,
-
-      start = debug_info.linedefined,
-      finish = debug_info.lastlinedefined,
-    })
+  for k, _ in pairs(builtin) do
```
tjdevries, line 504: Can you actually leave the debut stuff? It makes it so we can show a preview of the code, which I think is pretty cool!

s-zeng: Done. I also threw in an extra change to keep insert mode when selecting a builtin, if you don't mind

lua/telescope/actions.lua
```diff
@@ -165,6 +168,14 @@ actions.set_command_line = function(prompt_bufnr)
   vim.cmd(entry.value)
 end
 
+actions.run_builtin = function(prompt_bufnr)
+  local entry = actions.get_selected_entry(prompt_bufnr)
+
+  actions.close(prompt_bufnr, false)
```
tjdevries, line 174: instead of doing this (which doesn't follow the pattern of actions) you can just do `vim.cmd [[startinsert]]` right here.

tjdevries:
I see what you're doing there, but instead, we can just do the `startinsert` thing I mentioned in the comments.

After that, we can probably merge :) I'll look over one more time. Thanks again!

s-zeng:
So, I tried the startinsert thing, but it didn't actually work as expected - it always drops me in the new prompt in normal mode. Maybe I'm just doing it wrong; here's how I have it right now:

```lua
actions.run_builtin = function(prompt_bufnr)
  local entry = actions.get_selected_entry(prompt_bufnr)

  actions.close(prompt_bufnr)
  vim.cmd([[startinsert]])

  require('telescope.builtin')[entry.value.text]()
end
```

tjdevries:
Hmmm I'm not experiencing the same problem when I'm on this branch -- we can fix that part later if you want?

s-zeng:
That sounds good to me. I'll remove the stopinsert shenanigans

lua/telescope/actions.lua
```diff
@@ -165,6 +165,15 @@ actions.set_command_line = function(prompt_bufnr)
   vim.cmd(entry.value)
 end
 
+actions.run_builtin = function(prompt_bufnr)
+  local entry = actions.get_selected_entry(prompt_bufnr)
+
+  actions.close(prompt_bufnr)
+  vim.cmd [[startinsert]]
+
+  require('telescope.builtin')[entry.value.text]()
```
tjdevries, line 174: I thought this would be entry.text, not entry.value.text

s-zeng: Either one works on my machine. I'll change to to entry.text then

tjdevries:
Thanks!

PR #169, opened 2020-10-09
refactor: prompt->prompt_title ; prompt_prefix->prompt_text

pwntester:
Refactoring `prompt` to `prompt_title` which is consistent with `results_title` and `previewer_title` and `prompt_prefix` to `prompt_text`

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,7 +354,7 @@ Defaults:
 ```lua
 -- lua/telescope/pickers.lua
 Picker:new{
-  prompt             = "", -- REQUIRED
+  prompt_title       = "", -- REQUIRED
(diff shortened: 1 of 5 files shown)


tami5:
perhaps instead of prefix_text, prefix_header makes more sense, and the same goes for result.

pwntester:
> I think I prefer to keep it prompt prefix.
> 
> 
Should we keep the `prompt` to `prompt_title` change though?

tjdevries:
Yeah, prompt to prompt title is good.

pwntester:
ok, should be good to go

pwntester:
@tjdevries Can we merge this now?

Issue #163, opened 2020-10-09
Up/Down highlights multiple entries with LSP document symbols


krkhan:
![out](https://user-images.githubusercontent.com/697967/95544891-44056e00-09b1-11eb-97b8-aec49136cad2.gif)

```
NVIM v0.5.0-728-gf6ac37560
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc-5 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERA
TED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/travis/build/neovim/bot-ci/build/neovim/build/config -I/home/travis/build/neovim/bot-ci/build/neovim/src -I/home/travis/build/neovim/bot-ci/build/neovim/.deps/usr/include -I/
usr/include -I/home/travis/build/neovim/bot-ci/build/neovim/build/src/nvim/auto -I/home/travis/build/neovim/bot-ci/build/neovim/build/include
Compiled by travis@travis-job-7d98421d-d36d-46f4-bb40-7d1aaad0f65b

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/home/travis/build/neovim/bot-ci/build/neovim/build/nvim.AppDir/usr/share/nvim"
```

(On commit fe63e4f1defdc0d5f4d09a7026418c464c2e9f79 of telescope.vim)

krkhan:
Looks like the lsp_document_symbols picker isn't using bat for preview. Is there a way to make it use bat instead of the vim buffer for preview?

tjdevries:
Ah, sorry I didn't see this earlier.

You can pass a previewer in manually. I can probably also fix this to remove the highlights between each iteration (at some point)

tjdevries:
just do `previewer = <create previewer here>` in the options you pass to the builtin.

tami5:
[...]

PR #170, opened 2020-10-10
Add use_external_pager_for_lsp_document_symbols option

krkhan:
Adding as a workaround for #163

### DIFF
--- a/lua/telescope/builtin.lua
+++ b/lua/telescope/builtin.lua
@@ -184,13 +184,18 @@ builtin.lsp_document_symbols = function(opts)
     return
   end
 
+  local previewer = previewers.vim_buffer.new(opts)
+  if conf.use_external_pager_for_lsp_document_symbols then
+    previewer = previewers.qflist.new(opts)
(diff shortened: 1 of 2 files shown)


tjdevries:
You should be able to just pass in an override previewer to the builtin.

Why would we need this setting?

krkhan:
I've tried with both `previewer = previewers.qflist` and `previewer = previewers.bat_maker` in `defaults` of `telescope.setup` and it gives me those raw ANSI escape sequences again. This was the only way I could get it to work, but yes if there's a way of getting it done without a new config value that would be preferable.

clason:
You just need to add back the `-R` flag to https://github.com/nvim-lua/telescope.nvim/blob/6267ead134032b3746488d804b0e09a740d54ab6/lua/telescope/previewers.lua#L33 (which the latest commit already did for line 35, which isn't called for lsp symbols).

krkhan:
gotcha, #172 indeed helps. closing this one

PR #171, opened 2020-10-10
Use results_height and results_width from config

babariviere:
For this snippet of code:

```vim
lua <<EOF
require'telescope'.setup{
  defaults = require('telescope.themes').get_dropdown{}
}
EOF

nnoremap <leader><space> <cmd>lua require'telescope.builtin'.find_files{}<CR>
```

The results_height from `get_dropdown` was not applied to the `find_files` function.

This fix uses `results_height` and `results_width` from `config` as it has the same defaults.


tjdevries:
Awesome, thanks. I may end up changing some of these options later, but this is definitely the right fix for now. Thanks!

PR #172, opened 2020-10-10
Make all other previews not bad, too.

clason:
Adds the `-R` flag to `less` for the other code path as well, which is called for `grep` and `workspace_symbols`.


krkhan:
looks good :+1:

Issue #174, opened 2020-10-12
Wrong path in git file picker


p00f:
![Screenshot from 2020-10-12 17-36-27](https://user-images.githubusercontent.com/36493671/95746505-829e6f80-0cb4-11eb-9db3-71586a1ed9c6.png)

telescope can't find the files because it is looking in the wrong place

sorry for the handwriting

p00f:
I tried both
1) nvim `~/stuff/adwaita-dracula/Adwaita-Dracula/gnome-shell/gnome-shell.css`
2) open nvim, then `:e stuff/adwaita-dracula/Adwaita-Dracula/gnome-shell/gnome-shell.css`

neovim version `NVIM v0.5.0-741-gca6815115`

p00f:
Now updated to `NVIM v0.5.0-751-gd198aa511`, I still have this issue

tjdevries:
Yeah, you may want to use find files in the mean time. That should work for you.

p00f:
ohk

PR #180, opened 2020-10-18
Ensure git_files uses the git root directory as cwd

malramsay64:
This sets the cwd option of the git_files builtin to use the root of the
git directory when the cwd option doesn't already exist. When git lists
files, it is relative to the root of the git directory, rather than the
current working directory. This caused problems when using git_files in
a subdirectory of the git root (see #174).

This commit fixes the issue by always setting the cwd as the root of the
git directory. I feel there is probably a better solution to extracting shell output
than writing the utils.capture function, however that is the solution I could find.

### DIFF
--- a/lua/telescope/builtin.lua
+++ b/lua/telescope/builtin.lua
@@ -46,10 +46,14 @@ local builtin = {}
 builtin.git_files = function(opts)
   opts = opts or {}
 
-  opts.entry_maker = opts.entry_maker or make_entry.gen_from_file(opts)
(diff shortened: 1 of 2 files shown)


tjdevries:
I think we could probably use vim.fn.system or vim.fn.systemlist instead of the new util function.

malramsay64:
That would be exactly what I was looking for, thankyou.

tjdevries:
Awesome, thanks for the fix.

PR #184, opened 2020-10-20
Add a sorter using the fzy algorithm

swarn:
This adds the `fzy` algorithm to the sorters built in to `telescope`.

The [fzy](https://github.com/jhawthorn/fzy) algorithm is a fuzzy matcher with a number of nice characteristics. It's fast, and it tends to do [what you want](https://github.com/jhawthorn/fzy/blob/master/ALGORITHM.md).

Of course you can call out to a command-line fuzzy finder, like `fzy` itself, or `fzf`, or `sk`, etc. But it's nice to reduce external dependencies.

This includes the algorithm itself in `fzy.lua`, a wrapper to make the algorithm match the `Sorter` interface, and many additional tests.

I'm using the following code in my config to use the new sorter:
``` lua
function my_tscope_files()
  local find_command = {"rg", "-l", "--hidden", "--follow", "-g", "!.git", ""}
  local opts = get_my_theme()
  opts.entry_maker = make_entry.gen_from_file(opts)
  opts.finder = finders.new_oneshot_job(find_command)
  opts.previewer = previewers.cat.new(opts)
  opts.sorter = sorters.get_fzy_sorter()
  pickers.new(opts):find()
end
```

It's seemed quick to me. It doesn't slow down searching the neovim source tree.
shot


lua/telescope/algos/fzy.lua
```diff
@@ -0,0 +1,180 @@
...
+fzy.SCORE_GAP_TRAILING = SCORE_GAP_TRAILING
+fzy.SCORE_GAP_INNER = SCORE_GAP_INNER
+fzy.SCORE_MATCH_CONSECUTIVE = SCORE_MATCH_CONSECUTIVE
+fzy.SCORE_MATCH_SLASH = SCORE_MATCH_SLASH
+fzy.SCORE_MATCH_WORD = SCORE_MATCH_WORD
+fzy.SCORE_MATCH_CAPITAL = SCORE_MATCH_CAPITAL
+fzy.SCORE_MATCH_DOT = SCORE_MATCH_DOT
+fzy.SCORE_MAX = SCORE_MAX
+fzy.SCORE_MIN = SCORE_MIN
+fzy.MATCH_MAX_LENGTH = MATCH_MAX_LENGTH
```
tjdevries, line 178: Should probably not expose these if they aren't actually used:

You need to either use `fzy.CONSTANT` in the actual algorithm or don't expose them, because this doesn't allow people to configure them (which is why I'm assuming you're exposing them).

lua/telescope/sorters.lua
```diff
@@ -307,6 +307,49 @@ sorters.fuzzy_with_index_bias = function(opts)
...
+      -- telescope.Sorter "smaller is better" convention. Note that for exact
+      -- matches, fzy returns +inf, which when inverted becomes 0.
+      return 1 / (fzy_score + OFFSET)
+    end,
+
+    -- The fzy.positions function, which returns an array of string indices, is
+    -- compatible with telescope's conventions. It's moderately wasteful to
+    -- call call fzy.score(x,y) followed by fzy.positions(x,y): both call the
+    -- fzy.compute function, which does all the work. But, this doesn't affect
+    -- perceived performance.
```
tjdevries, line 346: if you're interested in making them share the same computation, it is possible to do so. The first parameter to the function is `self` which can be used to store some intermediate state if you wish.

Otherwise we could do this some other time if we needed to.

tjdevries:
Very cool! Thanks for the PR :) just a few small nit picks.

One other thing would be if you could add your name to the top of the fzy.lua file (so that if people have questions / issues / want to say you're awesome, they know who did all the great work) and if you could reindent to use two space indenting, that'd be great (which is what we use elsewhere in the project).

Thanks again!

tjdevries:
With commit f95fe484f744426f0e192d6aa61cd27aedeaf15a you can now set default sorters -- so as soon as we merge this, you can easily configure you own default sorter for files and generic searching.

swarn:
I'm happy to get any comments and advice; `fzy.lua` is the first thing I wrote in lua.

Removing the constants makes perfect sense. A few of them are useful outside of the module; does replacing them with functions that return them make sense? This way, it doesn't false look like you can change them.

I won't bother adding a cache for fzy results unless it looks like it's needed.

The extra space at beginning of the prompt surprised me. Type `a` and the `prompt` parameter at the matcher is ` a`. Is this expected?

I'll be waiting for the default sorter merge; thanks!

lua/telescope/algos/fzy.lua
```diff
@@ -0,0 +1,188 @@
...
+        j = j - 1
+      end
+    end
+  end
+
+  return positions
+end
+
+-- If strings a or b are empty or too long, `fzy.score(a, b) == fzy.SCORE_MIN()`.
+function fzy.SCORE_MIN()
```
tjdevries, line 171: let's call these `get_score_min()` instead (and for the ones below)

tjdevries:
Ah, it's a bug about the prompt thing -- I'll see if I can track that down for you.

tjdevries:
Awesome! I added a few small nitpick fixes that would not have been worth bothering you about. Thanks for the PR :)

swarn:
Nice of you to call it nitpicking; the `string.match` call without the `plain` argument was a flat-out bug. Try the fzy sorter some time, you might dig it :).

PR #188, opened 2020-10-24
feat: add default_icon option

### LINKED ISSUE #177: Option to enable default devicons
Do you think it's a good idea to give the user an option whether or not telescope use the default icon for a filetype that doesn't have an icon? In my case it's `svelte`

nvim-web-devicons has an option to enable the default icon if the icon is not available for that filetype. Personally, I prefer having the default icon rather than having an empty space. Here's a comparison between them, left one is using the default icon and the right one isn't.

Group 38

elianiva:
should resolve #177

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -9,12 +9,16 @@ local make_entry = {}
 
 local transform_devicons
 if has_devicons then
-  transform_devicons = function(filename, display, disable_devicons)
+  transform_devicons = function(filename, display, disable_devicons, default_icon)
     if disable_devicons or not filename then
       return display
     end
 
-    local icon_display = (devicons.get_icon(filename, string.match(filename, '%a+$')) or ' ') .. ' ' .. display
+    local icon_display = table.concat{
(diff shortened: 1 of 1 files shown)


tjdevries:
We should add a way to configure this globally.

Also, do you know if it's possible to do this using nvim devicons directly? Such that we can just pass default = true, and it will do the work for us?

tjdevries:
If we can do it with devicons, we can just add a how to in the Readme and pass default = true where we need to.

elianiva:
It's not currently possible but I've made a [PR](https://github.com/kyazdani42/nvim-web-devicons/pull/16) for devicons that allows you to do that.

tjdevries:
I think it makes sense to enable `default = true` all the time -- since, as you mentioned, it's a bit weird to be missing them.

Does latest commit ( a8e7807 ) work for you? If so, we can close these.

Thanks for the PR :) just ended up implementing this and something else at the same time :laughing:

elianiva:
Epic, it works! Gonna close this and #177

PR #192, opened 2020-10-26
WIP: Change pickers to stop working on previous prompt when prompt changes

tjdevries:
This should greatly increase "snappiness" feeling and also prevent us from doing tons of wasted work on fast keystroke situations.

Also a good opportunity to  clean up some things. We'll see how things go.

### DIFF
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -52,7 +52,13 @@ function config.set_defaults(defaults)
   set("border", {})
   set("borderchars", { '─', '│', '─', '│', '╭', '╮', '╯', '╰'})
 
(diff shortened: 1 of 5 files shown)


tjdevries:
Plan:

Make display & sorting scheduled in chunks on the event loop.
Check if the prompt / iteration is the same as when we started.
If it is it NOT the same, then just return out of the old sorting / displaying,
and then begin with the new one.

This should make telescope FEEL very very responsive (even if we're still working on huge queries and not done yet).

After we have things separated out like this, I hope to be able to (some day in the future when I can figure this out gud) put some of the sorting / processing into different threads. Once we do that, then we can do sorting even faster :) but that's for another day.

This is definitely a step in that direction.

tjdevries:
This doesn't make sense anymore, we have better ideas.

PR #197, opened 2020-10-27
rewrite command of telescope

glepnir:
@tjdevries  now the usage like `Telescope find_files`. `Telescope grep_string`, If want change some options, like this 

`Telescope find_files prompt_prefix=ssss` .

current can't get theme option in viml, so does not support add theme option in command .

### DIFF
--- a/plugin/telescope.vim
+++ b/plugin/telescope.vim
@@ -29,13 +29,28 @@ cnoremap <silent> <Plug>(TelescopeFuzzyCommandSearch) <C-\>e
         \ }"<CR><CR>
         
 " Telescope Commands
+function! s:telescope_complete(...)
+  let telescope_builtin = [
+      \ 'builtin','find_files','live_grep','grep_string','git_files',
+      \ 'oldfiles','quickfix','loclist','command_history','buffers',
+      \ 'lsp_references','lsp_document_symbol','lsp_workspace_symbol',
+      \ 'lsp_code_actions','treesitter','planets','help_tags','man_pages',
+      \ 'colorscheme','marks'
+      \]
+  return telescope_builtin
+endfunction
+
 function! s:load_command(builtin,...) abort
   let opts = {}
 
   " range command args
   for arg in a:000
     let opt = split(arg,'=')
-    let opts[opt[0]] = opt[1]
+    if opt[0] == 'theme'
+      let theme = opt[1]
+    else
+      let opts[opt[0]] = opt[1]
+    endif
   endfor
 
   let telescope = v:lua.require('telescope.builtin')
@@ -43,4 +58,4 @@ function! s:load_command(builtin,...) abort
 endfunction
 
 " Telescope Commands
-command! -nargs=+ Telescope          call s:load_command(<f-args>)
+command! -nargs=+ -complete=customlist,s:telescope_complete Telescope          call s:load_command(<f-args>)
--- a/plugin/telescope.vim
+++ b/plugin/telescope.vim
@@ -30,8 +30,7 @@ cnoremap <silent> <Plug>(TelescopeFuzzyCommandSearch) <C-\>e
         
 " Telescope builtin lists
 function! s:telescope_complete(...)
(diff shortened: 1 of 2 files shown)


tjdevries:
Can we add a completion function for the first item at least? So that it can complete to the correct builtin?

tjdevries:
Also, do we want to leave the commands? I'm not sure if anyone is using them or not currently :/

glepnir:
we need command. because it's useful for some layload feature like dein.  use `on_cmd` keyword  of dein to load telescope.just load telescope when call the command.

tjdevries:
Sorry, I meant, do you want to leave the old commands you made? Or should we delete them.

glepnir:
yep we should delete. they are hardcored and sees ugly.. done.

OkanEsen:
This'll be a breaking change though and "seems" ugly is kind of subjective, no? :)

I think this change should be documented on the README at least or should probably throw an error, that the old commands are deprecated in favor of the new syntax?

glepnir:
Sorry for the impact on you, I didn’t consider more settings when I made them, so I rewrote them to make them more friendly.

If you use some lazyload plugin management like dein just config `on_cmd = 'Telescope'`,  And If you want  call them in commandline just ` Telescope ` then press `tab` .and support custom options of telescope

OkanEsen:
Don't be sorry, it's all good and doesn't affect me, since I use my own mappings likes this: 

``` lua
utils.gmap(
  "n",
  "xxx",
  [[<cmd>lua require'plugin.telescope'.foo_bar()<CR>]],
  {...}
)
```

I was just making a comment for the people this change could affect and that's why I suggested, that this change should probably be documented on the README. :)

glepnir:
The command is only effective for most people who use lazy loading. If load telescope normally, Just need to map `lua require`.

tami5:
This is very useful change, I like it more then the previous way. Thanks @glepnir

glepnir:
@tjdevries cc/

tjdevries:
@glepnir can you add documentation to readme?

Also, for completion, you can use luaeval and vim.tbl_keys to get the keys, rather than hard coding. 

Then I will merge.

glepnir:
@tjdevries  Done.

plugin/telescope.vim
```diff
@@ -28,25 +28,30 @@ cnoremap <silent> <Plug>(TelescopeFuzzyCommandSearch) <C-\>e
...
-command! -nargs=0 -bar TelescopeLspCodeActions lua require'telescope.builtin'.lsp_code_actions{}
-command! -nargs=0 -bar TelescopeTreesitter lua require'telescope.builtin'.treesitter{}
-command! -nargs=0 -bar TelescopePlanets lua require'telescope.builtin'.planets{}
-command! -nargs=0 -bar TelescopeHelpTags lua require'telescope.builtin'.help_tags{}
-command! -nargs=0 -bar TelescopeManPages lua require'telescope.builtin'.man_pages{}
-command! -nargs=0 -bar TelescopeColorscheme lua require'telescope.builtin'.colorscheme{}
-command! -nargs=0 -bar TelescopeMarks lua require'telescope.builtin'.marks{}
+" Telescope builtin lists
+function! s:telescope_complete(...)
+  let l:telescope_builtin = luaeval('require("telescope.builtin")')
```
tjdevries, line 33: In luaeval, you should just return the list of keys. I don't know that this table will always serialize.

Use vim.tbl_keys function.

tjdevries (changes requested):
Small change, then it's good.

glepnir:
Done. what about this?

tjdevries:
Sorry, I meant within the Lua eval call, inside the string.

I'm on mobile so it's too hard to type it at the moment 😂

glepnir:
@tjdevries  haha It's Ok. I know your means. Done.

tjdevries:
Thanks for fixing those things up! As always, thanks for the good ideas 😁

PR #199, opened 2020-10-27
Fix find_command for find_files when using find executable

### LINKED ISSUE #193: find_files gives illegal option -- t: No such file or directory
Remap:
`nnoremap <Leader>p <cmd>lua require'telescope.builtin'.find_files{}<CR>
`

UI shows 
`cat: /Users/kai/find: illegal option -- t: No such file or directory                                
                                                                                   ││[Process exited 1]`

MacOS
NVIM v0.5.0-a22fe09
Build type: Release
LuaJIT 2.0.5
Screenshot 2020-10-26 at 21 46 43

Conni2461:
Fix #193

The MacOs [manpage](https://www.unix.com/man-page/osx/1/find/) says that a path is expected. This is not expected on Linux because `find` assumes `.`, if no starting-point is specified ([Ref](https://man7.org/linux/man-pages/man1/find.1.html)).

I can't test it because i don't own a Mac, so it must be confirmed by the author of the original issue.

### DIFF
--- a/lua/telescope/builtin.lua
+++ b/lua/telescope/builtin.lua
@@ -546,7 +546,7 @@ builtin.find_files = function(opts)
     elseif 1 == vim.fn.executable("rg") then
       find_command = { 'rg', '--files' }
     elseif 1 == vim.fn.executable("find") then
-      find_command = { 'find', '-type', 'f' }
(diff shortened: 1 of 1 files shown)


KaiSpencer:
Hello, thanks for this. I can confirm that this does fix the issue and find now works.

Issue #196, opened 2020-10-27
`bat` previewer doesn't respect BAT_THEME env var


spencergilbert:
https://github.com/nvim-lua/telescope.nvim/blob/master/lua/telescope/previewers.lua#L21

It seems like the that call above doesn't find the BAT_THEME variable you should be able to use to color your `bat` output. I've edited that line locally to include ``--theme`` and it properly colors the preview window then.

Is there something about calling it from within vim that changes the available env vars?

tjdevries:
It's probably that the env var doesn't exist in the subshell we spawn to display `bat`.

We could check for the env var though and use it in our command :) It'd just be a matter of using `os.getenv` on the lua side.

Would you be interested in making the PR?

spencergilbert:
I can definitely take a swing, haven't written too much lua - but was inspired to "lua-fy" my neovim setup over the weekend :)

spencergilbert:
@tjdevries it looks like the linked PR does properly source the theme but part of my problem is that my vim itself isn't coloring properly. So while now the preview window is _more_ accurate than vim :smile:

PR #200, opened 2020-10-28
feat: Get BAT_THEME var from environment

spencergilbert:
This PR checks the environment for a BAT_THEME variable and if present sets the `--theme` flag to the bat previewer command.

Related: #196


tjdevries:
LGTM! Thanks!

PR #204, opened 2020-10-29
Actions can now be summed up and center action

### LINKED ISSUE #182: [Feature Request] auto zz after jump
telescope can easily jump to definition, but that definition can be near the top or bottom of screen. Is there any to configure telescope to move current line to the center of screen after jump?

Conni2461:
Close #182

This PR enables adding up actions.
I have never overloaded operators in lua, so I'm not sure if this is what you had in mind.
I tested this with center.
So my test setup was i had a file open(middle of a big file so zz actually does something) with my cursor at the bottom of the screen. When i now run `live_grep` i was looking for a string which can be found at the top of the screen.
When I searched for a string outside the screen, it was already centered, even though actions.center was not set.

Missing:
- [x] Documentation
- [x] More actions (tab, sp, vsp) (I don't think we need more actions to be tables)
- [x] Remove preconfigured center (basically remove changes made in `lua/telescope/mappings.lua#L19`)
- [x] Make `action_mt` accessible outside of `actions.lua`

### DIFF
--- a/lua/telescope/actions.lua
+++ b/lua/telescope/actions.lua
@@ -12,6 +12,15 @@ local actions = setmetatable({}, {
   end
 })
 
+local action_mt = {
+  __add = function (lhs, rhs)
+    return setmetatable({}, { __call = function(_, ...)
+      lhs(...)
+      rhs(...)
+    end })
+  end
+}
+
 --- Get the current picker object for the prompt
 function actions.get_current_picker(prompt_bufnr)
   return state.get_status(prompt_bufnr).picker
@@ -51,7 +60,6 @@ end
 
 -- TODO: It seems sometimes we get bad styling.
 local function goto_file_selection(prompt_bufnr, command)
-  local picker = actions.get_current_picker(prompt_bufnr)
   local entry = actions.get_selected_entry(prompt_bufnr)
 
   if not entry then
@@ -90,7 +98,6 @@ local function goto_file_selection(prompt_bufnr, command)
       a.nvim_win_set_config(preview_win, {style = ''})
     end
 
-    local original_win_id = picker.original_win_id or 0
     local entry_bufnr = entry.bufnr
 
     actions.close(prompt_bufnr)
@@ -115,6 +122,10 @@ local function goto_file_selection(prompt_bufnr, command)
       end
     end
   end
+end
+
+function actions.center(_)
+  vim.cmd(':normal zz')
 end
 
 function actions.goto_file_selection_edit(prompt_bufnr)
@@ -184,5 +195,15 @@ actions.insert_value = function(prompt_bufnr)
 
   return entry.value
 end
+
+for k, v in pairs(actions) do
+  actions[k] = setmetatable({}, vim.tbl_extend("force", {
+    __call = function(_, ...)
+      return v(...)
+    end }, action_mt)
+  )
+end
+
+actions._action_mt = action_mt
 
 return actions
--- a/lua/telescope/actions.lua
+++ b/lua/telescope/actions.lua
@@ -13,13 +13,35 @@ local actions = setmetatable({}, {
 })
 
 local action_mt = {
-  __add = function (lhs, rhs)
-    return setmetatable({}, { __call = function(_, ...)
-      lhs(...)
-      rhs(...)
-    end })
+  __call = function(t, ...)
+    local values = {}
+    for _, v in ipairs(t) do
+      local result = {v(...)}
+      for _, res in ipairs(result) do
+        table.insert(values, res)
+      end
+    end
+
+    return unpack(values)
+  end,
+
+  __add = function(lhs, rhs)
+    local new_actions = {}
+    for _, v in ipairs(lhs) do
+      table.insert(new_actions, v)
+    end
+
+    for _, v in ipairs(rhs) do
+      table.insert(new_actions, v)
+    end
+
+    return setmetatable(new_actions, getmetatable(lhs))
   end
 }
+
+local transform_action = function(a)
+  return setmetatable({a}, action_mt)
+end
 
 --- Get the current picker object for the prompt
 function actions.get_current_picker(prompt_bufnr)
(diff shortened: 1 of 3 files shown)


lua/telescope/actions.lua
```diff
@@ -12,6 +12,15 @@ local actions = setmetatable({}, {
   end
 })
 
+local action_mt = {
```
Conni2461, line 15: We need to make the accessable for the user so that he can use it when defining his own actions. This is necessary if he wants to add his actions with actions.goto_file_selection_edit, etc.

lua/telescope/actions.lua
```diff
@@ -90,7 +98,6 @@ local function goto_file_selection(prompt_bufnr, command)
       a.nvim_win_set_config(preview_win, {style = ''})
     end
 
-    local original_win_id = picker.original_win_id or 0
```
Conni2461, line 93: Removed because its unused. After removing `original_win_id`, `local picker` is also no longer needed.

lua/telescope/actions.lua
```diff
@@ -12,6 +12,15 @@ local actions = setmetatable({}, {
   end
 })
 
+actions.action_mt = {
+  __add = function (lhs, rhs)
+    return { fn = function(prompt_bufnr)
+      lhs.fn(prompt_bufnr)
+      rhs.fn(prompt_bufnr)
+    end }
```
tjdevries, line 20: You should return a new table, that uses the metamethod `__call`, so you don't have to check for `fn`.

Are you familiar with `__call`?

lua/telescope/actions.lua
```diff
@@ -117,21 +124,35 @@ local function goto_file_selection(prompt_bufnr, command)
...
-end
-
-function actions.goto_file_selection_vsplit(prompt_bufnr)
-  goto_file_selection(prompt_bufnr, "vnew")
-end
-
-function actions.goto_file_selection_tabedit(prompt_bufnr)
-  goto_file_selection(prompt_bufnr, "tabedit")
-end
+actions.center = setmetatable({
```
tjdevries, line 127: instead of doing this, I think at the end of actions, right before we return, we should do something like the following:

```lua
for k, v in pairs(actions) do
  actions[k] = setmetatable({__call = function(_, ...) return k(...) }, action_mt)
end
```

Note, I wrote this code in the github review thing, so it probably doesn't work :laughing:

tjdevries: and then after that we can do `actions._action_mt = action_mt` or something like that.

Does that make sense?

lua/telescope/mappings.lua
```diff
@@ -201,7 +201,11 @@ mappings.execute_keymap = function(prompt_bufnr, keymap_identifier)
     )
   )
 
-  key_func(prompt_bufnr)
+  if type(key_func) == "table" then
+    key_func.fn(prompt_bufnr)
+  else
+    key_func(prompt_bufnr)
+  end
```
tjdevries, line 208: After you do the fix about `__call` I don't think you'll need this anymore :smile:

lua/telescope/actions.lua
Conni2461: Was not familiar with it but now i am. Thanks for the good hint.

lua/telescope/actions.lua
Conni2461: Yeah makes sense and makes things quiet easy

lua/telescope/mappings.lua
Conni2461: Yeah you are right, works like a charm.

lua/telescope/mappings.lua
```diff
@@ -16,7 +16,7 @@ mappings.default_mappings = config.values.default_mappings or {
       ["<Down>"] = actions.move_selection_next,
       ["<Up>"] = actions.move_selection_previous,
 
-      ["<CR>"] = actions.goto_file_selection_edit,
+      ["<CR>"] = actions.goto_file_selection_edit + actions.center,
```
Conni2461, line 19: Will remove this before merging like i said in my original post. Currently it is only there for testing purposes.
Would also wait until @smartding has confirmed that it really works :laughing:. 

If you wanna keep it we should do this for `normal` mapping as well. And maybe for `sp` and `vsp` as well.

lua/telescope/actions.lua
```diff
@@ -185,4 +196,14 @@ actions.insert_value = function(prompt_bufnr)
   return entry.value
 end
 
+for k, v in pairs(actions) do
+  actions[k] = setmetatable({}, vim.tbl_extend("force", {
+    __call = function(_, ...)
+      return v(...)
+    end }, action_mt)
+  )
+end
```
Conni2461, line 205: I had to change your suggestion a little bit. You had: 
```lua
actions[k] = setmetatable({__call = function(_, ...) return k(...) }, action_mt)
```

If I understand metatable correctly these operator overloads must be in the second parameter. Or am I wrong?
Also do i really need a `vim.tbl_extend` here? It did not work without one.

Conni2461: Thanks for the review btw. Helped me a lot :smile:

lua/telescope/mappings.lua
Conni2461: Confirmed working. `actions.center` is removed. Still would like to hear your thoughts about making `center` default or not.

tjdevries:
@Conni2461 The changes I had to make were pretty complicated, so I ended up implementing them in the commit I just pushed onto this branch.

Could you look over that and see if it makes sense? And then if it does, would you be willing to update the documentation?

We can leave the center in, it's pretty cool :)

Conni2461:
Makes sense for me. Seems a lot easier to define custom mapping functions than what i had :laughing:.
Thanks for the changes.
If I see this correctly we don't need to make `action_mt` visible with `actions._action_mt` anymore. I have removed it.
I rebased the pr and updated the documentation.

Issue #118, opened 2020-09-23
Lists with metadata would be less visually confusing with a column/table layout theme.


sunjon:
When a display entry has metadata (TreeSitter object-type, help-tag version, etc), I find it visually messy.
A theme with formatted columns (based on the length of longest entry) would be easier to read:

```sh
entry1 [v1.4]
longer_entry [v2.0]
even_longer_entry [v1.0]
shortest [v3.0]
```

vs
|   |   |
|  --- |  --- |
| entry1                     | [v1.4] |
| longer_entry           | [v2.0] |
| even_longer_entry | [v1.0] |
| shortest                  | [v3.0] |
```

sunjon:
lots of changes ave happened since this was opened and tables layouts are the norm now.

PR #208, opened 2020-10-30
Buffers rework (indicators and sorting)

### LINKED ISSUE #179: Sort buffers by fricensy of use
Hi, thanks for a wonderful project.

Can you add sort for `require'telescope.builtin'.buffers`?
When I use buffers. I would like to see the last files used at the bottom of the list.
981866b5182e84ddfbdd10a1e6d196be
I open Gemfile and then Procfile, but they find at the top of the list.

### LINKED ISSUE #245: builtin.buffers() does not show unnamed buffers.
While I don't wish to see 'special' buffers such as quickfix/location list, i _do_ want to be able to select unnamed buffers.

Conni2461:
Close #179
Close #245

~~This is currently WIP. Basically it is hacked together.~~

This reworks `builtin.buffers`, so that we also display the indicators.
I'm currently parsing the output of `:buffers`. I'm not sure if i can get the indicators on another way.
~~This currently shows all buffer, also unloaded buffers. (So `show_all_buffers = true` does nothing)~~(Should now work as before)

This also adds a function, which could help us display outputs in a more table like format, like suggested in #118.
This does not change the other builtins to be to a table format. (I plan to do this in another pr).

~~I also briefly looked into  #179 and i think we can use `unpack(vim.fn.getbufinfo(tonumber(bufnr))).lastused` to sort the buffers list. (I think this should be configurable. either pick the current solution or based on lastused.). I might do this in this PR aswell. If I can figure out how the whole sorting part works.~~
Previous Buffer is default selection. Should be configurable. Configurable with `sort_lastused = true`
 
Before PR | New
:-------------:|:-------------------------:
currently|new

@sunjon Would be great if you could comment the new layout(display output with the indicators).
~~Currently does not handle readonly buffers.~~
~~If no buffer has a write indicator the column will not be printed.~~

@Stanislav-Lapata and @sunjon when i sort the buffers after lastused, do you want the active buffer be displayed or should we filter out the current one?

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -612,16 +612,38 @@ builtin.fd = builtin.find_files
 builtin.buffers = function(opts)
   opts = opts or {}
 
-  local buffers = filter(function(b)
+  local bufnrs = filter(function(b)
     return
       (opts.show_all_buffers
       or vim.api.nvim_buf_is_loaded(b))
       and 1 == vim.fn.buflisted(b)
-
   end, vim.api.nvim_list_bufs())
 
+  local buffers = {}
+  local default_selection_idx = 1
+  for _, bufnr in ipairs(bufnrs) do
+    local flag = bufnr == vim.fn.bufnr('') and '%' or (bufnr == vim.fn.bufnr('#') and '#' or ' ')
+
+    if opts.sort_lastused and flag == "#" then
+      default_selection_idx = 2
+    end
+
+    local element = {
+      bufnr = bufnr,
+      flag = flag,
+      info = vim.fn.getbufinfo(bufnr)[1],
+    }
+
+    if opts.sort_lastused and (flag == "#" or flag == "%") then
+      local idx = ((buffers[1] ~= nil and buffers[1].flag == "%") and 2 or 1)
+      table.insert(buffers, idx, element)
+    else
+      table.insert(buffers, element)
+    end
+  end
+
   if not opts.bufnr_width then
-    local max_bufnr = math.max(unpack(buffers))
+    local max_bufnr = math.max(unpack(bufnrs))
     opts.bufnr_width = #tostring(max_bufnr)
   end
 
@@ -631,9 +653,9 @@ builtin.buffers = function(opts)
       results = buffers,
       entry_maker = make_entry.gen_from_buffer(opts)
     },
-    -- previewer = previewers.vim_buffer.new(opts),
-    previewer = previewers.vimgrep.new(opts),
+    previewer = previewers.vim_buffer.new(opts),
     sorter = conf.generic_sorter(opts),
+    default_selection_index = default_selection_idx,
   }):find()
 end
 
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -283,18 +283,16 @@ end
 function make_entry.gen_from_buffer(opts)
   opts = opts or {}
 
+  local displayer = entry_display.create {
+    separator = " ",
+    items = {
+      { width = opts.bufnr_width },
+      { width = 4 },
+      { remaining = true },
+    },
+  }
+
   local cwd = vim.fn.expand(opts.cwd or vim.fn.getcwd())
-
-  local get_position = function(entry)
-    local tabpage_wins = vim.api.nvim_tabpage_list_wins(0)
-    for k, v in ipairs(tabpage_wins) do
-      if entry == vim.api.nvim_win_get_buf(v) then
-        return vim.api.nvim_win_get_cursor(v)
-      end
-    end
-
-    return {}
-  end
 
   local make_display = function(entry)
     local display_bufname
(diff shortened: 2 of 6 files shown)


Stanislav-Lapata:
@Conni2461 
I think current buffer should be displayed but it should be disabled
Active line should be previous buffer
buffers@2x

Conni2461:
> @Conni2461
> I think current buffer should be displayed but it should be disabled
> Active line should be previous buffer
> buffers@2x

Thanks that made it clearer for me.
I'm still not sure what you mean with disable current buffer. Do you mean this buffer can't be selected at all? Because i can't really image how the workflow would be if a buffer in the middle of a big list is disabled.

I got the other think working(active line should be previous buffer), but i think @tjdevries would not be happy with this code. It's currently pretty hacky and i think it should only work with `sorting_strategy = 'descending'`.

But could you test it anyway @Stanislav-Lapata?

I would clean up the code tomorrow, @tjdevries.

Stanislav-Lapata:
@Conni2461 
I described fzf.vim logic
Install fzf.vim if you are comfortable. 
```vim
Plug 'junegunn/fzf', { 'dir': '~/.fzf', 'do': './install --bin' }
Plug 'junegunn/fzf.vim'
```
And run `:Buffer`

You shouldn't repeat fzf.vim logic fully. I would just do sorting for a start and show all buffers without disabling current buffer.

Sorry for my English

Conni2461:
@Stanislav-Lapata 
I haven't used `fzf` in a while. Okay i tested `fzf` and made some changes. I think this is more like `fzf`.
Could you try it out by running `:lua require'telescope.builtin'.buffers{ show_all_buffers = true }` and report back.
The selection of the current buffer is still possible and i think implementing a disable could be quiet tricky.

Stanislav-Lapata:
@Conni2461 
Looks nice. Thank you 👍

lua/telescope/pickers.lua
```diff
@@ -452,21 +453,28 @@ function Picker:find()
...
+          error('Unknown selection strategy: ' .. selection_strategy)
         end
-      elseif selection_strategy == 'reset' then
-        self:set_selection(self:get_reset_row())
       else
-        error('Unknown selection strategy: ' .. selection_strategy)
+        local idx = self.default_selection_index - 1
+        local base_row = self:get_reset_row()
+        local row = self.sorting_strategy == 'ascending' and (base_row + idx) or (base_row - idx)
+        self:set_selection(row)
```
Conni2461, line 477: We should only do this when we open the Picker. When we type we should not go here but rather follow the selection_strategy, otherwise `_selection_row` could exceed the maximum displayed results.

lua/telescope/pickers.lua
```diff
@@ -452,19 +454,43 @@ function Picker:find()
...
-          local follow_row = self:get_row(index)
-          self:set_selection(follow_row)
+        -- TODO(conni2461): Currently does not work with default_selection_index
+        -- find_entry(self:get_selection()) will return always 1 because inside find_entry
+        -- we iterate over all entry_states and compare based on display, which is a function
+        -- and points to the same function for all entry_states. Which makes it impossible to
+        -- get another result other than 1.
+        -- I tested it briefly with find_files and it seams to be broken aswell.
+        -- Maybe a comparison based on ordinal rather than display would be better.
+        -- Will take another look at this
```
Conni2461, line 475: @tjdevries I think `selection_strategy == 'follow'` is currently broken. I also tested this on current master with `find_files` and i think we can fix this if we would compare based on ordinal in `manager:find_entry`.

tjdevries:
@Conni2461 I will probably take some of the code you have there for making a display table and do something very similar and then you can rebase this branch off of it at some point.

I will take a look at the other parts, particularly the selection strategy stuff, later. If you want to submit a separate PR for the selection strategy fixes, I will definitely take a look. Thanks!

Conni2461:
The `selection_strategy` fix for `row` and `follow` is in #218.
I will continue to work on this one in the next couple of days. Have to do other stuff first :laughing:.
Thanks for picking up my display table stuff. I took a brief look at it and it looks good.

Conni2461:
@tjdevries Should be good to review. If you find some time.

lua/telescope/builtin/init.lua
```diff
@@ -645,17 +645,40 @@ builtin.fd = builtin.find_files
...
-      (opts.show_all_buffers
-      or vim.api.nvim_buf_is_loaded(b))
-      and 1 == vim.fn.buflisted(b)
+  local buffers_list = vim.split(vim.fn.execute('buffers'), "\n")
+
+  local buffers = {}
+  opts.bufnr_width = 0
+  local default_selection_idx = 1
+  for i = 2, #buffers_list do
+    local bufnr, indicator, bufname, lnum = string.match(
```
tjdevries, line 654: can we use `vim.api.nvim_list_bufs` instead or does it not do what you're looking for?

tjdevries: also with `getbufinfo` or whatever that function is.

lua/telescope/previewers.lua
```diff
@@ -288,52 +288,19 @@ previewers.vim_buffer = defaulter(function(_)
...
-      log.info("Previewing File:", filename)
-
-      local bufnr = vim.fn.bufnr(filename)
-      if bufnr == -1 then
-        -- TODO: Is this the best way to load the buffer?... I'm not sure tbh
-        bufnr = vim.fn.bufadd(filename)
+      if not vim.api.nvim_buf_is_loaded(bufnr) then
         vim.fn.bufload(bufnr)
-
-        vim.cmd([[doautocmd filetypedetect BufRead ]] .. filename)
```
tjdevries, line 315: does calling bufload do this?

lua/telescope/builtin/init.lua
Conni2461: looking into this tomorrow

lua/telescope/previewers.lua
Conni2461: yes i think so. I have highlighting for unloaded buffers and it seems like for loaded buffers i don't have to do anything

lua/telescope/builtin/init.lua
Conni2461: i used `:buffers` because i thought i can't get the indicators in another way. but i "compile" them now by myself. Thanks

lua/telescope/make_entry.lua
```diff
@@ -304,33 +302,35 @@ function make_entry.gen_from_buffer(opts)
...
+      entry.indicator,
+      display_bufname .. ":" .. entry.lnum,
+    }
   end
 
   return function(entry)
-    local bufnr_str = tostring(entry)
-    local bufname = path.normalize(vim.api.nvim_buf_get_name(entry), cwd)
-
+    local bufname = entry.info.name ~= "" and entry.info.name or '[No Name]'
```
Conni2461, line 313: That fixes #245 and is the reason why we no longer can use bat for previewing buffers, but i can't switch to unnamed buffers. Do you know why? I am getting `Vim(edit):E499: Empty file name for '%' or '#', only works with ":p:h"`.
Do i have to use something like `vim.api.nvim_win_set_buf(current_buffer, unnamed_bufnr)`

tjdevries:
What's left for this PR?

Conni2461:
Basically i also added @sunjon wish to view unnamed buffers but i haven't figured out a way to open them.
`:e #nr` will result in `Vim(edit):E499: Empty file name for '%' or '#', only works with ":p:h"`, so is my only option `vim.api.nvim_win_set_buf(winnr, unnamed_bufnr)`? Besides that i think done

Conni2461:
@tjdevries Should be done from my side. You can take a look at this.

PR #210, opened 2020-11-02
WIP: builtin add tagstack function

DevonMorris:
This PR adds a tagstack function. I marked it work in progress because we probably want a `make_entry.gen_from_tagstack` and I want it to push/pop appropriately based on the selection.

Also any other comments you have are appreciated.

### DIFF
--- a/lua/telescope/builtin.lua
+++ b/lua/telescope/builtin.lua
@@ -833,4 +833,34 @@ builtin.marks = function(opts)
   }):find()
 end
 
+builtin.tagstack = function(opts)
+  opts = opts or {}
+  local tagstack = vim.fn.gettagstack()
+  if vim.tbl_isempty(tagstack.items) then
+    return
+  end
+
+  for i, value in pairs(tagstack.items) do
+    value.text = value.tagname
+    value.lnum = value.from[2]
+    value.filename = vim.fn.bufname(value.from[1])
+  end
+
+  -- reverse the list
+  tags = {}
+  for i=#tagstack.items, 1, -1 do
+    tags[#tags+1] = tagstack.items[i]
+  end
+
+  pickers.new(opts, {
+      prompt_title = 'TagStack',
+      finder = finders.new_table {
+        results = tags,
+        entry_maker = make_entry.gen_from_quickfix(opts),
+      },
+      previewer = previewers.qflist.new(opts),
+      sorter = conf.generic_sorter(opts),
+    }):find()
+end
+
 return builtin


lua/telescope/builtin.lua
```diff
@@ -833,4 +833,34 @@ builtin.marks = function(opts)
   }):find()
 end
 
+builtin.tagstack = function(opts)
+  opts = opts or {}
+  local tagstack = vim.fn.gettagstack()
+  if vim.tbl_isempty(tagstack.items) then
+    return
```
tjdevries, line 840: probably should print a message "No tagstack available" or similar.

lua/telescope/builtin.lua
```diff
@@ -833,4 +833,34 @@ builtin.marks = function(opts)
...
+  end
+
+  for i, value in pairs(tagstack.items) do
```
tjdevries, line 843: you should use `ipairs` when iterating over an array-like table.

lua/telescope/builtin.lua
```diff
@@ -833,4 +833,34 @@ builtin.marks = function(opts)
...
+    value.text = value.tagname
+    value.lnum = value.from[2]
+    value.filename = vim.fn.bufname(value.from[1])
+  end
+
+  -- reverse the list
+  tags = {}
+  for i=#tagstack.items, 1, -1 do
+    tags[#tags+1] = tagstack.items[i]
```
tjdevries, line 852: you can use `table.insert(tags, tagstack.items[i])` for simpler lua.

tjdevries:
You had originally mentioned doing some stuff with manipulating the tagstack, are you still interested in doing that?

lua/telescope/builtin.lua
DevonMorris: af3c348ce302fa71f9287aa3a2ed3668290c7274

lua/telescope/builtin.lua
DevonMorris: af3c348ce302fa71f9287aa3a2ed3668290c7274

lua/telescope/builtin.lua
DevonMorris: af3c348ce302fa71f9287aa3a2ed3668290c7274

DevonMorris:
Yeah I'm definitely still interesting in manipulating the tagstack based on the chosen tag. Specifically, I'd like

- The picker to begin at your current position in the tagstack
- The tagstack to push/pop the appropriate number of entries based on your selection

I'm really unsure where to begin when it comes to implementing those features. I've taken a cursory look at `pickers.lua` but couldn't figure out how to do it in the 5 minutes I looked through it.

tjdevries:
So the first part (setting it to be your current position) would probably be kind of difficult (but perhaps possible).

The pushing / popping you could do. You'd need to override the default mapping of <CR> to be some new `action` that has to do with tagstack. You can find out what tagstack entry you have and pop the entries til you arrive at that one or something like that.

Does that make sense at all?

Conni2461:
@DevonMorris do you mean setting the selection at the current position, because i have a change in #208 where you can set a `default_selection_index` which can be passed to the Picker when creating with `new`.

DevonMorris:
@tjdevries yeah definitely makes sense. I'll see if I can get some time this weekend to try it out.

@Conni2461 thanks, I'll check it out

tjdevries:
Implemented in #744

PR #214, opened 2020-11-02
feat: Add exclusive text sort

tjdevries:
Does this work for you @clason https://github.com/nvim-lua/telescope.nvim/issues/212

Do you want it to be a global config option that you can choose by default or is an option for things that use `vimgrep` style stuff good enough?


clason:
How do I use this? I think it makes more sense on a per-picker level (e.g., you might want to use it for lsp workspace symbols but not for references).

tjdevries:
Closing cause of conflicts.

PR #217, opened 2020-11-03
feat: Add optional native compilation step

tjdevries:


tjdevries:
TODO:
- Proper attribution
- Readme updates
- Better native switching
- Might add re-using the matches as well... we'll see. Positions is not called for every match (but we could cache within one run)

tami5:
How can I test this, what additional steps to take? thanks @tjdevries

tjdevries:
I think I will implement extensions this week. Then I will implement fzy-native as an extension (it will be as easy as installing the repo with an installation step most likely to compile). 

Then you should be able to use fzy-native and it will be very fast :)

tami5:
@tjdevries looking forward for posting performance. as I intend to filter through a huge list of files . Thanks

tjdevries:
I actually did a readme update that shows how to do this the other way now. We made a new repo for this

PR #218, opened 2020-11-03
Fix selection_strategies row and follow

Conni2461:
Fixes the "wanky" behavior of 'row'. With that i mean, when i manually select entry 10 and than start typing and entry 10 is no longer available because only 8 elements are displayed, we will not have a selection and because of this
https://github.com/nvim-lua/telescope.nvim/blob/051aefdb8cea71294e6e62e8a4a1e051c48e4edd/lua/telescope/pickers.lua#L685-L688
we cant move the selection. So now it will stick to the last element when this happens.
I don't this we have to wrap `self:get_row(self.manager:num_results())` in `self:_handle_scroll_strategy`.

Fixes `follow` by changing `manager:find_entry`. I think we can even just compare elements, which would do a pointer comparison which is way faster than comparing with `ordinal`. I tested it and it works but let me know if you think it has some side effects i don't know about.

### DIFF
--- a/lua/telescope/entry_manager.lua
+++ b/lua/telescope/entry_manager.lua
@@ -53,7 +53,7 @@ function EntryManager:new(max_results, set_entry, info)
         local existing_entry = v.entry
 
         -- FIXME: This has the problem of assuming that display will not be the same for two different entries.
-        if existing_entry.display == entry.display then
+        if existing_entry == entry then
           return k
         end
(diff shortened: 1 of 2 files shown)


Conni2461:
@tjdevries Did you have time to take a closer look at this?

tjdevries:
This looks right to me! Thanks

PR #219, opened 2020-11-03
Add tags builtin

### LINKED ISSUE #209: :Tags command similar to Fzf
Is it possible to achieve something like the `:Tags` command similar to what FZF does?

It just basically displays what's inside of my `tags` file.

[asciicast](https://asciinema.org/a/FzTjMFeg7pmvdmW6jCJEPtVko)

Conni2461:
Close #209 

~~Written in a coffee break, so WIP.~~
~~Currently has to be generated with `--fields=n` to get the line numbers. So `ctags -R --fields=+n`~~

- [x] ~~Create `ctags` for user if file does not exists~~
- [x] Jump does not work
- [x] Write `make_entry` 
- [x] Create `display` with new table create
- [x] ~~Take another look at how lines numbers get retrieved in fzf, so we don't have to use `--fields=+n`. I decided to just force having `--fields=+n`. We could search for the lines but it is kinda unnecessary.~~
- [x] ~~Only `--fields=n` will not generate the other fields like `type` but using `--fields=+n` includes these fields aswell but does not work this current version~~
- [x] Some lines are not displayed. Happens in a file with tabs. has to be fixed by not only splitting with `\t` and don't split if the line has `\^\t`

default | hide_filename | show_line | hide_previewer
:-------------:|:-------------------------:|:-------------------:|:-----------------:
default|hide_filename|show_line|hide_previewer

Available options:
```lua
only_current_file = true -- default is false, setting this to true will only show tags for the current file
hide_filename = true     -- default is false
show_line = true           -- default is false
```

### DIFF
--- a/lua/telescope/pickers/entry_display.lua
+++ b/lua/telescope/pickers/entry_display.lua
@@ -43,8 +43,8 @@ end
 
 local function truncate(str, len)
   -- TODO: This doesn't handle multi byte chars...
-  if vim.fn.strdisplaywidth(str) > len - 1 then
-    str = str:sub(1, len)
+  if vim.fn.strdisplaywidth(str) > len then
+    str = str:sub(1, len - 1)
     str = str .. "…"
   end
   return str
@@ -68,8 +68,10 @@ entry_display.create = function(configuration)
 
   return function(self, picker)
     local results = {}
-    for k, v in ipairs(self) do
-      table.insert(results, generator[k](v, picker))
+    for i = 1, table.getn(generator) do
+      if self[i] ~= nil then
+        table.insert(results, generator[i](self[i], picker))
+      end
     end
 
     return table.concat(results, configuration.separator or "│")
--- a/lua/telescope/previewers.lua
+++ b/lua/telescope/previewers.lua
@@ -400,6 +400,46 @@ previewers.vimgrep = defaulter(function(opts)
   }
 end, {})
 
+previewers.ctags = defaulter(function(opts)
+  return previewers.new {
+    setup = function()
+      return {}
+    end,
+
+    teardown = function(self)
+      if self.state and self.state.hl_id then
+        pcall(vim.fn.matchdelete, self.state.hl_id, self.state.hl_win)
+        self.state.hl_id = nil
+      end
+    end,
+
+    preview_fn = function(self, entry, status)
+      with_preview_window(status, nil, function()
+        local scode = string.gsub(entry.scode, '[$]$', '')
+        scode = string.gsub(scode, [[\\]], [[\]])
+        scode = string.gsub(scode, [[\/]], [[/]])
+        scode = string.gsub(scode, '[*]', [[\*]])
+
+        local new_bufnr = vim.fn.bufnr(entry.filename, true)
+        vim.fn.bufload(new_bufnr)
+
+        vim.api.nvim_win_set_buf(status.preview_win, new_bufnr)
+        vim.api.nvim_win_set_option(status.preview_win, 'wrap', false)
+        vim.api.nvim_win_set_option(status.preview_win, 'winhl', 'Normal:Normal')
+        vim.api.nvim_win_set_option(status.preview_win, 'signcolumn', 'no')
+        vim.api.nvim_win_set_option(status.preview_win, 'foldlevel', 100)
+
+        pcall(vim.fn.matchdelete, self.state.hl_id, self.state.hl_win)
+        vim.fn.search(scode)
+        vim.cmd "norm zz"
+
+        self.state.hl_win = status.preview_win
+        self.state.hl_id = vim.fn.matchadd('Search', scode)
+      end)
+    end
+  }
+end, {})
+
 previewers.qflist = defaulter(function(opts)
   opts = opts or {}
 
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -923,4 +923,54 @@ builtin.keymaps = function(opts)
   }):find()
 end
 
+builtin.tags = function(opts)
+  opts = opts or {}
+
+  local ctags_file = opts.ctags_file or 'tags'
+
+  if not vim.loop.fs_open(vim.fn.expand(ctags_file), "r", 438) then
+    print('Tags file does not exists. Create one with ctags -R')
+    return
+  end
+
+  local fd = assert(vim.loop.fs_open(vim.fn.expand(ctags_file), "r", 438))
+  local stat = assert(vim.loop.fs_fstat(fd))
+  local data = assert(vim.loop.fs_read(fd, stat.size, 0))
+  assert(vim.loop.fs_close(fd))
+
+  local results = vim.split(data, '\n')
+
+  pickers.new(opts,{
+    prompt = 'Tags',
+    finder = finders.new_table {
+      results = results,
+      entry_maker = make_entry.gen_from_ctags(opts),
+    },
+    previewer = previewers.ctags.new(opts),
+    sorter = conf.generic_sorter(opts),
+    attach_mappings = function(prompt_bufnr)
+      actions._goto_file_selection:enhance {
+        post = function()
+          local selection = actions.get_selected_entry(prompt_bufnr)
+
+          local scode = string.gsub(selection.scode, '[$]$', '')
+          scode = string.gsub(scode, [[\\]], [[\]])
+          scode = string.gsub(scode, [[\/]], [[/]])
+          scode = string.gsub(scode, '[*]', [[\*]])
+
+          vim.cmd('norm! gg')
+          vim.fn.search(scode)
+          vim.cmd('norm! zz')
+        end,
+      }
+      return true
+    end
+  }):find()
+end
+
+builtin.current_buffer_tags = function(opts)
+  opts = opts or {}
+  return builtin.tags(vim.tbl_extend("force", {only_current_file = true, hide_filename = true}, opts))
+end
+
 return builtin
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -559,4 +559,73 @@ function make_entry.gen_from_vimoptions(opts)
   end
 end
 
+function make_entry.gen_from_ctags(opts)
+  opts = opts or {}
+
+  local cwd = vim.fn.expand(opts.cwd or vim.fn.getcwd())
+  local current_file = path.normalize(vim.fn.expand('%'), cwd)
+
+  local display_items = {
+    { width = 30 },
+    { remaining = true },
+  }
+
+  if opts.show_line then
+    table.insert(display_items, 2, { width = 30 })
+  end
+
+  local displayer = entry_display.create {
+    separator = " │ ",
+    items = display_items,
+  }
+
+  local make_display = function(entry)
+    local filename
+    if not opts.hide_filename then
+      if opts.shorten_path then
+        filename = path.shorten(entry.filename)
+      else
+        filename = entry.filename
+      end
+    end
+
+    local scode
+    if opts.show_line then
+      scode = entry.scode
+    end
+
+    return displayer {
+      filename,
+      entry.tag,
+      scode,
+    }
+  end
+
+  return function(line)
+    if line == '' or line:sub(1, 1) == '!' then
+      return nil
+    end
+
+    local tag, file, scode = string.match(line, '([^\t]+)\t([^\t]+)\t/^\t?(.*)/;"\t+.*')
+
+    if opts.only_current_file and file ~= current_file then
+      return nil
+    end
+
+    return {
+      valid = true,
+
+      ordinal = file .. ': ' .. tag,
+      display = make_display,
+      scode = scode,
+      tag = tag,
+
+      filename = file,
+
+      col = 1,
+      lnum = 1,
+    }
+  end
+end
+
 return make_entry


lua/telescope/builtin.lua
```diff
@@ -904,4 +904,55 @@ builtin.marks = function(opts)
   }):find()
 end
 
+builtin.tags = function(opts)
+  -- TODO(conni2461): Handle opts like no filename, etc
+  opts = opts or {}
+
+  -- TODO(conni2461): Do we have a function like this already?! In plenary for example
+  -- If not put in utils or plenary
+  local function file_exists(file)
```
tami5, line 913: There is builtin function 

```
if vim.fn.filereadable(vim.fn.exapnd(filepath)) == 1 then
   print("file exists, do stuff")
end

```

```

Comment (tami5) [reply lua/telescope/builtin.lua:913]: @Conni2461

Comment (Conni2461) [reply lua/telescope/builtin.lua:913]: Thanks that saved me a lot of time :smile:

File: lua/telescope/pickers/entry_display.lua [hunk 2]
```diff
@@ -43,7 +43,7 @@ end
 
 local function truncate(str, len)
   -- TODO: This doesn't handle multi byte chars...
-  if vim.fn.strdisplaywidth(str) > len - 1 then
+  if vim.fn.strdisplaywidth(str) > len then
```
Conni2461, line 46: Let me explain this one quick.
If i have a perfect match, string is 3 and width is 3 it will still be truncated.

lua/telescope/pickers/entry_display.lua
```diff
@@ -68,7 +68,7 @@ entry_display.create = function(configuration)
 
   return function(self, picker)
     local results = {}
-    for k, v in ipairs(self) do
+    for k, v in pairs(self) do
```
Conni2461, line 71: e.g. when `hide_filename` is set i would input a `nil` at index 1 and index 1 is not in the table. So the table looks like this:
```lua
{
  [2] = entry.tag,
  [3] = line
}
```
ipairs will not run if the table does not start with 1.

Conni2461:
@yujinyuz would be great if you could test this one.

lua/telescope/pickers/entry_display.lua
Conni2461: We could also do something like this.
```lua
for i=1, #self do
  if self[i] ~= nil then
    table.insert(results, generator[i](self[i], picker))
  end
end
```
I think this should be better

lua/telescope/pickers/entry_display.lua
```diff
@@ -43,8 +43,8 @@ end
 
 local function truncate(str, len)
   -- TODO: This doesn't handle multi byte chars...
-  if vim.fn.strdisplaywidth(str) > len - 1 then
-    str = str:sub(1, len)
+  if vim.fn.strdisplaywidth(str) > len then
+    str = str:sub(1, len - 1)
```
Conni2461, line 47: We also have to subtract 1 here because if we truncate the string is actually 1 to long, so the separator is shifted 1 to the right

yujinyuz:
@Conni2461 do you mind  guiding me how can I test this in my local?

Conni2461:
@yujinyuz i don't know where you need help exactly so i give you an full instruction. Feel free to ask questions if something does not work.
0. You have telescope.nvim installed, with all requirements mentioned in the readme 
1. You have to fetch my pr with `git fetch origin pull/219/head:tags`
2. Than checkout this branch with `git checkout tags`
3. Now you are good to go checkout the new tags feature with running `lua require'telescope.builtin'.tags{}`
4. Available options are:
```lua
only_current_file = true -- default is false, setting this to true will only show tags for the current file
hide_filename = true -- default is false
hide_line = true -- default is false
shorten_path = true
previewer = false -- default is true
```
these options can be combines freely. I would suggest checking out these ones:
- `lua require'telescope.builtin'.tags{ shorten_path = true, hide_line = true }`
    - This will the the line found in the tags file because the same line will be displayed in the previewer
- `lua require'telescope.builtin'.tags{ shorten_path = true, previewer = false }`
    - This will hide the previewer and show the line found in the tags file, most similar to `fzf`. If `shorten_path = false` this would be identical.
- `lua require'telescope.builtin'.tags{ hide_filename = true, hide_line = true, only_current_file = true }`
    - Will only show the tags for the current file, without showing the filename. Like before you can switch between `hide_line` and `previewer`

Conni2461:
@tjdevries Should be good for a review. Currently I force `ctags -R --fields+=n` but i'm not sure if we should do this.
An alternative would be to grep for the line in previewer and when confirming a selection. I tried this but it wasn't as precise and elegant as just having the line number in the tags file.
I also would appreciate some testers. I tested this with one c and one cpp project but i'm not sure if i parse the tags file 100% correctly.

lua/telescope/pickers/entry_display.lua
tjdevries: Makes sense to me

lua/telescope/builtin.lua
```diff
@@ -904,4 +904,62 @@ builtin.marks = function(opts)
...
+  opts = opts or {}
+
+  if not vim.loop.fs_open('tags', "r", 438) then
+    local answer = vim.fn.input('Tags file does not exists. Do you want to create one? [y/N]')
+    if answer:lower() == 'y' then
+      os.execute('ctags -R --fields=+n')
+    else
+      return
+    end
+  end
```
tjdevries, line 917: Hmmm I'm not sure I like that this is here. I think instead of doing input, let's just print that people should do do something like this.

If people don't have a tags file I don't think they should try and use `builtin.tags`. Does that make sense?

lua/telescope/builtin.lua
```diff
@@ -904,4 +904,62 @@ builtin.marks = function(opts)
...
+          line = scode,
+          lnum = tonumber(lnum)
+        })
+
+        if opts.tag_width < #tag then
+          opts.tag_width = #tag
+        end
+      end
+    end
+  end
```
tjdevries, line 952: I think this should be in the entry maker as well.

lua/telescope/pickers/entry_display.lua
tjdevries: Yeah, that makes sense as well.

lua/telescope/pickers/entry_display.lua
tjdevries: I think this is better. I think `table.getn` or something is the better way for this as well, rather than `#`? I can't remember off-hand.

lua/telescope/builtin.lua
Conni2461: I had this originally and then i took a look at what fzf is doing and they do this. I don't have any hard feelings for both solutions, so up to you.

tjdevries: I like `builtin` to not _do_ anything outside of making / doing pickers. I prefer just printing.

tjdevries:
OH!

The other thing is that if it has a command / search thing, we could use that to find the tag in the file, without the line numbers.

Isn't that what FZF does as well? I think we do it for helptags as well btw

Conni2461:
Yeah i mentioned this in a previous comment but i had problems with tags in Makefiles when it was a multiline tag. I will take another look at was FZF exactly does and ping you if i have a new version up and running

lua/telescope/pickers/entry_display.lua
Conni2461: Wait i didn't even test my other solution, I get 0 as result of the length for both `#` and `table.getn` if the table is not numbered. I will take another look at this tomorrow, maybe i miss something here.

Conni2461:
@tjdevries I don't know if this is good, but with my testing i have not encountered a line where i could not find the lnum.
I don't actually get what the guys in FZF do, because they use perl and a `shellescape` method to modify the tag file and than they can just spit the line with `'\t\zs'` and use `matchstr` to jump to the line and the get the line with `line('.')`. In the end i think they jump back at the original start point.
My viml skills are not good enough for this. :laughing: 

https://github.com/junegunn/fzf.vim/blob/53b3aea0da5e3581e224c958dbc13558cbe5daee/autoload/fzf/vim.vim#L867-L938

tami5:
Will this work with markdown headings @Conni2461

Conni2461:
@tami5 Yeah should work with all tags created by ctags. Just tested markdown headings, works flawless.

tami5:
I'm testing right now, (by the way if you don't know `pull fetch origin pull/219/head:name` lets you fetch and checkout `name` and test the pull request easily).

Opening a markdown file and running `Telescope tags` result in echomsg ="TAgs file does not exists. Create one with ctags -R"

Conni2461:
> Opening a markdown file and running `Telescope tags` result in echomsg ="Tags file does not exists. Create one with ctags -R"

Yeah tjdevries and i talked about this. I had it ask you if you want to create a tag file but we agreed that the user has to do this, so you get a message. You have to create a tag file by yourself now by running `ctags -R` in your root directory

tami5:
Hmmmmmmmm, I understand why tj would want the user to do that, but can we at least figure out some way to create temp (in /tmp/) tags for readme files, as it doesn't relate to a particular code base. I'm not sure but this might require adding a case for markdown files.

tami5:
Oh I see what tags is meant todo, I was thinking more as an outline for the markdown or current file.

Conni2461:
I think this is too much stuff and is not really a use case we should advertise. This can also be achieved by something like this `ctags README.md` or `ctags $(find -name '*.md')`. Should work haven't tested.
This will create a tags file with only tags for md files

tami5:
It would be cool to add an option to search for current buffer or another picker like current_buffer_fuzzy_find but current_buffer_tags. without the filename because it isn't needed for that case. 

If this is out of the scope of this pr, I can work on new pr for current_buffer_tags

Conni2461:
`only_current_file = true` is the option. I though we could maybe add an alias so create a new builtin function name it like `BTags` and call Tags with only that option added. So it is easier for users to use

tami5:
@Conni2461 Nice you covered that case very well, and I agree about creating an alias. How about this 

```lua

builtin.current_buffer_tags = function(opts)
  opts = opts or {}
  return builtin.tags(vim.tbl_extend("force", {only_current_file = true}, opts))
end

```

Conni2461:
https://github.com/nvim-telescope/telescope.nvim/pull/219#issuecomment-723844145 mentions all options that i implemented.

Yeah like that

tami5:
On another note, `:Telescope tags` when hitting enter on a selection, does jump to the file but not the line.

tami5:
Yes I think `hide_filename` should be part of current buffer tag alias

Conni2461:
> On another note, `:Telescope tags` when hitting enter on a selection, does jump to the file but not the line.

Okay i need a bunch of information now.  Let me explain.
The file create by tags looks like this:
```
Tag | file | scode for tag | useless stuff
```
So i actually have no line number and i was originally forcing people to use ctags with this option `--fields=+n` but tj and i agreed that we should not do this.
So my solution to this is i'm greping for the `scode` in the file the moment you select the line and the correct preview has to be displayed. At this time i also set the correct `lnum` in the entry, so the jump works. If `previewer` is disabled that `lnum` is never changed to something else than 1.
So i though i do this before `action.goto_file_selection` on the keymaps <CR>, <C-V>, <C-X>, <C-T> but if people have remappings this will not work. So i need to do a change in the api and add a `inject_action({action.goto_file_selection_edit, action.goto_file_selection_vsplit, etc}, {before: action.tags_get_lnum})` method. Which then works with all. (That part is more a note to myself).

So i need info which tags do not work. Maybe my grep is wrong. Because i have not found an occurrence which does not work.
Can you ran :Telescope tags show_line=true and show me what line actually is. And do you have a `previewer = true`?

tami5:
@Conni2461 Oh I forgot to mention that you can reproduce the but by decreasing the width of the window to a point preview window can't be drawn.

Telescope tags show_line=true did not fix it.

Telescope tags previewer=true broke my vim instance for some reason. (not specfically for tags but also other pickers when i pas previwer=true or false).

Do you think that jump is depended on the preview for some reason

Conni2461:
@tami5 the jump is definitely depended on previewer right now because i wrote an own preview function which will get the correct line on selection. Without previewer this will not be executed. I thought it is better to do the `grep` thing on selection to improve speed when opening tags. I have a new version that does this on confirmation but this does NOT respect user mappings. Which i think is not good and i will fix that.

Note: i will force push, give me 2 min. I also need to rebase

If you find a tag in which jump does not work i need the information of this tag with `show_line = true`. Just if you find one.

lua/telescope/builtin.lua
```diff
@@ -950,4 +950,79 @@ builtin.keymaps = function(opts)
...
+    --     selection.lnum = tonumber(output[1])
+    --   end)
+    --   inject({
+    --     actions.goto_file_selection_edit,
+    --     actions.goto_file_selection_split,
+    --     actions.goto_file_selection_vsplit,
+    --     actions.goto_file_selection_tabedit},
+    --     { before = update_lnum }
+    --   )
+    -- end
```
Conni2461, line 1019: @tjdevries Does this make sense. Rather than using attach_mappings, so we can respect user_mappings?

tami5:
> @tami5 the jump is definitely depended on previewer right now because i wrote an own preview function which will get the correct line on selection. Without previewer this will not be executed. I thought it is better to do the `grep` thing on selection to improve speed when opening tags. I have a new version that does this on confirmation but this does NOT respect user mappings. Which i think is not good and i will fix that.
> 
> 
> 
> Note: i will force push, give me 2 min. I also need to rebase
> 
> 
> 
> If you find a tag in which jump does not work i need the information of this tag with `show_line = true`. Just if you find one.


I'll do further testing again soon. But for the record themes like dropdown list (without preview window) wouldn't work with builtin.tags? 🤔 if so perhaps a different implementation is needed.

Conni2461:
No it works. Fix is up and running. Now we just have to respect user mappings. And i am working on a solution.

lua/telescope/builtin.lua
tjdevries: oh I see, the idea is basically to evolve certain actions into doing something special for this case?

So they map to `X` but we do `f(X)` or something.

tjdevries:
I will try and look at this today. Very interesting stuff.

lua/telescope/builtin.lua
```diff
@@ -950,4 +950,79 @@ builtin.keymaps = function(opts)
   }):find()
 end
 
+builtin.tags = function(opts)
+  opts = opts or {}
+
+  if not vim.loop.fs_open('tags', "r", 438) then
```
tjdevries, line 956: `'tags'` file should be an option for `opts`. `opts.tag_filepath` or similar, so that you could run this for other tags if you wanted.

lua/telescope/builtin.lua
```diff
@@ -950,4 +950,79 @@ builtin.keymaps = function(opts)
...
+    sorter = conf.generic_sorter(opts),
+    attach_mappings = function(prompt_bufnr, map)
+      local update_lnum = actions._transform_action(function(prompt_bufnr)
+        local selection = actions.get_selected_entry(prompt_bufnr)
+        local strip_endline = string.gsub(selection.scode, '[$]$', '')
+        local strip_doubleslash = string.gsub(strip_endline, [[\\]], [[\]])
+        local strip_c_comment = string.gsub(strip_doubleslash, [[\/]], [[/]])
+
+        local cmd = 'grep -Frn \'' .. strip_c_comment .. '\' ' .. selection.filename
+        local output = vim.split(utils.get_os_command_output(cmd), ':')
```
tjdevries, line 984: what's an example `scode` that you have in your ctags?

I thought `scodes` could almost just be `vim.fn.execute(scode)` or similar. I don't remember thougyh.

lua/telescope/previewers.lua
```diff
@@ -374,6 +374,33 @@ previewers.vimgrep = defaulter(function(opts)
   }
 end, {})
 
+previewers.ctags = defaulter(function(opts)
```
tjdevries, line 377: have you seen `previewers.help`? It might give you some ideas based on what `scode` looks like.

lua/telescope/builtin.lua
Conni2461: Works with `vim.fn.search` but not for all `scode` i have. E.g. an example where it does not work is when i have an c comment at the end of the line. Scode would be:
```
ClkClientWin, ClkRootWin, ClkLast }; \/* clicks *\/$
```

I thought i could just unescape the `\/` but that wasn't enough and escaping the * as well. I had to also remove the $ sign at the end. I think it is good now.

Thanks for the insight.

lua/telescope/previewers.lua
Conni2461: Thanks for the insight :)

lua/telescope/builtin.lua
Conni2461: @tjdevries So basically they have changed `split` to `[C-w]` or something and that will run `actions.goto_file_selection_split`. But if i have a new action now that has to run before, after or completely replace that one, i currently use `attach_mappings` and map this `[C-x]` because this is default. But if the user uses now his mapping it will not run my new action.

I think i need an example. For tags i have to open up the file with `actions.goto_file_selection_` and than i have to do the search. So i can do `actions.goto_file_selection_ + jump_to_tag`. And now i don't want this to happen only on that attached mapping but on any keymapping that runs `actions.goto_file_selection_`. Does this make sense?

Btw i ran into a problem when doing `actions.goto_file_selection_ + jump_to_line`. When my function would run i no longer can get the selection with `actions.get_selected_entry(prompt_bufnr)` because goto_file_selection already closes the picker. Maybe we should offer a lightway version of `goto_file_selection_` which does not close the picker.

Conni2461:
@tjdevries Thanks for the review. I updated the branch.
Hope you don't miss this: https://github.com/nvim-telescope/telescope.nvim/pull/219#discussion_r523770527

tjdevries:
I have big idea for actions. Give me a day or two to implement. I will ping. This is a great idea you're talking about.

tjdevries:
Implementation is so much cleaner now!

Also, implementation of action is very good. Thanks!

yujinyuz:
Why is it that it's not able to find my `tags` file?

```
❯ tree
.
├── bin
├── cron
│  ├── cleanbackup.crontab
│  └── README.md
├── emacs
├── fish
├── git
├── karabiner
├── README.md
├── ruby
├── syncdots
├── tags
├── tmux
└── vim
```

Running `:Telescope tags` I get

```
Tags file does not exists. Create one with ctags -R
```

tami5:
@yujinyuz sadly it needs to be generated via ctags -R it would be cool to havr auto generated and added to special place where they get manages by telephone "cache" . What do u think @Conni2461

yujinyuz:
@tami5 Im using universal ctags and a tags file already exists in my project root. What am I missing here?

tami5:
@yujinyuz ohh just make sure vim cwd matches that root directory

yujinyuz:
@tami5 If you have time, could you take a look at this:

[asciicast](https://asciinema.org/a/ZPfofX0ofwtdw8xCt7QR2rLTm)

My nvim version is

```
❯ nvim --version
NVIM v0.5.0-185732a1f
Build type: Release
LuaJIT 2.0.5
```

Also I'm running the latest telescope.nvim

tami5:
> @tami5 If you have time, could you take a look at this:
> 
> 
> 
> [asciicast](https://asciinema.org/a/ZPfofX0ofwtdw8xCt7QR2rLTm)
> 
> 
> 
> My nvim version is
> 
> 
> 
> ```
> 
> ❯ nvim --version
> 
> NVIM v0.5.0-185732a1f
> 
> Build type: Release
> 
> LuaJIT 2.0.5
> 
> ```
> 
> 
> 
> Also I'm running the latest telescope.nvim

This might a potential bug🧐, please open a bug report and I will try to reproduce and fix it after my morning coffee 😬

yujinyuz:
@tami5 I can't reproduce it with minimal settings. Might be because of my config. I'll try to debug this.

Anyway, thanks!

tami5:
@yujinyuz oh I hate when that happens, check you telescope settings then test if any plugin is intervening in some way

Conni2461:
@yujinyuz to be honest i'm really confused right now :laughing: 
We load the file a little different than fzf. We use `vim.loop.fs_open`.
Can you just run `lua print(vim.loop.fs_open(ctags, 'r', 438))` it should give you any value. If you get `nil` than something is wrong here.
Could you also run `lua print(vim.fn.filereadable('tags'))`.

@tami5 We don't wanna cache tags because ctags basically are a pregenerated cache. Does that make sense. All we do is read the file. But we could load that file async with a plenary job.

yujinyuz:
I just ran `:lua print(vim.loop.fs_open('tags', 'r', 438))` and I got 32 and I ran it again it became 36 and keeps increasing if I keep running it.

`:lua print(vim.fn.filereadable('tags'))` outputs `1`

yujinyuz:
But this might be one of the plugins I'm using that causes the conflict

Conni2461:
Okay so i don't get it. It is fine that it keeps increasing because we get a filehandle as result. 

But i don't get why another plugin might cause a conflict here. All i do is check if that handle isn't nil and if it is i will return that message :confused:

tami5:
> @tami5 We don't wanna cache tags because ctags basically are a pregenerated cache. Does that make sense. All we do is read the file. But we could load that file async with a plenary job.


I see what you mean @Conni2461, here's my thoughts

while some users and some use cases will rather have tags files pregenerated and maintained. There are some cases where you have a code base that doesn't have tags generated and you want to tun Telescope tags and get back results without having to generate it and then reeeerun the command. 

I suggest having vim.loop.fs_stat that check if tag file is valid in vim.loop.cwd and if not, ensure that there is one in .cache/...../tags-dirname and use it. 

What do you think? for the async part, :D i'm still in stages of not knowing how it works but see huge amount of value in it

yujinyuz:
Btw, I just tried commenting out all my plugins and it seems that it's not the case

Conni2461:
But you could also not reproduce it with a sami5 minimal rc? Could you remove the `fs_` check? 
https://github.com/nvim-telescope/telescope.nvim/blob/5513f85b59caefbd8a9cac9166f517489d748ba6/lua/telescope/builtin/init.lua#L942-L945. Just comment it out for a second. Maybe we have to check differently here. I'm will look into this in a sec, as well as what tami5 said. need to update that other pr first.

yujinyuz:
@Conni2461 

I finally found the cause of the error. I had this line in my .vimrc
```vim
 set wildignore+=tags,.*.un~,*.pyc
```

Idk why it can't find because I added `tags` to `wildignore`

Conni2461:
Okay. Hmm i think that should be fixed by me. Will provide a fix asap. 
How do you our tags builtin besides that :laughing: ? You never reported back

yujinyuz:
Oh, yeah I forgot to give feedback as I've only tested it today.

My only concern is that it doesn't exactly match what I want? For example in FZF, it matches correctly if I type `Badge`

Screen Shot 2020-11-22 at 5 46 33 PM


But when using telescope I get `xp` as the first match

Screen Shot 2020-11-22 at 5 47 05 PM

Conni2461:
I will look at that in a sec. Our default sorter is not as good as fzf. But you can change that to fzy
```
require('telescope').setup {
  defaults = {
    generic_sorter = require'telescope.sorters'.get_fzy_sorter,
    file_sorter = require'telescope.sorters'.get_fzy_sorter,
  }
}
```
Our idea is that we have a modular design and you basically can change everything.

Conni2461:
```help
A list of file patterns.  A file that matches with one of these
patterns is ignored when expanding |wildcards|, completing file or
directory names, and influences the result of |expand()|, |glob()| and
|globpath()| unless a flag is passed to disable this.
```
It is f****** `expand` which will expand `tags` into nothing and i can't check for a file `''`. :laughing: That is a super funny bug. Thanks for investigating

yujinyuz:
@Conni2461 welcome!

Btw, I just added this to my config

```
require('telescope').setup {
  defaults = {
    generic_sorter = require'telescope.sorters'.get_fzy_sorter,
    file_sorter = require'telescope.sorters'.get_fzy_sorter,
  }
}
```

But I'm still getting the same output as the one above. It seems that it matches the filename first rather than the content of the `tags` file

Conni2461:
I might need to change the `ordinal`. Will do it in @273. We should move the discussion over there

yujinyuz:
No problem. Thanks @Conni2461

Conni2461:
@tami5 I think that is too much we have to handle. We then basically become a ctags plugin. I think tj just wants builtin suff simple and clean. I am also not that familiar with plenary jobs. I am working with them since yesterday. I will try to implement that now. You should join the discussion over at #273.

PR #224, opened 2020-11-05
Cat will no longer display binary files

### LINKED ISSUE #223: telescope tries to show preview for binary files
Would be helpful if telescope knew not to try to preview arbitrary binary files as it leaves a bunch of garbage characters on the screen.   I've seen this functionality implemented by checking XDG MIME type, but unclear how general that solution can be on all platforms.

Conni2461:
Fix #223 
Will not affect systems without `file`

### DIFF
--- a/lua/telescope/previewers.lua
+++ b/lua/telescope/previewers.lua
@@ -51,7 +51,16 @@ local bat_maker = function(filename, lnum, start, finish)
 end
 
 -- TODO: Add other options for cat to do this better
-local cat_maker = function(filename, lnum, start, finish)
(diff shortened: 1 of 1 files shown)


tjdevries:
Oh... I just merged this but remembered that executable returns 0 or 1 I think.

We should chdck its return code.

Conni2461:
Oh you are absolutely right. That's my bad, i always forget you have to do this in lua. Are you going to fix that with a commit or should i put up a new pr?

tjdevries:
A new PR would be helpful. Thanks!

PR #225, opened 2020-11-06
Fixes previews for files beginning with ~/ and Previewer will now show directories

### LINKED ISSUE #226: Buffers preview: file path is not expanded correctly
Hi, I think there's an issue with how the file path is passed to `bat` for previewing.

This happens when I have a session with buffers open outside the current working directory, so the builtin `buffers{}` lists them as paths relative to my home dir (e.g. `~/Documents/whatever.txt`). This results in a `bat` error, as bat doesn't expand quoted paths (`bat ~/Documents/whatever.txt` works, but `bat '~/Documents/whatever.txt'` or `bat "~/Documents/whatever.txt"` don't). The error message I see in the preview window is exactly: 

```
[bat error]: '~/Documents/whatever.txt': No such file or directory (os error 2)
```

Some considerations:
- `bat` version is `0.16.0`
- `nvim` version is `NVIM v0.5.0-d17e50879 - Build type: Release`
- I'm on macOS `10.15.7` and my `$SHELL` is zsh, but it's not a shell issue, bash behaves exactly the same way
- I frequently update `telescope.nvim` to latest version
- well, obviously, the file exists ;)

Relevant config:

```
" Telescope {{{2
lua <<EOF
-- bat is not expanding '~', so let's try this.
-- UPDATE: no, it's not working
require('telescope').setup {
  defaults = {
    shorten_path = false -- currently the default value is true
  }
}
EOF

nnoremap <Leader>p <cmd>lua require'telescope.builtin'.find_files{}<CR>
nnoremap <Leader>/ <cmd>lua require'telescope.builtin'.live_grep{}<CR>
nnoremap <Leader>K <cmd>lua require'telescope.builtin'.grep_string{ search = cword }<CR>
nnoremap <Leader>B <cmd>lua require'telescope.builtin'.buffers{}<CR>
" }}}
```

Let me know if I can provide any further info.

Thanks a lot!

Conni2461:
Fix #226 

Edit: I will hijack this pr for #226 .
It also fixes the if condition. Its so easier than opening a new pr :laughing: 
It will also show directories. If i recall correctly find_files will only return files but when a directory buffer is opened buffers will show this buffer and preview the directory with `ls -la`

We could change find_files back to also show directories but i don't think its usefull. I don't know. Kinda need input on the whole directory suff @tjdevries

### DIFF
--- a/lua/telescope/previewers.lua
+++ b/lua/telescope/previewers.lua
@@ -21,7 +21,15 @@ Previewer.__index = Previewer
 local bat_options = {"--style=plain", "--color=always", "--paging=always"}
 local has_less = (vim.fn.executable('less') == 1) and config.values.use_less
 
+local get_file_stat = function(filename)
+  return assert(vim.loop.fs_stat(vim.fn.expand(filename)))
+end
+
 local bat_maker = function(filename, lnum, start, finish)
+  if get_file_stat(filename).type == 'directory' then
+    return { 'ls', '-la' }
+  end
+
   local command = {"bat"}
   local theme = os.getenv("BAT_THEME")
 
@@ -46,23 +54,26 @@ local bat_maker = function(filename, lnum, start, finish)
   end
 
   return flatten {
-    command, bat_options, "--", filename
+    command, bat_options, "--", vim.fn.expand(filename)
   }
 end
 
 -- TODO: Add other options for cat to do this better
 local cat_maker = function(filename, _, _, _)
-  if vim.fn.executable('file') then
-    local handle = io.popen('file --mime-type -b ' .. filename)
-    local mime_type = vim.split(handle:read('*a'), '/')[1]
-    handle:close()
+  if get_file_stat(filename).type == 'directory' then
+    return { 'ls', '-la' }
+  end
+
+  if 1 == vim.fn.executable('file') then
+    local output = utils.get_os_command_output('file --mime-type -b ' .. filename)
+    local mime_type = vim.split(output, '/')[1]
     if mime_type ~= "text" then
       return { "echo", "Binary file found. These files cannot be displayed!" }
     end
   end
 
   return {
-    "cat", "--", filename
+    "cat", "--", vim.fn.expand(filename)
   }
 end
 
--- a/lua/telescope/builtin.lua
+++ b/lua/telescope/builtin.lua
@@ -821,9 +821,7 @@ builtin.man_pages = function(opts)
 
   local cmd = opts.man_cmd or "apropos --sections=1 ''"
 
-  local f = assert(io.popen(cmd, 'r'))
-  local pages = assert(f:read('*a'))
-  f:close()
+  local pages = utils.get_os_command_output(cmd)
 
   local lines = {}
   for s in pages:gmatch("[^\r\n]+") do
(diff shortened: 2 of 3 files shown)


lua/telescope/previewers.lua
```diff
@@ -21,7 +21,24 @@ Previewer.__index = Previewer
 local bat_options = {"--style=plain", "--color=always", "--paging=always"}
 local has_less = (vim.fn.executable('less') == 1) and config.values.use_less
 
+local function_preview_director = function(filename)
+  if 1 == vim.fn.executable('file') then
+    local handle = io.popen('file --mime-type -b ' .. filename)
+    local mime_type = vim.fn.split(handle:read('*a'), '\n')[1]
```
tjdevries, line 27: Can just vim.split

lua/telescope/previewers.lua
```diff
@@ -21,7 +21,24 @@ Previewer.__index = Previewer
...
+    handle:close()
```
tjdevries, line 28: Let's make a util function for executing command and not having to remember to close handles.

tjdevries:
Can you try and do the file operations using vim.loop? There are a ton of `fs_` functions we can use there. I need to explore them more myself 😂

lua/telescope/previewers.lua
```diff
@@ -21,7 +21,15 @@ Previewer.__index = Previewer
 local bat_options = {"--style=plain", "--color=always", "--paging=always"}
 local has_less = (vim.fn.executable('less') == 1) and config.values.use_less
 
+local get_file_stat = function(filename)
+  return assert(vim.loop.fs_stat(vim.fn.expand(filename)))
+end
```
Conni2461, line 26: We could remove this but i think it is better readable than doing twice
```lua
if assert(vim.loop.fs_stat(vim.fn.expand(filename))).type == 'directory' then
  return { 'ls', '-la' }
end
```

lua/telescope/previewers.lua
```diff
@@ -46,23 +54,26 @@ local bat_maker = function(filename, lnum, start, finish)
...
-    local mime_type = vim.split(handle:read('*a'), '/')[1]
-    handle:close()
+  if get_file_stat(filename).type == 'directory' then
+    return { 'ls', '-la' }
+  end
+
+  if 1 == vim.fn.executable('file') then
+    local output = utils.get_os_command_output('file --mime-type -b ' .. filename)
+    local mime_type = vim.split(output, '/')[1]
     if mime_type ~= "text" then
```
Conni2461, line 70: I don't see a way to get the `mime_type` with `vim.loop.fs_`. At least `fs_stat` will return me file for binaries and text files, and the other functions don't look promising. And i think running `file` isn't that bad. Because this code will only run if a new entry gets selected.

lua/telescope/previewers.lua
Conni2461: Done. Also used this new function for a `popen` call in `builtin.man_pages`

Conni2461:
Thanks for the great tip with those `fs_` functions. Some of them look pretty interesting. And this cleans up my code a lot so thanks

tjdevries:
I am working on this: https://github.com/nvim-lua/plenary.nvim/pull/22 Give me a bit of time and we can try and do all of the path stuff with this instead.

Should be much easier to manage. Probably can then remove all path related utilties from telescope and just usse `Path:new(...)` on paths.

tjdevries:
OK, later we can build more things in plenary path and remove some of these things (and they will all be implemented with `uv.fs_*` functions so they can be called in API_FAST events and will just be better all around)

PR #227, opened 2020-11-06
Add vimspector picker

sharksforarms:
This allows the user to select a [vimspector](https://github.com/puremourning/vimspector) configuration and launch with that configuration

### DIFF
--- a/lua/telescope/builtin.lua
+++ b/lua/telescope/builtin.lua
@@ -904,4 +904,39 @@ builtin.marks = function(opts)
   }):find()
 end
 
+builtin.vimspector = function(opts)
+  opts = opts or {}
+
+  if vim.g.loaded_vimpector == nil then
+    error("Vimspector is not loaded")
+  end
+
+  local configurations = vim.api.nvim_call_function("vimspector#GetConfigurations", {})
+
+  if vim.tbl_isempty(configurations) then
+    return
+  end
+
+  pickers.new(opts, {
+    prompt_title = 'Vimspector Launch',
+    finder    = finders.new_table {
+      results = configurations
+    },
+    sorter = sorters.get_generic_fuzzy_sorter(),
+    attach_mappings = function(prompt_bufnr, map)
+      local start_debugger = function()
+        local selection = actions.get_selected_entry(prompt_bufnr)
+        actions.close(prompt_bufnr)
+
+        vim.cmd("call vimspector#LaunchWithSettings(#{ configuration: '" .. selection.value .. "'})")
+      end
+
+      map('i', '<CR>', start_debugger)
+      map('n', '<CR>', start_debugger)
+
+      return true
+    end
+  }):find()
+end
+
 return builtin


bew:
I'm sorry, but to me it does not make sense to add pickers for plugins that may not be installed...
Wouldn't it be better to add this to your own config, or if you want to share it, to make a separate plugin that anyone can use to configure their telescope?

sharksforarms:
> I'm sorry, but to me it does not make sense to add pickers for plugins that may not be installed...
> Wouldn't it be better to add this to your own config, or if you want to share it, to make a separate plugin that anyone can use to configure their telescope?

Makes sense to me! :+1:

sharksforarms:
For anyone searching, here's a function:

```lua
function picker_vimspector(opts)
    local finders = require('telescope.finders')
    local pickers = require('telescope.pickers')
    local sorters = require('telescope.sorters')
    opts = opts or {}

    local configurations = vim.api.nvim_call_function("vimspector#GetConfigurations", {})

    if vim.tbl_isempty(configurations) then
        return
    end

    pickers.new(opts, {
        prompt_title = 'Vimspector Launch',
        finder = finders.new_table {
            results = configurations
        },
        sorter = sorters.get_generic_fuzzy_sorter(),
            local start_debugger = function()
                local selection = actions.get_selected_entry(prompt_bufnr)
                actions.close(prompt_bufnr)

                vim.cmd("call vimspector#LaunchWithSettings(#{ configuration: '" .. selection.value .. "'})")
            end

            map('i', '<CR>', start_debugger)
            map('n', '<CR>', start_debugger)

            return true
        end
    }):find()
end
```

tami5:
Why did you close it? I think it is a great addition to the library unless @tjdevries think otherwise. 

Personally I was thinking of using vimspector so thanks so much for that 😍🙏

sharksforarms:
Added a check to see if the plugin is loaded, this should address your concerns @bew ?

bew:
> Edit: But there are support for other plugins, such as treesitter, so maybe?

treesitter is not a plugin, it is builtin to neovim 0.5.0 (not released yet)

tami5:
Damn @bew you have a good point here. But still thanks to the design and how modular telescope is, each function or builtin add-on is its own world. So maintaining them or perhaps fixing bugs "in them" wouldn't be more overhead on any user or rather any contributor who is also using the function daily. 

Of course not every possible function or picker should be added to builtin, but still stuff that support development and developer workflow and that makes it easier for other users to adopt telescope in their workflow is always welcomed.

> I'm not author / not a maintainer of this project

So what? Nether am I, but that should't holds us from expressing what might be of top interest  of something that we like a lot, plan to use long-term and contribute to when we can. Your thoughts matter a lot and thank you for shearing them. Telescope is a community driven project. (or as :smile: @tjdevries puts it sarcastically: a community driven project with me getting all the credit :laughing:)  

> Take my opinion with a grain of salt though,

I do and hope you do to :D, thanks again


Either ways

eth3lbert:
Maybe its better not to add into code base, but into the wiki?

rockerBOO:
I think the idea of having broad support for builtins is not the true goal of the project. It is to have a system of systems that work together and builtin is an example of those. In terms of builtin I think they should be of broad interest to most users.

In terms of Treesitter, that I think is with the idea that treesitter will eventually be included without an external plugin. Maybe I am wrong here. Having it separate now allows more rapid development outside of sticking to master. 

I think in this example this plugin is maybe used by a minority of users, requires a separate plugin. This would make it a good example of using telescope as a base for another plugin. 

I think we should list all extending plugins on the README to help with discoverability.

tjdevries:
I was thinking it would be cool to include these kinds of things in the repo.

However, if it requires some other plugin, we should put it somewhere besides builtin. For example, we can make `telescope.extensions` or similar.

Then each of these plugins can be developed there (if it makes sense to include them -- which at the moment just means "does TJ think this is cool"). 

Other plugins can of course also add things to `telescope.extensions` as a way of saying "Hey, I'm a telescope plugin".

I am happy to make "Code owners" or similar items for `telescope.extensions` so that people can contribute to those easily and not have to wait on me constantly.

tjdevries:
The other idea I've been having is maybe we should move `telescope.nvim` to its own org... people are using it a lot more than I thought they would :laughing: 

This way, we can have "official" plugins. `github.com/telescope.nvim/vimspector` or similar then could include these and help people work together on them.

tjdevries:
OK, https://github.com/nvim-telescope/telescope-vimspector.nvim here's where you can make the PR :)

I'm adding you to that repo @sharksforarms

PR #230, opened 2020-11-10
Rework helptags builtin

parmort:
This grabs the tags from the `&runtimepath`, which means that plugin helptags will be searchable. We do lose the ability to grab the version of the tags; if that functionality is necessary, we can create a `runtime_help_tags` builtin or something.

This will probably end up being slower, although I did not see a noticeable slow-down, so this is probably equivalent to file searching or similar operations.


tami5:
Works really well, great work @parmort , finally I can access tags from plugins

PR #237, opened 2020-11-13
Allow finder to find from multiple dirs

### LINKED ISSUE #238: Feature: Allow finders to get files from multiple directories
Basically what the title says. I find this useful to have a picker to select among my dotfiles, which are distributed among multiple directories

anihm136:
Adds functionality for finders to get files from multiple directories. Currently tuned only for the three supported finders, but shouldn't be too hard to add more, as long as the command itself supports searching from multiple directories. This PR adds a search_dirs option, which is injected into the command args to search from multiple directories
Fixes #238

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -53,40 +53,50 @@ files.grep_string = function(opts)
   }):find()
 end
 
--- NOTE(tj): Get clarification on comment and put this in the right place
-local function insert_search_dirs(find_command, search_dirs)
-  local command = find_command[1]
-  local insert_pos = -1
-  if command == 'fd' or command == 'fdfind' then
-    table.insert(find_command, '.')
-  elseif command == 'find' then
-    for idx,val in ipairs(find_command) do
-      if val=='.' then
-        insert_pos = idx
-        break
-      end
-    end
-    table.remove(find_command, insert_pos)
-  end
-  for _,v in pairs(search_dirs) do
-    table.insert(find_command, insert_pos==-1 and (#find_command+1) or insert_pos, vim.fn.expand(v))
-  end
-end
-
 -- TODO: Maybe just change this to `find`.
 --          Support `find` and maybe let people do other stuff with it as well.
 files.find_files = function(opts)
   local find_command = opts.find_command
+  local search_dirs = opts.search_dirs
+
+  if search_dirs then
+    for k,v in pairs(search_dirs) do
+      search_dirs[k] = vim.fn.expand(v)
+    end
+  end
 
   if not find_command then
     if 1 == vim.fn.executable("fd") then
       find_command = { 'fd', '--type', 'f' }
+      if search_dirs then
+        table.insert(find_command, '.')
+        for _,v in pairs(search_dirs) do
+          table.insert(find_command, v)
+        end
+      end
     elseif 1 == vim.fn.executable("fdfind") then
       find_command = { 'fdfind', '--type', 'f' }
+      if search_dirs then
+        table.insert(find_command, '.')
+        for _,v in pairs(search_dirs) do
+          table.insert(find_command, v)
+        end
+      end
     elseif 1 == vim.fn.executable("rg") then
       find_command = { 'rg', '--files' }
+      if search_dirs then
+        for _,v in pairs(search_dirs) do
+          table.insert(find_command, v)
+        end
+      end
     elseif 1 == vim.fn.executable("find") then
       find_command = { 'find', '.', '-type', 'f' }
+      if search_dirs then
+        table.remove(find_command, 2)
+        for _,v in pairs(search_dirs) do
+          table.insert(find_command, 2, v)
+        end
+      end
     end
   end
 
@@ -96,15 +106,9 @@ files.find_files = function(opts)
     return
   end
 
-  local search_dirs = opts.search_dirs
-  if search_dirs then
-    insert_search_dirs(find_command, search_dirs)
-  end
-
   if opts.cwd then
     opts.cwd = vim.fn.expand(opts.cwd)
   end
-  P(find_command)
 
   opts.entry_maker = opts.entry_maker or make_entry.gen_from_file(opts)
 
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -53,6 +53,26 @@ files.grep_string = function(opts)
   }):find()
 end
 
+-- NOTE(tj): Get clarification on comment and put this in the right place
+local function insert_search_dirs(find_command, search_dirs)
+  local command = find_command[1]
+  if command == 'fd' or command == 'fdfind' then
+    table.insert(find_command, '.')
+    for _,v in pairs(search_dirs) do
+      table.insert(find_command, v)
+    end
+  elseif command == 'rg' then
+    for _,v in pairs(search_dirs) do
+      table.insert(find_command, v)
+    end
+  elseif command == 'find' then
+    table.remove(find_command, 1)
+    for _,v in pairs(search_dirs) do
+      table.insert(find_command, 2, v)
+    end
+  end
+end
+
 -- TODO: Maybe just change this to `find`.
 --          Support `find` and maybe let people do other stuff with it as well.
 files.find_files = function(opts)
@@ -74,6 +94,10 @@ files.find_files = function(opts)
     print("You need to install either find, fd, or rg. " ..
           "You can also submit a PR to add support for another file finder :)")
     return
+  end
+
+  if search_dirs then
+    insert_search_dirs(find_command, search_dirs)
   end
 
   if opts.cwd then
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -56,20 +56,20 @@ end
 -- NOTE(tj): Get clarification on comment and put this in the right place
 local function insert_search_dirs(find_command, search_dirs)
   local command = find_command[1]
+  local insert_pos = -1
   if command == 'fd' or command == 'fdfind' then
     table.insert(find_command, '.')
-    for _,v in pairs(search_dirs) do
-      table.insert(find_command, v)
-    end
-  elseif command == 'rg' then
-    for _,v in pairs(search_dirs) do
-      table.insert(find_command, v)
-    end
   elseif command == 'find' then
-    table.remove(find_command, 1)
-    for _,v in pairs(search_dirs) do
-      table.insert(find_command, 2, v)
+    for idx,val in ipairs(find_command) do
+      if val=='.' then
+        insert_pos = idx
+        break
+      end
     end
+    table.remove(find_command, insert_pos)
+  end
+  for _,v in pairs(search_dirs) do
+    table.insert(find_command, insert_pos==-1 and (#find_command+1) or insert_pos, vim.fn.expand(v))
   end
 end
 
@@ -104,6 +104,7 @@ files.find_files = function(opts)
   if opts.cwd then
     opts.cwd = vim.fn.expand(opts.cwd)
   end
+  P(find_command)
 
   opts.entry_maker = opts.entry_maker or make_entry.gen_from_file(opts)
 
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -96,6 +96,7 @@ files.find_files = function(opts)
     return
   end
 
+  local search_dirs = opts.search_dirs
   if search_dirs then
     insert_search_dirs(find_command, search_dirs)
   end
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -77,7 +77,7 @@ do
 
     mt_file_entry.cwd = cwd
     mt_file_entry.display = function(entry)
-      local display, hl_group = entry.value
+      local display, hl_group = path.make_relative(entry.value, cwd), nil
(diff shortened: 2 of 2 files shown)


tjdevries:
One thing is I think we should probably make sure these paths are absolute when we search, otherwise it might not work as expected? I'm not 100% sure... I'll have to think about this works exactly.

tjdevries:
Perhaps instead we could make `cwd` take a list of inputs as well, and we can transofmr the commands if `cwd` is a list instead of a string as well?... that might be more re-usable generally.

anihm136:
> One thing is I think we should probably make sure these paths are absolute
> Perhaps instead we could make cwd take a list of inputs as well
I'll try to answer both of these together. Right now paths can be relative (am using it that way right now), but they are relative to the cwd that has been provided (which defaults to nvim's cwd if nothing is explicitly provided). If we make cwd a list, then we would have to enforce something like the first entry being the working directory, which I feel is not easy to remember or use (plus it misses the case where you might want cwd not to be searched itself, but only subdirectories through relative paths). Also the uv function used to spawn the process explicitly needs a cwd, while the search dirs are just injected as args into the find command. Considering all this, I would suggest keeping them separate as far as possible

lua/telescope/builtin.lua
```diff
@@ -623,12 +623,32 @@ builtin.builtin = function(opts)
 end
 
 
+local function insert_search_dirs(find_command, search_dirs)
+  local command = find_command[1]
+  if command=='fd' or command=='fdfind' then
```
tjdevries, line 628: can you add spacing between the `==`?

lua/telescope/builtin.lua
```diff
@@ -647,6 +667,10 @@ builtin.find_files = function(opts)
     return
   end
 
+  if search_dirs then
+    insert_search_dirs(find_command, search_dirs)
```
tjdevries, line 671: Instead of doing this here, can we do this above? I think it will be much more obvious what we're searching for. Change the `find_command` declarations above for each of the different executables.

anihm136: I didn't understand exactly what you mean here. Do you want me to do the `insert_search_dirs` immediately after each `find_command` declaration in the if statement?

lua/telescope/builtin.lua
anihm136: Done

tami5:
@anihm136 what is pending here? do we need any assistance?

anihm136:
There was a comment from TJ that I didn't fully understand, so I wqas just waiting on some clarification. Otherwise it's working fine, I've been using it regularly in my setup

tami5:
@anihm136 Wonderful, I just tried to understand what @tjdevries meant by that. But for now, just just fix the conflict and add a comment `NOTE(tj): ` with his comment.

Then I think we we are good to merge after a quick and generous review from @Conni2461 :P

anihm136:
Done

Conni2461:
No you didn't fix the conflicts. the code you were looking for was moved to `lua/telescope/builtin/files.lua`. You basically reverted the changes from the last few days. If it is correctly rebased i will take a look at it.

anihm136:
Whoops, looks like my rebase did not do what I expected. I'll sort it out in a couple of minutes

anihm136:
Alright, I think I did the right thing now

Conni2461:
I think i know what tj meant. Why do we need another `if` block for this and don't to this in the original `find_command` `if` block. Would be this L79-L91 for you.

anihm136:
Ah okay I understand now. The reason I did it that way is mainly because there would essentially be a call to the same function (with the same args) in every branch of the `if` block, so I just moved it out. That way I felt it would also be easier to support more find commands if they came up. But if it helps readability I can move it back into the `find_command` block

anihm136:
Also I just noticed, the `find_command` block is only used when the find command is not specified in the opts. If it is, then the search dirs would be ignored if they are placed there. The `insert_search_dirs` function would also need to handle the case of having an empty `search_dirs`, otherwise it would be an additional `if` block whenever it is called

tami5:
@anihm136 I think search_dirs is not defined in the function, also paths in search_dirs need to be expanded in order to work.

anihm136:
Stupid mistake from my side while rebasing. Should be fixed now

lua/telescope/builtin/files.lua
```diff
@@ -53,6 +53,26 @@ files.grep_string = function(opts)
...
+    table.insert(find_command, '.')
+    for _,v in pairs(search_dirs) do
+      table.insert(find_command, v)
+    end
+  elseif command == 'rg' then
+    for _,v in pairs(search_dirs) do
+      table.insert(find_command, v)
+    end
+  elseif command == 'find' then
+    table.remove(find_command, 1)
```
tami5, line 69: Here you need to expand dir paths, otherwise it results in error by users who add relative path to home dir

anihm136: Okay I did that, and found another issue. Even when an absolute path is given(relative to home), the search path tries to find it relative to the cwd after expanding. I'm looking into how to fix that now

anihm136: I'm having some trouble finding the source of the issue. At some point, the cwd is being prepended to the search directories, but I cannot figure out exactly where

tami5: Is it because we've set cwd to some value perhaps?

```diff
   if command == 'fd' or command == 'fdfind' then
     table.insert(find_command, '.')
     for _,v in pairs(search_dirs) do
-      table.insert(find_command, v)
+      table.insert(find_command, vim.fn.expand(v))
     end
   elseif command == 'rg' then
     for _,v in pairs(search_dirs) do
-      table.insert(find_command, v)
+      table.insert(find_command, vim.fn.expand(v))
     end
   elseif command == 'find' then
     table.remove(find_command, 1)
     for _,v in pairs(search_dirs) do
-      table.insert(find_command, 2, v)
+      table.insert(find_command, 2, vim.fn.expand(v))
     end
   end
 end

```
Works well with me. Also it would be better if you moved the conditions inside the for loop so you don't end up with repeated code, don't you think?

windwp:
No description provided.

lua/telescope/builtin/files.lua
anihm136: Sorry, was busy for a couple of days. I still cannot figure out where it's happening though. Do you think it's okay to ignore the absolute path case for now, and only allow relative paths to the cwd?
I also rewrote the loops and conditions as you suggested

tami5: Nice looking forward to seeing the changes. For not using relative path case, I'm not sure why it didn't work for you, like what kinda of behavior are you getting?

The above diff works flawlessly for me.

```lua 
relativepath1 = "~/doc"
relativepath2 = "~/telescope"
builtin.fd({prompt_prefix = ">" , search_dirs =  [relativepath1 relativepath2] })

```

I'm sure I misunderstood what issue you are facing.

anihm136: I notice the following issues - 

1. If there are only absolute paths in the search dirs _and_ the cwd is the home directory, then the preview fails since it is getting the wrong path, but the correct file is opened
2. If there are a mixture of absolute and relative paths (or only absolute paths), and the cwd is the desired root of the relative paths (or anything other than the home directory), then bothe preview and file opening fails for the absolute paths

For example, running the following from my home directory
```lua
		find_command={ 'rg', '--files'},
		search_dirs = {"~/.config/sxhkd", "~/.config/alacritty"},
		width = .25
}
```
results in the following error in the preview (but the correct files open when I press `<CR>` on them)-
image
If the cwd in neovim is not my home directory, for the same case I get the same errors in preview, and the wrong file opens
For the case you mentioned above, try running it from any directory other than your home directory. Does it work?

anihm136: As for relative paths with the correct cwd set it works perfectly fine. Since that is the only way I was using it, I did not notice this issue until you mentioned the expansion for supporting absolute dirs

anihm136:
Meanwhile I've also fixed some issues with the location of insertion of the search dirs for other commands (specifically for `find`, the flags need to come before the search paths so I needed to make a small change in the logic)

lua/telescope/builtin/files.lua
tami5: Oh you ware right, I see what you mean, I just test it out with preview enabled.

tjdevries:
Do you want me to review?

anihm136:
Yes please. It's mostly working, only absolute paths are not handled yet. If you can give me a pointer to where all the path stuff is happening, I can take a look and fix it. Otherwise I can work on it in a separate PR, since all the other stuff is ready

lua/telescope/builtin/files.lua
```diff
@@ -53,6 +53,26 @@ files.grep_string = function(opts)
   }):find()
 end
 
+-- NOTE(tj): Get clarification on comment and put this in the right place
+local function insert_search_dirs(find_command, search_dirs)
+  local command = find_command[1]
+  local insert_pos = -1
+  if command == 'fd' or command == 'fdfind' then
+    table.insert(find_command, '.')
```
tjdevries, line 61: I think we should not have a separate function like this. but in the `find_files` section, we can change what the `find_command` is directly.

So on line 82 right now, you can do:

```lua
if search_dirs then
  find_command = { 'fd, '--type', 'f', '.' }
else
  find_command = {'fd', '--type', 'f'}
end
```
or similar.

Does that make sense? I don't like how this function lives out here but has exactly the same `if/elseif/else` idea as in `find_files`.

lua/telescope/builtin/files.lua
```diff
@@ -75,9 +95,15 @@ files.find_files = function(opts)
     return
   end
 
+  local search_dirs = opts.search_dirs
```
tjdevries, line 98: So up here, in the `if not find_command` that's when you should add it

lua/telescope/builtin/files.lua
anihm136: Okay got it

anihm136:
Hey @tjdevries I think I figured out where the path stuff happens. The issue is that, assuming all paths returned by the find commands are relative (I think), we concatenate the cwd to the paths in `lua/telescope/make_entry.lua` over here (line 98-101) -
```lua
      if k == "path" then
        return t.cwd .. path.separator .. t.value
      end
```
I'd like to update this to handle absolute paths, but I'm not sure if this is the right place to do it. Can I just make a simple check, something like if the path begins with a certain set of characters, consider it to be an absolute path and handle it accordingly, or do I need a more elaborate check?

lua/telescope/builtin/files.lua
anihm136: Hmmm. Doing it hard-coded that way makes it impossible to use `search_dirs` for custom commands. Or did you mean it figuratively, as in, should I just replicate the logic I have used in the function, just in the main `if` block in `find_files`? It might lead to some repeated code

tjdevries: if you pass a custom command, you should make it work with search_dirs. I think that is the fair way to implement. I would not worry about custom commands.

tjdevries: (by you, I mean, the user, in the above comment).

I prefer the hard coded way.

tjdevries:
Hmmm can we do something with `vim.loop` to determine whether a path is absolute? It seems like that might already be implemented.

anihm136:
I can think of one simple way to use `vim.loop` - check if the path exists when cwd is concatenated, and if not, treat it as an absolute path. The catch is that we assume the user only gives valid paths (although I think they need to give valid paths anyway)

tami5:
This might be better in term of performance rather than using `vim.loop.fs_stat`

```lua
local function is_absolute(path) 
  if path:find("~") or path:find("^%.") then 
    return false 
  else
    return true 
  end
end
```

anihm136:
Hmm. I've used `vim.loop.fs_access`, not sure how much that affects performance. Also @tami5 correct me if I'm wrong, but won't that solution work only on *Nix-like paths?

tami5:
> Hmm. I've used `vim.loop.fs_access`, not sure how much that affects performance. Also @tami5 correct me if I'm wrong, but won't that solution work only on *Nix-like paths?

Oh I'm not really sure, unfortunately I don't consider windows platforms when writing code at all. So this might be better.

tami5:
@anihm136 pleas resolve conflicts so I can push

anihm136:
Hmmm. There's a lint error, but I'm not sure if removing the variable assignment will break anything. Could you take a look and let me know? If it cannot be removed, I can probably disable the linter for that line if everything else is fine

lua/telescope/make_entry.lua
```diff
@@ -77,7 +77,7 @@ do
 
     mt_file_entry.cwd = cwd
     mt_file_entry.display = function(entry)
-      local display, hl_group = entry.value
+      local display, hl_group = path.make_relative(entry.value, cwd), nil
```
Conni2461, line 80: Just do `local dispay = path...` and do `local hl_group` before we actually need `hl_group`. Please don't ignore linting errors. Its here for a reason

anihm136: Done. Didn't mean to ignore the linter at the outset, only as a last resort if there was no other alternative

Conni2461:
LGTM, but i didn't follow the conversation so i don't know if @tjdevries has any more remarks.

tami5:
I'm going todo one final test tom. Thanks @anihm136

PR #239, opened 2020-11-13
Fix git files in non-git directories

anihm136:
It appears that the `git_files` picker is supposed to throw an error in a non-git directory, but is not doing so correctly (see commit message for why it happens). This PR fixes the issue. 
Also fixes a consequential issue that the error function was called with the wrong parameters (since the condition was never being met, it was probably missed)


tjdevries:
Ah yes, thank you. I have done this like.... 1 million times. haha thanks!

PR #243, opened 2020-11-15
fix lsp_code_action error when no code action available

smartding:
I get an error message executing builtin.lsp_code_action when there's no code action available:

E5108: Error executing lua ...telescope.nvim/lua/telescope/builtin.lua:232: attempt to get length of local 'results' (a nil value)

The reason is that the results variable can be nil when there's no code action available,

check for nil results before checking the length


tami5:
Yeb I can confirm I'm getting the same error

PR #216, opened 2020-11-03
Complete rewrite of README

tami5:
- Changes
  - New structure:
    - Overview
      - Installation (add examples for three plugin managers)
      - Dependencies (dependencies sounds better then requirements :D < opinionated)
      - Quick Start (to test telescope)
    - Configuration (contains a quick overview of configuration options and mappings)
    - FAQ (How to questions with answers)
    - API (oh boy needs review, and where most of todos is located. @tjdevries help :D)
    - Media (moved all links and videos here)
    - Recipes (some recipes I thought will be make it easier to configure telescope)

  - Strip previews, I feel telescope preview and screenshot should be located in
    gallery.  It might be good to include a signle preview in the introduction,
    but I feel it may constrian the expectations of new users, "This what I
    expect telescope offer", where as having previews only in gallery, "Oh
    telescope can be configured to look in different ways". :D < opinionated

  - Add faq section, any question answered in issues should be written here.
    (future pr I guess.) Add some questions I previously had, but more is needed
    to be added here.

  - Add configuration recipes based on how I configure telescope, @tjdevries
    please consider having the second recipe a builtin function. (perhaps with
    introducing an additional parameter for the dict)

TODO:

  - Fix links
  - Proofreading introduction

PR #249, opened 2020-11-16
Update attach_mappings to use new api

Conni2461:
- [x] Update readme. Should be out of date now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,11 +201,14 @@ end
 ["<C-i>"] = test_action,
 
 -- If you want your function to run after another action you should define it as follows
-local test_action = actions._transform_action(function(prompt_bufnr)
-  print("This function ran after another action. Prompt_bufnr: " .. prompt_bufnr)
-  -- Enter your function logic here. You can take inspiration from lua/telescope/actions.lua
-end)
-["<C-i>"] = actions.goto_file_selection_split + test_action
+local transform_mod = require('telescope.actions.mt').transform_mod
+local test_action = transform_mod {
+  x = function()
+    print("This function ran after another action. Prompt_bufnr: " .. prompt_bufnr)
+    -- Enter your function logic here. You can take inspiration from lua/telescope/actions.lua
+  end,
+}
+["<C-i>"] = actions.goto_file_selection_split + test_action.x
 
 ```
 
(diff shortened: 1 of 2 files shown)


lua/telescope/builtin.lua
```diff
@@ -504,37 +493,18 @@ builtin.help_tags = function(opts)
...
-      map('i', '<C-x>', hhelp)
-      map('n', '<C-x>', hhelp)
-      map('i', '<C-t>', thelp)
-      map('n', '<C-t>', thelp)
+        if cmd == 'edit' or cmd == 'new' then
+          vim.cmd('help ' .. selection.value)
+        elseif cmd == 'vnew' then
+          vim.cmd('vert bo help ' .. selection.value)
+        elseif cmd == 'tabedit' then
+          vim.cmd('tab help ' .. selection.value)
```
Conni2461, line 505: I think this can be done even better.
But this new api is amazing. So much cleaner

lua/telescope/builtin.lua
```diff
@@ -854,17 +821,19 @@ builtin.man_pages = function(opts)
...
-      end
-
-      map('i', '<CR>', view_manpage)
-      map('n', '<CR>', view_manpage)
+        if cmd == 'edit' or cmd == 'new' then
+          vim.cmd('Man ' .. selection.value)
+        elseif cmd == 'vnew' then
+          vim.cmd('vert bo Man ' .. selection.value)
+        elseif cmd == 'tabedit' then
+          vim.cmd('tab Man ' .. selection.value)
```
Conni2461, line 834: Same as help_tags. Could be better

lua/telescope/builtin.lua
```diff
@@ -97,24 +97,20 @@ builtin.commands = function()
       end
     },
     sorter = conf.generic_sorter(),
-    attach_mappings = function(prompt_bufnr, map)
-      local run_command = function()
-        local selection = actions.get_selected_entry(prompt_bufnr)
+    attach_mappings = function(prompt_bufnr)
+      actions.goto_file_selection_edit:replace(function()
```
Conni2461, line 101: Most of the time i replace only `actions.goto_file_selection_edit` because all other mappings do not make sense. Or should i just replace `actions._goto_file_selection` so that `<C-v>`, etc does the same as `<CR>`

Conni2461:
@tami5 tjdevries changed the actions api in 985856946e30a7d93eb3b8aac6b5b5d7d589a768, to make it even more awesome. I updated the current readme (probably not as good as you would do it). So take a look at the changes in readme in this pr and please implement it in #216. Thanks :)

tami5:
@Conni2461 sure thing.

tjdevries:
OK, LGTM. If you wanted to change anything else you can, otherwise feel free to merge

PR #251, opened 2020-11-16
Fix #250

### LINKED ISSUE #250: Previewer panics on invalid files
```
Error executing vim.schedule lua callback: ...nvim/plugged/telescope.nvim/lua/telescope/previewers.lua:29: ENOENT: n
o such file or directory:
```

https://imgur.com/EIxSYFs

Conni2461:


tjdevries:
Ah yeah, we should be more careful about our asserts. Better to just skip a result or something for those cases.

PR #252, opened 2020-11-17
check opts.cwd if it's a git directory, not opts

smartding:
We should check if opts.cwd is a git directory, not opts


Conni2461:
LGTM. Thanks for fixing my mistake :)

PR #256, opened 2020-11-17
refactor set_opts_cwd function so that git is called only once if cwd is a git dir

smartding:
The original implementation of set_opts_cwd calls git command twice if current dir is a git dir:
1. `git rev-parse HEAD` : to determine if current dir is a git dir
2. `git rev-parse --show-toplevel`: to find root of git dir

Can we remove the first call and just call `git rev-parse --show-toplevel` instead? It  returns an error message if not we are not in a git dir anyway.

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -144,20 +144,18 @@ git.status = function(opts)
 end
 
 local set_opts_cwd = function(opts)
-  local is_git_dir = function(path)
-    vim.fn.system('cd ' .. path .. ' && git rev-parse HEAD >/dev/null 2>&1')
-    if vim.v.shell_error ~= 0 then
-      error(path .. ' is not a git directory')
-    end
-  end
-
   if opts.cwd then
     opts.cwd = vim.fn.expand(opts.cwd)
-    is_git_dir(opts.cwd)
   else
-    is_git_dir(vim.fn.expand('%:p:h'))
-    --- Find root of git directory and remove trailing newline characters
-    opts.cwd = vim.fn.systemlist("git rev-parse --show-toplevel")[1]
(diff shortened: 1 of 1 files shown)


Conni2461:
I see what you mean, but we still have to do a check if `opt.cwd` is set.

lua/telescope/builtin/git.lua
```diff
@@ -144,20 +144,18 @@ git.status = function(opts)
...
-    opts.cwd = vim.fn.systemlist("git rev-parse --show-toplevel")[1]
+    opts.cwd = vim.fn.expand('%:p:h')
+  end
+
+  -- Find root of git directory and remove trailing newline characters
+  opts.cwd = vim.fn.systemlist("git -C " .. opts.cwd .. " rev-parse --show-toplevel")[1]
+
+  -- opts.cwd contains the error message from git command if not in a git dir
+  if 1 ~= vim.fn.isdirectory(opts.cwd) then
+    error(opts.cwd)
```
Conni2461, line 158: This results in a "not" good error message. For me
```
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

Conni2461: And checking for `vim.v.shell_error` is way better than isdirectory.
Could you keep the path around so we can get `error(path .. ' is not a git directory')` and if ret val is 0 than set opts.cwd?
Do you know what i mean

smartding: ok, the new commit uses `vim.v.shell_error` now, I also kept path around so that the error message is better.

Conni2461:
Thanks for that. Works expected.
Also thanks for basically reviewing my code :)

PR #260, opened 2020-11-18
Various previewer fixes

### LINKED ISSUE #228: how to set a environment variable for bat
telescope and fzf.vim have the same problem when using bat for previewing: the color in preview looks more saturated than running bat directly in command line.  See [this issue](https://github.com/junegunn/fzf.vim/issues/1065) in fzf.vim repo for  screenshots from many users.

The short term solution (as discussed [here](https://github.com/sharkdp/bat/issues/634#issuecomment-524525661) ) is to set the environment variable `COLORTERM=truecolor` each time we run bat. How can we do this in telescope?

### LINKED ISSUE #253: Preview Scrolling doesn't work
### Description

Scrolling with `<C-d>` and `<C-u>` doesn't work for me.

**Expected Behavior**

`<C-d> <C-u>` move preview up down without any issues

**Actual Behavior**

`<C-d> <C-u>` doesn't navigate preview window. Additionally with git_commits I
get additional error: `Error executing lua Vim:Can't send data to closed stream`

### Details

1. nvim -NU test.vim
2. :Telescope live_grep or git_commits 
3. `<C-d>/<C-u>`

</details>

- nvim version: v0.5.0-828-g0a95549d6
- Operating system :Archlinux
- Telescope commit: #749c865

</details>

<p>
 

```viml

endif

source /tmp/plug.vim
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
call plug#end()
" the command ran, also try Telescope fd
autocmd VimEnter * PlugClean! | PlugUpdate --sync | close | Telescope git_commits
lua << EOF
local actions = require'telescope.actions'
require('telescope').setup{
  defaults = {
    vimgrep_arguments = {
      'rg',
      '--color=never',
      '--no-heading',
      '--with-filename',
      '--line-number',
      '--column',
      '--smart-case'
    },
    prompt_position = "top", --// change
    prompt_prefix = ">",
    selection_strategy = "reset",
    sorting_strategy = "ascending", --// change
    layout_defaults = {},
    file_sorter =  require'telescope.sorters'.get_fuzzy_file ,
    generic_sorter =  require'telescope.sorters'.get_generic_fuzzy_sorter,
    file_ignore_patterns = {},
    shorten_path = true,
    winblend = 0,
    width = 0.75,
    preview_cutoff = 120,
    results_height = 1,
    results_width = 0.8,
    border = {},
    borderchars = { '─', '│', '─', '│', '╭', '╮', '╯', '╰'},
    color_devicons = true,
    use_less = true,
  }
}
EOF
```
</p>
</details>

Conni2461:
Fix #228
Fix #253

- Set env vars for previewer with set_env({ key = value } #228)
- Set pager for git previewes (default is GIT_PAGER=less so users can scroll see #253, but we can't just use the configured pager from `gitconfig` because pager's like `delta` which have a `split` view. That does not look good)
- Use less for people who don't have bat but less installed (see https://github.com/nvim-telescope/telescope.nvim/pull/216#discussion_r525067979)
- Reset less env to only '-RS' because '-X' will break scrolling(#253 thanks @eth3lbert)
- Scrolling for buffer previewer(thanks @tami5 for helping)
  - Also new function to create buffer previewers (which has scrolling by default). Used for tags and help_tags previewer.
- Add new setting set_env to readme

@smartding @ThomasFeher
Add `set_env = { ['COLORTERM'] = 'truecolor' }` to your telescope config. Does this work

@tami5 My autocmd has removed so many trailing whitespaces.

### DIFF
--- a/lua/telescope/previewers.lua
+++ b/lua/telescope/previewers.lua
@@ -20,6 +20,7 @@ Previewer.__index = Previewer
 -- TODO: Should play with these some more, ty @clason
 local bat_options = {"--style=plain", "--color=always", "--paging=always"}
 local has_less = (vim.fn.executable('less') == 1) and config.values.use_less
+local termopen_env = vim.tbl_extend("force", { ['GIT_PAGER'] = (has_less and 'less' or ''), LESS = '-RS' }, config.values.set_env)
 
 local get_file_stat = function(filename)
   return vim.loop.fs_stat(vim.fn.expand(filename)) or {}
@@ -39,9 +40,9 @@ local bat_maker = function(filename, lnum, start, finish)
 
   if has_less then
     if start then
-      table.insert(command, {"--pager", string.format("less -RS +%s", start)})
+      table.insert(command, {"--pager", string.format("'less +%s'", start)})
     else
-      table.insert(command, {"--pager", "less -RS"})
+      table.insert(command, {"--pager", "less"})
     end
   else
     if start and finish then
@@ -54,14 +55,14 @@ local bat_maker = function(filename, lnum, start, finish)
   end
 
   return flatten {
-    command, bat_options, "--", vim.fn.expand(filename)
+    command, bat_options, "--", "'" .. vim.fn.expand(filename) .. "'"
   }
 end
 
 -- TODO: Add other options for cat to do this better
-local cat_maker = function(filename, _, _, _)
+local cat_maker = function(filename, _, start, _)
   if get_file_stat(filename).type == 'directory' then
-    return { 'ls', '-la', vim.fn.expand(filename) }
+    return { 'ls', '-la', "'" .. vim.fn.expand(filename) .. "'" }
   end
 
   if 1 == vim.fn.executable('file') then
@@ -72,9 +73,17 @@ local cat_maker = function(filename, _, _, _)
     end
   end
 
-  return {
-    "cat", "--", vim.fn.expand(filename)
-  }
+  if has_less then
+    if start then
+      return { 'less', string.format('+%s', start), "'" .. vim.fn.expand(filename) .. "'" }
+    else
+      return { 'less', "'" .. vim.fn.expand(filename) .. "'" }
+    end
+  else
+    return {
+      "cat", "--", "'" .. vim.fn.expand(filename) .. "'"
+    }
+  end
 end
 
 local get_maker = function(opts)
@@ -167,6 +176,41 @@ previewers.new = function(...)
   return Previewer:new(...)
 end
 
+previewers.new_buffer_previewer = function(opts)
+  opts = opts or {}
+
+  assert(opts.preview_fn, "preview_fn is required function")
+
+  if not opts.scroll_fn then
+    function opts.scroll_fn(self, direction)
+      local input = direction > 0 and "d" or "u"
+      local count = math.abs(direction)
+
+      self:send_input({ count = count, input = input })
+    end
+  end
+
+  if not opts.send_input then
+    function opts.send_input(self, input)
+      if not self.state then
+        return
+      end
+
+      local bufnr = vim.api.nvim_win_get_buf(self.state.hl_win)
+      local max_line = vim.fn.getbufinfo(bufnr)[1].linecount
+      local line = vim.api.nvim_win_get_cursor(self.state.hl_win)[1]
+      if input.input == 'u' then
+        line = (line - input.count) > 0 and (line - input.count) or 1
+      else
+        line = (line + input.count) <= max_line and (line + input.count) or max_line
+      end
+      vim.api.nvim_win_set_cursor(self.state.hl_win, { line, 1 })
+    end
+  end
+
+  return Previewer:new(opts)
+end
+
 previewers.new_termopen_previewer = function(opts)
   opts = opts or {}
 
@@ -234,13 +278,25 @@ previewers.new_termopen_previewer = function(opts)
 
     local term_opts = {
       cwd = opts.cwd or vim.fn.getcwd(),
+      env = termopen_env
     }
 
+    -- TODO(conni2461): Workaround for neovim/neovim#11751.
+    local get_cmd = function(status)
+      local shell = vim.o.shell
+      if string.find(shell, 'powershell.exe') or string.find(shell, 'cmd.exe') then
+        return opts.get_command(entry, status)
+      else
+        local env = {}
+        for k, v in pairs(termopen_env) do
+          table.insert(env, k .. '=' .. v)
+        end
+        return table.concat(env, ' ') .. ' ' .. table.concat(opts.get_command(entry, status), ' ')
+      end
+    end
+
     with_preview_window(status, bufnr, function()
-      set_term_id(
-        self,
-        vim.fn.termopen(opts.get_command(entry, status), term_opts)
-      )
+      set_term_id(self, vim.fn.termopen(get_cmd(status), term_opts))
     end)
 
     vim.api.nvim_buf_set_name(bufnr, tostring(bufnr))
@@ -274,7 +330,7 @@ previewers.new_termopen_previewer = function(opts)
 end
 
 previewers.vim_buffer = defaulter(function(_)
-  return previewers.new {
+  return previewers.new_buffer_previewer {
     setup = function()
       return {
         last_set_bufnr = nil
@@ -333,6 +389,7 @@ previewers.vim_buffer = defaulter(function(_)
         -- print("LNUM:", entry.lnum)
       end
 
+      vim.api.nvim_win_set_option(status.preview_win, 'scrolloff', 999)
       log.info("Previewed bufnr", bufnr)
     end,
   }
@@ -342,7 +399,7 @@ previewers.git_commit_diff = defaulter(function(_)
   return previewers.new_termopen_previewer {
     get_command = function(entry)
       local sha = entry.value
-      return { 'git', '--no-pager', 'diff', sha .. '^!' }
+      return { 'git', '-p', 'diff', sha .. '^!' }
     end
   }
 end, {})
@@ -350,8 +407,8 @@ end, {})
 previewers.git_branch_log = defaulter(function(_)
   return previewers.new_termopen_previewer {
     get_command = function(entry)
-      return { 'git', 'log', '--graph',
-               '--pretty=format:%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr)%Creset',
+      return { 'git', '-p', 'log', '--graph',
+               "--pretty=format:'%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr)%Creset'",
                '--abbrev-commit', '--date=relative', entry.value }
     end
   }
@@ -360,7 +417,7 @@ end, {})
 previewers.git_file_diff = defaulter(function(_)
   return previewers.new_termopen_previewer {
     get_command = function(entry)
-      return { 'git', '--no-pager', 'diff', entry.value }
+      return { 'git', '-p', 'diff', entry.value }
     end
   }
 end, {})
@@ -400,8 +457,8 @@ previewers.vimgrep = defaulter(function(opts)
   }
 end, {})
 
-previewers.ctags = defaulter(function(opts)
-  return previewers.new {
+previewers.ctags = defaulter(function(_)
+  return previewers.new_buffer_previewer {
     setup = function()
       return {}
     end,
@@ -428,10 +485,10 @@ previewers.ctags = defaulter(function(opts)
         vim.api.nvim_win_set_option(status.preview_win, 'winhl', 'Normal:Normal')
         vim.api.nvim_win_set_option(status.preview_win, 'signcolumn', 'no')
         vim.api.nvim_win_set_option(status.preview_win, 'foldlevel', 100)
+        vim.api.nvim_win_set_option(status.preview_win, 'scrolloff', 999)
 
         pcall(vim.fn.matchdelete, self.state.hl_id, self.state.hl_win)
         vim.fn.search(scode)
-        vim.cmd "norm zz"
 
         self.state.hl_win = status.preview_win
         self.state.hl_id = vim.fn.matchadd('Search', scode)
@@ -469,7 +526,7 @@ previewers.qflist = defaulter(function(opts)
 end, {})
 
 previewers.help = defaulter(function(_)
-  return previewers.new {
+  return previewers.new_buffer_previewer {
     setup = function()
       return {}
     end,
@@ -542,9 +599,12 @@ previewers.help = defaulter(function(_)
 
         log.trace([[lua vim.fn.search("]], search_query, [[")]])
 
+        -- Clear previous search
+        pcall(vim.fn.matchdelete, self.state.hl_id, self.state.hl_win)
         vim.cmd "norm! gg"
         vim.fn.search(search_query, "W")
-        vim.cmd "norm  zt"
+
+        vim.api.nvim_win_set_option(status.preview_win, 'scrolloff', 999)
 
         self.state.hl_win = status.preview_win
         self.state.hl_id = vim.fn.matchadd('Search', search_query)
@@ -681,7 +741,7 @@ previewers.man = defaulter(function(_)
 end)
 
 previewers.display_content = defaulter(function(_)
-  return previewers.new {
+  return previewers.new_buffer_previewer {
     preview_fn = function(self, entry, status)
       with_preview_window(status, nil, function()
         local bufnr = vim.fn.bufadd("Preview command")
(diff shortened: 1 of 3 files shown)


lua/telescope/previewers.lua
```diff
@@ -39,9 +40,9 @@ local bat_maker = function(filename, lnum, start, finish)
 
   if has_less then
     if start then
-      table.insert(command, {"--pager", string.format("less -RS +%s", start)})
+      table.insert(command, {"--pager", string.format("'less -RS +%s'", start)})
     else
-      table.insert(command, {"--pager", "less -RS"})
+      table.insert(command, {"--pager", "'less -RS'"})
```
Conni2461, line 45: Required because this is no longer 1 argument. So we have to add quotes

tami5:
@Conni2461 it works really well. This totally fixed #228. I wonder if there is any other uses for the new option `set_env` .... 

For #253 it didn't solve the issue, but guess what <c-u> works now for both git and grep but `<c-d>` seems to break the preview all together.  I'm currently on the hunt for something in my env (zsh, tmux ...) that maybe causing this, since I guess preview is term split right?

*FYI the cursorline highlight seem a bit off, or should I say little darker then my config. Any suggestions?*

lua/telescope/previewers.lua
tami5: use [[]] for that one. I used when ever there is `\\` `' '` `" "` involved in the string.

Conni2461:
> I wonder if there is any other uses for the new option `set_env`

Besides `GIT_PAGER`? I don't know right now but different users  will find different things.

> For #253 it didn't solve the issue

It should not fix it for you. But scroll up and down works for me now in `git_builtin`s. I thought about your env and had a suspicion that maybe your terminal emulator catches some special key combinations. Did you try it without tmux or different emulator?

tami5:
I just tested kitty term same issue (BTW I use st). 

For the result of `print(input)` it did show 30d and 30u, here is a preview of the behavior https://asciinema.org/a/NGOw7cdknQfB5Eb39VqwRAyC7, you can see that after pressing c-d everything breaks

lua/telescope/previewers.lua
```diff
@@ -413,6 +429,30 @@ previewers.ctags = defaulter(function(opts)
...
+        if input.input == 'u' then
+          line = (line - count) > 0 and (line - count) or 1
+        else
+          line = (line + count) <= max_line and (line + count) or max_line
+        end
+        vim.api.nvim_win_set_cursor(self.state.hl_win, { line, 1 })
+        vim.cmd "norm zz"
+      end
+    end,
+
```
Conni2461, line 455: @tjdevries Does this make sense for implementing scrolling for vimbuffer in preview? Because i would love to just do `vim.cmd(':exe norm <c-d>')` but this does nothing. Even testing this just in nvim. 
`zz` does nothing aswell because im in a different window. How can i just pass this to my correct window?

lua/telescope/previewers.lua
Conni2461: Will not work here because we are not escaping anything. These `'` are meant for `bat` so that `bat` knows that `-RS` is meant for `less` and not `bat`.  Also required when running this is `bash` or `zsh`

Conni2461:
> I just tested kitty term same issue (BTW I use st).

st works fine for me. But it seems like scrolling works for you. Just in a weird way. 

I also added scrolling to tags preview (that no longer a bat process but a vim buffer). Its not perfect yet because `zz` does not work but could you try that out. Thanks

tami5:
> I also added scrolling to tags preview (that no longer a bat process but a vim buffer). Its not perfect yet because `zz` does not work but could you try that out. Thanks

Just tested it well done!!, it works really well. But there is an issue when switching between up and down preview (my env I'm sure :smile: ): first time pressing `<c-d>` nothing moves only the second time the scroll down start working okay, then on first time pressing `<c-u>` nothing moves only the second time the scroll go up. However if I open new tag window and press `<c-u> it works on first click .

tami5:
@Conni2461 it turns out that the issue I previously mentioned is not related to my issue with less, (which by the way was solved using `set_env {LESS "iMRS"}` here's another useful usecase :D).

Basically the issue is whenever I switch between up/down I need to press it twice .

Conni2461:
I think we can actually just set `LESS = ''` by default. This way we can clean the `LESS` env for the user and make sure that scrolling works for him. We don't need `i` because they can't search. `-M` (long prompt) is not really needed here. We set `R` when we build our output command. Same goes for `S`. Great that you have scrolling working now.

About the vim buffer scrolling. The problem is that we are in a different window. We have focus on the search window. I would have loved to just do something like `vim.cmd(':norm! <C-D>')` to scroll but this does nothing. Everywhere.
So what i'm doing is just manually calculating the new cursor line and than doing a `zz`. But i haven't figure out how to do a `zz` in a different window. Thats why you have to press twice when switching. Basically the cursor line sits at the bottom after scrolling down and when you scroll up the curosor sits now at the top and is just not out of window.

tami5:
> I think we can actually just set LESS = '' by default. This way we can clean the LESS env for the user and make sure that scrolling works for him. 

Great idea, it would be tedious to mention in the doc to add this option. 

> We don't need i because they can't search. -M (long prompt) is not really needed here. We set R when we build our output command. Same goes for S. Great that you have scrolling working now.

Okay, Okay I see where this going, :smile: taking over user env configuration :smiling_imp: . No serriouslly that would make us avoid many issues, and perhaps integrate term with telescope somehow with taking advantage of this. 


> But i haven't figure out how to do a zz in a different window. Thats why you have to press twice when switching. Basically the cursor line sits at the bottom after scrolling down and when you scroll up the curosor sits now at the top and is just not out of window which means no scrolling.

I will look into it

tami5:
@Conni2461 try this, It fixed the issue while centering the view, not sure though if this what you was looking for

```diff 
@@ -449,7 +449,7 @@ previewers.ctags = defaulter(function(_)
           line = (line + count) <= max_line and (line + count) or max_line
         end
         vim.api.nvim_win_set_cursor(self.state.hl_win, { line, 1 })
-        vim.cmd "norm zz"
+        -- vim.cmd "norm zz"
       end
     end,
 
@@ -468,6 +468,7 @@ previewers.ctags = defaulter(function(_)
         vim.api.nvim_win_set_option(status.preview_win, 'winhl', 'Normal:Normal')
         vim.api.nvim_win_set_option(status.preview_win, 'signcolumn', 'no')
         vim.api.nvim_win_set_option(status.preview_win, 'foldlevel', 100)
+        vim.api.nvim_win_set_option(status.preview_win, 'scrolloff', 999)
 
         pcall(vim.fn.matchdelete, self.state.hl_id, self.state.hl_win)
         vim.fn.search(scode)
-        vim.cmd "norm zz"
+        -- vim.cmd "norm zz"
```

Conni2461:
Note to everyone who is using this. I just forced pushed :). Sry. But now there is working scrolling for tags and help_tags :) Thanks to @tami5

tami5:
> Note to everyone who is using this. I just forced pushed :). Sry. But now there is working scrolling for tags and help_tags :) Thanks to @tami5

Oh that solved the issue you ware talking about, I was like it can't be that simple for our amazing @Conni2461,  (although it took me some time to figure it out :$). Am glad I could help @Conni2461 . I'm going to test it in few hours. But I'm so excited to see to what extent we can take this as it is wayyyyyyyyyyyyyyyyyy better then bat!!. 

Idea

- scroll indicator 
- smooth scrolling 
- QF functionalities << tooo far

tami5:
Just test it out lgmt, I think we should merge @Conni2461

Conni2461:
I think we should at least wait till @tjdevries had a chance to look at it.
How we add env vars is super hacky and just temporary until neovim/neovim#11751 is fixed.
@tjdevries should give his okay on this one. Also it doesn't hurt if this needs two more days until this is merged.

> * scroll indicator
> * smooth scrolling

Add to wishlist. These are good proposals. And i can start working on this sometime in the future :)

> * QF functionalities << tooo far

I have no idea what this is. We need a chatroom, like it is proposed by clason, to talk about stuff like this without spamming at a random PR. :laughing:

lua/telescope/previewers.lua
```diff
@@ -236,10 +245,17 @@ previewers.new_termopen_previewer = function(opts)
...
+      table.insert(env, k .. '=' .. v)
+    end
+    env = table.concat(env, ' ')
+
     with_preview_window(status, bufnr, function()
       set_term_id(
         self,
-        vim.fn.termopen(opts.get_command(entry, status), term_opts)
+        vim.fn.termopen(env .. ' ' .. table.concat(opts.get_command(entry, status), ' '),
+                        term_opts)
```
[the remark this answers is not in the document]
tjdevries: So this can be removed when the bug in neovim is fixed?

My only concern is this the same syntax for windows and/or other shells?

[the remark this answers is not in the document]
Conni2461: Yes i would remove this if the bug upstream is fixed.

According to this: [cmd](https://superuser.com/questions/223104/setting-and-using-variable-within-same-command-line-in-windows-cmd-exe) and [powershell](https://stackoverflow.com/questions/1420719/powershell-setting-an-environment-variable-for-a-single-command-only) will both not work.

lua/telescope/previewers.lua
```diff
@@ -236,11 +280,46 @@ previewers.new_termopen_previewer = function(opts)
...
+        for k, _ in pairs(termopen_env) do
+          output = output .. "$env: " .. k .. "=''; "
+        end
+        return output
+      else
+        local env = concat_env()
+        return table.concat(env, ' ') .. ' ' ..
+               table.concat(opts.get_command(entry, status), ' ')
+      end
+    end
```
Conni2461, line 318: I closed the old conv. This is the new "shit".
It already worked with `bash`, `zsh`, and `dash`.
Handles `cmd.exe` and `powershell.exe`, if these are actually the correct names :laughing:. Are there any other shells i am currently not thinking about?
My implementation is based on these two threads. And i printed the resulting outputs to check if everything works theoretically but i can't test it.
Its a super bad workaround but it would be great if we could manage to get this working for all platforms.

Conni2461: Just testing inside a windows vm. Both not yet working.

Conni2461: So it works for fd with both `cmd.exe` and `powershell.exe`.
For `git_commit` etc preview does not work with `set shell=cmd.exe`. It fails with `terminal is not fully functional`. Does also not work on `master` so i guess that is a upstream/cmd.exe bug?! Just running something like `lua vim.fn.termopen({'git', 'log'})` will also give that result. So i won't bother as long as no windows user will report it :laughing: 

With `set shell=powershell.exe` git does not work at all. Just running `!git` will result in `An unknown error was received. This can happen if the remote process closed or ended abnormally. shell returned 4000`. Seems like an typical windows error. Maybe i did something wrong. When setting up all this stuff inside a vm.

sunjon:
> Idea
    scroll indicator
    smooth scrolling
    QF functionalities << tooo far

I have some mostly working scrollbar code I made a while ago  I'd wanted to make it more API-like so it could be used by other plugins, but got distracted by other projects. It uses ascii half-blocks and looks quite pretty.
It has a couple of outstanding problems like some rounding bugs which can make the scroll-thumb change size slightly when scrolling with mouse. You're welcome to the code if you want to do something more with it.

preview

lua/telescope/previewers.lua
Conni2461: After I slept over it for a night, I think it's stupid. We should just not support env on powershell and cmd for now. And write something like this.
```lua
local get_cmd = function(status)
  if vim.o.shell == 'powershell.exe' or vim.o.shell == 'cmd.exe' then -- we should probably do a string.find here
    if table.getn(config.values.set_env) ~= 0 then
      print('Env is currently not supported for your shell')
    end
    return opts.get_command(entry, status)
  else
    return table.concat(env, ' ') .. ' ' .. table.concat(opts.get_command(entry, status), ' ')
  end
end
```

Conni2461:
@sunjon i'm kinda interested right now. But that preview doesn't work :cry:. Im getting `Non-Image content-type returned`

tami5:
@Conni2461 I think this PR should be renamed to Enhance Preview, because it is getting out of the scope of simple fixes.

sunjon:
@Conni2461 fixed the link to the preview

tami5:
> > Idea
> > scroll indicator
> > smooth scrolling
> > QF functionalities << tooo far
> 
> I have some mostly working scrollbar code I made a while ago I'd wanted to make it more API-like so it could be used by other plugins, but got distracted by other projects. It uses ascii half-blocks and looks quite pretty.
> It has a couple of outstanding problems like some rounding bugs which can make the scroll-thumb change size slightly when scrolling with mouse. You're welcome to the code if you want to do something more with it.
> 
> [preview](https://streamable.com/5plgxw) - (ignore the code in the bg, it was rewritten in lua already)

Oh that looks really nice @sunjon

sunjon:
@Conni2461  re: scrollbars, there's [this PR](https://github.com/neovim/neovim/pull/13117) that landed recently that can remove the `&iconstring` hack that I'd used

Conni2461:
Thank you very much for the code and all the insight :heart:. I will start working on this this weekend.

Conni2461:
@tami5 @sunjon I got it working :smile: Kinda.
I don't use autocmd to start it i just use the `setup` and `teardown` functions from previewers and `update` in the `scroll` function.
That attaching happens kinda too early so i have to fix that and i have to make the bar hight fixed. I will open a new PR of this version tomorrow. This pr will be based on this one.
Btw sunjon I reduced the code to 120 lines. Basically everything that has to do with handling multiple windows and viewport changes are not relevant for our use case. But thank you for the code base i really didn't have to write anything new and it saved me soooo much time :smile:.

[asciicast](https://asciinema.org/a/374268)

tami5:
> @tami5 @sunjon I got it working  Kinda.
> I don't use autocmd to start it i just use the `setup` and `teardown` functions from previewers and `update` in the `scroll` function.
> That attaching happens kinda too early so i have to fix that and i have to make the bar hight fixed. I will open a new PR of this version tomorrow. This pr will be based on this one.
> Btw sunjon I reduced the code to 120 lines. Basically everything that has to do with handling multiple windows and viewport changes are not relevant for our use case. But thank you for the code base i really didn't have to write anything new and it saved me soooo much time .
> 
> [asciicast](https://asciinema.org/a/374268)

Great work @Conni2461 Its how far your going!! . Thanks a lot, I'm going to test it soon.

tami5:
@Conni2461 smooth!!

lua/telescope/previewers.lua
tjdevries: Instead of printing, let's silently ignore and make a note in the readme. No need to spam people with the same message every time. (we can of course still pass the dict into `env`, just in case that ever gets fixed in master)

I am fine with just making sure that we do not attempt to run for invalid shells. That is good with me.

TY for all the investigation.

tjdevries:
@Conni2461 do you want me to review now?

Conni2461:
@tjdevries i am going to update the `get_cmd`. Just read your comment. And than you can take a deeper look. I'm kinda exited about some of these changes. Would be great if you could get these out today :)

lua/telescope/previewers.lua
```diff
@@ -167,6 +176,41 @@ previewers.new = function(...)
...
+
+      local bufnr = vim.api.nvim_win_get_buf(self.state.hl_win)
+      local max_line = vim.fn.getbufinfo(bufnr)[1].linecount
+      local line = vim.api.nvim_win_get_cursor(self.state.hl_win)[1]
+      if input.input == 'u' then
+        line = (line - input.count) > 0 and (line - input.count) or 1
+      else
+        line = (line + input.count) <= max_line and (line + input.count) or max_line
+      end
+      vim.api.nvim_win_set_cursor(self.state.hl_win, { line, 1 })
```
Conni2461, line 207: I'm still not happy about this "thing". I would love to just do `vim.api.nvim_feedkeys(key, 'n', true)`, but i am in the wrong window and `:exe nr .. wincmd w` does not work. Same goes for `windo`. Do you have any idea?

tjdevries: Later bfreld and I will add `vim.api.nvim_ex_call` which will allow setting this like `noautocmd`, `windo` etc.

As of now, this is probably best

tjdevries:
LGTM assuming you're done! Go ahead and merge

lua/telescope/previewers.lua
Conni2461: I just looked at the pr. Looks really interesting. Will keep an eye on that one. Thanks for the review

Conni2461:
I will merge soon. I found the breaking change :sob:

Issue #255, opened 2020-11-17
Wishlist


tami5:
Here we can post ideas of *features*, *improvement* and *new pickers*. 

Later we can further refine an idea and update progress on it or link it  to a more detailed "enhancement/feature issue", or even reference the comment in a PR.

Try to leave discussions out of here, but if necessary or your have something in mind post it then mark your replies and mentions as out of scope to keep the Wishlist clean.

tami5:
Feature: combine pickers #213 . I agree with @wbthomason , I think its really useful in cases where you want to combine things like 

- quickfix + loclist 
- old_files + buffers
- dotfiles + nvim + bin files ...

tami5:
Enhancement: add additional parameter to determine the depth of a given cwd #176 . Useful when you want to set the cwd but you don't want to go further than 2 levels

tami5:
Feature: Create a picker for [vista](https://github.com/liuchengxu/vista.vim). First suggested by tj in #138

tami5:
Idea: with `actions._goto_file_selection(prompt_bufnr, command)` if no entry, instead of `print("[telescope] Nothing currently selected")` edit any way, sometime I'd like to create a file if it doesn't exists.

tjdevries:
documentation generator :)

tamago324:
builtin add 'filetypes'

tami5:
> documentation generator :)

I can work on this @tjdevries  using https://stevedonovan.github.io/ldoc/  it may take few weeks to go over the code base but I will be happy to do it. Later we can use the output from ldoc, manipulate it and create doc/telescope.txt or even webpage :D

sharksforarms:
builtin for `:highlights` :)

sunjon:
A lot of the ideas here  are relocated(prioritized?) existing issues.

The ability to scroll through the search results is a frequent pain for me. I don't think fuzzy finders are intended to filter results down to 5-10 items;  Often the choice is offscreen due to sorting, and knowing it exists momentarily via a page-down, is enough for me to make the descision to edit the query further or press "down" a couple of times, as I already know further refinement wouldn't be that useful.

weilbith:
Yank history

sunjon:
> Yank history

Yank history could be a wrapper around existing functionality in something like [nvim-miniyank](https://github.com/bfredl/nvim-miniyank), but I personally don't think Telescope should do the legwork itself.

[...]

PR #263, opened 2020-11-18
Add filetypes builtin

Conni2461:
https://github.com/nvim-telescope/telescope.nvim/issues/255#issuecomment-729303825 can be deleted or hidden when this PR is merged.

@tamago324 please test it out and report back thanks :)

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -888,6 +888,28 @@ builtin.keymaps = function(opts)
   }):find()
 end
 
+builtin.filetypes = function(opts)
+  opts = opts or {}
+
+  local filetypes = vim.fn.getcompletion('', 'filetype')
+
+  pickers.new({}, {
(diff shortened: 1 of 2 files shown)


lua/telescope/builtin/init.lua
```diff
@@ -913,10 +935,10 @@ builtin.tags = function(opts)
     },
     previewer = previewers.ctags.new(opts),
     sorter = conf.generic_sorter(opts),
-    attach_mappings = function(prompt_bufnr)
+    attach_mappings = function(_)
```
tjdevries, line 938: you don't need any param here if you don't want. Only time to use `_` as a param is if you have some parameter you actually want later in the function. `function(_, _, param)` or something like that.

tjdevries:
You can merge if it works, not somewhere I can test this easily atm

Conni2461:
Works for me. And if it is not working, they can still complain :)

PR #264, opened 2020-11-19
fix cwd detection of builtin.git_files

smartding:
Let's say I'm working in a git dir, and I open a vim help buffer.

If I run `builtin.git_files()` now, it will report we are not in a git dir because `git_files` detects cwd by `expand('%:p:h')`, which is expanded to `/usr/share/nvim/runtime/doc`

`getcwd` is a better way to detect current working dir.

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -147,7 +147,7 @@ local set_opts_cwd = function(opts)
   if opts.cwd then
     opts.cwd = vim.fn.expand(opts.cwd)
   else
-    opts.cwd = vim.fn.expand('%:p:h')
+    opts.cwd = vim.fn.getcwd()
   end
 
   -- Find root of git directory and remove trailing newline characters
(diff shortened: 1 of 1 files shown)


lua/telescope/builtin/git.lua
```diff
@@ -147,7 +147,7 @@ local set_opts_cwd = function(opts)
   if opts.cwd then
     opts.cwd = vim.fn.expand(opts.cwd)
   else
-    opts.cwd = vim.fn.expand('%:p:h')
+    opts.cwd = vim.fn.getcwd()
```
tami5, line 150: Use `vim.loop.cwd` better, (`vim.fn.getcwd = 0.03267 msecs,  vim.loop.cwd = 009026 msecs`)

tami5 (commented):
I think its a great idea, unless there is some side effect @Conni2461

lua/telescope/builtin/git.lua
smartding: wasn't aware there's a faster version of cwd in vim.loop, thanks

tami5: FYI: `vim.fn.expand('%:p:h')` is 0.09 - 0.1 mescs, slowest of them all :smile:

Conni2461:
No seems like a great change. Thank you :)

PR #266, opened 2020-11-19
fix: builtin runner insert issue

### LINKED ISSUE #262: Telescope builtin: prompt in normal mode instead of insert mode
after running `:Telescope builtin` and selecting some picker, prompt is in normal mode instead of being in insert mode

tami5:
Fixes #262

I don't like the fact that I added an additional parameter,

@doubleloop please tested by checking out this branch after doing git fetch


Conni2461:
> I don't like the fact that I add an additional parameter,

Does `vim.fn.mode()` help? Right now insert should be nil because you haven't added the second parameter in line 175.

tami5:
h

> > I don't like the fact that I add an additional parameter,
> 
> Does `vim.fn.mode()` help? Right now insert should be nil because you haven't added the second parameter in line 175.

Good catch, hmm let me see, this function return a string that indicates the current mode. the close action actually kills the process along with any additional cmd in `run_builtin` so I'm not sure.

tami5:
This case is tricky since the run_builtin doesn't renew the content of the float but rather close it then open a fresh new one. action.close insure that the when the window close it go to normal mode.

@Conni2461 I don't know do you have better idea to skip the additional param, or should we keep ?? since it doesn't bother any other picker :laughing:

Conni2461:
yeah i think `mode()` will not work. Adding a new parameter in lua isn't that bad because it will be nil if it is not present. Let's wait for tj review and what he thinks

tjdevries:
I think for the `builtin`, what we can do is give `action.close` a `post` action of `startinsert` again, which will put you back in insert mode.

Does that make sense? @Conni2461 knows about the `post` stuff, and can link to wherever we just used it :)

tami5:
@Conni2461 how I can I use post stuff tj mentioned ? or should I just merge since its just a minor thing? Thanks

Conni2461:
Don't merge. We will fix this. I am kinda here and can walk you through.
You wanna look at `enhance`. You can find docs in the readme and examples in `builtin.tags` and `builtin.current_buffer_fuzzy_find`.

tami5:
> Don't merge. We will fix this. I am kinda here and can walk you through.
> You wanna look at `enhance`. You can find docs in the readme and examples in `builtin.tags` and `builtin.current_buffer_fuzzy_find`.
@Conni2461 
Got it

So I couldn't understand how I can extend with replace?

```
    attach_mappings = function(_)
      actions.goto_file_selection_edit:replace(actions.run_builtin)
      return true
    end

      actions.run_builtin:enhance {
        post = function()
          vim.cmd [[startinsert]]
        end
      }


```

Conni2461:
I think tj suggested you should extend `actions.close`. And you would need to put this inside attach_mappings after the replace

tami5:
@Conni2461 
Like that?
```lua
      actions.goto_file_selection_edit:replace(actions.run_builtin)
      actions.close:extend{
        post = function() vim.cmd [[startinsert]] end
      }
        return true
```

got `Error executing lua ...packer/opt/telescope.nvim/lua/telescope/builtin/init.lua:559: attempt to call method 'extend' (a nil value)`

tami5:
thanks so much @Conni2461  please review

lua/telescope/actions/init.lua
```diff
@@ -1,4 +1,4 @@
--- Actions functions that are useful for people creating their own mappings.
+jjjjj-- Actions functions that are useful for people creating their own mappings.
```
Conni2461, line 1: :laughing:

lua/telescope/builtin/init.lua
```diff
@@ -556,9 +556,12 @@ builtin.builtin = function(opts)
...
       actions.goto_file_selection_edit:replace(actions.run_builtin)
-      return true
-    end
-  }):find()
+      actions.close:enhance{
+        post = function() vim.cmd [[startinsert]] end
+      }
+        return true
+      end
+    }):find()
```
Conni2461, line 564: indent doesn't look right. Apart from that. LGTM

lua/telescope/actions/init.lua
tami5: @Conni2461 hhhhhhhhhhhhhhhhh forgot that i deleted stopinsert and that happens :D

lua/telescope/builtin/init.lua
tami5: Nice thanks

Conni2461:
I love that action api. Does this new version fix the bug?

tami5:
Hmm not quite, it only works if I am in normal mode on I hit enter, while in insert mode and enter the picker goes back to normal mode @Conni2461 same issue

Conni2461:
We could also replace `actions.close()` and do what we actually wanna do. Just an idea. i haven't looked into that issue

tami5:
@Conni2461 Yeb but endup with duplication. I think I will look at it again tom.

tami5:
Fixed with #287

cempassi:
I found a way to go back in insert after closing/opening a builtin from lua with `vim.api.nvim_feedkeys('i', 'n', false)`. Maybe that can solve this problem for now?

tami5:
@cempassi it did, but apparently there is one more issue left with builtin runner picker. backspace on clear prompt brakes the sorter

tami5:
Wrong solution fixed by #300

tjdevries:
Hmm, I don't like having the bonus parameter. All `actions` have had exactly the same signature up until this point. I can try and take a look at this later.

tami5:
@tami5 yep me neither, couldn't find a solution that doesn't create undesired side effect. I'm interested to see how you endup solving it without introducing new param

tjdevries:
:+1:

Issue #233, opened 2020-11-11
TelescopeMatching resets after TelescopeSelection


OkanEsen:
The matching highlights reset after the selection is moved.

See: https://i.imgur.com/uYzUCmo.mp4

Shouldn't matter but these are the highlight groups:
``` vim
TelescopeMatching xxx gui=bold guifg=#83a598 guibg=none
TelescopeSelection xxx gui=bold guifg=#ebdbb2 guibg=#1d2021
```

rockerBOO:
This also clears the devicon colors too.

tjdevries:
Yeah, I'm not sure if I want to implement the fix here or if buffer highlights should be fixed in core to retain previous highlights when overridden. We will have to see.

tjdevries:
@bfredl do is there anything I can do to solve this problem:

1. Namespace 1 applies highlights on line X
2. Namespace 2 applies highlights on lin X
3. Clear namespace 2.
4. Now there are no highlights.

Conni2461:
Fixed in #355

PR #267, opened 2020-11-19
Add highlights builtin

Conni2461:
@sharksforarms @clason 

Please try it out thanks :)

Adds a preview window in which we can put any content which is specified in a `preview_command` function.
Is useful for `dap` extensions, so we can basically display any lua table in a preview window.

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -910,6 +910,30 @@ builtin.filetypes = function(opts)
   }):find()
 end
 
+builtin.highlights = function(opts)
+  opts = opts or {}
+
+  local highlights = vim.fn.getcompletion('', 'highlight')
+
+  pickers.new({}, {
+    prompt_title = 'Highlights',
+    finder = finders.new_table {
+      results = highlights,
+      entry_maker = make_entry.gen_from_highlights(opts)
+    },
+    sorter = conf.generic_sorter(),
+    attach_mappings = function(prompt_bufnr)
+      actions.goto_file_selection_edit:replace(function()
+        local selection = actions.get_selected_entry()
+        actions.close(prompt_bufnr)
+        vim.cmd('hi ' .. selection.value)
+      end)
+      return true
+    end,
+    previewer = previewers.display_content.new(opts),
+  }):find()
+end
+
 builtin.tags = function(opts)
   opts = opts or {}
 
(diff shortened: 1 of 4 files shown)


sharksforarms:
Tried it out, very nice! Would be nice to preview the actual color in the preview area, or on the results picker list

Conni2461:
The `xxx` is now highlighted with the specific highlighting group.
Is the action on enter fine? I don't really know what you had in mind here.

Note: Force push

sharksforarms:
Looks great! Thanks for adding that!

Would something like this be possible?

image

Conni2461:
The thing is when i create these results entries i don't have the `bufnr` so i could set a highlighting. I don't think this is possible without some "big" changes to the picker. But we should better ask @tjdevries.

sharksforarms:
Maybe a thought for the future :+1: this looks great for now!

lua/telescope/builtin/init.lua
```diff
@@ -888,6 +888,43 @@ builtin.keymaps = function(opts)
...
+    finder = finders.new_table {
+      results = highlights,
+      entry_maker = function(entry)
+        return {
+          value = entry,
+          ordinal = entry,
+          display = entry,
+          preview_command = function(bufnr)
+            vim.api.nvim_buf_set_option(bufnr, 'filetype', 'vim')
+            local output = vim.split(vim.fn.execute('hi ' .. entry), '\n')
```
tjdevries, line 907: you can probably use things like `nvim_get_hl_by_name` or similar to get some of this information.

tjdevries:
There are a couple places you can add hooks to do additional highlighting.
1. `sorter.highlight` -> this gets run after a sorter matches a row. Usually used to do the highlighting for matched chars (or approx matched chars)
2. `entry.display` -> This can generate highlights from the entry. I don't think I've handled `entry.display` generating highlights AND the `.create` stuff, but it might work. But anyway, any entry can generate its own highlights that it wants applied (this is now nvim-devicons gets highlighted, by the way)

tjdevries:
Oh, but you could just make the text be highlighted the way the color actually is. That would not be too hard. You'd just need to make a new entry maker for  the highlights.

lua/telescope/builtin/init.lua
Conni2461: I don't think this is useful for the previewer. This output will be 100% in the previewer and i think it is better if we present the user with something like this rather than something like
```lua
{
  foreground = 16711680,
  bold
}
```

Conni2461:
> Oh, but you could just make the text be highlighted the way the color actually is. That would not be too hard. You'd just need to make a new entry maker for the highlights.

Just looking at this. I think i see how it works thanks :)

Conni2461:
@sharksforarms Does this work for you? #233 will also apply here. But we got highlighting :smile: 

@tjdevries That api is super simple. Good to know that you programmed something amazing like this

sharksforarms:
Very cool :+1:

PR #268, opened 2020-11-20
Preview scrollbar

Conni2461:
Based on branch from @260.
Fancy new scrollbar based on the work from @sunjon. Thank you very much. Basically send all praise for this feature to him.
Sometimes kinda works good sometimes not.
Idea originally proposed by @tami5 [here](https://github.com/nvim-telescope/telescope.nvim/pull/260#issuecomment-730299139)

- [x] Fix scrollbar sometimes not going to the end of the file. (I think this is fixed)
- [x] Fix scrollbar window creating too early. So we don't have any legit result until we scroll with `<c-d>`, `<c-u>`.
- [x] Update scrollbar when changing buffer. (I think we can do all this without autocmd. But please correct me if we should use them)
- [ ] Maybe maybe smooth scrolling. Need to take a look at how people make this possible.

[asciicast](https://asciinema.org/a/374300)

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -4,6 +4,7 @@ local utils = require('telescope.utils')
 local putils = require('telescope.previewers.utils')
 local Previewer = require('telescope.previewers.previewer')
 local conf = require('telescope.config').values
+local scrollbar = require('telescope.previewers.scrollbar')
 
 local pfiletype = require('plenary.filetype')
 local pscan = require('plenary.scandir')
@@ -151,6 +152,9 @@ previewers.new_buffer_previewer = function(opts)
       opt_teardown(self)
     end
 
+    if self.state.sb then
+      self.state.sb:close()
+    end
     local last_nr
     if opts.keep_last_buf then
       last_nr = global_state.get_global_key('last_preview_bufnr')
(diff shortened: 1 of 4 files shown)


tjdevries:
I wonder if we could think of a way to make scrollbar become a telescope extension.... that is the kind of thing that will really make us think about how to make interfaces for extensions.

tami5:
> I wonder if we could think of a way to make scrollbar become a telescope extension.... that is the kind of thing that will really make us think about how to make interfaces for extensions.

😅 @tjdevries plz plz let this one slide 🙏☺️  + extension are for support for external plugins

tjdevries:
@Conni2461 Do you want to implement the scroll bar thingy in plenary instead? I think a bunch of libraries might like to have it and consume it within plenary's windows.

What do you think?

Conni2461:
Some more work here. tj mentioned this today, still one small issue so almost done.

Conni2461:
No more small issues. I think. We should add some configuration for this. People can start test this now :)

@tjdevries do you still want the scrollbar module in plenary?

Conni2461:
I think something like this should live in plenary in the new popup library that @l-kershaw plans. Maybe we can talk about this if that new popup interface is usable and integrated here. Maybe this can be used as inspiration maybe not (the math is probably a little bit off. i havent looked at this since forever) :laughing: 

I am closing this. I dont have the time right now to fix the remaining issue.

PR #271, opened 2020-11-21
Fixed minor typos

Xearty:


sunjon:
Now that I looked at it, I noticed a grammar error in the same line:
> "it is long"

Should be "it has long"

PR #272, opened 2020-11-22
docs: fix builtin table formatting

elianiva:
I noticed that the builtin table seems broken so I fixed it. I also added period to every items because I saw some items have it but some others don't and it triggers me :p


tami5:
lgtm, thanks @elianiva

PR #273, opened 2020-11-22
If wildignore is set to tags expand will return '' unless true is passed

Conni2461:
We could also move to a plenary job @tami5 
And fix ordinal. So don't merge
Thanks for the bug report @yujinyuz

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -938,13 +938,13 @@ builtin.tags = function(opts)
   opts = opts or {}
 
   local ctags_file = opts.ctags_file or 'tags'
+  local fd = vim.loop.fs_open(vim.fn.expand(ctags_file, true), "r", 438)
 
-  if not vim.loop.fs_open(vim.fn.expand(ctags_file), "r", 438) then
+  if fd == nil then
     print('Tags file does not exists. Create one with ctags -R')
     return
   end
 
-  local fd = assert(vim.loop.fs_open(vim.fn.expand(ctags_file), "r", 438))
   local stat = assert(vim.loop.fs_fstat(fd))
   local data = assert(vim.loop.fs_read(fd, stat.size, 0))
   assert(vim.loop.fs_close(fd))
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -641,7 +641,7 @@ function make_entry.gen_from_ctags(opts)
     return {
       valid = true,
 
-      ordinal = file .. ': ' .. tag,
+      ordinal = tag .. ' : ' .. scode,
       display = make_display,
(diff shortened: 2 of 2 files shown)


yujinyuz:
Looks good to me!

But maybe it should be ordered by the result and then their length? e.g. `Badge` should come first before `BadgeLabel`

or maybe I need to set it in my config?


Screen Shot 2020-11-22 at 6 15 31 PM

tami5:
@yujinyuz agree 
@Conni2461 I kinda agree with you on it becoming a more like a ctags plugin, but such an automation is important. Is it worth it create an extension for telescope that does that in addition to relocating the tags picker there "finally a test case for tj extension feature😂"

tami5:
@tjdevries we neeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeed gitter setup for telescope, our prs are becoming chatrooms

Conni2461:
I am not that familiar with what is going on in sorters. I haven't worked on that part :smile:.
I think we calculate a score for each entry and if the score is higher than we will rank it higher. It seems that `BadgeLabel` has a higher score because it also matches the `ge` from the file.

@tami5 You could work on a extension for that but i think we should keep tags here and than extend it in an extension. Also Tj already has two possible extension he can work with vimspector and i wrote a dap extension.
Tami i'm sooooooo sorry. I already have the jobs thing up and running but if you want some work to do. I am fine with not pushing and you can have that PR :smile: I can also walk you through

Conni2461:
@tami5 You actually don't have to do almost anything. Look at `lua/telescope/builtin/init.lua` and what tj does with `finders.new_oneshot_job`

tami5:
> @tami5 You actually don't have to do almost anything. Look at `lua/telescope/builtin/init.lua` and what tj does with `finders.new_oneshot_job`

I'm lost , 😂 you need to accomplish something with function?

Conni2461:
That was on sync. I just updated my last post. Tj did the heavy lifting with `new_oneshot_job`. You just have to read the file. So a fitting command would be cat. The plenary job will than add each element to the table when it is printed to std out.
You might wanna take a look at [jobs specs](https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/job.lua).
And don't forget to add the entry_maker to opts. So `new_oneshot_job` can use it. 
If that all works you will get an error pointing to `entry_display.lua` because we are not allowed to call any vimL in an async job so you wanna replace the `vim.fn` with the lua style to get the string length.
I am back in 90 mins. Sorry

tami5:
Sorry, still lost :$ on what is need to be accomplished 
FYI: invited you on gitter

Conni2461:
Closing in favor of #274

PR #274, opened 2020-11-22
Enhance: async ctags loading 

tami5:
- If wildignore is set to tags expand will return '' unless true is passed 
- async loading of ctags.

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -939,12 +939,13 @@ builtin.tags = function(opts)
 
   local ctags_file = opts.ctags_file or 'tags'
 
-  if not vim.loop.fs_open(vim.fn.expand(ctags_file), "r", 438) then
+  local fd = vim.loop.fs_open(vim.fn.expand(ctags_file, true), "r", 438)
+
+  if fd == nil then
     print('Tags file does not exists. Create one with ctags -R')
     return
   end
 
-  local fd = assert(vim.loop.fs_open(vim.fn.expand(ctags_file), "r", 438))
   local stat = assert(vim.loop.fs_fstat(fd))
   local data = assert(vim.loop.fs_read(fd, stat.size, 0))
   assert(vim.loop.fs_close(fd))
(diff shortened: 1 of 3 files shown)


lua/telescope/pickers/entry_display.lua
```diff
@@ -43,7 +43,7 @@ end
 
 local function truncate(str, len)
   -- TODO: This doesn't handle multi byte chars...
-  if vim.fn.strdisplaywidth(str) > len then
+  if #str > len then
```
Conni2461, line 46: Can can no longer do `strdisplaywidth` because we are not allowed to call `vimL function in a lua loop callback`, but `#str` is not the same. Is there another lua solution for this

lua/telescope/builtin/init.lua
```diff
@@ -936,27 +936,18 @@ end
...
-  local results = vim.split(data, '\n')
-
   pickers.new(opts,{
     prompt = 'Tags',
-    finder = finders.new_table {
-      results = results,
-      entry_maker = make_entry.gen_from_ctags(opts),
-    },
+    finder = finders.new_oneshot_job(
+      { 'cat', ctags_file },
```
Conni2461, line 948: We don't expand here maybe we should do `local ctags_file = vim.fn.expand(opts.ctags_file, true) or 'tags'` in line 940. Then we could just have on expand.

tami5: Done

Conni2461: @tami5 won't work. because `fn.expand` will return null rather than nil which does not work with or. Could you please figure out another way :smile: thanks

tami5: Oh you mean for the lua loop issue?

tami5: @Conni2461

tami5: Also @Conni2461 I noticed a lot of repo locate their tag file in doc/tag, i think we should add the additional step of checking ctags_file again

Conni2461: We cannot do this because if opts.ctags_file is not specified `expand(...)` will return `null` and not `nil` so we never get into the second condition. Try it out

tami5: Updated ` opts.ctags_file and opts.ctags_file:gsub("~", vim.loop.os_homedir()) or 'tags'`

lua/telescope/pickers/entry_display.lua
Conni2461: @tjdevries

tjdevries: vim.api.nvim_strwidth() ? I think that will work.

tami5: @tjdevries nope that didn't cut it for the lua loop. ... hmmmm

tami5:
@Conni2461 is this still relevant to you? I'm thinking of closing it

Conni2461:
Its picked up here: https://github.com/nvim-telescope/telescope.nvim/pull/288

PR #275, opened 2020-11-22
Register finder

sunjon:
Not super happy with the user experience so far.
Need nicer ways for editing, and more control over how you paste.

Putting the finder specific actions in `actions/init.lua` also feels strange, when everything else there is relatively generic.

Suggestions welcome

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -899,8 +899,8 @@ builtin.registers = function(opts)
     -- use levenshtein as n-gram doesn't support <2 char matches
     sorter = sorters.get_levenshtein_sorter(),
     attach_mappings = function(_, map)
+      map('i', '<CR>', actions.paste_register)
       map('i', '<C-e>', actions.edit_register)
-      map('i', '<C-p>', actions.paste_register)
 
       return true
     end,
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -899,9 +899,6 @@ builtin.registers = function(opts)
     -- use levenshtein as n-gram doesn't support <2 char matches
     sorter = sorters.get_levenshtein_sorter(),
     attach_mappings = function(_, map)
-      -- TODO: Find a way to insert the text... it seems hard.
-      -- map('i', '<C-i>', actions.insert_value, { expr = true })
-      map('i', '<CR>', actions.insert_value)
       map('i', '<C-e>', actions.edit_register)
       map('i', '<C-p>', actions.paste_register)
 
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -892,7 +892,6 @@ builtin.registers = function(opts)
     table.insert(registers_table, {kind = "Read-Only",    key = readonly[i]})
   end
 
-
   pickers.new(opts,{
     prompt_title = 'Registers',
     finder = finders.new_table {
@@ -900,7 +899,17 @@ builtin.registers = function(opts)
       entry_maker = make_entry.gen_from_registers(opts),
     },
     previewer = previewers.vimgrep.new(opts),
-    sorter = sorters.get_generic_fuzzy_sorter(),
+    -- use levenshtein as n-gram doesn't support <2 char matches
+    sorter = sorters.get_levenshtein_sorter(),
+    attach_mappings = function(_, map)
+      -- TODO: Find a way to insert the text... it seems hard.
+      -- map('i', '<C-i>', actions.insert_value, { expr = true })
+      map('i', '<CR>', actions.insert_value)
+      map('i', '<C-e>', actions.edit_register)
+      map('i', '<C-p>', actions.paste_register)
+
+      return true
+    end,
   }):find()
 end
 
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -862,7 +862,7 @@ end
 builtin.registers = function(opts)
   opts = opts or {}
 
-  local registers_table = {"\"", "_", "#", "=", "_", "/",-"*", "+", ":", ".", "%"}
+  local registers_table = {"\"", "_", "#", "=", "_", "/", "*", "+", ":", ".", "%"}
 
   -- named
   for i = 0, 9 do
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -862,17 +862,7 @@ end
 builtin.registers = function(opts)
   opts = opts or {}
 
-  local registers_table = {}
-  table.insert(registers_table, "\"")
-  table.insert(registers_table, "_")
-  table.insert(registers_table, "#")
-  table.insert(registers_table, "=")
-  table.insert(registers_table, "_")
-  table.insert(registers_table, "/")
-
-  -- selection
-  table.insert(registers_table, "*")
-  table.insert(registers_table, "+")
+  local registers_table = {"\"", "_", "#", "=", "_", "/",-"*", "+", ":", ".", "%"}
 
   -- named
   for i = 0, 9 do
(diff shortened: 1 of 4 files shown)


sunjon:
screenshot

lua/telescope/builtin/init.lua
```diff
@@ -859,6 +859,60 @@ builtin.marks = function(opts)
   }):find()
 end
 
+builtin.registers = function(opts)
+  opts = opts or {}
+
+  local registers_table = {}
+  table.insert(registers_table, {kind = "Unnamed",      key = "\""})
```
Conni2461, line 866: Do we need that kind entry? You are not doing anything with it

lua/telescope/builtin/init.lua
```diff
@@ -859,6 +859,60 @@ builtin.marks = function(opts)
...
+  table.insert(registers_table, {kind = "Alternate",    key = "#"})
+  table.insert(registers_table, {kind = "Expression",   key = "="})
+  table.insert(registers_table, {kind = "Black Hole",   key = "_"})
+  table.insert(registers_table, {kind = "Last Search",  key = "/"})
+
+  -- clipboaord
+  local selection = {"*", "+"}
+  for i in pairs(selection) do
+    table.insert(registers_table, {kind = "Selection",    key = selection[i]})
+  end
```
Conni2461, line 877: A for loop is here not necessary

lua/telescope/builtin/init.lua
```diff
@@ -859,6 +859,60 @@ builtin.marks = function(opts)
...
+    table.insert(registers_table, {kind = "Read-Only",    key = readonly[i]})
+  end
+
+  pickers.new(opts,{
+    prompt_title = 'Registers',
+    finder = finders.new_table {
+      results = registers_table,
+      entry_maker = make_entry.gen_from_registers(opts),
+    },
+    previewer = previewers.vimgrep.new(opts),
```
Conni2461, line 901: Does this work for you? I think you might have to remove that line and not enable a previewer at all

lua/telescope/builtin/init.lua
```diff
@@ -859,6 +859,60 @@ builtin.marks = function(opts)
...
+    -- use levenshtein as n-gram doesn't support <2 char matches
+    sorter = sorters.get_levenshtein_sorter(),
+    attach_mappings = function(_, map)
+      -- TODO: Find a way to insert the text... it seems hard.
+      -- map('i', '<C-i>', actions.insert_value, { expr = true })
+      map('i', '<CR>', actions.insert_value)
```
Conni2461, line 907: You should use `actions.goto_file_selection_edit:replace(actions.insert_value)`. That is a cool new api done by tj. In that way we respect user mappings. Take a look at what other `builtins` do with `replace`

lua/telescope/make_entry.lua
```diff
@@ -453,6 +453,34 @@ function make_entry.gen_from_marks(_)
...
+  end
+
+  return function(entry)
+    local reg_content = vim.fn.getreg(entry.key)
+    return {
+      value = entry.key,
+      ordinal = entry.key,
+      content = reg_content,
+      regtype = vim.fn.getregtype(entry.key),
+      display = make_display(entry, reg_content),
```
Conni2461, line 479: You should not all make_entry. The picker does that on demand which is better for performance. The `entry` which you will get would be yourself. So in make entry `register` becomes `entry` and `content` would be `entry.content`.

lua/telescope/builtin/init.lua
```diff
@@ -788,7 +788,7 @@ builtin.man_pages = function(opts)
   end
 
   pickers.new(opts, {
-    prompt_tile = 'Man',
+    prompt_title = 'Man',
```
Conni2461, line 791: Nice catch :+1:

Conni2461:
@sunjon Great to see you working on another PR :+1:. Looks really good. I just left some nitpicking comments but besides that i am exited for that picker :smile:

lua/telescope/builtin/init.lua
sunjon: Not needed. I'd pondered if it had any use in the display, then forgot to remove it.

lua/telescope/builtin/init.lua
sunjon: There's currently no previewer at all. I'd disabled it in my `require` call.

For most 90% of entries, the preview is going to be the same one-line as the register content (In my workflow anyway), the latter 10% you may well want to see the entire content.

The only other (silly) idea  I had, was to copy the original buffer, and preview what the buffer would look like _after_ the paste/action.

I think it will  just be preview-less.

lua/telescope/builtin/init.lua
sunjon: I'll look into this  and update tomrrow.

lua/telescope/builtin/init.lua
Conni2461: Yeah i agree that we don't need a previewer. I just thought this previewer might be broken because it expects a `filename` entry.
I just tested it and i got just a bat error. So i suggest removing that line, so people with a previewer configured will not be confused about that error.

lua/telescope/make_entry.lua
Conni2461: Reading over this makes no sense. What i mean is we don't call `make_display` when we return this table. What you actually want to do is just do `display = make_display` and your `make_display` function signature would be just `(entry)`, which is the table, that we return. `key` would be `entry.value` and `content` would be `entry.content`

lua/telescope/builtin/init.lua
sunjon: To be honestr, that mapping wasn't used and I'm not sure how `:replace` fits in with the other mappings. If it's still applicable, let me know.

sunjon:
Perhaps not for this itteration, but I think a new sorter would suit these results better.
Currently, if my query is `A`, the fixed order of the list remains, but the "A" entry jumps to the top. What I think would be better, is if the results "rotated" to the match, ie. If  "3" is the top entry, "4", "5","6", etc would follow.

sunjon:
Remaining issues (not necessarily blocking merge):

1. Hide registers with no content?
    I'd have done this, but if [this issue](https://github.com/nvim-telescope/telescope.nvim/issues/276) is resolved, then a user might want to use the edit functionality to put something into an empty register.

2. Custom ordering. 
    It may be that a user is only interested in the `[0-9]` registers.  Perhaps configration to apply a "hidden_registers = { ...}` filter.

lua/telescope/builtin/init.lua
Conni2461: The thing about `actions.goto_file_selection_edit:replace` is that we basically replacing the users main action.
So if a user doesn't like `<CR>` as mapping and likes to use idk `<space>` he will hit `<space>` and we would call `actions.goto_...`. With that replace we basically just follow what the user expects for all individual pickers, even though they do something differently like checking out a branch or stuff. Does that makes sense.

lua/telescope/builtin/init.lua
```diff
@@ -859,6 +859,54 @@ builtin.marks = function(opts)
...
+  table.insert(registers_table, "\"")
+  table.insert(registers_table, "_")
+  table.insert(registers_table, "#")
+  table.insert(registers_table, "=")
+  table.insert(registers_table, "_")
+  table.insert(registers_table, "/")
+
+  -- selection
+  table.insert(registers_table, "*")
+  table.insert(registers_table, "+")
```
Conni2461, line 875: Oh we could just do `local registers_table = { "\"", "_", "#" ...}`. That would save us a lot of inserts. I am not talking about the for loops btw. These are good.

sunjon: done

Conni2461:
>     1. Default action. Should the default action be put()?

Yes i think the default should be pasting the content inside the file at the current cursor pos.

>     3\. It may be that a user is only interested in the `[0-9]` registers.  Perhaps configration to apply a "hidden_registers = { ...}` filter.

We can also just say, that a user can define a table of registers in which he is interested.

Besides my remaining comments. I think this LGTM and when you done i think you can squash and merge.
Oh and could you also add the new builtin to the readme builtin table. That everyone sees that we have that cool new picker. Thanks

tami5:
well done @sunjon awesome work

PR #279, opened 2020-11-23
Filter the completion of the command

tamago324:
I like it better when the first letter filters out the candidates, don't you?

*before*:
before

*after*:
after


tami5:
Oh thats what I remember talking about with glepnir . That useful thanks
cc @glepnir

tjdevries:
Ah, yes, this is the correct way to do it. Thanks.

tamago324:
Thank you for merging!

glepnir:
yep

PR #282, opened 2020-11-24
Set and unset report to what's needed when deleting buffers.

### LINKED ISSUE #105: 1 buffer deleted messages when closing telescope
Hey there, I'm ready to switch to telescope and I think its awesome. I have one issue though, everytime I close the buffer I get few lines of 1 buffer deleted. Any way to omit this messages, thanks

Julian:
Fixes #105.

This does the "blunt" seeming just set-and-unset report whilst deleting buffers, which seems to fix the issue for me (in that no message for buffer deletion is shown).

Lemme know if this makes sense to y'all or if there's some obvious issue with it.

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -141,9 +141,15 @@ end
 function utils.buf_delete(bufnr)
   if bufnr == nil then return end
 
+  -- Suppress the buffer deleted message for those with &report<2
+  start_report = vim.o.report
+  if start_report < 2 then vim.o.report = 2 end
+
   if vim.api.nvim_buf_is_valid(bufnr) and vim.api.nvim_buf_is_loaded(bufnr) then
     vim.cmd(string.format("bdelete! %s", bufnr))
   end
+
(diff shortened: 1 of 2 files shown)


lua/telescope/utils.lua
```diff
@@ -141,9 +141,14 @@ end
 function utils.buf_delete(bufnr)
   if bufnr == nil then return end
 
+  -- Suppress the buffer deleted message for those with &report<2
+  start_report, vim.o.report = vim.o.report, 2
```
Julian, line 145: It seems like what I see is `plenary`'s `context_manager` would be useful here but I don't see one implemented for setting-and-then-resetting a vim option. Maybe that's generically useful to add there?

tjdevries: Can we only set it if it's not a good value?

And yeah, it could be used with the context manager. It's not a bad idea at all.

Julian: Done.

tjdevries:
Thanks, this is better than just erroring :)

PR #283, opened 2020-11-24
add back filetypedetect on edit

mhartington:
Similar to https://github.com/nvim-telescope/telescope.nvim/pull/116 and https://github.com/nvim-telescope/telescope.nvim/issues/115

It seems that this was removed in some refactoring. Without this, any autocmds for filetypes will not run.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -113,6 +113,7 @@ function actions._goto_file_selection(prompt_bufnr, command)
         vim.cmd(string.format(":%s %s", command, filename))
         bufnr = vim.api.nvim_get_current_buf()
         a.nvim_buf_set_option(bufnr, "buflisted", true)
+        vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))
       end
 
       if row and col then
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -113,7 +113,6 @@ function actions._goto_file_selection(prompt_bufnr, command)
         vim.cmd(string.format(":%s %s", command, filename))
         bufnr = vim.api.nvim_get_current_buf()
         a.nvim_buf_set_option(bufnr, "buflisted", true)
-        vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))
       end
 
       if row and col then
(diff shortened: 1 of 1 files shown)


tami5:
```lua
vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))
```

@sunjon I'm trying to find out if this is moved somewhere else or removed to fix another issue before merging, I've mentioned you because from git history, the code block was last edited by so I'm not sure.

sunjon:
If it's vanished in one of my commits, then it's happened unintentionally in refactoring/merging and not as an intend fix.

Conni2461:
@sunjon don't worry it vanished [here](https://github.com/nvim-telescope/telescope.nvim/commit/2ce3ad61aa9342fff4d8be69fc9e019e393bf161).

lua/telescope/actions/init.lua
```diff
@@ -113,6 +113,7 @@ function actions._goto_file_selection(prompt_bufnr, command)
         vim.cmd(string.format(":%s %s", command, filename))
         bufnr = vim.api.nvim_get_current_buf()
         a.nvim_buf_set_option(bufnr, "buflisted", true)
+        vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))
```
Conni2461, line 116: Move down so `buffer`/`sbuffer` is also effected

Conni2461 (changes requested):

mhartington:
Hmm, actually, this doesn't seem to solve it fully.  The filetype is getting set, but even with the filetypedetect in place, folds still are not being processed.

Conni2461:
It looks good to me. It just does not work for me and i don't know why. I will take a closer look at this as soon as possible.

lua/telescope/actions/init.lua
```diff
@@ -122,6 +122,7 @@ function actions._goto_file_selection(prompt_bufnr, command)
         end
       end
     end
+    vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))
```
tjdevries, line 125: This should probably not be here, but in the `filename` part of the else block above. `bufnr` means we already have a buffer for it I think (which means we would call this stuff again, which would be weird I think)

tjdevries (changes requested):
Small nit, then I'll merge.

lua/telescope/actions/init.lua
Conni2461: What about unloaded buffers? If we have a old session? With `show_all_buffers` we will still display them and are able to change to them

tjdevries:
OK, works for me. Maybe we can fix the stuff in core to actually make this work without weird hax.

PR #284, opened 2020-11-24
Fix the bug report template to suggest -u

Julian:
I *assume* this is what the intention here is? I.e. to say to use the test.vim as the config.

I don't think `-N` does anything either, but let me know if I'm wrong.


tami5:
You right it doesn't @Julian thanks for catching that, can you add `n` in place of `N`. 

`n` to have no swap files

Julian:
Done.

PR #286, opened 2020-11-24
Use noremap when mapping.

Julian:
Previously, if e.g. `:` was remapped, Telescope would be unclosable with its mappings (`<Esc>`).

### DIFF
--- a/lua/telescope/mappings.lua
+++ b/lua/telescope/mappings.lua
@@ -94,9 +94,9 @@ local telescope_map = function(prompt_bufnr, mode, key_bind, key_func, opts)
     return
   end
 
-  opts = opts or {
-    silent = true
-  }
+  opts = opts or {}
+  if opts.noremap == nil then opts.noremap = true end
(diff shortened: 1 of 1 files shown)


lua/telescope/mappings.lua
```diff
@@ -97,6 +97,7 @@ local telescope_map = function(prompt_bufnr, mode, key_bind, key_func, opts)
   opts = opts or {
     silent = true
   }
+  if opts.noremap == nil then opts.noremap = true end
```
tjdevries, line 100: we should do this check for both:

```lua
opts = opts or {}
if opts.silent == nil then opts.silent = true end
if opts.noremap == nil then opts.noremap = true end
```

Something like that.

Then we can merge, looks like you were right about that!

Julian: Done, thanks.

tjdevries:
Cool, thanks for the fix. Good call.

Issue #277, opened 2020-11-23
Fixup/Enhancement: We should make overriding default pickers work more consistently


tjdevries:
We should make it easier to override the default sorters, since not every builtin currently respects the config flags.

tami5:
Our leader is back :smile: , yes I agree @tjdevries

tjdevries:
Outline (for someone else to do):
- make all `sorter = ...` for the pickers take some function
- function that takes the `opts`, checks if `opts.sorter` is in keys, uses that if so
- otherwise, uses the `file_sorter`, `generic_sorter` or whatever type of sorter as specified.

We should add more documentation about this somewhere as well.

nc1337:
Can't we just reuse `utils.get_default` from `utils.lua`? I am noob in lua so please don't mind it this is the wrong way?

a420gmoon:
where from I will start Enhancement default pickers work more consistently.

PR #287, opened 2020-11-24
Refactor builtin into smaller modules

Conni2461:
Should be good but someone should take a look at this so i don't wake up with a lot of angry messages. New structure:
```
builtin.live_grep = require('telescope.builtin.files').live_grep
builtin.grep_string = require('telescope.builtin.files').grep_string
builtin.find_files = require('telescope.builtin.files').find_files
builtin.fd = builtin.find_files
builtin.treesitter = require('telescope.builtin.files').treesitter
builtin.current_buffer_fuzzy_find = require('telescope.builtin.files').current_buffer_fuzzy_find
builtin.tags = require('telescope.builtin.files').tags
builtin.current_buffer_tags = require('telescope.builtin.files').current_buffer_tags

builtin.git_files = require('telescope.builtin.git').files
builtin.git_commits = require('telescope.builtin.git').commits
builtin.git_bcommits = require('telescope.builtin.git').bcommits
builtin.git_branches = require('telescope.builtin.git').branches
builtin.git_status = require('telescope.builtin.git').status

builtin.builtin = require('telescope.builtin.internal').builtin
builtin.planets = require('telescope.builtin.internal').planets
builtin.commands = require('telescope.builtin.internal').commands
builtin.quickfix = require('telescope.builtin.internal').quickfix
builtin.loclist = require('telescope.builtin.internal').loclist
builtin.oldfiles = require('telescope.builtin.internal').oldfiles
builtin.command_history = require('telescope.builtin.internal').command_history
builtin.vim_options = require('telescope.builtin.internal').vim_options
builtin.help_tags = require('telescope.builtin.internal').help_tags
builtin.man_pages = require('telescope.builtin.internal').man_pages
builtin.reloader = require('telescope.builtin.internal').reloader
builtin.buffers = require('telescope.builtin.internal').buffers
builtin.colorscheme = require('telescope.builtin.internal').colorscheme
builtin.marks = require('telescope.builtin.internal').marks
builtin.registers = require('telescope.builtin.internal').registers
builtin.keymaps = require('telescope.builtin.internal').keymaps
builtin.filetypes = require('telescope.builtin.internal').filetypes
builtin.highlights = require('telescope.builtin.internal').highlights

builtin.lsp_references = require('telescope.builtin.lsp').references
builtin.lsp_document_symbols = require('telescope.builtin.lsp').document_symbols
builtin.lsp_code_actions = require('telescope.builtin.lsp').code_actions
builtin.lsp_workspace_symbols = require('telescope.builtin.lsp').workspace_symbols
```


Conni2461:
Oh btw @tjdevries I made sure that the sorter is always loaded with `conf.` So this would partially fix #277. (Everything besides `registers` because we only grep for one char there.)

tami5:
Oh wow, may I suggest an alternative for that, also how would it be possible to push to this pr @Conni2461 ?

Conni2461:
@tjdevries Need your help for a sec with one change. Because we do `apply_checks(mod)` for all builtin modules now and overwrite all functions `mod[k]`, the previewer of builtin.builtin will no longer work correctly. The line number will always be where `mod[k] = function...` happens. Do you know what i mean?
I thought maybe some metatable magic could solve this but i haven't figured it out yet. Also we could just get the line number with `vim.fn.search` or treesitter (but i don't like requiring treesitter for that)

tjdevries:
Ah, I hadn't thought of that part of it.

One thing we could do is we could just save that as some metadata with the functions. But I think that's a bit hacky and perhaps error prone.

`vim.fn.search` does not seem like a bad plan, I will have to think of it a bit more though.  Perhaps a layer of indirection between the two could make it work, but I'm not 100% sure yet. We can always do something different for the previewer as well.

For example, something cool would be we could write some documentation for the functions or something and then show the corresponding help.... I should really work on my doc generator :)

Conni2461:
I hadn't thought of that either. I would like to merge this today, so that we can keep the number of people who need to rebase as small as possible.
So either we merge without a fix, i will create an issue and someone fixes it later, or i will write a new preview using `vim.fn.search()`. Up to you.

Oh and yeah i have seen some of that doc generator based on treesitter, looks very promising and really helpful.

Conni2461:
@tjdevries Fixed the builtin previewer. Uses vim buffer and `vim.fn.search()`. We might have to clean up previewers and enable vim buffers for more pickers (as opt in). Might do that next.

@tami5 Could you test it for a second. Thanks.

tjdevries:
If it works, then you can merge. Don't have time to test today really. We can always poke around the previewer later. I agree should not keep this branch open for long.

tami5:
Oh wow, it works really well. You've also fixed a lot of issues we had previously with number of pickers. LGTM, let me know if there some final touches you need to get done before I merge. Thanks

PR #288, opened 2020-11-24
Add support for &tags (via tagfiles())

Julian:
I.e., search the locations specified for possible locations of tag files,
stopping on the first found tag file.

C.f. e.g. https://github.com/Shougo/denite.nvim/blob/0639e67a33df9f62065889a2feda8c5648c633b3/rplugin/python3/denite/source/tag.py#L98-L102

This isn't ready to merge yet though, or if it is there's a follow-up issue to fix, which is paths aren't resolved relative to the tag file location, so my tag files (which contain e.g. `../foo/bar`) don't get that relative path resolved relative to where the found tagfile lives (they're instead resolved relative to cwd).

I think I see where to fix that as well, but putting this up for review because I'm sure there may be some even more idiomatic lua way to do this.

EDIT: also... I see that tagfiles() contains all found tag files, so I shouldn't be breaking on the first one, I guess you get to union all the tag files... So yeah this really isn't correct yet.

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -251,7 +251,9 @@ files.tags = function(opts)
     return
   end
 
-  local results = {}
+  opts.entry_maker = opts.entry_maker or make_entry.gen_from_ctags(opts)
+  local finder = finders.new_incremental(opts)
+
   for _, each in ipairs(tagfiles) do
     local tags_directory = vim.fn.fnamemodify(each, ":h")
     local resolve_filename = function(filename)
@@ -260,22 +262,15 @@ files.tags = function(opts)
     end
     path.read(each, function(err, data)
       for _, line in ipairs(vim.split(data, '\n')) do
-        table.insert(results, { line = line, resolve_filename = resolve_filename })
+        finder:feed({ line = line, resolve_filename = resolve_filename })
       end
     end)
+    finder:finish()
   end
 
   pickers.new(opts,{
     prompt = 'Tags',
-    finder = finders.new_job(function(prompt)
-      print(prompt)
-      if not prompt or prompt == "" then
-        return nil
-      end
-      return prompt
-    end,
-    opts.entry_maker or make_entry.gen_from_ctags(opts),
-    opts.max_results),
+    finder = finder,
     previewer = previewers.ctags.new(opts),
     sorter = conf.generic_sorter(opts),
     attach_mappings = function()
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -254,6 +254,8 @@ files.tags = function(opts)
   opts.entry_maker = opts.entry_maker or make_entry.gen_from_ctags(opts)
   local finder = finders.new_incremental(opts)
 
+  local remaining = #tagfiles
+
   for _, each in ipairs(tagfiles) do
     local tags_directory = vim.fn.fnamemodify(each, ":h")
     local resolve_filename = function(filename)
@@ -264,8 +266,8 @@ files.tags = function(opts)
       for _, line in ipairs(vim.split(data, '\n')) do
         finder:feed({ line = line, resolve_filename = resolve_filename })
       end
+      if remaining == 1 then finder:finish() else remaining = remaining - 1 end
     end)
-    finder:finish()
   end
 
   pickers.new(opts,{
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -262,7 +262,7 @@ files.tags = function(opts)
         return path.exists(filename) and filename or
                path.join(tags_directory, filename)
     end
-    path.read(each, function(err, data)
+    path.read_file_async(each, function(data)
       for _, line in ipairs(vim.split(data, '\n')) do
         finder:feed({ line = line, resolve_filename = resolve_filename })
       end
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -242,19 +242,24 @@ files.current_buffer_fuzzy_find = function(opts)
 end
 
 files.tags = function(opts)
-  local ctags_file = opts.ctags_file or 'tags'
-
-  if not vim.loop.fs_open(vim.fn.expand(ctags_file, true), "r", 438) then
-    print('Tags file does not exists. Create one with ctags -R')
-    return
+  local tagfiles
+  if opts.ctags_file then
+    tagfiles = {opts.ctags_file}
+  else
+    tagfiles = vim.fn.map(vim.fn.tagfiles(), 'expand(fnamemodify(v:val, ":p"))')
+    if vim.tbl_isempty(tagfiles) then
+      print('No tags file found. Create one with ctags -R or verify your &tagfiles option is correctly set.')
+      return
+    end
   end
 
-  local fd = assert(vim.loop.fs_open(vim.fn.expand(ctags_file, true), "r", 438))
-  local stat = assert(vim.loop.fs_fstat(fd))
-  local data = assert(vim.loop.fs_read(fd, stat.size, 0))
-  assert(vim.loop.fs_close(fd))
-
-  local results = vim.split(data, '\n')
+  local results = {}
+  for _, each in ipairs(tagfiles) do
+    tags_directory = vim.fn.fnamemodify(each, ":h")
+    for line in io.lines(each) do
+      table.insert(results, { line = line, tags_directory = tags_directory })
+    end
+  end
 
   pickers.new(opts,{
     prompt = 'Tags',
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -1,6 +1,7 @@
 local actions = require('telescope.actions')
 local finders = require('telescope.finders')
 local make_entry = require('telescope.make_entry')
+local path = require('telescope.path')
 local pickers = require('telescope.pickers')
 local previewers = require('telescope.previewers')
 local utils = require('telescope.utils')
@@ -242,22 +243,23 @@ files.current_buffer_fuzzy_find = function(opts)
 end
 
 files.tags = function(opts)
-  local tagfiles
-  if opts.ctags_file then
-    tagfiles = {opts.ctags_file}
-  else
-    tagfiles = vim.fn.map(vim.fn.tagfiles(), 'expand(fnamemodify(v:val, ":p"))')
-    if vim.tbl_isempty(tagfiles) then
-      print('No tags file found. Create one with ctags -R or verify your &tagfiles option is correctly set.')
-      return
-    end
+  local tagfiles = opts.ctags_file and { opts.ctags_file } or
+                   vim.fn.map(vim.fn.tagfiles(), 'expand(fnamemodify(v:val, ":p"))')
+
+  if vim.tbl_isempty(tagfiles) then
+    print('No tags file found. Create one with ctags -R or verify your &tags option is correctly set.')
+    return
   end
 
   local results = {}
   for _, each in ipairs(tagfiles) do
-    tags_directory = vim.fn.fnamemodify(each, ":h")
+    local tags_directory = vim.fn.fnamemodify(each, ":h")
+    local resolve_filename = function(filename)
+        return path.exists(filename) and filename or
+               path.join(tags_directory, filename)
+    end
     for line in io.lines(each) do
-      table.insert(results, { line = line, tags_directory = tags_directory })
+      table.insert(results, { line = line, resolve_filename = resolve_filename })
     end
   end
 
@@ -271,6 +273,10 @@ files.tags = function(opts)
     sorter = conf.generic_sorter(opts),
     attach_mappings = function()
       actions._goto_file_selection:enhance {
+        pre = function()
+          local entry = actions.get_selected_entry()
+          entry.filename = entry.resolve_filename(entry.filename)
+        end,
         post = function()
           local selection = actions.get_selected_entry()
 
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -258,17 +258,24 @@ files.tags = function(opts)
         return path.exists(filename) and filename or
                path.join(tags_directory, filename)
     end
-    for line in io.lines(each) do
-      table.insert(results, { line = line, resolve_filename = resolve_filename })
-    end
+    path.read(each, function(err, data)
+      for _, line in ipairs(vim.split(data, '\n')) do
+        table.insert(results, { line = line, resolve_filename = resolve_filename })
+      end
+    end)
   end
 
   pickers.new(opts,{
     prompt = 'Tags',
-    finder = finders.new_table {
-      results = results,
-      entry_maker = opts.entry_maker or make_entry.gen_from_ctags(opts),
-    },
+    finder = finders.new_job(function(prompt)
+      print(prompt)
+      if not prompt or prompt == "" then
+        return nil
+      end
+      return prompt
+    end,
+    opts.entry_maker or make_entry.gen_from_ctags(opts),
+    opts.max_results),
     previewer = previewers.ctags.new(opts),
     sorter = conf.generic_sorter(opts),
     attach_mappings = function()
--- a/lua/telescope/finders.lua
+++ b/lua/telescope/finders.lua
@@ -28,10 +28,10 @@ function IncrementalFinder:new(opts)
 
   obj._find = coroutine.wrap(function(finder, _, process_result, process_complete)
     repeat
-      finder, _, process_result, process_complete = coroutine.yield()
-      for index, result in ipairs(self.results) do
+      for _, result in ipairs(self.results) do
         process_result(finder.entry_maker(result))
       end
+      finder, _, process_result, process_complete = coroutine.yield()
     until self.completed
 
     process_complete()
--- a/lua/telescope/finders.lua
+++ b/lua/telescope/finders.lua
@@ -15,6 +15,40 @@ local _callable_obj = function()
 end
 
 
+local IncrementalFinder = _callable_obj()
+
+function IncrementalFinder:new(opts)
+  opts = opts or {}
+  local obj = setmetatable({
+    entry_maker = opts.entry_maker or make_entry.from_string,
+  }, self)
+
+  self.results = {}
+  self.finished = false
+
+  obj._find = coroutine.wrap(function(finder, _, process_result, process_complete)
+    repeat
+      finder, _, process_result, process_complete = coroutine.yield()
+      for index, result in ipairs(self.results) do
+        process_result(finder.entry_maker(result))
+      end
+    until self.completed
+
+    process_complete()
+  end)
+
+  return obj
+end
+
+function IncrementalFinder:feed(result)
+  table.insert(self.results, result)
+end
+
+function IncrementalFinder:finish(result)
+  self.finished = true
+end
+
+
 --[[ =============================================================
 
     JobFinder
@@ -272,6 +306,11 @@ finders._new = function(opts)
   end
 
   return JobFinder:new(opts)
+end
+
+finders.new_incremental = function(opts)
+  opts = opts or {}
+  return IncrementalFinder:new(opts)
 end
 
 finders.new_job = function(command_generator, entry_maker, maximum_results, cwd)
--- a/lua/telescope/finders.lua
+++ b/lua/telescope/finders.lua
@@ -32,7 +32,7 @@ function IncrementalFinder:new(opts)
         process_result(finder.entry_maker(result))
       end
       finder, _, process_result, process_complete = coroutine.yield()
-    until self.completed
+    until self.finished
 
     process_complete()
   end)
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -828,17 +828,21 @@ function make_entry.gen_from_ctags(opts)
     }
   end
 
-  return function(line)
-    if line == '' or line:sub(1, 1) == '!' then
+  return function(result)
+    if result.line == '' or result.line:sub(1, 1) == '!' then
       return nil
     end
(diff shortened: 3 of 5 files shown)


Julian:
OK this is potentially still deficient but is probably ready for first review (don't tell my mom I NIH'ed random path functions!? because Lua... let me know if those exist in some accessible place because I didn't find them).

lua/telescope/builtin/files.lua
```diff
@@ -186,19 +186,24 @@ files.current_buffer_fuzzy_find = function(opts)
...
-  local ctags_file = opts.ctags_file or 'tags'
-
-  if not vim.loop.fs_open(vim.fn.expand(ctags_file), "r", 438) then
-    print('Tags file does not exists. Create one with ctags -R')
-    return
+  local tagfiles
+  if opts.ctags_file then
+    tagfiles = {opts.ctags_file}
+  else
+    tagfiles = vim.fn.map(vim.fn.tagfiles(), 'fnamemodify(v:val, ":p")')
```
Conni2461, line 193:
```lua
local tagfiles = opts.ctags_file 
                        and { opts.ctags_file } 
                        or vim.map....
```

lua/telescope/builtin/files.lua
```diff
@@ -186,19 +186,24 @@ files.current_buffer_fuzzy_find = function(opts)
...
-  local fd = assert(vim.loop.fs_open(vim.fn.expand(ctags_file), "r", 438))
-  local stat = assert(vim.loop.fs_fstat(fd))
-  local data = assert(vim.loop.fs_read(fd, stat.size, 0))
-  assert(vim.loop.fs_close(fd))
-
-  local results = vim.split(data, '\n')
+  local results = {}
+  for _, each in ipairs(tagfiles) do
+    tags_directory = vim.fn.fnamemodify(each, ":h")
+    for line in io.lines(each) do
```
Conni2461, line 203: Could we not use `io.lines`? Its super slow compared to `vim.loop.fs_`
I was think about either this
```lua
local fd = vim.loop.fs_open(path, "r", 438)
local stat = vim.loop.fs_fstat(fd)
for i = 0, stat.size do -- might be a off by one issue. Not sure right now. Writing here in github
  local line = assert(vim.loop.fs_read(fd, 1, i))
end
vim.loop.fs_close(fd)
```
Or we could do it async with finder.new_oneshot_job like in #274. As command you could do something like this:
`flatten { 'grep', "", tagfiles }` that would give use 'filename:line' in `entry_maker` as input. So you might have to change the regex to get the tagfile_name. 
Benefit: Picker would be there asap and lines will come after one by one.

lua/telescope/make_entry.lua
```diff
@@ -655,12 +655,16 @@ function make_entry.gen_from_ctags(opts)
...
+    if result.line == '' or result.line:sub(1, 1) == '!' then
       return nil
     end
 
-    local tag, file, scode = string.match(line, '([^\t]+)\t([^\t]+)\t/^\t?(.*)/;"\t+.*')
+    local tag, file, scode = string.match(result.line, '([^\t]+)\t([^\t]+)\t/^\t?(.*)/;"\t+.*')
+
+    if not path.exists(file) then
+      file = path.join(result.tags_directory, file)
+    end
```
Conni2461, line 667: We should not do this for every line. Hitting the filesystem for every entry.
We just have to do this when displaying the preview and when opening a entry.
So you wanna add a `pre = function()` on action which will update the filename. (Pre will run first then `_goto_file_selection` and then post, which does the jump). Take a look at `attach_mappings` in `lua/telescope/builtin/files.lua:216-218`, also [docs](https://github.com/nvim-telescope/telescope.nvim#mappings). Third code snippet in this section.
For the previewer you wanna take a look [here](https://github.com/nvim-telescope/telescope.nvim/blob/4a8ea7763ec3c1c0672398418da682c1e0d4017d/lua/telescope/previewers.lua#L459-L483).

Also you have to carry the `tags_directory` with you so you wanna `return` it in this function as well

lua/telescope/builtin/files.lua
Julian: I can do the whole `if` block and return from the function from within the right side of the `or` there?

lua/telescope/path.lua
```diff
@@ -4,6 +4,15 @@ local path = {}
...
 path.home = vim.fn.expand("~")
 
+path.exists = function(filepath)
+  return vim.loop.fs_stat(filepath) ~= nil
+end
+
+-- TODO: This is obviously super naive and shouldn't exist in a random library
+path.join = function(filepath, child)
+  return filepath .. path.separator .. child
+end
```
Conni2461, line 14: Currently we just do this in code. There is a WIP path module in `plenary.nvim` which is the library which we rely on.

lua/telescope/builtin/files.lua
Julian: I figured there was a reason the code didn't use `io.lines` already but that someone'd tell me what it was if I used it :).

Will have a look at #274.

Julian: What's the crazy magic 438 that appears all over the place?

lua/telescope/path.lua
```diff
@@ -4,6 +4,15 @@ local path = {}
 path.separator = package.config:sub(1, 1)
 path.home = vim.fn.expand("~")
 
+path.exists = function(filepath)
+  return vim.loop.fs_stat(filepath) ~= nil
+end
```
Conni2461, line 9: I think using `vim.loop.fs_stat` is not optimal because it will not only check if the file exists but collect almost everything about the file and returns it as a table.

lua/telescope/builtin/files.lua
Conni2461: It's the mode. There are also some `luv.constants` which can be used for that. But i don't know i just used what they documented.
Take a look at [luv docs](https://github.com/luvit/luv/blob/master/docs.md#file-system-operations) which is `vim.loop` for more information. I am about to go to sleep. So good luck and have fun fixing my all my nit pickings :)
Also thanks for the PR. I didn't know about `&tagfiles` when i implemented this. I basically just took fzf as inspiration. So i appreciate you fixing my stuff.

lua/telescope/builtin/files.lua
Conni2461: No you do that or after. `condition and a or b` is basically just the ternary operator for lua.

Julian: OK, yeah I saw that that's how lua spells ternary -- that has slightly different behavior then than what's here but guess the minor extra check is fine then.

lua/telescope/builtin/files.lua
Julian: Oh right decimal-octal, fun.

And yeah all good! Thanks for the comments.

Conni2461: Oh i just realized. When you do it with `grep` you should ues `-H` option, so the filename will always be displayed. Even for just one file.

lua/telescope/path.lua
Julian: If `fs_stat` is calling `stat` that's what e.g. `os.path.exists` does in Python -- if you have a better suggestion though obviously I'm all ears.

Conni2461: `fs_stat` just gathers and returns all [stats](https://github.com/luvit/luv/blob/master/docs.md#uvfs_statpath-callback), which i feel like is not optimal but it should really matter if we do that only on `_goto_file_selection`

Julian: That says it's just calling `stat(2)`, so yeah at least as far as I know, that's as efficient as you can be.

Of course yeah creating all the Lua objects is useless work so you can FFI things to avoid that I guess but I doubt it matters much.

lua/telescope/builtin/files.lua
```diff
@@ -210,6 +217,10 @@ files.tags = function(opts)
     sorter = conf.generic_sorter(opts),
     attach_mappings = function()
       actions._goto_file_selection:enhance {
+        pre = function()
+          local entry = actions.get_selected_entry()
+          entry.filename = entry.resolve_filename(entry.filename)
```
Julian, line 222: Not sure if there was a non-mutate-the-world way of doing this or if this was what you were suggesting but yeah let me know.

>

Conni2461: That looks good. Does it work as well?

Julian: Indeed.

Julian: (I also don't 100% recall how method syntax works in lua, I'm sure there's a way to write that as `entry:resolve_filename` right? but my attention span is quickly waning since there are 10 other things I need to fix after this, so yeah this was as good as I came up with for now)

Conni2461: No it looks good. `:` is used if we actually do something like a class. E.g. `lua/telescope/pickers.lua`. This just assign's a function pointer to a table entry. Not sure if it is called function pointer but thats basically it.

lua/telescope/builtin/files.lua
Julian: I can't seem to get this working with finder.new_oneshot_job.

I get yelled at for calling vim functions in a libuv callback.

It seems part of #274 fixed that but even with the fix there I get the same warning.

It also seems a bit silly to me to not have a way to read files from Lua. I was able to just straight up copy a function from luvit to do so -- will push that in a bit to get your thoughts, if there was a way to just populate a finder incrementally in a callback I think this'd work. I thought that'd be `finder.new_job` but maybe I'm not using it right.

Conni2461: The thing is we can't use any vimL inside a `coroutine` but if we would do a `new_oneshot_job` we would use a `coroutine`.
Inside `entry_display.lua` (which is used to dynamically generate tables) we also truncate and to determine the string display width we use `vim.fn.strdisplaywidth(str)` which is vimL. Tj thought maybe `vim.api.nvim_strwidth` would help but it did not. So right now we could use `#str` to determine the string size but that does not work for any charakter that requires two cells(emojis or chinese characters i think).

I will take a look at this tomorrow. Its already to late for me :smile:.

I will open new converstation if there is something to talk about so you can close this one for now. Thanks for your work

lua/telescope/builtin/files.lua
```diff
@@ -186,30 +187,47 @@ files.current_buffer_fuzzy_find = function(opts)
...
+    end)
+  end
 
   pickers.new(opts,{
     prompt = 'Tags',
-    finder = finders.new_table {
-      results = results,
-      entry_maker = make_entry.gen_from_ctags(opts),
-    },
+    finder = finders.new_job(function(prompt)
```
Julian, line 214: (This obviously doesn't work anymore now but I'm assuming you'll have advice for how to make it work, if you approve of the way I went forward with the above)

Conni2461: We are doing a real discussion

Conni2461:
So first of all you can't use `finder.new_job` because everything with `job` refers to `plenary.nvim`. And jobs refers to [this doc/spec](https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/job.lua#L56-L69). The idea behind job is give it a os command and run it either async or sync and we are able to do processing `on_stdout`, `on_stderr` and `on_exit`. Which we do in `finders.new_oneshot_job`. This is why i suggested lets use a external command `grep`. Thats kinda the easy/lazy way. But i like the idea which you pursue. We just need to do stuff differently.

I am currently thinking we write a `finders.new_fs_reader(files, opts)` with files being either a filepath or a table of filepaths and opts working like in `finders.new_oneshot_job`. Basically we are cloning `finders.new_oneshot_job` and writing a `obj._find = coroutine.wrap(function(finder, _, process_result, process_complete)` with `process_result` is called `on_line` and `process_complete` `on_exit`. Does this makes sense?

CC @tami5 do you think we can make this work? Also could you look at that luvit stuff in this PR. I don't know about this stuff 🤣

tjdevries:
Do you want me to check some of this out?

Julian:
That would definitely be super helpful -- if it helps more than it hurts the last commit I pushed is the incomplete work I started on a finder that just takes coroutines and lets a picker stream data into it, but obviously feel free to ignore it too.

lua/telescope/builtin/files.lua
```diff
@@ -186,30 +187,43 @@ files.current_buffer_fuzzy_find = function(opts)
...
 
-  local results = vim.split(data, '\n')
+  local results = {}
+  for _, each in ipairs(tagfiles) do
+    local tags_directory = vim.fn.fnamemodify(each, ":h")
+    local resolve_filename = function(filename)
+        return path.exists(filename) and filename or
+               path.join(tags_directory, filename)
+    end
+    path.read(each, function(err, data)
```
tjdevries, line 209: Hmmm... is path.read sync or async? Didn't read the rest.

If async, probably need to wait til it's complete before calling picker, or implement the streaming to add values.

Julian: Async.

Julian:
This PR oddly... functions now, after some minor additional effort.

I doubt what's here is fully ideal but probably comments welcome again on approach?

There's also a minor issue where no entry is selected initially, unlike in other finders. Would love a suggestion on how to fix that and have the first entry get selected once it's fed in.

lua/telescope/previewers.lua
```diff
@@ -648,7 +648,7 @@ previewers.ctags = defaulter(function(_)
         scode = string.gsub(scode, [[\/]], [[/]])
         scode = string.gsub(scode, '[*]', [[\*]])
 
-        file_maker_sync(entry.filename, self.state.bufnr, self.state.bufname)
+        file_maker_sync(entry.resolve_filename(entry.filename), self.state.bufnr, self.state.bufname)
```
Conni2461, line 651: This needs also be updated at line 641. So the cache works and we only load a file max once and all further file accesses are just reusing the already loaded buffer.

Conni2461:
I need to have a closer look at the new finder and what you are actually doing in path. I will do this as soon as possible. But just one remark why not using `path.read_file_async` and doing the chunk thing. `path.read_file_async` implements exactly whats mentioned in the [luv docs](https://github.com/luvit/luv/blob/master/docs.md#file-system-operations). And would not jump that much between file system access and lua code. Also i don't get why most of the time errors are just passed to `noop()` rather then handled.

lua/telescope/finders.lua
```diff
@@ -15,6 +15,40 @@ local _callable_obj = function()
...
+  self.results = {}
+  self.finished = false
+
+  obj._find = coroutine.wrap(function(finder, _, process_result, process_complete)
+    repeat
+      for _, result in ipairs(self.results) do
+        process_result(finder.entry_maker(result))
+      end
+      finder, _, process_result, process_complete = coroutine.yield()
+    until self.completed
```
Conni2461, line 35: This should probably be `self.finished` which could be the reason that `process_complete()` is not called and a selection is not set. But i could be wrong. I just can't find another `self.completed`

lua/telescope/builtin/files.lua
```diff
@@ -187,30 +188,42 @@ files.current_buffer_fuzzy_find = function(opts)
...
+      for _, line in ipairs(vim.split(data, '\n')) do
+        finder:feed({ line = line, resolve_filename = resolve_filename })
+      end
+    end)
+    finder:finish()
```
Conni2461, line 213: I also think this `finish` is placed at the wrong position. We don't want to `finish` after the first `tagfile` is processed. We want files be loaded

lua/telescope/previewers.lua
Julian: Aha. Cool. Fixed.

lua/telescope/finders.lua
Julian: Sigh hooray nil... Thanks, nice catch.

lua/telescope/builtin/files.lua
Julian: Thanks, I had in mind to do that (a counter) but forgot to actually implement it -- there now I think.

Julian:
> lso i think @tjdevries mentioned he had a working new finder draft which would allow us to do something like this for all passed result tables.

Sounds good to me -- definitely what I'm doing here doesn't seem to be specific to tags, to me yeah it seems all finders should work this way and then ones that either fork subprocesses or have all their results at once should be special-case versions of the one I have here (or a better one) which is incremental.

> But just one remark why not using path.read_file_async and doing the chunk thing. path.read_file_async implements exactly whats mentioned in the luv docs. And would not jump that much between file system access and lua code. Also i don't get why most of the time errors are just passed to noop() rather then handled.

I don't know what that is but will have a look now :) -- as for `noop`, this is copied exactly without modification from `luvit` (other than renaming the event loop variable), so not sure really.

> Edit: Also just ask if you need any help with the linter :)

Definitely wouldn't mind a tip on how to fix the last linter error yeah.

>  Edit 2: I tried it and i don't get tags every single time i open it. Sometimes just nothing appears. Is this a known issue

Hm, it works for me -- will have a more careful look though.

Thanks for the tips!

lua/telescope/finders.lua
```diff
@@ -15,6 +15,40 @@ local _callable_obj = function()
...
+  }, self)
+
+  self.results = {}
+  self.finished = false
+
+  obj._find = coroutine.wrap(function(finder, _, process_result, process_complete)
+    repeat
+      for _, result in ipairs(self.results) do
+        process_result(finder.entry_maker(result))
+      end
```
Conni2461, line 33: Thinking about this. This should not work. Like on each repeat run you will process all results again, not only the newly added once. What about doing something like:
```lua
while table.getn(self.results) > 0 do
  -- pop element and do process_result
  -- But depending on how table is implemented this could be a bad perf hit.
done
```

Conni2461:
I did a recording. Its kinda weird. When fuzzy finding entries, they don't disappear. If i don't have any entries left i just, they still get displayed but i can't scroll. If i have only some left, i still have a full screen of entries but only 10 entries will cycle. I hope that recording will show that. Also i don't like that there are no entries when the picker gets drawn first. I have to type that they appear.

https://streamable.com/chvhqq

lua/telescope/builtin/files.lua
```diff
@@ -217,30 +218,44 @@ files.current_buffer_fuzzy_find = function(opts)
...
+  for _, each in ipairs(tagfiles) do
+    local tags_directory = vim.fn.fnamemodify(each, ":h")
+    local resolve_filename = function(filename)
+        return path.exists(filename) and filename or
+               path.join(tags_directory, filename)
+    end
+    path.read(each, function(err, data)
+      for _, line in ipairs(vim.split(data, '\n')) do
+        finder:feed({ line = line, resolve_filename = resolve_filename })
+        if remaining == 1 then finder:finish() else remaining = remaining - 1 end
```
Conni2461, line 243: Should not be in the loop

tjdevries:
Hey, sorry, I'm trying to work through the backlog of telescope PRs that we have.

Do you need anything from me? Did I accidentally leave everyone hanging? Sorry about that.

delphinus:
`IncrementalFinder` should be useful for other pickers / extensions. Can this PR be merged? Still have some issues?

Julian:
There's still at least one issue here yeah, one that requires `path.normpath` implemented to fix (because as-is, `builtins.current_buffer_tags` would be broken). I haven't gotten much time to get back to this (and more honestly, it's a bit tedious to reimplement things like that instead of just having access to them via an existing lua library, which demotivates a bit). I checked `plenary.path`, it doesn't look like `normpath` is in it quite yet.

I agree though that `IncrementalFinder` seems useful, so perhaps it's worth splitting off and merging for wherever else.

Conni2461:
A IncrementalFinder is cool but it didn't work at all the last time i checked in on this work and i couldn't figure out why so no offense :laughing: . 
I also don't like that its currently way to specific to vim.loop.fs_read and i don't like the way how we read the file async. With reading chunk, doing lua, reading chunk, etc. Might be worse then just one fs for file.

@Julian we already have a `normalize` in `telescope.path` and if i recall you have already seen the PR to move that function :)

Julian:
> @Julian we already have a normalize in telescope.path and if i recall you have already seen the PR to move that function :)

That's a different function than the one that's needed (or it needs more work) -- I forget specifically what, but I did of course know about that one and tried to use it but yeah it doesn't do what's needed. I can probably remind myself in what specific way that was by probably taking another look myself though my TODO list has a few things on it first :/

And definitely no offense taken, we definitely shouldn't merge broken stuff :D.

Conni2461:
@Julian I am super sorry that this took a month but i was super busy and this PR always seemed like such a big task :laughing: 

Okay i just figured it out how to make it work :) I can send you a diff if you want and you would need to rebase (i also did this locally :laughing:) and probably remove the telescope path additions and use plenary.path instead.

I was already reading what python normalize does and i think you are looking for vim.loop.fs_realpath (man 3 realpath)

If you want you can also write me on gitter :) but i think you are using irc so thats not an option :laughing: 
[diff](https://gist.github.com/Conni2461/3af965ab1e60da4f03518834caf8f7a9) you might also wanna clean up some things

elianiva:
>  i think you are using irc so thats not an option

FWIW you can use irc.gitter.im to chat with people on Gitter

Conni2461:
I edited the diff i found some more things when i tried to do [fun things (async scandir picker, doesn't need to hide from fd and its all lua)](https://gist.github.com/Conni2461/8d78a1ebea35f8ccee6880e6684719e8) with that finder.

Julian:
> @Julian I am super sorry that this took a month but i was super busy and this PR always seemed like such a big task 😆

Absolutely no worries at all... I went back to denite for a bit so if anything I'm the one behind on seeing whether I can finish this off. Will have another look myself, the diff you sent looks definitely helpful.

May ping you later, but yeah the `fs_realpath` hint is also likely to be helpful thanks!

Conni2461:
I will close the PR now, so new people can have a shot at it. Someone already added support for &tags and i think i will look into streaming in new items (sometime soon).

Feel free to reopen if you still work on it :)

Issue #276, opened 2020-11-22
Ability to perform an action on an entry (edit) and display the updated value in the original finder 


sunjon:
I'm creating a [register finder](https://github.com/nvim-telescope/telescope.nvim/pull/275).

One of the actions I would like to have is `edit`. I'm using input() for that at the moment and it works okay-ish.

I don't call action.close(), because I want the user to be able to continue using the finder, and maybe perform an action on another entry.

Currently, the register is updated, `entry.value` is updated and the finder is refocused, however it still shows the original pre-edited entry.

l-kershaw:
I am guessing that `Picker:refresh()` is sufficient for this, so will close this now.

Feel free to reopen if you disagree 🙂

smackesey:
`Picker:refresh()` works for this, but is there a way to restore the selection position? When I run `Picker:refresh()` with a new finder, my input text at the prompt stays the same, but the selection leaps back to the top of the results list. I tried saving the selected row index before and restoring it after I call `Picker:refresh()`, but this isn't working-- I'm assuming because Picker:refresh() runs asynchronously?

How would I restore the selection (it's not a multi selection, just the selected row)? Is there a way to "await" intil Picker:refresh finishes, or provide it a callback?

PR #240, opened 2020-11-13
git builtins

Conni2461:
Close #49

Just putting this up here, so that people can i see that i am working on this.

Command that will be here:
- `commits`: with diff preview and checkout action
- `bcommits`: for current buffer with diff preview and checkout action
- `buffers`: with log preview(not sure about this) and checkout branch
- `git_status`: with diff preview and add on action (maybe multiselection would be great for this)

Should work, so please give it a try and report back :)

PR #289, opened 2020-11-25
Add git staging toggle to git_status builtin

cempassi:
This pull request creates three actions: "git_restore_staged_worktree", "git_restore_staged" and "git_restore_worktree".

I based my work on git add and go_to_file_selection

@Conni2461 Let me know how i can improve this PR. I also think that the git add/restore actions shouldn't close the prompt_buffer? This would allow multiple edits of the index.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -243,11 +243,20 @@ actions.git_checkout = function(prompt_bufnr)
   os.execute('git checkout ' .. val)
 end
 
-actions.git_add = function(prompt_bufnr)
+actions.git_staging_toggle = function(prompt_bufnr)
   local selection = actions.get_selected_entry(prompt_bufnr)
+
+  -- If parts of the file are staged and unstaged at the same time, stage
+  -- changes. Else toggle between staged and unstaged if the file is tracked,
+  -- and between added and untracked if the file is untracked.
+  if selection.status:sub(2) == ' ' then
+    os.execute('git restore --staged ' .. selection.value)
+  else
+    os.execute('git add ' .. selection.value)
+  end
   actions.close(prompt_bufnr)
-  local val = selection.value
-  os.execute('git add ' .. val)
+  require('telescope.builtin').git_status()
+  vim.api.nvim_feedkeys('i', 'n', false)
 end
 
 -- ==================================================
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -106,18 +106,9 @@ git.branches = function(opts)
 end
 
 git.status = function(opts)
-  local output = vim.split(utils.get_os_command_output('git status -s'), '\n')
-  local results = {}
-  for _, v in ipairs(output) do
-    if v ~= "" then
-      local mod, fname = string.match(v, '(..)%s(.+)')
-      if mod ~= 'A ' and mod ~= 'M ' and mod ~= 'R ' and mod ~= 'D ' then
-        table.insert(results, { mod = mod, file = fname })
-      end
-    end
-  end
+  local output = utils.get_os_command_output('git status -s')
 
-  if vim.tbl_isempty(results) then
+  if output == '' then
     print('No changes found')
     return
   end
@@ -125,19 +116,23 @@ git.status = function(opts)
   pickers.new(opts, {
     prompt_title = 'Git Status',
     finder = finders.new_table {
-      results = results,
+      results = vim.split(output, '\n'),
       entry_maker = function(entry)
+        if entry == '' then return nil end
+        local mod, file = string.match(entry, '(..).*%s[->%s]?(.+)')
         return {
-          value = entry.file,
-          ordinal = entry.mod .. ' ' .. entry.file,
-          display = entry.mod .. ' ' .. entry.file,
+          value = file,
+          status = mod,
+          ordinal = entry,
+          display = entry,
         }
       end
     },
     previewer = previewers.git_file_diff.new(opts),
     sorter = conf.file_sorter(opts),
-    attach_mappings = function()
-      actions.goto_file_selection_edit:replace(actions.git_add)
+    attach_mappings = function(_, map)
+      map('i', '<tab>', actions.git_staging_toggle)
+      map('n', '<tab>', actions.git_staging_toggle)
       return true
     end
   }):find()


Conni2461:
Overall looks pretty good :+1:. Do you plan to update the builtin to include those actions? You currently only changed `lua/telescope/actions/init.lua`

About the multiple edits, i think we should close the picker on `<enter>` or the main action, whatever it is mapped to. But i also want multiselection, so similar to fzf, where you can select multiple entries with <Tab>. Tj started working on this and you can enable this by commenting in `lua/telescope/mappings.lua:28` but it is not done yet. Selecting works but on confirm nothing happens with any of the selected items in the list. Tj never finished that, so i was planning to look into that, but haven't found the time yet.

cempassi:
I only changed the actions so that i could ask questions about the modification of the builtin:

I agree that we should close the picker on `<enter>`. I was thinking that we could map git add/restore to some `<C->` keys. That would allow us to extend the git actions with 'add to .gitignore', 'git rm', 'git stash', ect. Then on `<enter>` we could either jump to the file as the defaults or create a new commit, as it's the logical step after editing the index IMO.

I was thinking about multiple edits in sequence, which should work fine with just the actions mapped to something else than `<enter>`. But multiple edit by selection should sure be the natural improvement for this!

tami5:
> Overall looks pretty good . Do you plan to update the builtin to include those actions? You currently only changed `lua/telescope/actions/init.lua`
> 
> About the multiple edits, i think we should close the picker on `<enter>` or the main action, whatever it is mapped to. But i also want multiselection, so similar to fzf, where you can select multiple entries with . Tj started working on this and you can enable this by commenting in `lua/telescope/mappings.lua:28` but it is not done yet. Selecting works but on confirm nothing happens with any of the selected items in the list. Tj never finished that, so i was planning to look into that, but haven't found the time yet.

Regarding closing the picker on enter, I feel it might be better for this use case that enter just toggles between stuff. I really want to replace fugitive :$ @Conni2461

Conni2461:
My idea was we list the stuff that has changes. You can look at the diff and say okay file 2 and 3 should be in the commit. So you press `<Tab>` twice and confirm. Files on the staging index. Super simple and i think fzf does that the same way. Or am i wrong?
And i agree that we could do the same thing with getting them of the staging index.

But all that other fancy stuff like `add to .gitignore`, `git rm`, `git stash`, `making a commit` would be out of scope imo. Because that would mean that we would become a `git plugin` as well.
At least for core. But we could do it similar to what we discussed with the filesystem. As an extension.

Also also `git rm` does only makes sense for `git_files`.

Also also also maybe i should like link [comment](https://github.com/nvim-telescope/telescope.nvim/pull/240#issuecomment-727571912) and my [response](https://github.com/nvim-telescope/telescope.nvim/pull/240#issuecomment-727573452). We had the end of the discussion somewhere else but for what its worth.

tami5:
I see what you mean, and I agree with you, (so tab stage/unstage files?) but in the idea of hitting enter would close the git commit buffer would limiting, like I'd want to open git commit picker have a key to stage and unstage files and a key to restore, that simple :D. I suggest the key for staging and unstaging file would be enter, escape to close the instance of git picker? 

This is just my thought on this matter. and maybe tab would do the job and we can keep the functionality of enter the same

cempassi:
I agree that an extension would work better for advanced things. 

Just like @tami5, I would like to replace fugitive, and have a seamless git integration with telescope. I would be very interested to help in any capacity on this git extension.

Having said that, I think a toogle stage/ unstage action is the way to go. The git restore --workspace may already be out of scope, as we'd either git add or git restore --staged. Should I remove those?

I'll do the toogle, and map it to tab for defaults, leaving enter unchanged, would that work?

For the toogle I need the staged files in the results. Should I wait as you assigned yourself to the issue @Conni2461 or should I do the change in this PR? 

I'm fairly new to contributing, so there are many things I'm not sure of 😅

Conni2461:
You are very welcome to help with that. But replacing fugitive inside core would be insane. 
Yeah i think just `git restore --staged` is good.

You can map it to tab. But you might have to write a little logic around to figure out if you wanna `add` it or `restore`. The core code picker code can be found here: https://github.com/nvim-telescope/telescope.nvim/blob/0f9fd84a9742e22294cfaf46f32bad9b29de0b31/lua/telescope/builtin/git.lua#L108-L144
And how to attach a mapping without replacing can be found [here](https://github.com/nvim-telescope/telescope.nvim#picker). You are interested in the `map()` function.
You might have a difficult time to update the list after a toggle. So you might wanna close and reload the picker. We currently don't have a redraw.

I am not working on the issue right now. I unassigned myself the second after you said you have interested in working on it :laughing:. Feel free to ask here or on gitter. But gitter might be better to keep this one here "cleaner"

cempassi:
@Conni2461 I get the expected behaviour but the way I refresh the window feels fairly hacky to me. Do you think there is a better way to refresh the result window?

Conni2461:
As i said we currently have no redraw. So yes it is super hacky. But so many people want it. sunjoin wants it, me too and now you :) https://github.com/nvim-telescope/telescope.nvim/issues/276 is the issue which tracks that request.

Will `<CR>` no open the file?
Also i am not sure if that works with all mods. Something that is super annoying and was broken with my version as well is `RM`. Which happens if you rename a file. And do modifications. I think it looks something like this `RM oldfile -> newfile`. With my regex i think filename will be `oldfile -> newfile` which would result in something broken

lua/telescope/actions/init.lua
```diff
@@ -250,6 +250,22 @@ actions.git_add = function(prompt_bufnr)
...
+  -- changes. Else toggle between staged and unstaged if the file is tracked,
+  -- and between added and untracked if the file is untracked.
+  if selection.status:sub(2) == ' ' then
+    os.execute('git restore --staged ' .. selection.value)
+  else
+    os.execute('git add ' .. selection.value)
+  end
+  actions.close(prompt_bufnr)
+  vim.cmd [[lua require('telescope.builtin').git_status()]]
+  vim.api.nvim_feedkeys('i', 'n', false)
```
Conni2461, line 266: That might be the fix @tami5 and i were looking for the past week. Will this put you in insert mode after toggling? Because `vim.cmd('startinsert')` would not.

lua/telescope/builtin/git.lua
```diff
@@ -129,15 +127,16 @@ git.status = function(opts)
...
           display = entry.mod .. ' ' .. entry.file,
         }
       end
     },
     previewer = previewers.git_file_diff.new(opts),
     sorter = conf.file_sorter(opts),
-    attach_mappings = function()
-      actions.goto_file_selection_edit:replace(actions.git_add)
+    attach_mappings = function(_, map)
+      map('i', '<tab>', actions.git_staging_toggle)
```
Conni2461, line 139: We should also map normal mode

lua/telescope/builtin/git.lua
```diff
@@ -111,9 +111,7 @@ git.status = function(opts)
   for _, v in ipairs(output) do
     if v ~= "" then
       local mod, fname = string.match(v, '(..)%s(.+)')
-      if mod ~= 'A ' and mod ~= 'M ' and mod ~= 'R ' and mod ~= 'D ' then
         table.insert(results, { mod = mod, file = fname })
```
Conni2461, line 114: indentation should be wrong now

lua/telescope/actions/init.lua
cempassi: Yes, that puts you in insert mode after toggling. The hackiness of that line made me feel a bit guilty so I'm glad it helps haha

lua/telescope/builtin/git.lua
cempassi: For me Indentation still works, as the space is significant: without colors, it's currently the only way to distinguish between staged and unstaged files

Conni2461: i meant the line indentation from table.insert. Two spaces too much. Sorry about nitpicking

cempassi:
@Conni2461 Ah true, I forgot about the Rename and Copied cases. I think there's also some specifics during a merge conflict. I'll look into all of that!

lua/telescope/builtin/git.lua
cempassi: Don't be sorry, I'm here for that kind of precision! I have to fix my indentation for lua files, but when you spot blunders like that, I'm thankful more than anything ^^

lua/telescope/builtin/git.lua
```diff
@@ -110,10 +110,12 @@ git.status = function(opts)
...
     if v ~= "" then
-      local mod, fname = string.match(v, '(..)%s(.+)')
-      if mod ~= 'A ' and mod ~= 'M ' and mod ~= 'R ' and mod ~= 'D ' then
-        table.insert(results, { mod = mod, file = fname })
+      local mod, file = string.match(v, '(..)%s(.+)')
+      local old = ''
+      if mod:sub(1, 1) == 'R' or mod:sub(1, 1) == 'C' then
+        old, file = string.match(file, '(.+)%s.+%s(.+)')
       end
+      table.insert(results, { mod = mod, old = old, file = file })
```
Conni2461, line 118: We should just move that into entry_maker. So we remove that whole `results` part, and just do `results = output` at line 130. That way you don't have to concat the lines back together you can just do `ordinal = line` and `display = line`.
Does that make sense to you?

Conni2461: But you should do a `if entry == '' then return nil end` check inside the entry_maker because we will end up with one empty line and that would be added. You won't see it but it is selectable.

cempassi: I'm totally for that. I asked about the empty result to actually go even further and rely on finders.new_oneshot_job instead of `vim.split(utils.get_os_command_output('git status -s'), '\n')`. But moving everything we can in entry_maker is fine already.

Conni2461: But moving to new_oneshot_job will always open the window and pull results async. We can't just not open it with that solution. Because we don't even know if there will be results at this point in this. Also doing this upfront is no real "performance" hit.

lua/telescope/builtin/git.lua
```diff
@@ -107,37 +107,35 @@ end
...
   pickers.new(opts, {
     prompt_title = 'Git Status',
     finder = finders.new_table {
-      results = results,
+      results = output,
       entry_maker = function(entry)
+        if entry == '' then return nil end
+        local mod, file = string.match(entry, '(..)%s(.+)')
+        if mod:sub(1, 1) == 'R' or mod:sub(1, 1) == 'C' then
+          _, file = string.match(file, '(.+)%s.+%s(.+)')
```
Conni2461, line 124: Whats matched inside the `()` will be returned. Just not setting the first `()` lets you get rid of `_, `.

And i think we can do this in one regex. With an optional.

lua/telescope/builtin/git.lua
```diff
@@ -107,37 +107,35 @@ end
...
-    if v ~= "" then
-      local mod, fname = string.match(v, '(..)%s(.+)')
-      if mod ~= 'A ' and mod ~= 'M ' and mod ~= 'R ' and mod ~= 'D ' then
-        table.insert(results, { mod = mod, file = fname })
-      end
-    end
-  end
 
-  if vim.tbl_isempty(results) then
+  if vim.tbl_isempty(output) then
```
cempassi, line 111: This check fails, i can't find the syntax to handle the empty result right. Any ideas? with vim.inspect, output is { "" }

lua/telescope/builtin/git.lua
cempassi: I'll try to figure this one out, it's interesting!

lua/telescope/builtin/git.lua
Conni2461: Okay i see. Because of the newline the terminal prints. Vim split will create an entry for each new line and this will include that one. I think we should just be able to get the output do a string check for `""` and than split

lua/telescope/builtin/git.lua
Conni2461: Let me know if you need help with it. Its actually quite easy. But its feels good to finally figure out a regex :laughing:

cempassi: I figured out the PCRE version of the regex. I though i could just replace `\` by `%` to have it work in lua, but the lua regex engine doesn't seem to allow logical 'or', so I throw the towell haha. Let me know how! The PCRE Regex I ended up with : `([RC].)\s.+\s\-\>\s(.+)|(..)\s(.+)`

Conni2461: Okay i wasn't thinking about `or` and i don't even know how to do that in lua.
I ended up doing `'(..).*%s[->%s]?(.+)'`

cempassi: Hahaha this is wizardry ! 😂

Can you please explain to me how it works? 🙏🏿

Conni2461: Sure 
- `(..)` = will match the first two chars.
- `.*` = will do anything or nothing. Nothing is important
- `%s` = one white space. Thats the whitespace between `xx file` or `xx oldfile `(last one). why the last one because `.*` would match oldfile.
- `[->%s]?` = will match these characters optional. So this only applies when we have an arrow. for `xx file` we still only matched `xx ` and for `xx oldfile -> newfile` we have matched `xx oldfile -> ` at this point.
- `(.+)` = will just match till the end of the string.

Conni2461:
@cempassi I will test this pr tomorrow and merge if it works. Right now i am f****** exhausted with all that less debugging 🤣

cempassi:
Haha i saw that from afar, it looked rough! There's no hurry 👍🏿

 Thank you for your time and support. Please be sure to rest and take care 🙏🏿

tami5:
> @cempassi I will test this pr tomorrow and merge if it works. Right now i am f****** exhausted with all that less debugging 

:bow: :heart:

lua/telescope/actions/init.lua
tami5: @cempassi nice love it, going to steal it to finish my pr :$

lua/telescope/actions/init.lua
```diff
@@ -250,6 +250,22 @@ actions.git_add = function(prompt_bufnr)
...
+  -- If parts of the file are staged and unstaged at the same time, stage
+  -- changes. Else toggle between staged and unstaged if the file is tracked,
+  -- and between added and untracked if the file is untracked.
+  if selection.status:sub(2) == ' ' then
+    os.execute('git restore --staged ' .. selection.value)
+  else
+    os.execute('git add ' .. selection.value)
+  end
+  actions.close(prompt_bufnr)
+  vim.cmd [[lua require('telescope.builtin').git_status()]]
```
Conni2461, line 265: Calling vimL out of lua just to call lua out of vimL is so next level inception.
Its just `require('telescope.buitlin').git_status()`.

Also can you remove actions.git_add()? Its not used anymore

Conni2461:
Okay everything LGTM. I still just not 100% sold of the new functionality because personally i expect something else, but if @tami5 says its good, he can merge :smile:
Thanks for your work. @cempassi

tami5:
@Conni2461 delegating stuff as always 😂😂😝😝. @cempassi It looks good to me, I'm going to squash and merge as soon as I'm on my computer but you can do that with good commit message. 

Thanks

PR #290, opened 2020-11-25
Add file browser

Conni2461:
File browser. Currently navigating works. We need a redraw as suggested in #276 to remove the flickering.

- [x] Add basic functions like create file/mkdir (use `<C-e>`)
- [x] Create nested directories
- [x] ~~Not only depend on `ls`. `find` and `fd` should be possible as well. For that need to inject `..` item.~~
- [x] Same issue as described in this issue #262. Absolutely no idea how to fix it.
- [x] Replace if implemented 🎉 

Edit: For anyone interesed. `Picker:refresh` is done https://streamable.com/v79mjb

Edit 2: Requires https://github.com/nvim-lua/plenary.nvim/pull/59

Edit 3: I think i wanna do some other things in that PR as well 🤔 

CC @rafcamlet

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -245,16 +245,11 @@ actions.git_staging_toggle = function(prompt_bufnr)
   local cwd = action_state.get_current_picker(prompt_bufnr).cwd
   local selection = action_state.get_selected_entry()
 
-  -- If parts of the file are staged and unstaged at the same time, stage
-  -- changes. Else toggle between staged and unstaged if the file is tracked,
-  -- and between added and untracked if the file is untracked.
   if selection.status:sub(2) == ' ' then
     utils.get_os_command_output({ 'git', 'restore', '--staged', selection.value }, cwd)
   else
     utils.get_os_command_output({ 'git', 'add', selection.value }, cwd)
   end
-  do_close(prompt_bufnr, true)
-  require('telescope.builtin').git_status()
 end
 
 local entry_to_qf = function(entry)
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -1,3 +1,4 @@
+local actions = require('telescope.actions')
 local action_state = require('telescope.actions.state')
 local action_set = require('telescope.actions.set')
 local finders = require('telescope.finders')
@@ -6,6 +7,10 @@ local pickers = require('telescope.pickers')
 local previewers = require('telescope.previewers')
 local utils = require('telescope.utils')
 local conf = require('telescope.config').values
+
+local scan = require('plenary.scandir')
+local Path = require('plenary.path')
+local os_sep = Path.path.sep
 
 local flatten = vim.tbl_flatten
 
@@ -191,6 +196,84 @@ local function prepare_match(entry, kind)
   return entries
 end
 
+files.file_browser = function(opts)
+  opts = opts or {}
+
+  opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or vim.loop.cwd()
+
+  local gen_new_finder = function(path)
+    opts.cwd = path
+    local data = {}
+
+    scan.scan_dir(path, {
+      add_dirs = true,
+      depth = 1,
+      on_insert = function(entry, typ)
+        table.insert(data, typ == 'directory' and (entry .. os_sep) or entry)
+      end
+    })
+    table.insert(data, 1, '../')
+
+    return finders.new_table {
+      results = data,
+      entry_maker = (function()
+        local tele_path = require'telescope.path'
+        local gen = make_entry.gen_from_file(opts)
+        return function(entry)
+          local tmp = gen(entry)
+          tmp.ordinal = tele_path.make_relative(entry, opts.cwd)
+          return tmp
+        end
+      end)()
+    }
+  end
+
+  pickers.new(opts, {
+    prompt_title = 'Find Files',
+    finder = gen_new_finder(opts.cwd),
+    previewer = conf.file_previewer(opts),
+    sorter = conf.file_sorter(opts),
+    attach_mappings = function(prompt_bufnr, map)
+      action_set.select:replace_if(function()
+        return action_state.get_selected_entry().path:sub(-1) == os_sep
+      end, function()
+        local new_cwd = vim.fn.expand(action_state.get_selected_entry().path:sub(1, -2))
+        local current_picker = action_state.get_current_picker(prompt_bufnr)
+        current_picker.cwd = new_cwd
+        current_picker:refresh(gen_new_finder(new_cwd), { reset_prompt = true })
+      end)
+
+      local create_new_file = function()
+        local current_picker = action_state.get_current_picker(prompt_bufnr)
+        local file = action_state.get_current_line()
+        if file == "" then
+          print('To create a new file or directory(add ' .. os_sep .. ' at the end of file) ' ..
+                'write the desired new into the prompt and press <C-e>. ' ..
+                'It works for not existing nested input as well.' ..
+                'Example: this' .. os_sep .. 'is' .. os_sep .. 'a' .. os_sep .. 'new_file.lua')
+          return
+        end
+
+        local fpath = current_picker.cwd .. os_sep .. file
+        if string.sub(fpath, -1) ~= os_sep then
+          actions.close(prompt_bufnr)
+          Path:new(fpath):touch({ parents = true })
+          vim.cmd(string.format(':e %s', fpath))
+        else
+          Path:new(fpath:sub(1, -2)):mkdir({ parents = true })
+          local new_cwd = vim.fn.expand(fpath)
+          current_picker.cwd = new_cwd
+          current_picker:refresh(gen_new_finder(new_cwd), { reset_prompt = true })
+        end
+      end
+
+      map('i', '<C-e>', create_new_file)
+      map('n', '<C-e>', create_new_file)
+      return true
+    end,
+  }):find()
+end
+
 files.treesitter = function(opts)
   opts.show_line = utils.get_default(opts.show_line, true)
 
@@ -295,7 +378,7 @@ files.tags = function(opts)
     attach_mappings = function()
       action_set.select:enhance {
         post = function()
-          local selection = action_set.get_selected_entry()
+          local selection = action_state.get_selected_entry()
 
           if selection.scode then
             local scode = string.gsub(selection.scode, '[$]$', '')
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -1,4 +1,5 @@
 local actions = require('telescope.actions')
+local action_state = require('telescope.actions.state')
 local finders = require('telescope.finders')
 local make_entry = require('telescope.make_entry')
 local pickers = require('telescope.pickers')
@@ -108,22 +109,35 @@ git.branches = function(opts)
 end
 
 git.status = function(opts)
-  local output = utils.get_os_command_output({ 'git', 'status', '-s' }, opts.cwd)
+  local gen_new_finder = function()
+    local output = utils.get_os_command_output({ 'git', 'status', '-s' }, opts.cwd)
 
-  if table.getn(output) == 0 then
-    print('No changes found')
-    return
+    if table.getn(output) == 0 then
+      print('No changes found')
+      return
+    end
+
+    return finders.new_table {
+      results = output,
+      entry_maker = opts.entry_maker or make_entry.gen_from_git_status(opts)
+    }
   end
+
+  local initial_finder = gen_new_finder()
+  if not initial_finder then return end
 
   pickers.new(opts, {
     prompt_title = 'Git Status',
-    finder = finders.new_table {
-      results = output,
-      entry_maker = make_entry.gen_from_git_status(opts)
-    },
+    finder = initial_finder,
     previewer = previewers.git_file_diff.new(opts),
     sorter = conf.file_sorter(opts),
-    attach_mappings = function(_, map)
+    attach_mappings = function(prompt_bufnr, map)
+      actions.git_staging_toggle:enhance {
+        post = function()
+          action_state.get_current_picker(prompt_bufnr):refresh(gen_new_finder(), { reset_prompt = true })
+        end,
+      }
+
       map('i', '<tab>', actions.git_staging_toggle)
       map('n', '<tab>', actions.git_staging_toggle)
       return true
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -23,6 +23,7 @@ builtin.live_grep = require('telescope.builtin.files').live_grep
 builtin.grep_string = require('telescope.builtin.files').grep_string
 builtin.find_files = require('telescope.builtin.files').find_files
 builtin.fd = builtin.find_files
+builtin.file_browser = require('telescope.builtin.files').file_browser
 builtin.treesitter = require('telescope.builtin.files').treesitter
 builtin.current_buffer_fuzzy_find = require('telescope.builtin.files').current_buffer_fuzzy_find
 builtin.tags = require('telescope.builtin.files').tags
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -380,10 +380,9 @@ function Picker:find()
       prompt_prefix = prompt_prefix .. " "
     end
     vim.fn.prompt_setprompt(prompt_bufnr, prompt_prefix)
-
-    a.nvim_buf_add_highlight(prompt_bufnr, ns_telescope_prompt_prefix, 'TelescopePromptPrefix', 0, 0, #prompt_prefix)
   end
   self.prompt_prefix = prompt_prefix
+  self:_reset_prefix_color()
 
   -- Temporarily disabled: Draw the screen ASAP. This makes things feel speedier.
   -- vim.cmd [[redraw]]
@@ -443,6 +442,8 @@ function Picker:find()
       log.warn("Failed with msg: ", msg)
     end
   end
+
+  self.__on_lines = on_lines
 
   on_lines(nil, nil, nil, 0, 1)
   status_updater()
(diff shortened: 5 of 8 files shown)


lua/telescope/builtin/init.lua
```diff
@@ -1037,4 +1037,48 @@ builtin.current_buffer_tags = function(opts)
...
+builtin.file_browser = function(opts)
+  opts = opts or {}
+
+  local find_command = opts.find_command
+
+  if not find_command then
+    if 1 == vim.fn.executable("ls") then
+      find_command = { 'ls', '-Fa', '--group-directories-first' }
+    end
+  end
```
tjdevries, line 1049: I think we might even be able to do this with `vim.loop`... perhaps we should find ways to make `vim.loop` handle more of this work. It would be cool to make some of these completely system & command independent as well as not having to start up a new shell or anything like that.

tami5: Agree, might this shade some light

```lua
local readdir = function(path)
  local files = {}
  local handle = uv.fs_scandir(path)
  if handle == nil then
    return {}
  end

  while true do
    local name, typ = uv.fs_scandir_next(handle)
    if name == nil then
      break
    end
    local is_dir = false
    if typ == 'directory' or typ == 'link' or typ == 'linked' then
      is_dir = true
    end

    local icon, highlight_name = Devicons.get_devicons(name, is_dir)
    icon = (icon and icon ~= '' and icon .. ' ' or '')

    table.insert(files, {
      value = name,
      display = ' ' .. icon .. name .. (is_dir and '/' or ''),
      devicons = {
        icon = icon,
        highlight_name = highlight_name,
      },
      is_dir = is_dir,
    })
  end
  return files
end


```

Conni2461: I am currently reading the luv logs, but thanks :smile:. I was just lazy and used `ls` and kinda forgot about `vim.loop` again

lua/telescope/actions/init.lua
```diff
@@ -225,6 +225,48 @@ actions.run_builtin = function(prompt_bufnr)
...
+-- Or maybe tami5 is interested.
+local create_file = function(file)
+  vim.loop.fs_open(file, "w", 420, vim.schedule_wrap(function(err, fd)
+    if err then
+      vim.api.nvim_err_writeln('Could not create file: ' .. file)
+    else
+      vim.loop.fs_close(fd)
+    end
+  end))
+end
```
Conni2461, line 238: @tami5 Do i do that correctly. Also comment

tami5: Yeb that looks correct. Also regarding the comment, I was under the impression that we will create a new filesystem repo for telescope. But if you want us to temporary have action/fs.lua I can work on it

lua/telescope/builtin/files.lua
```diff
@@ -107,6 +107,58 @@ local function prepare_match(entry, kind)
...
+    attach_mappings = function(prompt_bufnr, map)
+      actions._goto_file_selection:replace(function(_, cmd)
+        local path = actions.get_selected_entry().path
+        actions.close(prompt_bufnr)
+        if string.sub(path, -1) == '/' then
+          require'telescope.builtin'.file_browser(vim.tbl_extend('force', opts, {cwd = path}))
+        else
+          vim.cmd(string.format(':%s %s', cmd, path))
+        end
+      end)
```
Conni2461, line 152: @tjdevries I think i have a cool new idea for actions. So if i have a dir i wanna open it and if not i just wanna call `_goto_file_selection` as usual. I have two ideas how to do that, interface wise:
```lua
actions._goto_file_selection:enhance ({
  pre = function()
    local path = actions.get_selected_entry().path
    actions.close(prompt_bufnr)
    if string.sub(path, -1) == '/' then
      require'telescope.builtin'.file_browser(vim.tbl_extend('force', opts, {cwd = path}))
      actions.stop_action(prompt_bufnr) -- An action that can always just stop all actions which would run later
    end
    -- If not true we just proceed as usual with pre functions calling _goto_file_selection
  end,
})
```
or
```lua
actions._goto_file_selection:replace_if(function()
  local path = actions.get_selected_entry().path
  return string.sub(path, -1) == '/'
end, function(_, cmd)
  local path = actions.get_selected_entry().path
  actions.close(prompt_bufnr)
  require'telescope.builtin'.file_browser(vim.tbl_extend('force', opts, {cwd = path}))
end)
```
Basically having two function parameters, the first one will return a bool and if that bool is true we will call the second function, if false we will `_goto_file_selection`. So a replace based on a condition.

What do you think?

lua/telescope/actions/init.lua
Conni2461: I don't want `actions/fs.lua`. We had discussed just opening and creating here so its usable and all that other fancy stuff in a extension. But i was thinking doing that `fs_` stuff in a plenary module and than just using that abstraction for `telescope-fs.nvim`

tami5: hmm 🤔, okay I’ll create a pr in plenary but will this pr depend on it because I only can work on it tomorrow.

My suggestion is to merge asap and the missing basic fs functions can be added when plenery pr is merged

Conni2461: Thats https://github.com/nvim-lua/plenary.nvim/pull/22 the current work of tj. (You are also probably only interested in the first commit. Everything else is github actions)
That was also just an idea. Does not need to be rushed. We can replace that raw `fs_` stuff i am currently doing here later. This PR one is also not done yet. Kinda wanna be able to create nested directories and directories + file. See lines 253 - 260

tami5: @Conni2461 plenary has a full set of fs related stuff, is there something in particular we want me to implement in plenary path module?

Conni2461: Oh. i should have looked before writing. I thought we only had a WIP module. My bad

tami5: Check it and tell me if something you like on it, so I can create it :smile:

lua/telescope/builtin/files.lua
tjdevries: oh, replace if is very cool! I think it's a good idea!

lua/telescope/actions/init.lua
```diff
@@ -225,6 +225,49 @@ actions.run_builtin = function(prompt_bufnr)
...
+
+actions.create_file = function(prompt_bufnr)
+  local file = vim.fn.input('Create file or directory(add / at the end of file). Enter name: ')
+  if file == "" then return end
+
+  local fpath = actions.get_selected_entry(prompt_bufnr).path
+  fpath = string.sub(fpath, -1) == '/' and vim.fn.fnamemodify(fpath, ':h:h') or vim.fn.fnamemodify(fpath, ':h')
+  fpath = fpath .. '/' .. file
```
tjdevries, line 246: should use path.sep. We can add `path.join` or something to the utils.

lua/telescope/actions/init.lua
```diff
@@ -225,6 +225,49 @@ actions.run_builtin = function(prompt_bufnr)
...
+
+  actions.close(prompt_bufnr)
+
+  if string.sub(fpath, -1) ~= '/' then
```
tjdevries, line 250: path.sep

Also, I think `fs_stat` can tell if it's a dir?

lua/telescope/actions/init.lua
```diff
@@ -225,6 +225,49 @@ actions.run_builtin = function(prompt_bufnr)
...
+    -- Would be cool.
+    -- Also if dir is last entry. Open file_browser in last dir
+    -- Or if file is last entry. Open file
+    local fd = vim.loop.fs_mkdir(fpath, 493)
+    if not fd then
+      vim.api.nvim_err_writeln('Could not create directory: ' .. fpath)
+      return
+    end
+    -- TODO(conni2461): We should keep track of the opts. We don't have them here right? So we keep opts.show_hidden
+    require'telescope.builtin'.file_browser({ cwd = fpath })
```
tjdevries, line 266: we could save the opts for the picker if we wanted, the same way we do the for the last selected entry.

lua/telescope/actions/init.lua
Conni2461: Yes but if i recall correctly that is about creating a file/dir and if the user left a `/` at the end we wanna create a dir. But we should definitly use `path.sep` thanks

Conni2461:
I use that PR to play around with a refresh interface :)

- [x] file_browser
- [x] git_status

lua/telescope/actions/init.lua
```diff
@@ -57,6 +61,10 @@ function actions.get_current_line()
   return state.get_global_key('current_line')
 end
 
+function actions.get_original_opts()
+  return state.get_global_key('original_opts')
+end
```
tjdevries, line 66: this seems kind of weird as an action. Not necessarily opposed to it, but it seems a little weird.

Perhaps `_` prefix might make sense here.

lua/telescope/actions/init.lua
```diff
@@ -257,6 +265,44 @@ actions.insert_symbol = function(prompt_bufnr)
...
+  else
+    fpath = vim.fn.fnamemodify(fpath, ':h')
+  end
+
+  fpath = fpath .. path.separator .. file
+
+  if string.sub(fpath, -1) ~= path.separator then
+    actions.close(prompt_bufnr)
+    path.create_file(fpath)
+    vim.cmd(string.format(':e %s', fpath))
```
tjdevries, line 286: Missing ways to do split, etc.

Could do similar strategy we did for selecting files here. Maybe we need some sort of abstraction over that to make it easier generally.

lua/telescope/actions/init.lua
```diff
@@ -257,6 +265,44 @@ actions.insert_symbol = function(prompt_bufnr)
...
+  else
+    do_close(prompt_bufnr, true)
+    -- TODO(conni2461): Recursive
```
tjdevries, line 289: `vim.fn.mkdir` I think has an option that will auto create all the required directories

lua/telescope/builtin/git.lua
```diff
@@ -114,24 +114,43 @@ git.branches = function(opts)
...
     previewer = previewers.git_file_diff.new(opts),
     sorter = conf.file_sorter(opts),
     attach_mappings = function(_, map)
-      map('i', '<tab>', actions.git_staging_toggle)
-      map('n', '<tab>', actions.git_staging_toggle)
+      local toggle_staging = function(prompt_bufnr)
+        local selection = actions.get_selected_entry()
+
+        if selection.status:sub(2) == ' ' then
+          os.execute('git restore --staged ' .. selection.value)
```
tjdevries, line 144: Does this work from other directories... just noticed this may not do that.

You can use `Job:new { ... }:sync()` to run a command in process.

lua/telescope/pickers.lua
```diff
@@ -756,6 +760,14 @@ function Picker:reset_selection()
   self.multi_select = {}
 end
 
+function Picker:refresh(finder)
+  local pbnr = self.prompt_bufnr
```
tjdevries, line 764: don't love the name :)

lua/telescope/pickers.lua
```diff
@@ -756,6 +760,14 @@ function Picker:reset_selection()
...
+  vim.api.nvim_buf_set_lines(pbnr, 0, 1, false, {})
```
tjdevries, line 765: to be safe, `0, -1` as the lines.

lua/telescope/pickers.lua
```diff
@@ -756,6 +760,14 @@ function Picker:reset_selection()
...
+  vim.api.nvim_buf_add_highlight(pbnr, ns_telescope_prompt_prefix, 'TelescopePromptPrefix', 0, 0, #self.prompt_prefix)
```
tjdevries, line 766: Doe we already have some code that does this? We should call the same function if so.

lua/telescope/pickers.lua
```diff
@@ -783,6 +795,7 @@ function Picker:set_selection(row)
   end
 
   state.set_global_key("selected_entry", entry)
+  state.set_global_key("original_opts", self.original_opts)
```
tjdevries, line 798: It worries me keeping around the original opts because we may have weird cases of leaving memory cycles & therefor not garbage collecting stuff. Not sure if it will happen though.

lua/telescope/actions/init.lua
Conni2461: Hmmm i don't think so this is a function that only creates files. Not that anything is final. I was just playing around with it a bit more but how i thought this should work:

- keypress -> vim.fn.input filename
  - if filename == test/ ends with slash directory and navigate into that dir
  - if filename == test.lua create file and open it. We don't really have a way to specifiy the split/vsplit here
  - if filename == test/nested/ create both dirs
  - if filename == test/nested/test.lua create dirs and open file

lua/telescope/actions/init.lua
Conni2461: Kinda want to do it with plenary.path and uv :) you already wrote the test but its still pending. I just haven't found the motivation to do it

lua/telescope/builtin/git.lua
Conni2461: I know all about Job :) but good point

lua/telescope/pickers.lua
Conni2461: https://github.com/nvim-telescope/telescope.nvim/pull/290#discussion_r534726522 kinda was your idea keeping them 😆

lua/telescope/pickers.lua
Conni2461: It lays somewhere above 😆 we could do a local function for it

lua/telescope/pickers.lua
Conni2461: But thinking about it now i think i don't need to keep the opts. I will try something the next time i work on it :)

lua/telescope/builtin/files.lua
```diff
@@ -185,6 +189,82 @@ local function prepare_match(entry, kind)
...
+          actions.close(prompt_bufnr)
+          Path:new(fpath):touch({ parents = true })
+          vim.cmd(string.format(':e %s', fpath))
+        else
+          Path:new(fpath:sub(1, -2)):mkdir({ parents = true })
+          actions.refresh(prompt_bufnr, gen_new_finder(vim.fn.expand(fpath)), { reset_prompt = true })
+        end
+      end
+
+      map('i', '<C-e>', create_new_file)
```
delphinus, line 260: Can I use this feature in not only `file_browser` but also `find_files` or `git_files`? I want to create a new file frequently in searching by that pickers.

And it is cool that the `Enter name:` prompt has the same words in Telescope prompt initially, I think.

Conni2461: We can do this. Need to abstract some thinks to make this work. But i think the new `replace_mod` think can make this possible. Good point.

I was also thinking like doing somethink like this:
- Write some text in prompt buffer
- Now we have maybe entries maybe not
- So <C-e> doesn't open the selection but takes prompt text and creates new file.
- So basically getting rid of the vim.fn.input thingy

Conni2461: @delphinus I am moving that sometimes else. It is now a really cool workflow but the fact that i am calling refresh makes it not usable more general. Maybe with enhanced. 

I need to think about this more but it needs to be merged soon because sunjon needs refresh and more api stuff that this PR provides.

Or you try it after its merged :)

lua/telescope/builtin/files.lua
```diff
@@ -295,7 +378,7 @@ files.tags = function(opts)
     attach_mappings = function()
       action_set.select:enhance {
         post = function()
-          local selection = action_set.get_selected_entry()
+          local selection = action_state.get_selected_entry()
```
Conni2461, line 381: Typo :sob:

lua/telescope/builtin/files.lua
tjdevries: I think we could consider trying to do something like `actions.file_create` or something. I don't know how we  want to expose more of these actions to people so that they can be re-used easily.

lua/telescope/builtin/git.lua
```diff
@@ -62,7 +63,7 @@ git.bcommits = function(opts)
     prompt_title = 'Git BCommits',
     finder = finders.new_table {
       results = results,
-      entry_maker = opts.entry_maker or make_entry.gen_from_git_commits(opts),
+      entry_maker = opts.entry_maker or make_entry.gen_from_git_commits(),
```
tjdevries, line 66: I think it's good to leave `opts` in there in case `gen_from_git_commits` ever needs them.

Just my opinion about it.

lua/telescope/builtin/git.lua
```diff
@@ -108,24 +109,42 @@ git.branches = function(opts)
...
-    finder = finders.new_table {
-      results = output,
-      entry_maker = make_entry.gen_from_git_status(opts)
-    },
+    finder = initial_finder,
     previewer = previewers.git_file_diff.new(opts),
     sorter = conf.file_sorter(opts),
     attach_mappings = function(_, map)
-      map('i', '<tab>', actions.git_staging_toggle)
-      map('n', '<tab>', actions.git_staging_toggle)
```
tjdevries, line 128: Why did you want to remove these from `actions`? I'm OK with doing it, but maybe there is some way we can still keep these around.

Conni2461: Its now a local function so we are able to generate a `new_finder`. Otherwise we don't have information about that finder. I can still keep it action around tho

lua/telescope/builtin/git.lua
Conni2461: I am adding them back in makes sense.

tjdevries:
What about ways to refresh only the current entry? Why do we need to refresh the entire finder?

I'm not in love with the fact that we've broken our contract of all `actions` always take just `prompt_bufnr` as the only parameter (or at least, that's what I thought we were doing).

I also don't think we necessarily need to _always_ delete the old finder and get a new one -- I'm wondering if there is a way we can refresh just the current entry.

Not sure :) just thinking some thoughts. We can always merge this for now as well and think about how we can improve them later.

Conni2461:
It would be cool to be able to just update one entry. But that way we would need some information where this entry can be found in the table. (In case of finder.new_table).

For me / filebrowser (and @sunjon) i need to read the whole next directory so its the easiest to just replace that finder. For `git stauts` we could just update the display and the underlying stuff but that would require some logic on our side so its easier to get the new information from git.

For why i added refresh to actions. I thought it would be an easier alias to handle it that way. If you don't like it, we can also just get the `current_picker` with action_state and call the function that way. I don't mind :)

tjdevries:
I'm not sure yet :) We can merge the way that is working and think about it some more in the future.

lua/telescope/builtin/git.lua
Conni2461: Maybe we can use `enhanced post` here. Just don't close telescope

Conni2461:
I am removing the `actions.refresh` helper function because we are transforming it and it isn't that useful that that happens.
Also its not an action, and it shouldn't be used more general right now. So its probably better that its "hidden" behind picker.

FabioRamatis:
This feature is available for master?

How can I use it? I really liked the feature and wanna try but I'm kinda noob

Conni2461:
`:Telescope file_browser` and then you can navigate like always. Directories have a `/` at the end.
You can create new files with typing filename in prompt and pressing <C-e>. Works recursive and creates dirs if they don't exist.
If you are only interested to create a dir not a file. Put a `/` at the end of your query.

It will creates those from cwd. So if you are in git root the dir is created in git root. If you are in `lua/` it will be created in `lua/` etc

Example:
- this/is/a/new/lua/file.lua
- asdf/new/dir/

We will add more documentation for this picker and everything else soon

FabioRamatis:
Oh ok! thanks

FabioRamatis:
It would be nice to have a feature to rename files/folders, etc...

Conni2461:
We talked about this briefly in gitter some time ago. Our idea was provide basic actions (only create file/dir) and than write an extension that fills in the gaps but i am not sure how we should realize that now.

You can just open an issues because i wanna do something like that. And an issue is easier to track.
Keep it simple you don't have to fill out the feature request form. I know what you mean :)

FabioRamatis:
Ok! nice :)

PR #294, opened 2020-11-27
fix: normal mode toggle issue

cempassi:
Fixes the bug I introduced with #289. I tested normal, insert mapping for regular staging, move and conflicted.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -265,9 +265,8 @@ actions.git_staging_toggle = function(prompt_bufnr)
   else
     os.execute('git add ' .. selection.value)
   end
-  actions.close(prompt_bufnr)
+  do_close(prompt_bufnr, true)
   require('telescope.builtin').git_status()
-  vim.api.nvim_feedkeys('i', 'n', false)
 end
 
 -- ==================================================


tami5:
@cempassi It got it fixed, put one other issue left and I think it may be the result of using feedkeys. Basically if you press double backspace the sorting should stop functioning.

FIY: I'm having the same issue with #266 after using feedkeys.

cempassi:
I'm not sure it comes from feedkeys: from a fresh start, try any other picker, type some keys, go in normal, back in insert, and try to backspace. Here, backspace stops working for me? To me it looks like the edit buffer doesn't have `backpase=indent,eol,start` and behaves as vim default?

tami5:
Ohh let me see

Conni2461:
@tami5 tbh i don't even get why we have this bug. we should close the old picker which puts us back in normal and opening a new one which should put us back in insert. I am totally lost about this one.

cempassi:
How I see it there seems to be two distinct bugs:
- The prompt behavior when we go back between insert and normal after something is already typed. It feels like backspace isn't set for the buffer. 
- The fact that closing and opening aren't switching mode as expected. I suspect some state is shared between the closed and opened pickers.
Those are assumptions, I'll try to look into it!

cempassi:
I tracked the source of the problem for the prompt to pickers.lua, line 357: `a.nvim_buf_set_option(prompt_bufnr, 'buftype', 'prompt')`

 The way the prompt is set is problematic for the way we use the prompt buffer. With that line commented, I don't have prompt anymore, but i can enter text, go in normal mode, back in insert, backspace to erase. Can you try it out to see if the source of this bug is correct?

Conni2461:
I don't get why this line should be problematic. Which issue should this fix. The first or second. Because i am still getting the second issue. And i don't really get whats the first issue.
Maybe @tjdevries has to take a look at it.

cempassi:
Ah my bad, I should have been more specific:
I'm talking about the first issue I see, the fact that we can't edit the line back if we type a bit, go into normal mode, and go back in insert. 
The line I reported don't fix the issue yet, but tracks the origin of this problem: the prompt filetype used to set the prompt prefix.

tami5:
@tjdevries any thoughts

tami5 (approved):

Conni2461:
Can't we solve this with `do_close` if we make it public. I don't like having multiple workarounds for the same problem

cempassi:
@Conni2461 I agree about having one way to solve one problem. Where can I find informations about `do_close` ? I'll look into it.

Conni2461 (approved):
LGTM. We can also discuss making `do_close` public with `actions._do_close` so its available for users as well if one creates an action function outside of this file. Like i do for `file_browser` don't want to move that function inside `actions`

PR #297, opened 2020-11-27
feat: add lsp range code actions

windwp:
https://github.com/neovim/neovim/pull/12962


tami5:
LGMT, very useful. Thanks

PR #301, opened 2020-11-29
fix: pass the options to pickers

tamago324:
This will allow for customization.

Thank you!

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -88,7 +88,7 @@ internal.planets = function(opts)
 end
 
 internal.commands = function(opts)
-  pickers.new({}, {
+  pickers.new(opts, {
     prompt_title = 'Commands',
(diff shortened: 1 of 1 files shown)


tamago324:
The following settings did not work

```lua
require('telescope.builtin').commands {
    local edit_command = function(prompt_bufnr)
      local entry = actions.get_selected_entry()
      actions.close(prompt_bufnr)
      local val = entry.value
      local cmd = string.format([[:%s ]], val.name)
      vim.cmd [[stopinsert]]
      vim.fn.feedkeys(cmd)
    end
    map('i', '<C-e>', edit_command)
    return true
  end}

```

I want to add a new mapping in built-in commands.
To do so, I need to add opts as an argument to pickers.
It would make me very happy to merge!

tamago324:
Thank you for merging!

tjdevries:
Thanks for fixing 😁

PR #302, opened 2020-11-29
new builtin - autocommands

sunjon:
aucmds


tjdevries:
This looks fine. After we do our PR in neovim, we can change this up I think.

You ready for it to be merged? if so, you can go ahead and do that.

PR #303, opened 2020-11-29
Symbol Picker

Conni2461:
~~Modified this [file](https://www.unicode.org/Public/emoji/13.1/emoji-test.txt). Is there a better file?~~

More sources (open for suggestions):
- [x] [emoji](https://www.unicode.org/Public/emoji/13.1/emoji-test.txt)
- [x] [kaomoji](https://github.com/kuanyui/kaomoji.el/blob/master/kaomoji-data.el)
- [x] [math](https://raw.githubusercontent.com/wspr/unicode-math/ef5688f303d7010138632ab45ef2440d3ca20ee5/unicode-math-table.tex)
- [x] [latex definitions](https://raw.githubusercontent.com/wspr/unicode-math/ef5688f303d7010138632ab45ef2440d3ca20ee5/unicode-math-table.tex)

Fix Me:
- [ ] Don't use curl. Maybe socket.http
- [x] Support multiple source from picker side. Default all sources, and opts for just some.
- [ ] Docs

Usage:
`:lua require'telescope.builtin'.symbols{} -- all sources`
`:lua require'telescope.builtin'.symbols{ sources = { 'math' } } -- only math`
`:lua require'telescope.builtin'.symbols{ sources = { 'emoji' } } -- only emoji`
`:lua require'telescope.builtin'.symbols{ sources = { 'kaomoji' } } -- only kaomoji`
`:lua require'telescope.builtin'.symbols{ sources = { 'math', 'emoji' } } -- both math and emoji`

Download:
`:lua require'telescope.script'.get_all_sources()`
`:lua require'telescope.script'.get_x_source() -- with x ∈ { 'emoji', 'math', 'kaomoji' }`

Source file format(3 spaces are used because of the fact that kaomoji can include spaces):
`symbol   description`

~~Also can someone explain what kaomoji is and is there a file similar to the used one. We could also make that work.~~

CC @clason

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -86,6 +86,55 @@ internal.planets = function(opts)
   }:find()
 end
 
+internal.symbols = function(opts)
+  local files = vim.api.nvim_get_runtime_file('data/telescope-sources/*.json', true)
+  if table.getn(files) == 0 then
+    print("No sources found! Check out https://github.com/nvim-telescope/telescope-symbols.nvim " ..
+          "for some prebuild symbols or how to create you own symbol source.")
+    return
+  end
+
+  local sources = {}
+  if opts.sources then
+    for _, v in ipairs(files) do
+      for _, s in ipairs(opts.sources) do
+        if v:find(s) then
+          table.insert(sources, v)
+        end
+      end
+    end
+  else
+    sources = files
+  end
+
+  local results = {}
+  for _, source in ipairs(sources) do
+    local data = vim.fn.json_decode(path.read_file(source))
+    for _, entry in ipairs(data) do
+      table.insert(results, entry)
+    end
+  end
+
+  pickers.new(opts, {
+    prompt_title = 'Symbols',
+    finder    = finders.new_table {
+      results     = results,
+      entry_maker = function(entry)
+        return {
+          value = entry,
+          ordinal = entry[1] .. ' ' .. entry[2],
+          display = entry[1] .. ' ' .. entry[2],
+        }
+      end
+    },
+    sorter = conf.generic_sorter(opts),
+    attach_mappings = function(_)
+      actions.goto_file_selection_edit:replace(actions.insert_symbol)
+      return true
+    end
+  }):find()
+end
+
 internal.commands = function(opts)
   pickers.new(opts, {
     prompt_title = 'Commands',
--- a/README.md
+++ b/README.md
@@ -391,6 +391,7 @@ Built-in function ready to be bound to any key you like :smile:.
 | `builtin.git_bcommits`              | Lists buffer's git commits with diff preview and checkouts it out on enter.                 |
 | `builtin.git_branches`              | Lists all branches with log preview and checkout action.                                    |
 | `builtin.git_status`                | Lists current changes per file with diff preview and add action. (Multiselection still WIP) |
+| `builtin.symbols`                   | Lists symbols inside a file `data/telescope-sources/*.json` found in your rtp. More info and symbol sources can be found [here](https://github.com/nvim-telescope/telescope-symbols.nvim) |
 | ..................................  | Your next awesome finder function here :D                                                   |
 
 #### Built-in Previewers
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -230,6 +230,12 @@ actions.run_builtin = function(prompt_bufnr)
   require('telescope.builtin')[entry.text]()
 end
 
+actions.insert_symbol = function(prompt_bufnr)
+  local selection = actions.get_selected_entry()
+  actions.close(prompt_bufnr)
+  vim.api.nvim_put({selection.value[1]}, '', true, true)
+end
+
 -- TODO: Think about how to do this.
 actions.insert_value = function(prompt_bufnr)
   local entry = actions.get_selected_entry(prompt_bufnr)
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -36,6 +36,7 @@ builtin.git_status = require('telescope.builtin.git').status
 
 builtin.builtin = require('telescope.builtin.internal').builtin
 builtin.planets = require('telescope.builtin.internal').planets
+builtin.symbols = require('telescope.builtin.internal').symbols
 builtin.commands = require('telescope.builtin.internal').commands
 builtin.quickfix = require('telescope.builtin.internal').quickfix
 builtin.loclist = require('telescope.builtin.internal').loclist


clason:
@Conni2461 Thanks for tackling this! This seems to be the most official plaintext file. But I'd recommend not bundling this in with the repo (people tend to complain about ~0.5Mb text files in their plugins) and downloading this on first use to `stdpath('data')`. (This will make it easier to update to 13.2 or whatever comes next.) Possibly this would be even better as a plugin? (@tjdevries ?)

"Kaomoji" are the older Japanese emoticons (a contraction from "kao", Japanese for "face", and "emoji", which originated as the Japanese pronunciation/spelling of "emoticon"). Compared to the western smileys, they are upright, e.g., `^_^` and `;_;`. There's a stunning variety that Google will be only too happy to show you, but the most famous is the table flip: (╯°□°)╯︵ ┻━┻

Unfortunately, I haven't found a good *annotated* (in English!) data file for them yet; the only one is https://github.com/kuanyui/kaomoji.el/blob/master/kaomoji-data.el (which at least contains the table flip ;))

Conni2461:
Thanks for the insight.

I also don't feel good about including that file, i am currently thinking doing a script to download the original source and apply the modifications. We have two options
- The lazy way would be sh script and using curl and sed. Which would not work for most windows users.
- The correct way would do a `'telescope.scripts'` module which downloads the file and uses `gsub`.

Then we write the modified files into `data/symbols` or similar and put this folder in `.gitignore`.
I tend to the second options and would do this tomorrow.

I like the math symbols idea and this can easily be done with the same picker if we just modify every file the same way: 
`symbol description`

I currently tend to include this in core. The picker is not complex and we could include different sources(emojis, kaomoji, math) which the user can install. On default we just display all installed sources and the user can then also specify a specific source with opts.

clason:
That makes sense, yes. If the data is not in the plugin directory but `stdpath('data')` (in which case `.gitignore` wouldn't be needed), core would be fine with me.

I like the idea of separate sources that can be specified via an option! This would make this a more general `symbols` picker, and users could even include their own custom symbols by putting them in a separate file in the `stdpath('data')` directory.

Conni2461:
@clason I think this should work pretty good now. I updated my original post with all information.

We might need a different math symbol source because there are a bunch of duplications with emojis. 
But not all emojis are in the math source and the descriptions are also lacking so we can't replace the emoji source with the math source.

Currently we also depend on `curl` but we could use `socket.http`. But that would require the user to install it with `neorocks` because afaik there is no plenary module for downloading stuff only this feature request https://github.com/nvim-lua/plenary.nvim/issues/27

clason:
> Currently we also depend on `curl` but we could use `socket.http`. But that would require the user to install it with `neorocks` because afaik there is no plenary module for downloading stuff only this feature request [nvim-lua/plenary.nvim#27](https://github.com/nvim-lua/plenary.nvim/issues/27)

`luarocks` integration in packer is coming :)

An alternative would be to have the emoji functionality itself in core, but put the sources themselves in a plugin?

Then we could put a "cleaned" list of unicode math symbols in there. (The Julia list is indeed redundant, being a subset of the full unicode table.)

A better list may be https://www.fileformat.info/info/unicode/category/Sm/list.htm (would have to be supplemented by others, like Greek letters and symbols). I also found https://gist.github.com/jaantollander/b49132a36b035cdfe05345dd0c7edf0f which may be the best source yet.

clason:
Upstream for that math source gist: http://milde.users.sourceforge.net/LUCR/Math/

Conni2461:
Thanks for all that symbol sources. :+1: Really helpful.
I think we should use [this](https://gist.github.com/jaantollander/b49132a36b035cdfe05345dd0c7edf0f) as a math source which will return the actual unicode symbol and [this](https://github.com/wspr/unicode-math/blob/master/unicode-math-table.tex) as latex source which will return the latex code. Currently implemented as `alt-math` and `latex`

So here is what i am currently thinking. We have two options and we probably want @tjdevries opinion on this whole picker.
1. I think `curl` as a downloader dependency is not that crazy. Currently everyone who uses that plugin must also use `bat`/`cat` and probably `less`. So i think everybody who doesn't have `curl` yet will find a way to get a `curl` binary. Also keeping set of functions here isn't a big problem as well.
2. We create a plugin with that script module. We also use curl there but we write a github action, which will run on each pr. This action will generate all sources and put them into a `data` directory. So anybody who wanna use it does not need `curl`.

clason:
I think you can generate the unicode symbols themselves using one of the functions from https://stackoverflow.com/questions/7983574, so you don't have to combine lists.

Both options sound good!

tjdevries:
This might end up being a good extension:

I could see several interations, like filetype icons being added, and maybe more. We can leave some of the utilties here that might be helpful for this and/or other extensions thouh. What do you think @Conni2461 ?

Conni2461:
@tjdevries Clason and i also did not really know how we should proceed. I think we kinda wanted your decision. 

> So here is what i am currently thinking. We have two options and we probably want @tjdevries opinion on this whole picker.
> 
> 1. I think `curl` as a downloader dependency is not that crazy. Currently everyone who uses that plugin must also use `bat`/`cat` and probably `less`. So i think everybody who doesn't have `curl` yet will find a way to get a `curl` binary. Also keeping set of functions here isn't a big problem as well.
> 
> 2. We create a plugin with that script module. We also use curl there but we write a github action, which will run on each pr. This action will generate all sources and put them into a `data` directory. So anybody who wanna use it does not need `curl`.

That was my whole take. Either keep everything here, or keep the picker here and sources as an extension. And filetype icons could be a cool source too. Thanks for that

tjdevries:
I think I prefer symbol picker itself as an extension. I can see many many additions to the API there and think it'd be better handled separately. At least some the utilities that you've added here though seem to be worthwhile keeping, so feel to merge in utilties that you think will be useful.

Conni2461:
Moved to https://github.com/nvim-telescope/telescope-symbol.nvim

I don't think we don't need to include here anything. Tami is working on a curl wrapper for plenary so we also don't have to include that download function here.

lua/telescope/script.lua
```diff
@@ -0,0 +1,130 @@
+local Job = require'plenary.job'
+
+local download_file = function(source, callback)
```
sharksforarms, line 3: Didn't check the size of the files, but would be nice to have a caching option to avoid unnecessary network requests

Conni2461: We would have to download the file upfront with `lua require'telescope.script'.get_emoji_source()` download a file, modify it and then save the file as `.local/share/nvim/telescope-sources/emoji.txt`.
I agree downloading the file every time when running the picker is "stupid", that is just a workaround to not provide the files in the repo and a easy way to do version bumps.

But now that we will end up with an extension i think i will keep the script and do a github action that will create these files and push them to the repo. Maybe all maybe just some. I will ping you when i created the new PR for the extensions. I am currently porting

clason:
I think it makes sense to move the emoji/math/etc. symbols to a separate extension.

However, I feel that the "bare" symbol picker functionality would be reasonable as a code function. By "bare", I mean:

1. a picker that takes a table from an external `json` (in a specified format) of pairs (triples?) "symbol"/"description"/("preview"?) and feeds it to telescope
2. an action that inserts the picked symbol in the buffer at the cursor position.

So if people want to just use it for their own list of symbols or abbreviations (or maybe snippets?), they don't need an extra plugin; it's just for the big, premade lists.

Conni2461:
Thats what i also suggested but @tjdevries said

> I think I prefer symbol picker itself as an extension. 

So i thought he wanted the picker in the extension as well. Honestly i really don't know how to continue with this one. 🤣

lua/telescope/builtin/internal.lua
```diff
@@ -87,6 +87,71 @@ internal.planets = function(opts)
...
+      end
+    },
+    sorter = conf.generic_sorter(opts),
+    attach_mappings = function(prompt_bufnr)
+      actions.goto_file_selection_edit:replace(function()
+        local selection = actions.get_selected_entry()
+        actions.close(prompt_bufnr)
+        local symbol = selection.value[1]
+        vim.api.nvim_put({symbol}, '', true, true)
+      end)
```
tjdevries, line 149: Let's make this it's own action.

Then you can do `actions.goto_file_selection_edit:replace(actions.insert_symbol)`

tjdevries (changes requested):

lua/telescope/builtin/internal.lua
Conni2461: Wait a sec i am currently explaining everything. But yeah i can do that

Conni2461:
@tjdevries Read it when ever you are fully awake. I just summarize what @clason and i tried to do. I also updated the source code so you can try it out.

1. We want a picker that reads a or multiple json files. I just changed the decoding to json, we had txt before that.
2. The user can specify that json file, he just has to follow a specific schema. `[["symbol", "description"], ["😀", "grinning face"], ["Λ", "capital lambda, greek"]]`. Basically anything he wants. He just has to put the file inside `vim.fn.stdpath('data') .. '/telescope-sources/'`.
3. Now he can fuzzy find with the description and enter will put the symbol in the current buffer.
4. Because not every user wants to create his own source, clason and i thought lets just provide some sources and to not bloat the repo, do it in a script.
5. So that not everyone has to run the script once i did a github action to just generate them and push them to the repo. So that happens here https://github.com/nvim-telescope/telescope-symbols.nvim/pull/1
6. Now the user can just use that `telescope-symbols.nvim` repo and has all sources or he can just pick one, e.g. the emoji source and put it inside `vim.fn.stdpath('data') .. '/telescope-sources/'`

On default all available sources are read but you can specify one or more sources with an opt

lua/telescope/builtin/internal.lua
```diff
@@ -87,6 +87,66 @@ internal.planets = function(opts)
   }:find()
 end
 
+internal.symbols = function(opts)
+  -- TODO(conni2461): Is their a better way to get the extension path?
+  local sourced_file = require('plenary.debug_utils').sourced_filepath()
+  local base_directory = vim.fn.fnamemodify(sourced_file, ":h:h:h:h:h")
+  local extension_path = vim.fn.globpath(base_directory .. '/telescope-symbols.nvim/data/', '*', true, true)
```
Conni2461, line 94: Note to myself. We could just create a function inside extension which returns the `extension_data_files` and call that function with pcall

tjdevries: `vim.api.nvim_get_runtime_file('data/telescope-symbols/', false)`

tjdevries: just change where it gets saved and use that API probably.

tjdevries: Files should just always go in `data/telescope-symbols/*.json`

Then we can always find them easily with `vim.api.nvim_get_runtime_file()` you just add your plugin with the correct data things.

lua/telescope/builtin/internal.lua
```diff
@@ -87,6 +87,66 @@ internal.planets = function(opts)
...
+  else
+    sources = files
+  end
+
+  local results = {}
+  for _, v in ipairs(sources) do
+    local fd = assert(vim.loop.fs_open(v, "r", 438))
+    local stat = assert(vim.loop.fs_fstat(fd))
+    local data = assert(vim.loop.fs_read(fd, stat.size, 0))
+    assert(vim.loop.fs_close(fd))
```
tjdevries, line 123: should be a util function somewhere already probably (oir if not, then we can just get it)

PR #304, opened 2020-12-01
Displayer now supports highlights

Conni2461:
@sunjon Does this help?

@tjdevries Do we wanna talk about the interface. I think this is the best way to do this, so people can change the `hl_group` based on the specific value. Like sunjon wants to do: `if x == 'opts' then hl_group1 else hl_group2 end`. 
The other idea i had was to specify the `hl_group` when we create the `displayer`. But that would limit our flexibility.

Separator Example: When creating the displayer
```lua
local hl = 'Visual'
  separator = " ",
  separator_hl = hl, -- New
  items = {
    { width = opts.bufnr_width },
    { width = 4 },
    { remaining = true },
  },
}
```

Elements Example: When submitting data
```lua
local hl_group = x == y then 'Normal' or 'Visual'

return displayer {
  entry.bufnr,
  { entry.indicator, hl_group },
  display_bufname .. ":" .. entry.lnum,
}
```

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -329,7 +329,7 @@ function make_entry.gen_from_buffer(opts)
       bufnr = entry.bufnr,
       filename = bufname,
 
-      lnum = entry.info.lnum and entry.info.lnum or 1,
+      lnum = entry.info.lnum ~= 0 and entry.info.lnum or 1,
       indicator = indicator,
     }
   end
--- a/lua/telescope/pickers/entry_display.lua
+++ b/lua/telescope/pickers/entry_display.lua
@@ -57,24 +57,52 @@ entry_display.create = function(configuration)
       local justify = not v.right_justify and "-" or ""
       local format_str = "%" .. justify .. v.width .. "s"
       table.insert(generator, function(item)
-        return string.format(format_str, truncate(item, v.width))
+        if type(item) == 'table' then
+          return string.format(format_str, truncate(item[1], v.width)), item[2]
+        else
+          return string.format(format_str, truncate(item, v.width))
+        end
       end)
     else
       table.insert(generator, function(item)
-        return item
+        if type(item) == 'table' then
+          return item[1], item[2]
+        else
+          return item
+        end
       end)
     end
   end
 
   return function(self, picker)
     local results = {}
+    local highlights = {}
     for i = 1, table.getn(generator) do
       if self[i] ~= nil then
-        table.insert(results, generator[i](self[i], picker))
+        local str, hl = generator[i](self[i], picker)
+        if hl then
+          local hl_start = 0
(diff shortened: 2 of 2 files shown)


lua/telescope/make_entry.lua
```diff
@@ -329,7 +329,7 @@ function make_entry.gen_from_buffer(opts)
       bufnr = entry.bufnr,
       filename = bufname,
 
-      lnum = entry.info.lnum and entry.info.lnum or 1,
+      lnum = entry.info.lnum ~= 0 and entry.info.lnum or 1,
```
Conni2461, line 332: I got a nasty buffers previewer bug sometimes. That should fix it.

sunjon:
as discussed on gitter, this change makes [223](https://github.com/nvim-telescope/telescope.nvim/issues/233) rear its head as soon as you change selection:

highlights_lost

When I applied the highlighter to the registers finder, it simply didn't highlight some of the entries at the initial state (not sure if this one has the same cause):

no_hl

Conni2461:
Hmmm. I only tested this with `builtin.buffers` but just looking at that screenshot i think there might be a bug. maybe i need to add a `tostring` somewhere when calculating the string width. Because it only affects the numbers. Will take a look at it.

tjdevries:
This seems to be a strict improvement over what we had before. I think you can merge (after investigating what sunjon mentioned) and I'll see if I can get #233 fixed in neovim core or develop some other strategy at some point.

Conni2461:
@tjdevries I am just investigating but it seams like that is not a bug on my side. I am generating the correct highlight table, for all elements.

Conni2461:
@tjdevries Figured it out. Don't know how to solve it yet. We don't make it every time in the if. Sometimes `set_ok` is false

https://github.com/nvim-telescope/telescope.nvim/blob/f3caae5ddad10eed8209cd3dcdd020620d0b9f7f/lua/telescope/pickers.lua#L847-L853

Conni2461:
@tjdevries @sunjon i think i know the root of this problem. Its the content of the register picker. It does not work when the display of the entry contains new lines because `nvim_buf_set_lines` does not allow new lines in the passed content table.

tjdevries:
OK, we can just do replace "\n" with "\\n" or something when we set the lines. Maybe don't check set ok for displaying the highlights. Either way, we can probably merge with the newline problem unfixed and fix it shortly after if you want. The pictures look very good from the PR and the code makes sense I think!

Conni2461:
I will just merge this now. Just because the packer extension is currently broken and we actually don't use this anywhere else. @sunjon was just playing around with it more. 
So when someone actually uses it for a builtin, we have to find a solution how to make sure that no `\n` are there. Optimal way would be to do a `gsub` in the builtin.

@p00f have fun with the packer extension

PR #315, opened 2020-12-08
Add test ci

Conni2461:
It actually runs: https://github.com/Conni2461/telescope.nvim/actions/runs/409082452

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,69 @@
+name: Tests
+
+on: [push, pull_request]
+
+jobs:
+  x64-ubuntu:
+    name: X64-ubuntu
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - run: date +%F > todays-date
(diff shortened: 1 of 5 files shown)


tjdevries:
I have changed plenary.nvim setup & execution. No more dependencies. We can chat about it tomorrow perhaps if you don't solve it before I'm awake.

Conni2461:
Works now almost out of the box. Tested it with no luarocks setup and `say` was missing. After installing that with: `luarocks install --lua-version 5.1 --local say` everything worked as expected. Thanks for updating it :)
https://github.com/Conni2461/telescope.nvim/runs/1522350679?check_suite_focus=true

tjdevries:
I'm wondering if we even need the first thing on ubuntu... It takes a LONG time to build nvim from scratch. So maybe we can just do that only for mac or something (and maybe only on master or something... I don't know how many free minutes I get per open source repo)

tjdevries:
You can merge whenever you're ready, we can tinker with it more later!

PR #316, opened 2020-12-09
fix: cycle scroll when num_results is small

aca:
Fixes the cycle scroll when num_results is smaller than max_results.

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -690,10 +690,18 @@ end
 
 function Picker:_handle_scroll_strategy(row)
   if self.scroll_strategy == "cycle" then
-    if row >= self.max_results then
-      row = 0
-    elseif row < 0 then
-      row = self.max_results - 1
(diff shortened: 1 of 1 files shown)


aca:
image

Cycle scroll doesn't work as expected when the result is small like this.

tjdevries:
Thanks! Appreciate it :)

tjdevries:
Ah, this actually broke for me when I started trying it more. What do you expect to have happen currently that isn't? I can try and fix it for you :)

tjdevries:
Actually I figured it out, ty for your PR it made it obvious now what was going wrong :) thanks!

aca:
@tjdevries Thanks for the kind fix. 

But things got really strange right now.  I can't select anything. I will check if other things are wrong.

Conni2461:
@aca Yeah i know here is the fix: https://github.com/nvim-telescope/telescope.nvim/pull/327

PR #322, opened 2020-12-09
[documentation] fix typo

efx:
👋 
I found this while trying to override the default finder so I can search hidden files in my project.
Would you mind providing an example for `.vimrc`?

I'm quite ignorant of how nvim works to load the file. I tried `lua :Telescope find_files find_command=rg,--ignore,--hidden,--files` but that didn't quite work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -575,7 +575,7 @@ supports tab completions and settings options.
 
 " If option is table type in lua code ,you can use `,` connect each command string eg:
 " find_command,vimgrep_arguments they are both table type. so config it in commandline like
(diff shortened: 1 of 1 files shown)


rockerBOO:
In this case there is a small vimscript layer that adds the `Telescope` command. So in the command mode in neovim you can do 

```
:Telescope find_files find_command=rg,--ignore,--hidden,--files
```
and that should work.

To change the command you can do 

```
:Telescope find_files find_command=my find command
```
or 
```
:lua require('telescope.builtin').find_files(find_command="my find command")

PR #324, opened 2020-12-10
Lots of README changes

rockerBOO:
- Add back prior intro line
- Add What is Telescope? 
_(saw multiple people read the intro line, then it has a image, then there is a block of text about what it does. This directs it before the image.)_
- Add Theme image to encourage people to play with their layouts more. 
- Added all the sections to the ToC and renamed some of the built in redundancies 
_(We can be more specific of external commands instead of saying everything is builtin.)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Gaze deeply into unknown regions using the power of the moon.
 
 `telescope.nvim` is a highly extendable fuzzy finder over lists.   Built on top of latest
 awesome features from `neovim` core.  Telescope is centered around
-modularity allowing for easy customization.
+modularity, allowing for easy customization.
 
 ![by @glepnir](https://user-images.githubusercontent.com/41671631/100819597-6f737900-3487-11eb-8621-37ec1ffabe4b.gif)
 
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ modularity, allowing for easy customization.
 
 ![by @glepnir](https://user-images.githubusercontent.com/41671631/100819597-6f737900-3487-11eb-8621-37ec1ffabe4b.gif)
 
-`telescope.nvim` has a number of [extensions](https://github.com/nvim-telescope) for supporting specific plugins. 
 `telescope.nvim` comes with a growing number of community driven [built-in pickers](#built-in-pickers),
-covering a `vim, file, git, lsp, treesitter`. The popup interface can be customized using options like `borderchars` and [themes](#themes).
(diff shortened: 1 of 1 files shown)


tjdevries (approved):
LGTM. Tami can make any comments required.

README.md
```diff
@@ -1,36 +1,31 @@
...
 
-`Telescope.nvim` is a next generation library for creating floating pickers
-with advanced features.  It is written in `lua` and is built on top of latest
-awesome features from `neovim` core.  `Telescope.nvim` is centered around
-modularity so much so that each picker is its own world, (meaning it can be
-configured in isolation from other pickers, such presentation, algorithm,
-mappings ... etc).
+`telescope.nvim` is a highly extendable fuzzy finder over lists.   Built on top of latest
+awesome features from `neovim` core.  Telescope is centered around
+modularity, allowing for easy customization.
```
tami5, line 10: Here (after around modularity) perhaps we need to mention to what extent telescope is modular. Modularity is very general feature that could mean different things.

README.md
```diff
@@ -1,36 +1,31 @@
...
 
-`Telescope.nvim` was built with the intention of becoming a library, but it has long grown to
-be much bigger than that.  In addition to
-[extensions](https://github.com/nvim-telescope) `Telescope.nvim` comes with a
-growing number of community driven [built-in pickers](#built-in-pickers),
-covering a wide range of use cases and tools, and offers a customizable user
-interface.
+![by @glepnir](https://user-images.githubusercontent.com/41671631/100819597-6f737900-3487-11eb-8621-37ec1ffabe4b.gif)
 
+`telescope.nvim` has a number of [extensions](https://github.com/nvim-telescope) for supporting specific plugins.
```
tami5, line 14: Builtin picker should be mentioned before the external extensions.

tami5 (changes requested):
LGMT @rockerBOO, I agree the introduction was a bit bulky

README.md
rockerBOO: I think its all modular so everything can be swapped out? From the end user, this should be accurate.

README.md
rockerBOO: I get that added!

README.md
rockerBOO: Maybe we say the sorters, pickers, finders can all be composed modularly?

tami5: Oh thats it. that should do it

rockerBOO:
@tami5 we can refine that phrase in another pull if you have any suggestions.

PR #325, opened 2020-12-11
refactor: Move selector logic out of pickers

tjdevries:
Still WIP.


tjdevries:
Conni did this better

PR #329, opened 2020-12-11
Allow to list files in git submodules

mredaelli:
I need to be able to list files in git submodules, and probably this would be useful for someone else.

Please note that:
- this is my first lua code ever
- I also changed the default to `show_untracked` to `false`. Was it a mistake? If it was intended (but seems counterintuitive to me) I will switch it back

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -11,6 +11,10 @@ local git = {}
 
 git.files = function(opts)
   local show_untracked = utils.get_default(opts.show_untracked, true)
+  local recurse_submodules = utils.get_default(opts.recurse_submodules, false)
+  if show_untracked and recurse_submodules then
+    error("Git does not suppurt both --others and --recurse-submodules")
+  end
 
   -- By creating the entry maker after the cwd options,
   -- we ensure the maker uses the cwd options when being created.
@@ -19,7 +23,11 @@ git.files = function(opts)
   pickers.new(opts, {
     prompt_title = 'Git File',
     finder = finders.new_oneshot_job(
-      { "git", "ls-files", "--exclude-standard", "--cached", show_untracked and "--others" },
+      vim.tbl_flatten( {
+        "git", "ls-files", "--exclude-standard", "--cached",
+        show_untracked and "--others" or nil,
+        recurse_submodules and "--recurse-submodules" or nil
+      } ),
       opts
     ),
     previewer = conf.file_previewer(opts),
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -7,6 +7,20 @@ utils.get_separator = function()
   return pathlib.separator
 end
 
+utils.as_bool = function(x)
+  local t = type(x)
+  if t == "string" then
+    if x == "true" then return true
+    elseif x == "false" then return false
+    else return nil end
+  elseif t == "boolean" then
+    return x
+  elseif t == "number" then
+    return x ~= 0
+  end
+  return nil
+end
+
 utils.if_nil = function(x, was_nil, was_not_nil)
   if x == nil then
     return was_nil
@@ -16,6 +30,9 @@ utils.if_nil = function(x, was_nil, was_not_nil)
 end
 
 utils.get_default = function(x, default)
+  if type(default) == "boolean" then
+    return utils.if_nil(x, default, utils.as_bool(x))
+  end
   return utils.if_nil(x, default, x)
 end
 
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -7,20 +7,6 @@ utils.get_separator = function()
   return pathlib.separator
 end
 
-utils.as_bool = function(x)
-  local t = type(x)
-  if t == "string" then
-    if x == "true" then return true
-    elseif x == "false" then return false
-    else return nil end
-  elseif t == "boolean" then
-    return x
-  elseif t == "number" then
-    return x ~= 0
-  end
-  return nil
-end
-
 utils.if_nil = function(x, was_nil, was_not_nil)
   if x == nil then
     return was_nil
@@ -30,9 +16,6 @@ utils.if_nil = function(x, was_nil, was_not_nil)
 end
 
 utils.get_default = function(x, default)
-  if type(default) == "boolean" then
-    return utils.if_nil(x, default, utils.as_bool(x))
-  end
   return utils.if_nil(x, default, x)
 end
 


lua/telescope/builtin/git.lua
```diff
@@ -10,16 +10,25 @@ local conf = require('telescope.config').values
...
+  end
+  if recurse_submodules then
+    table.insert(args, "--recurse-submodules")
+  end
+
   pickers.new(opts, {
     prompt_title = 'Git File',
     finder = finders.new_oneshot_job(
-      { "git", "ls-files", "--exclude-standard", "--cached", show_untracked and "--others" },
+      args,
```
Conni2461, line 31: Lets just do this in place, similar to what we had before. `recurse_subodules and "--recurse-submodules"`. 
You can split up the table over multiple lines, if CI doesn't like the length

Conni2461:
Thanks for the PR :) Really good idea.
Welcome to lua. 

> I also changed the default to show_untracked to false. Was it a mistake?

I am not sure about that one. Could bother some people who expect that right now. @tjdevries needs to decide that one

lua/telescope/builtin/git.lua
mredaelli: I started with that, but I think if the `show_untracked` condition is false,  the `recurse_submodule` one is not passed to `git`. Possibly because the function stops getting arguments when finding a `false`. Can that be?

Conni2461: You are right. Ignore that :)

Conni2461: Okay it actually does work. You just have to do `show_untracked and "--others" or nil`. If you don't do this that would put `false` into the table. After that the table looks like that:
```
{ "git", "ls-files", "--exclude-standard", "--cached",
  [6] = "--recurse-submodules"
}
```
Now all we have to do is `vim.tbl_flatten`. Put that in front of the table. After that `[5]` will no longer be `nil`. Now its `--recurse-submodules`.

mredaelli: I see! Question: would it be better to use `tbl_flatten` directly inside the `oneshot` function (I forget the name now), or maybe even discard `false` values to avoid using it at all? Semantically it would seam to make sense, and I'm assuming this kind of pattern might recur often

Conni2461: We have
```lua
    finder = finders.new_oneshot_job(
      vim.tbl_flatten { conf.vimgrep_arguments, opts.word_match, search},
      opts
    ),
```
for `live_grep` for example. So just define your table as argument for `vim.tbl_flatten`.

mredaelli: Sorry it's taking me a while, I was on another PC.

Another small roadblock. `utils.get_default` seems problematic. If I run
`:Telescope git_files show_untracked=false` or `:Telescope git_files show_untracked=0`, that parameter is treated as a string, not a bool, and so when I do `show_untracked and "--others" or nil` I still get `"--others"`.

Should the ex-command invocation syntax be different?

Conni2461: No problem. Take all the time you need.
Oh that is an interesting problem. I never use `:Telescope`, only `:lua require` but we should fix that, probably inside utils.get_default. You wanna do it? And thanks for discovering this one :)

mredaelli: Hmmm I wonder how to approach this.

We might still want to pass `false` as a string in some cases, so automatically converting the string to the bool feels wrong.

My hunch would be, for boolean flags, to accept these syntaxes:
```vim
:Telescope git_files show_untracked
:Telescope git_files !show_untracked
```
and convert them to `true` and `false` respectively.

But I get an error:
```
Error detected while processing function <SNR>54_load_command:
line   11:                                                                                                                                        
E684: list index out of range: 1
E15: Invalid expression: opt[1]
```
which feels more "internal" than `telescope` :p What do you think? Is it even `vim`-ish?

mredaelli: Or maybe, more simply, a utility function that converts the value;  and then the called function, which is the only one who knows what type to expect, calls the appropriate one (`utils.to_bool` or `utils.to_num`)?

Conni2461: `:Telescope git_files !show_untracked` that would just translate to writing `:Telescope git_files false` its not associated with a key. Lets just ping @tjdevries here. I don't know how to fix that from vim side. My take would be have `utils.get_default` have to handle this one but like 90% of the builtin dont even use `utils.get_default` so those are broken as well.
I personally always just do: `opts.show_untracked = opts.show_untracked or true`.

Conni2461: Yeah lets just do what you said. Did read that before writing my last comment. Make sure to check the type with `type()` so it does not run for bools

mredaelli: So I'd say we can't do:
```
  local show_untracked = utils.get_default(utils.as_bool(opts.show_untracked), true)
```
because if no option is passed we get a `nil` in `as_bool` and we can only return `false`, which would prevent the default of `true` to count.
So we could switch to 
```
  local show_untracked = utils.as_bool(utils.get_default(opts.show_untracked, true))
```
which works but is a bit silly (the default is already a bool, so why convert it).

So I'd rather suggest that we check the default value, and if it is a bool, then convert the passed parameter to bool, too.

This seems to work for all 4 combinations of `true`/`false` except when both flags are `true`, in which case I get:
```
[bat error]: '/home/t.../fatal: ls-files --recurse-submodules unsupported mode': No such file or directory (os error 2)
```

No idea where that comes from. Is `opts` being passed around too much? Should I remove the `opt` that are only for `git` before passing it elsewhere?

Conni2461: I would say `utils.as_bool` could return a nil. When passed in a value that does not convert or passed in nil.
I would say just push what you have right now.
I will look at it later. I can't right now. But i would say because its a [bat error] its the previewer.

Conni2461: Okay that is a git error. Just try running: `git ls-files --exclude-standard --cached --others --recurse-submodules`.
`--others` and `--recurse-submodules` does not work together.

mredaelli: Wow I thought I tried that. Weird though, I don't see the issue myself :P 

Anyway, I'll fix `as_bool` to pass `nil` through. 
Do you also want it to accept only `true`, `false`, `"true"`, `"false"`, `0`, `1`, `"0"`, `"1"` (or a subset of that) and return `nil` otherwise? I think standard `vim`'s behaviour would be "every string that does not start with a digit greater than 0 is falsy", but that won't work here because `"true"` need to be `true` (I think), so... I'll leave that choice to you :)

But that's another reason why this should probably be handled at a lower level

Conni2461: I would say if type bool return input.
If type string then return true if "true", elif "false" then return false otherwise return nil.
if type int (i actually don't know that type 🤣 ) if 0 false, else true.
else return nil

mredaelli: There :)

Conni2461: Okay [man page](https://git-scm.com/docs/git-ls-files#Documentation/git-ls-files.txt---recurse-submodules) actually says that there is currently only support for `--cached` mode, so no `--others`. So we need to make sure only one of those is true after that i think i will just merge. Unless you are not yet done :)

mredaelli: Hmm I thought about that, but
1. if they add support later, we'd have to remember and remove the check. Besides, it's the user's problem that they specifically ask for something that won't work, no? :P But you are right, they'd probably end up here opening an issue, just like I didn't know it either ^_^
2. not sure what to do. Just print and exit? Is there a pattern for echo-ing errors?

Conni2461: We can't have it error out without a good error message.
So you could print a hint with `print(msg)` and just default to one of those or error out with `error(msg)`

mredaelli: Done

Conni2461:
LGTM. Thanks :) 
We should probably update all builtin pickers to use the utils function. So opts works with `:Telescope `. 
We could do this now or later. Depends on what you want

Conni2461:
@mredaelli https://github.com/nvim-telescope/telescope.nvim/pull/398 does the converting on a more general level now. So can you please rebase,  remove your converting, check if everything still works and then i merge?

Thanks for your work and patience :) And I'm sorry that some of your work will not end up in this repo now

Conni2461:
I haven't received an reply from you in over a month so i decide to just finish it myself :)
Still thanks for your work and for that great feat :)

mredaelli:
Wow sorry, I completely forgot :/
Thanks for the guidance!

PR #331, opened 2020-12-13
change quickfix make_entry to use column displayer

sunjon:
This is a small tidyup to make the finder more usable for me.
Just noticed that there's another quickfix branch active; If this is already taken care of, feel free to close.

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -233,24 +233,31 @@ function make_entry.gen_from_quickfix(opts)
   opts = opts or {}
   opts.tail_path = get_default(opts.tail_path, true)
 
-  local make_display = function(entry)
-    local to_concat = {}
+    local displayer = entry_display.create {
+    separator = "▏",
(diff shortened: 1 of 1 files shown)


lua/telescope/make_entry.lua
```diff
@@ -233,24 +233,31 @@ function make_entry.gen_from_quickfix(opts)
...
-      table.insert(to_concat, ":")
     end
 
-    table.insert(to_concat, entry.text)
-
-    return table.concat(to_concat, "")
+    return displayer {
+      entry.lnum .. ":" .. entry.col,
+      entry.text:gsub(".* | ", ""),
+      filename or "",
```
Conni2461, line 259: If one `filename` is `nil` it will be completely ignored. So if you will pass in `""` it will still draw the separator, but if you just do `filename` the column will just disappear if `hide_filename = true`.

Conni2461:
LGTM besides my nitpick :) Thanks sunjon

PR #332, opened 2020-12-13
Update autocommands/highlights and help_tags previewers

Conni2461:
@sunjon 
I updated the highlights previewer as you suggested.
I also updated the autocommand previewer so we don't have to set all lines for every new entry, we just have to set them once per autocmd group and then use that buffer and jump back and forth. Result more speeeeeeeed 🤣 
Edit: Also you didn't jump before, so if the group was huge we ended up not even showing the correct result. Example `filetypedetect`

Edit 2: Also fixed a crash when `find_files` or similar wants to read a directory. Happens with `git_files` and submodules.
Edit 3: Also make help_tags super fast :)

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -616,7 +616,7 @@ internal.highlights = function(opts)
       end)
       return true
     end,
-    previewer = previewers.display_content.new(opts),
+    previewer = previewers.highlights.new(opts),
   }):find()
 end
 
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -321,22 +321,55 @@ internal.vim_options = function(opts)
 end
 
 internal.help_tags = function(opts)
+  local all_tag_files = {}
+  local all_help_files = {}
+  for _, v in ipairs(vim.split(vim.fn.globpath(vim.o.runtimepath, 'doc/*', 1), '\n')) do
+    local split_path = vim.split(v, path.separator, true)
+    local filename = split_path[#split_path]
+    if filename == 'tags' then
+      table.insert(all_tag_files, v)
+    else
+      all_help_files[filename] = v
+    end
+  end
+
+  local delim = string.char(9)
   local tags = {}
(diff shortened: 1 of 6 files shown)


sunjon:
Thanks for taking care of the autocommand viewer. Works great!

The Highlights previewer currently breaks for me as it's not accomodating for the `link` syntax, eg:
```
TermCursor     xxx cterm=reverse gui=reverse
                   links to Cursor
```
The second line fails to be matched by the regex, and it will try to calculate `startPos` from a nil value.

Conni2461:
Hmm why is it not all one line for you. So the output of `highlights`? For me at least it is. So one line and then new hlgroup.
But i can try to fix it. Give me like 10 mins. Thanks for testing :)

sunjon:
The multi-line formatting is very similar to the autocommands.
I've always know it to be that. I wonder what's coercing your setup to display it to single line?

Conni2461:
Is also a single line with `nvim -u NORC`. But does the new version help

sunjon:
Works fine now!

Conni2461:
> On this particular finder, the vim syntax setting leads to a bit of confusion. Keywords, operators and numbers are highlighted with the exact same colors you may be visually trying to track down. Ideally only selection, the highlights xxx column and the results window would be colorized.

Yeah wasn't sure about that one

Conni2461:
@sunjon btw if you write new vim buffer previewer, feel free to ask any question considering the new interface. 
Basically you no longer need to create a buffer and define window opts. You can just access it and we keep track of it so we are able to delete them after closing the picker.

Conni2461:
@sunjon The last commit should make help_tags super fast as well. Could you give it a try. Could you give it a try? Also does the new code make sense? I think you wrote the first iteration? Am i correct?

sunjon:
It's instant now. Nice work

Conni2461:
Thanks @sunjon :) If you don't found any issues I would like @tjdevries to take a short look at it and then we can merge.

Conni2461:
@sunjon I might have an idea how we can do this even better.

Issue #334, opened 2020-12-14
Feature: recursive tree-like menus


oberblastmeister:
I added the ability to create recursive tree-like menus in telescope. It is kinda of like vim menus but uses lua tables. I don't know if it is the right choice but I have forked the repo [here](https://github.com/oberblastmeister/telescope.nvim) (maybe I should have created a separate extension?). I am pretty new to lua so I am not sure how the code is. I added a test builtin function to showcase the feature.

Conni2461:
That a really neat idea. I briefly looked at the branch and i think you are doing some things a little bit too complicated, if i am allowed to say that :) 
I have a file browser https://github.com/nvim-telescope/telescope.nvim/pull/290, which also works recursive, i haven't finished all there but it might give you an idea on how to do it a little bit easier.

weilbith:
Any progress here? Supporting trees in general is pretty amazing. Many useful cases. Language servers can return document symbols as tree. Language servers support call hierarchies. And many more. 
I was thinking if this might be worth a new plugin. But I think the intersection with Telescope is too big. And then Telescope is just already super optimized, well tested, good community, many well interacting features.

PR #337, opened 2020-12-14
Feature: recursive tree like menus

oberblastmeister:
from [issue](https://github.com/nvim-telescope/telescope.nvim/issues/334)


lua/telescope/builtin/init.lua
```diff
@@ -35,6 +35,8 @@ builtin.git_branches = require('telescope.builtin.git').branches
 builtin.git_status = require('telescope.builtin.git').status
 
 builtin.builtin = require('telescope.builtin.internal').builtin
+builtin.menu = require('telescope.builtin.internal').menu
+builtin.test_menu = require('telescope.builtin.internal').test_menu
```
Conni2461, line 39: We need a way to provide these tables. I know its currently just here for testing purposes but i have absolutely no idea how a user can define them in a good way. I am open for all suggestions.

lua/telescope/builtin/internal.lua
```diff
@@ -45,6 +45,73 @@ internal.builtin = function(opts)
   }):find()
 end
 
+do
```
Conni2461, line 48: This do block does nothing.

lua/telescope/builtin/internal.lua
```diff
@@ -45,6 +45,73 @@ internal.builtin = function(opts)
...
+  internal.menu = function(opts)
+    if opts.callback == nil then
+      vim.cmd [[echoerr 'There is no callback']]
+      return
+    end
```
Conni2461, line 53: Idea what if each leaf can have a callback attached to itself. Because i believe we don't want to do the same for every single leaf, or do we.

lua/telescope/builtin/internal.lua
```diff
@@ -45,6 +45,73 @@ internal.builtin = function(opts)
...
+            -- it is a leaf
+            opts.callback(key)
+            actions.close(prompt_bufnr)
+          elseif type(value) == "table" then
+            -- it is a node
+            actions.close(prompt_bufnr)
+            -- recurse
+            internal.menu{t = value, callback = opts.callback}
+            -- sometimes does not start insert for some reason
+            vim.api.nvim_input('i')
```
Conni2461, line 76: There is an actually workaround for this issue. There is a local function inside `lua/telescope/actions/init.lua` which is used for `actions.run_builtin` and is called `do_close`. We could make that one public like this: `actions._do_close` similar to `actions._goto_file_selection` so you can do `actions._do_close(prompt_bufnr, true)` rather the `actions.close` and `nvim_input`

lua/telescope/builtin/internal.lua
```diff
@@ -45,6 +45,73 @@ internal.builtin = function(opts)
...
+          else
+            vim.cmd [[echoerr 'value must be leaf or table']]
```
Conni2461, line 78: `error(msg)`

lua/telescope/finders.lua
```diff
@@ -248,6 +248,38 @@ function StaticFinder:new(opts)
...
+      -- entry.index = counter
+      table.insert(results, entry)
+    end
+
+    counter = counter + 1
+  end
+
+  return setmetatable({ results = results }, self)
+end
+
```
Conni2461, line 282: I believe we don't need a new finder for this. Just do that in `internal.menu`

lua/telescope/make_entry.lua
```diff
@@ -57,6 +57,40 @@ do
...
+      else
+        -- it is a node
+        return setmetatable({
+          key,
+          value,
+        }, mt_tree_entry)
+      end
+    end
+  end
+end
```
Conni2461, line 92: Also can be realized easier without a metatable.

lua/telescope/builtin/internal.lua
Conni2461: You also might wanna consider insert a `..` value on each iteration. So we are able to go up one table. Obviously not on the top level :)

Conni2461:
Looks like a really cool feature. Thanks :)

Left some comments. Feel free to disagree with anything that i mentioned. Especially with my opinion that you don't need a new finder, because i am not sure about that one.

lua/telescope/builtin/internal.lua
oberblastmeister: Definitely a good idea, I was thinking of the exact same thing about the going back. We could have a global callback and then a callback for each leaf to override.

lua/telescope/builtin/internal.lua
oberblastmeister: I should have put a comment for the do block. I was preparing to make the function have some state. For example, there would be a table of all user selections, and a stack for previous user selections so going back would just pop and selecting would push. This would allow me to implement going back to the previous menu like in your next comment.

lua/telescope/finders.lua
oberblastmeister: I agree with not using another finder. I was not very familiar with the code base so I was trying out different ways to implement this. I think I have settled on using a different entry maker for the tree menu instead of creating a whole new finder.

lua/telescope/make_entry.lua
oberblastmeister: It may be a bit confusing using timetables, I agree. I was trying to keep the same style as the other entry maker functions use them too.

oberblastmeister:
Thanks for your comments, they are very helpful. This project is very cool and is also helping me learn some lua!

lua/telescope/make_entry.lua
Conni2461: These metatables are only used for builtins like `find_files`, `live_grep`, buitlins which eventually deal with 10000 and more entries and having display and ordinal the same for each entry requires a lot more ram usage, so it makes sense to have it with those `lookup_keys`. But you will end up with lets say 50 entrys max and so a simpler version is good enough in my opinion. You can find these simples once after the metatables. On `master` they start at line: 216. Or they can be found at the builtin picker definition itself, if it is even simpler (example: `telescope.builtin.internal.keymaps` starting line 553 on master)

lua/telescope/finders.lua
oberblastmeister: I have an issue though, how would I use the regular static finder instead creating a new one. My entry maker accepts two arguments while the entry maker in the regular static finder can only accept one argument. Also what would I set the index to because the regular static finder sets the index to the value which is not correct for the tree.

Conni2461: I though you can do something like this before the picker. Correct me if i am wrong i am just rambling.
```lua
local level = opts.menu
if opts.passed_level then
  level = opts.menu[opts.passed_level] -- that needs some more work for levels below 1. Maybe passed_level should be a table like that. [ level1_key, level2_key, level3_key ] and we just table.insert newest level when a branch is selected and if branch == '..' then remove last level.
end
local results = {}
for k, v in pairs(opts.menu) do
  if type(v) == "string" then -- leaf
    table.insert(results, { leaf = v, menu = opts.menu }) -- and have each element have the table. So you also have it when we hit enter
  elseif type(k) == "string" then
    table.insert(results, { branch = k, menu = opts.menu }) -- when selecting you can then just check if branch is not nil. if thats the case the restart menu
  else
    -- we should never get here. Am i correct?
  end
end
table.insert(results, { branch = '..', menu = opts.menu })
```

And in entry maker you could do
```lua
function(entry)
  return {
    value = entry,
    display = display.leaf or display.branch,
    ordinal = display.leaf or display.branch,
  }
end
```
Does that makes sense to you or should i elaborate a little bit more. Also i don't get what you mean with index 😆

oberblastmeister: Wow that is much better. I will start implementing that. Just so I am understanding we are going to preprocess the tree from something more user-friendly to an internal representation that we can use with an entry maker.

oberblastmeister: Also the passed level thing I am implementing right now. We should use a stack like you said so it can remember multiple `..`

Conni2461: Basically i was thinking we just have to reduce the table to the current level we are looking at. That was my idea. But my code might not work i just wrote that inside github. So let me know if you run into issues :)

I am also thinking we can do something cool and can display the rest of the tree of the current selected branch in the previewer. There is an interface so that every entry can have a defined function which then will be called and allows us to do `vim.api.nvim_buf_set_lines` for the previewer. We don't use that interface here anymore so there is no example. So feel free to ask me for information if you wanna do it.

oberblastmeister: Don't worry your comments are not annoying :smile:. They are helping me a lot especially since I am new to all this stuff. Actually this is my first ever pull request so your guidance is really helpful.

lua/telescope/actions/init.lua
```diff
@@ -226,7 +226,7 @@ end
 actions.run_builtin = function(prompt_bufnr)
   local entry = actions.get_selected_entry(prompt_bufnr)
 
-  do_close(prompt_bufnr, true)
+  _do_close(prompt_bufnr, true)
```
Conni2461, line 229: You have to make the `local do_close` public like this: `actions._do_close` and then update the current impl. But you can wait a little bit. Someone else was also doing this, to fix a bug within `buitlin.git_status`

Conni2461:
Okay i am just playing around with this. And i feel like its kinda unnecessary to go through 2 classes. It doesn't make things easier for us or for the person who creates the tree.
Also can you please follow the way how builtins are definied. `function(opts)` because without outs users can specify a specific theme for that picker. So if someone wants to use this as dropdown right now this isn't possible.

Because i played around with it. I have it working without any classes and so, basically just with what i proposed earlier. Is there any reason why we need that abstraction level?

oberblastmeister:
Thanks for your comment. You are right that it is unnecessary to go through 2 classes so I just used functions instead and used `table.insert` and `table.remove` for the stack. I will put the opts in but I have a question about how to add the root tree. Should I make the function accept two arguments like this `function(root, opts)` or should I do `function(opts)` and have the root tree be inside opts as a key value pair?

Conni2461:
> function(opts)

Would follow the current style.

Okay we are getting close. But i still don't like the function abstraction stuff at all. Because i don't actually get why its there. Maybe you need to help me.
I think its way easier to define a tree like that:
```lua
  {
    "top level leaf",
    "another top level leaf",
    ["a node"] = {
      "second level leaf",
      "another second level leaf",
      ["second level node"] = {
        ["another third level leaf with specific callback"] = function() print('this is a specific callback') end,
        "third level leaf",
      },
    }
  }
```
So i as user i don't need to follow the weird stuff where i have to define the root, or where i ave to define a node. So its way more readable and we have all the stuff we need. If we iterate over a level we have `k, v`. If `v` is a table than we will have a node and `k` is the name of the node, if `v` is a function then we will have a leaf with a specific callback and `k` is the name of the leaf and if v is a string then we just have a leaf with fallback callback. And then at the beginning of top we just have to do the `preprocess` thing. And then all we have to do is keeping track of the current level chain. And if we have a level chain and the level chain is bigger then 0 then we need to insert `..`

lua/telescope/finders.lua
```diff
@@ -231,13 +231,20 @@ function StaticFinder:new(opts)
 
   local entry_maker = opts.entry_maker or make_entry.gen_from_string()
 
+  -- print('entry_maker')
+  -- dump(entry_maker)
+
```
Conni2461, line 236: Please remove everything in this file at some point. Thanks :)

oberblastmeister:
You are right, it is a little bit clunky. Removing the functions is probably ok. I have some concerns though because adding new options would be much harder. The code that you gave there is no way to set specific options for each node or leaf (only one option for the leaf, callback). For example, what if the user wants to add a description for a leaf *and* a specific callback for that leaf. The user might also want to override the title for a specific node. Removing the functions is probably ok but I don't think we can make it that simple. Maybe this is better:
```lua
menu.better_test = function()
  menu.open {
    tree = {
      "top level leaf",
      "another top level leaf",
      ["a node"] = {
        -- instead of directly setting the key to the value, we set it to a table of options, [1] is the contents
        {
          "second level leaf",
          "another second level leaf",
          ["second level node"] = {
            {
              -- first way, can only have one option
              ["another third level leaf with specific callback"] = function()
                print('this is a specific callback')
              end,
              -- second way, we can have multiple options
              {
                "third level leaf with a specific callback different",
                -- for example we might want to set the description
                description = 'this is a description',
                callback = function()
                  print("this is a specific callback")
                end,
              },
              "third level leaf",
            },
            -- because it is a table inside of a table, we can set options as the contents of the tree will be node[1]
            -- other options will be anything except [1]
            title = 'this title overrides the defualt one'
          }
        }
      }
    },
  }
end
```
The functions are gone and it is a little bit better. The nodes are not direct tables and can also fit options. I don't think I like setting the name of the leaf equal to the callback in the first way because 1. there cannot be other options 2. there is not a distinction between nodes and leaf. I like that leafs are number value pairs while nodes are string value pairs.

Conni2461:
Yes thats good as well. Thanks for clarifying things

oberblastmeister:
I have implemented the above using the `process` function that is recursive and walks the entire user tree, adds `..` where needed, and also makes the tree entry friendly. This was contrasted with my previous implementation that only walked the processed the tree when the user selected the node. The drawbacks are not immediately known but processing the entire tree first allows to define methods on that tree like `:iter` method and a `:display` method. The display method is not implemented yet but in the future will probably be used for the previewer, so you can have a tree view of the entire node in the preview window as you said earlier. I am thinking that the display function should print something similar to the unix `tree` command. I will probably need some help with implementing the previewer because I am not familiar with any of that code. This would not be possible without fully walking the tree. The code is a lot but I think that it is worth it because it allows maximum configuration as in the user can configure each node or tree if they want to and also offers powerful manipulation of the internal tree that was converted from the user tree. It also allows the user to specify the leaf as a number string pair or a number table pair and a node as a string table pair. I have also added some automated tests for the tree processor.

tjdevries:
Hey, I have some suggestions for this PR but don't have time to type them up at the moment (and need to think them through more clearly). We may want to put a pause on this so you don't go too far down a direction that might not make sense for the project.

Is that fine with you? I don't want to waste your time :)

Sorry for any delay, I'm applying for new jobs and it's Christmas time :laughing:

oberblastmeister:
Yes that is okay. My time isn't very valuable anyway :laughing:

tjdevries:
Let's make a new external extension for making menus with this PR so that @oberblastmeister can keep doing super cool stuff with that .

https://github.com/nvim-telescope/telescope-menu.nvim

I've made you an admin @oberblastmeister .

Once you move that over there we can close this PR.

tjdevries:
Spoke w/ @oberblastmeister . Will move to other repo :)

PR #339, opened 2020-12-15
Fix preview scrolling

eth3lbert:
This pr should fix preview scrolling issue when using less 😄.


Conni2461:
Did i seriously break that one again. Get me out of here 🤣 
Thanks @eth3lbert. And i am soooooo sorry

PR #342, opened 2020-12-17
Fix builtin.builtin again.

Conni2461:
I didn't break it this time :)
Lets no longer use quickfix to generate the entries because we no longer have lnum and col

CC @cempassi


cempassi (approved):
Tested locally, LGTM!

Conni2461:
I would just hijack that pr to change the hlgroups like @clason suggested. Don't feel like opening a new PR for renaming two strings 😆 After that i merge.

Issue #335, opened 2020-12-14
Scrolling (C-p C-n) over results makes them loose color until the query changes (adding or removing)


felipesere:
### Description
When I trigger `find_files` or any other picker, they initially show with colored icons just fine.
When I then scroll to the any item, items I have scrolled over loose their color.

**Expected Behavior**
I expect the the currently selected result to look different, but all other results to use the appropriate colored icons.

**Actual Behavior**
Items scrolled over turn white until changing the query refreshes the screen.

### Details

<details><summary>Reproduce</summary>
1. Open vim
2. :Telescope find_files
3. Ctrl-P a couple of times
4. Bang, colors in the search result are gone at the bottom. (screenshots)

Settings for `telescope.nvim`:

</details>

<details><summary>Environment</summary>

nvim --version
NVIM v0.5.0-nightly-131-g104749440
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/bin/clang -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNDEBUG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/tmp/neovim-20201125-14291-1afp395/build/config -I/tmp/neovim-20201125-14291-1afp395/src -I/usr/local/include -I/tmp/neovim-20201125-14291-1afp395/deps-build/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX11.0.sdk/usr/include -I/usr/local/opt/gettext/include -I/tmp/neovim-20201125-14291-1afp395/build/src/nvim/auto -I/tmp/neovim-20201125-14291-1afp395/build/include
Compiled by felipesere@Felipes-MacBook-Pro.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/Cellar/neovim/HEAD-1047494/share/nvim"

Run :checkhealth for more info

macOS BigSur
[...]

Issue #190, opened 2020-10-25
"Use colors from nvim-devicons" cause high CPU usage, sometimes nvim process gets killed


smartding:
Right after I pulled the latest update of telescope, I noticed a high CPU usage when running `telescope.builtin.git_files` in source code tree with 7k+ files.  And sometimes the whole nvim process is killed, dropping me back to command line.

Then I locally deleted the commit `a8e780708698a4` `feat: Use colors from nvim-devicons`, everything returns normal. 

Can we revert this feature until performance is improved? Or provide an option to disable this feature.

tjdevries:
Huh, that's very weird -- I can't figure out why that would happen. Does it happen immediately when running git files? can you paste your setup as well?

I did add a way to turn it off though: 113a516 you can use `color_devicons = false` in your configuration to turn the feature off.

tjdevries:
(but we shouldn't close this issue til I can track down the performance issue)

smartding:
> Huh, that's very weird -- I can't figure out why that would happen. Does it happen immediately when running git files? can you paste your setup as well?
> 
> I did add a way to turn it off though: [113a516](https://github.com/nvim-lua/telescope.nvim/commit/113a516c63fb6e1ae1fb01173ed8fefdf450b08a) you can use `color_devicons = false` in your configuration to turn the feature off.

It doesn't happen immediately after I run git_files, cpu usage doesn't go up much after I type in a query. But once I start deleting some characters with backspace, cpu usage goes through the roof.

Here's my config for telescope:
```lua
local telescope = {}

telescope.edit_dotfiles = function()
  local opts = require('telescope.themes').get_dropdown{
    cwd = '~/.config',
    -- add symbolic links too, most of my dotfiles are symlinked
    find_command = { 'fd', '--type', 'f', '--type', 'l' },
    prompt_title = 'dotfiles',
    previewer = false,
    borderchars = {
      prompt = {"─", "│", " ", "│", "┌", "┐", "│", "│"},
      results = {"─", "│", "─", "│", "├", "┤", "┘", "└"},
    }
  }

  require('telescope.builtin').find_files(opts)
end

telescope.lsp_workspace_symbols =  function()
  local input = vim.fn.input('Query: ')
  vim.api.nvim_command('normal :esc<CR>')
  vim.api.nvim_command('echom "' .. input .. '"')
  if not input or #input == 0 then return end
  require('telescope.builtin').lsp_workspace_symbols{
      query = input
  }
[...]

PR #344, opened 2020-12-18
fix: Better highlights

tjdevries:
Related: #190

Closes: #233
Closes: #335

Could use a bit of other people testing as well, but should work. Does add a bit of performance overhead but it's very small. It only calls sorting on one extra item, and displaying on one extra item... so.... that's like nothing :)


Conni2461:
I have figured it out why it was broken for @clason and me.
When `color_devicons` is set to `false` it doesn't work. With `true` it works as expected.

PR #346, opened 2020-12-18
Add more colour

sunjon:
add more colors to finders that would benefit from it now that highlighting is fixed/improved.

Previously I've assigned hl_groups as I'd seen appropriate -  Comment for less important text, Number for numbers..
This isn't ideal, and I'm wondering if/how the current Telescope* groups can be expanded to allow a reasonable amount of customisation without finder specifc configs.

lsp_symbols
I've altered the LSP_*_symbol finder to be a bit more readable and tried to relate the display colour to the hl_group used in Vim. As this will change per langauge, I'm not sure what the best way to do this will be. If we decide on a way forward, the Treesitter and perhaps ctags finders can use this display method.

All input welcome

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -34,7 +34,7 @@ internal.builtin = function(opts)
     prompt_title = 'Telescope Builtin',
     finder    = finders.new_table {
       results     = objs,
-      entry_maker = make_entry.gen_from_quickfix(opts),
+      entry_maker = make_entry.gen_from_builtin(opts),
     },
     previewer = previewers.builtin.new(opts),
     sorter = conf.generic_sorter(opts),
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -33,17 +33,15 @@ internal.builtin = function(opts)
   pickers.new(opts, {
     prompt_title = 'Telescope Builtin',
     finder    = finders.new_table {
-      results     = objs,
-      entry_maker = function(entry)
-            return {
-              value = entry,
-              text = entry.text,
-              display = entry.text,
-              ordinal = entry.text,
-              filename = entry.filename
-            }
-          end
-    },
+      results = objs,
+      entry_maker = function(entry)return {
+        value = entry,
+        text = entry.text,
+        display = entry.text,
+        ordinal = entry.text,
+        filename = entry.filename
+        }end
+      },
     previewer = previewers.builtin.new(opts),
     sorter = conf.generic_sorter(opts),
     attach_mappings = function(_)
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -34,7 +34,15 @@ internal.builtin = function(opts)
     prompt_title = 'Telescope Builtin',
     finder    = finders.new_table {
       results     = objs,
-      entry_maker = make_entry.gen_from_builtin(opts),
+      entry_maker = function(entry)
+            return {
+              value = entry,
+              text = entry.text,
+              display = entry.text,
+              ordinal = entry.text,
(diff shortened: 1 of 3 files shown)


clason:
> This isn't ideal, and I'm wondering if/how the current Telescope* groups can be expanded to allow a reasonable amount of customisation without finder specifc configs.

I wouldn't worry about "finder specific configs" -- other plugins also have specific highlight groups, and as long as they have some half-reasonable default link, that is the best way.

lua/telescope/make_entry.lua
```diff
@@ -283,6 +283,77 @@ function make_entry.gen_from_quickfix(opts)
...
+      ["Function"] = "Function",
+      ["Property"] = "Operator",
+      ["Variable"] = "SpecialChar",
+    }
+
+    return displayer {
+      {entry.lnum .. ":" .. entry.col, "LineNr"},
+      entry.symbol_name,
+      {"[", "TelescopeBorder"},
+      {entry.symbol_type, type_highlight[entry.symbol_type],type_highlight[entry.symbol_type]},
```
Conni2461, line 325: `type_highlight[entry.symbol_type]` is here twice

Conni2461:
Thanks sunjon :). As always looks great and more colors is always nice to have.

> If we decide on a way forward, the Treesitter and perhaps ctags finders can use this display method.

I can't comment for treesitter but ctags will not be possible with that method because we don't have a lnum at that point, so we need to store the code line coming from the tags file. Also we do the line splitting inside the entry_maker. But we could update ctags entry_maker to support colors. It already uses the displayer so it should be trivial

lua/telescope/make_entry.lua
sunjon: that was for the "override match colors" . I'm not sure that it's needed here, but that's what was the intention.

Conni2461: Oh i have not jet implemented that. So right now i only process the first element(string) and the second element hlgroup. But if highlighting fix is about to get merged we can think about this.

sunjon:
I added a make_entry() function for `builtins`. Althought it had been taken care of since I forked for this PR, I think that as this is an area we control it would be nice to introduce a single line description string for finders as standrard,(so I've left `gen_from_builtin()` in there)

Conni2461:
Yeah i introduced a super simple one yesterday because it was broken with the changes you made to gen_from_quickfix.
For a standard one we have `gen_from_string` which already will be used if no `entry_maker` is set. So i don't think we need another one. Most of the times its just easier to define the function which returns the correct table anyways.

sunjon:
I'm done with the changes for now. I'll make further improvements once the highlights PR has been merged.

lua/telescope/builtin/internal.lua
```diff
@@ -35,14 +35,14 @@ internal.builtin = function(opts)
...
-        }
-      end
+            return {
+              value = entry,
+              text = entry.text,
+              display = entry.text,
+              ordinal = entry.text,
+              filename = entry.filename
+            }
+          end
```
Conni2461, line 45: Now i am really nitpicking. 🙈 spacing

lua/telescope/make_entry.lua
```diff
@@ -311,10 +387,10 @@ function make_entry.gen_from_buffer(opts)
...
 
     return displayer {
-      entry.bufnr,
-      entry.indicator,
-      display_bufname .. ":" .. entry.lnum,
-    }
+        {entry.bufnr, "Number"},
+        {entry.indicator, "Comment"},
+        display_bufname .. ":" .. entry.lnum,
+      }
```
Conni2461, line 393: spacing

Conni2461 (approved):
LGTM. Feel free to merge if you are done. I think its a straight improvement. Thank you very much :)

PR #351, opened 2020-12-19
More colours

sunjon:
added colours for git hashes in relevant finders and converted it to use displayer.
more work on LSP/TS colors.
Restoration of filename for workspace finder and addition of inline symbol preview for document view (as filename is not useful).

### DIFF
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -55,11 +55,12 @@ lsp.document_symbols = function(opts)
     return
   end
 
+  opts.ignore_filename = true
   pickers.new(opts, {
     prompt_title = 'LSP Document Symbols',
     finder    = finders.new_table {
       results = locations,
-      entry_maker = make_entry.gen_from_symbols(opts)
+      entry_maker = make_entry.gen_from_lsp_symbols(opts)
     },
     previewer = conf.qflist_previewer(opts),
     sorter = conf.generic_sorter(opts),
@@ -169,7 +170,7 @@ lsp.workspace_symbols = function(opts)
     prompt_title = 'LSP Workspace Symbols',
     finder    = finders.new_table {
       results = locations,
-      entry_maker = make_entry.gen_from_symbols(opts)
+      entry_maker = make_entry.gen_from_lsp_symbols(opts)
     },
     previewer = conf.qflist_previewer(opts),
     sorter = conf.generic_sorter(opts),
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -55,7 +55,7 @@ lsp.document_symbols = function(opts)
     return
   end
 
-  opts.ignore_filename = true
+  opts.ignore_filename = opts.ignore_filename or true
   pickers.new(opts, {
     prompt_title = 'LSP Document Symbols',
     finder    = finders.new_table {
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -166,8 +166,9 @@ lsp.workspace_symbols = function(opts)
     return
   end
 
-  opts.ignore_filename = opts.ignore_filename or false
-  opts.hide_filename = opts.hide_filename or false
+
+  opts.ignore_filename = utils.get_default(opts.ignore_filename, false)
+  opts.hide_filename = utils.get_default(opts.hide_filename, false)
 
   pickers.new(opts, {
     prompt_title = 'LSP Workspace Symbols',
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -166,6 +166,9 @@ lsp.workspace_symbols = function(opts)
     return
   end
 
+  opts.ignore_filename = opts.ignore_filename or false
+  opts.hide_filename = opts.hide_filename or false
+
   pickers.new(opts, {
     prompt_title = 'LSP Workspace Symbols',
     finder    = finders.new_table {
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -238,14 +238,14 @@ function make_entry.gen_from_quickfix(opts)
   opts = opts or {}
   opts.tail_path = get_default(opts.tail_path, true)
 
-    local displayer = entry_display.create {
+  local displayer = entry_display.create {
     separator = "▏",
     items = {
       { width = 8 },
       { width = 50 },
-      { remaining = true },
-    },
-  }
+      { remaining = true }
+      }
+    }
 
   local make_display = function(entry)
     local filename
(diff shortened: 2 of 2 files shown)


lua/telescope/builtin/lsp.lua
```diff
@@ -55,11 +55,12 @@ lsp.document_symbols = function(opts)
     return
   end
 
+  opts.ignore_filename = true
```
Conni2461, line 58: Lets do `opts.ignore_filename = opts.ignore_filename or true` so its still configurable

lua/telescope/make_entry.lua
```diff
@@ -238,14 +254,14 @@ function make_entry.gen_from_quickfix(opts)
...
     separator = "▏",
     items = {
       { width = 8 },
       { width = 50 },
-      { remaining = true },
-    },
-  }
+      { remaining = true }
+      }
+    }
```
Conni2461, line 264: Move these two two spaces back :)

lua/telescope/make_entry.lua
```diff
@@ -290,30 +306,45 @@ function make_entry.gen_from_quickfix(opts)
...
-      { remaining = true },
-      { width = 1 },
-    },
-  }
+      { width = 8 },
+      { width = 25 },
+      { width = 8 },
+      { remaining = true }
+      }
+    }
```
Conni2461, line 321: Same. spacing

lua/telescope/make_entry.lua
```diff
@@ -290,30 +306,45 @@ function make_entry.gen_from_quickfix(opts)
...
+
+    -- what to show in the last column: filename or symbol information
+    if opts.ignore_filename then
+      msg = vim.api.nvim_buf_get_lines(
+          bufnr,
+          entry.lnum - 1,
+          entry.lnum,
+          false
+        )[1] or ''
+      msg = vim.trim(msg)
```
Conni2461, line 334: Hmm. I am not sure if i like this. Especially with having a previewer. I think its fine if i can enable and disable this. I do something like this for ctags

Conni2461: You can also define the `displayer.items` dynamically based on the opts. I am also doing this with ctags

lua/telescope/make_entry.lua
```diff
@@ -331,10 +362,8 @@ function make_entry.gen_from_symbols(opts)
     return displayer {
       {entry.lnum .. ":" .. entry.col, "LineNr"},
       entry.symbol_name,
-      {"[", "TelescopeBorder"},
-      {entry.symbol_type, type_highlight[entry.symbol_type], type_highlight[entry.symbol_type]},
-      {"]", "TelescopeBorder"},
-      filename,
+      {entry.symbol_type:lower(), type_highlight[entry.symbol_type], type_highlight[entry.symbol_type]},
```
Conni2461, line 365: Do we no longer have borders? Or do i need to add them back in my PR.

lua/telescope/make_entry.lua
Conni2461: Also for workspace symbols with filename i think something like this would be optimal: `filename:lnum:col ...`

sunjon: I think it's hard to know what the user will want.
I personally avoid previewers where I can, and I'd certainly use it. For document specific finders (lsp_document, treesitter) it makes more sense than having the filename repeated in the column N times.

If it were to be configurable to the point the column would be empty, then I'd do the dynamic sizing and reclaim the space.

sunjon: Agree on the reformatting of the filename:lnum:..

lua/telescope/make_entry.lua
```diff
@@ -424,23 +455,44 @@ function make_entry.gen_from_treesitter(opts)
...
+    ["associated"] = "TSConstant",
+    ["constant"]   = "TSConstant",
+    ["field"]      = "TSField",
+    ["function"]   = "TSFunction",
+    ["method"]     = "TSMethod",
+    ["parameter"]  = "TSParameter",
+    ["property"]   = "TSProperty",
+    ["struct"]     = "Struct",
+    ["var"]        = "TSVariableBuiltin",
+  }
```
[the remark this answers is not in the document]
sunjon: I intended to, but the LSP and Treesitter definitions actually differ quite a lot. I wasn't sure how to go forward so shipped it as is.

lua/telescope/make_entry.lua
Conni2461: I agree that it makes more sense for the `lsp_document` to not have the filename but it would also be awesome to not have the line shown if the user wishes it. I think we can do something like this to realize it
```lua
local displayer_items = {
  { width = 6 }, -- filename and lnum og would be 6 spacing for only lnum:col
  { remaining = true }, -- default for our symbol, using the remaining space
}

if opts.show_filename then
  displayer_items[1].width = 25 -- make it take enough space for filename. Don't know how much.
end

if opts.show_line then
  table.insert(displayer_items, 2, { width = 9 }) -- restrict symbols to 9 width and show_line will now take the remaining space.
end

  separator = " ",
  items = display_items
}
```

Now if show_line is set then we can construct it and if now we just pass in nil as third element. And it will be ignored. Does this make sense for you. I just like to provide the user with the most amount of freedom to create his picker like he wants to.

lua/telescope/make_entry.lua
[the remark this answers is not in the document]
Conni2461: Just give the table different names. Its just unnecessary to create the table for each entry. Lets just create it when the file gets required :)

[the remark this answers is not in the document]
sunjon: ah, I see what you mean. Done.

lua/telescope/make_entry.lua
sunjon: I'm not going to use them for the LSP/TS finders. Too visually noisy.
I do intend to use them in the registers finder and others though

lua/telescope/builtin/lsp.lua
```diff
@@ -55,11 +55,12 @@ lsp.document_symbols = function(opts)
     return
   end
 
+  opts.ignore_filename = opts.ignore_filename or true
```
Conni2461, line 58: We can use `utils.get_default()` here as well. Someone made a fix for it so it also works with `:Telescope fn opts=false`. Because currently we are getting a string and that breaks things.

lua/telescope/make_entry.lua
```diff
@@ -344,6 +344,7 @@ function make_entry.gen_from_lsp_symbols(opts)
 
   local displayer = entry_display.create {
     separator = " ",
+    hl_chars = { ['['] = 'TelescopeBorder', [']'] = 'TelescopeBorder' },
```
Conni2461, line 347: We can do this but this is not in master right now. So it will currently not apply the right color. But if this is merged i will rebase my PR immediately so everything comes together.

Conni2461 (approved):
Approved from me which is not worth as much as an approval from @tami5
Because he has admin rights 🤣

PR #354, opened 2020-12-19
More faster previewers

Conni2461:
Missing:
- [x] git diff without termopen
- [x] document how to add custom filetypes
- [x] `previewers.vim_buffer_vimgrep` and `previewers.vim_buffer_qflist` is now async
- [x] All `buffer` previewer are now async 🎉 
- [x] Refactor `file_maker` maybe make it configurable for rockerBoo
- [x] Make wrap configurable
- [x] Maybe just replace the old git previewers with the new one. Maybe even replace git log and do coloring with `nvim_buf_add_highlight`
- [x] `keep_last_buf` for pwntester
- [x] `job_maker` public with previewer utils

Also requires https://github.com/nvim-lua/plenary.nvim/pull/37 to work perfectly

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,21 +6,21 @@ Gaze deeply into unknown regions using the power of the moon.
 ## What Is Telescope?
 
 `telescope.nvim` is a highly extendable fuzzy finder over lists. Built on the latest
-awesome features from `neovim` core. Telescope is centered around 
-modularity, allowing for easy customization. 
+awesome features from `neovim` core. Telescope is centered around
+modularity, allowing for easy customization.
 
 Community driven built-in [pickers](#pickers), [sorters](#sorters) and [previewers](#previewers).
 
 ### Built-in Support:
-- [Vim](#vim-pickers) 
-- [Files](#file-pickers) 
+- [Vim](#vim-pickers)
+- [Files](#file-pickers)
 - [Git](#git-pickers)
 - [LSP](#lsp-pickers)
 - [Treesitter](#treesitter-pickers)
 
 ![by @glepnir](https://user-images.githubusercontent.com/41671631/100819597-6f737900-3487-11eb-8621-37ec1ffabe4b.gif)
 
- 
+
 <!-- You can read this documentation from start to finish, or you can look at the -->
 <!-- outline and directly jump to the section that interests you most. -->
 
@@ -162,6 +162,9 @@ require('telescope').setup{
     file_previewer = require'telescope.previewers'.cat.new, -- For buffer previewer use `require'telescope.previewers'.vim_buffer_cat.new`
     grep_previewer = require'telescope.previewers'.vimgrep.new, -- For buffer previewer use `require'telescope.previewers'.vim_buffer_vimgrep.new`
     qflist_previewer = require'telescope.previewers'.qflist.new, -- For buffer previewer use `require'telescope.previewers'.vim_buffer_qflist.new`
+
+    -- Developer configurations: Not meant for general override
+    buffer_previewer_maker = require'telescope.previewers'.buffer_previewer_maker
   }
 }
 ```
@@ -202,6 +205,12 @@ EOF
 | `file_previewer`       | What telescope previewer to use for files.            | [Previewers](#previewers)  |
 | `grep_previewer`       | What telescope previewer to use for grep and similar  | [Previewers](#previewers)  |
 | `qflist_previewer`     | What telescope previewer to use for qflist            | [Previewers](#previewers)  |
+
+
+### Options for extension developers
+| Keys                   | Description                                           | Options                    |
+|------------------------|-------------------------------------------------------|----------------------------|
+| `buffer_previewer_maker` | How a file gets loaded and which highlighter will be used. Extensions will change it | function             |
 
 ### Options affecting Sorting
 
@@ -356,7 +365,7 @@ require('telescope.builtin').fd({ -- or new custom picker's attach_mappings fiel
 
 Built-in functions. Ready to be bound to any key you like. :smile:
 
-```vim 
+```vim
 :lua require'telescope.builtin'.planets{}
 
 :nnoremap <Leader>pp :lua require'telescope.builtin'.planets{}
@@ -441,17 +450,24 @@ Built-in functions. Ready to be bound to any key you like. :smile:
 | `previewers.cat.new`               | Default previewer for files. Uses `cat`/`bat`                   |
 | `previewers.vimgrep.new`           | Default previewer for grep and similar. Uses `cat`/`bat`        |
 | `previewers.qflist.new`            | Default previewer for qflist. Uses `cat`/`bat`                  |
-| `previewers.vim_buffer_cat.new`        | Experimental previewer for files. Uses vim buffers              |
-| `previewers.vim_buffer_vimgrep.new`    | Experimental previewer for grep and similar. Uses vim buffers   |
-| `previewers.vim_buffer_qflist.new`     | Experimental previewer for qflist. Uses vim buffers             |
+| `previewers.vim_buffer_cat.new`    | Experimental previewer for files. Uses vim buffers              |
+| `previewers.vim_buffer_vimgrep.new`| Experimental previewer for grep and similar. Uses vim buffers   |
+| `previewers.vim_buffer_qflist.new` | Experimental previewer for qflist. Uses vim buffers             |
 | .................................. | Your next awesome previewer here :D                             |
 
 By default, telescope.nvim uses `cat`/`bat` for preview. However after telescope's new experimental previewers
 are stable this will change. The new experimental previewers use tree-sitter and vim buffers, provide much
-better performance and are ready for daily usage (`vim_buffer_cat` more than the others), but there might be
-cases where it can't detect a Filetype correctly, thus leading to wrong highlights.
-Also `vimgrep` and `qflist` might be slower due to synchronous file loading.
+better performance and are ready for daily usage, but there might be cases where it can't detect a Filetype
+correctly, thus leading to wrong highlights. This is because we can't determine the filetype in the traditional way
+(we don't do `bufload`. We read the file async with `vim.loop.fs_` and attach only a highlighter), because we can't
+execute autocommands, otherwise the speed of the previewer would slow down considerably.
+If you want to configure more filetypes take a look at
+[plenary wiki](https://github.com/nvim-lua/plenary.nvim#plenaryfiletype).
 
+If you want to configure the `vim_buffer_` previewer, e.g. you want the line to wrap do this:
+```vim
+autocmd User TelescopePreviewerLoaded setlocal wrap
+```
 
 ## Sorters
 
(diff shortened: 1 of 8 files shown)


Conni2461:
@rockerBOO I tried to detect a binary file fast. Currently just `find_files`, `git_files` but it might work. I tried it with a random png file for testing purpose and it worked. Would be great if you could give it a try, if you find some spare time 😄.

Basically what we are doing is after reading the file, before we dump the content to a buffer, i iterate over the file once and look for bad chars. According to stackoverflow 🤣 bad is `char > string.char(0) and char < string.char(8)` or `char > string.char(13) and char < string.char(26)`. The good thing is for binary files we detect a string pretty fast, the bad thing for text files we will never find one and have to iterate over the whole file once, so i might make it optional.

rockerBOO:
What about like https://github.com/wbond/puremagic/blob/master/puremagic.lua#L306 which checks the first 8-12 characters for most cases. We can maybe also do extension inference  as a fast option. IMO it might be worth it to allow people to set their detection functions so they can customize which process they want to use or make their own. 

For instance if I am always on linux, I personally can call out to linux mime checkers or do my own expensive checks for certain projects with a lot of large binary files. 

Binaries are generally faster to load then worth to check every character? I was also suggesting doing a filesize check, then you can just not preview files > like 10MB or whatever. Most binaries will be quite large. 

For images we can do a file extension check likely for the common ones to speed them up.

Conni2461:
Yeah it was just an idea. It also doesn't work for `lua/telescope/utils.lua`. We can postpone it. I just tried an idea because we read the content of the file anyway right now. So i might just remove it for now. 

I also thought about your issue with don't wanna highlight `*.min.js` with treesitter because its slow.
We have a highlighter function right now which does that. Either we could extend the current one like that. User can define a table with regex and we match each entry with filename and if true we do regex highlighting. Or we could let you write a own highlighter and you/user can do it.
First option would be more user friendly, second option would be harder to do but more flexible. We can also add an example highlighter to the wiki so people can find some inspiration.

The hard part would be giving the highlighter the correct input. The default highlighter needs `(bufnr, filetype)` you also want `filename` but `content` would also be great but we only have the `content` when we read the file the first time. The second time we need to retrieve the `content` with `nvim_buf_get_lines` because the cache would kick in and the file will not be read again.

Also also also all `vim_buffer` previewer are now `async`. Which means super sonic speed for everyone. No more bullying 🤣 I think we can deprecate `bat` now and make them default.

rockerBOO:
Cool cool. Mostly I wanted to say my perspective and share my ideas but its all a lower priority. 

The function would take the filename and the function would handle a lot of the heavy lifting, hopefully using telescope/nvim API's 

They would need to read or not read the file, do stuff with the filename, check caches, stuff like that. Like it may be worth it for me to have it cache certain files and not preview by default. Then I can re-check through some action.

```lua
function(bufnr, filename) 
    if io.filesize(filename) > 10MB then
        nvim_write_to_buffer(bufnr, "file too big")
    end
end
```

I think the async previewer will help a lot. 

Great work Conni :). Thanks for all you do :).

Conni2461:
> I think the async previewer will help a lot.

I had issues with doing the jump for `vimgrep` or `vim.fn.search` for ctags but now i realized that the issue was the the moment the jump would happen the `self.state.bufnr` could already be changed. So i just passed in `bufnr` to the callback function and used that one. And for `vim.fn.search` i wrapped everything in `vim.api.nvim_buf_call` sometimes the solution is just soooo simple. 🤣 

Thats the current highlighter
```lua
local run_highlighter = function(bufnr, ft)
  if ft ~= '' then
    if has_ts and ts_parsers.has_parser(ft) then
      ts_highlight.attach(bufnr, ft)
    else
      vim.cmd(':ownsyntax ' .. ft)
    end
  end
end
```

Conni2461:
I can also let you overwrite this `file_maker`.
```lua
local file_maker_async = function(filepath, bufnr, bufname, winid, use_ft_detect, callback)
  use_ft_detect = use_ft_detect or true
  local ft = use_ft_detect and pfiletype.detect(filepath)

  if bufname ~= filepath then
    path.read_file_async(filepath, vim.schedule_wrap(function(data)
      vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, vim.split(data, "\n"))

      if callback then callback(bufnr, winid) end
    end))
  else
    if callback then callback(bufnr, winid) end
  end

  if use_ft_detect then run_highlighter(bufnr, ft) end
end
```

But i think that will be not optimal. Basically thats too much power for the average user 🤣.
I think we need to find a extension interface that can hook in easily and we can do binary file detect, file size detect and people can just configure what they want. And we maintain it

rockerBOO:
Whoa whoa give us the power Connii :)

I think `function(filepath, bufnr, opts)` is good since the first 2 are required. Unless `winid` is? 

Then we can expose that signature and the list of opts possibly available.

Conni2461:
`callback` is necessary too now. The `jump to line` will happen there. I probably will have
`opts, filepath, bufnr, callback`

rockerBOO:
That list seems good. Maybe callback can take opts and you can just extend it that way. Takes the same opts passed down from the file maker.
```
callback(bufnr, opts)
```

opts should be last and optional in the function signature. 

Looks cool Conni :)

Conni2461:
Hmm my previewers are the only ones who that function anyways so i already pass in those options so i know them from the calling scope anyway. I am just not allows to use `self.state.bufnr` because that one can already be outdated. But i should be able to get rid of `winid` because that one stays the same for the whole "run". Gotta test that 😆
I will talk with tj about all that. What he thinks about being able to redefine that function.

But thanks for your insight. It will happen either in this or the next PR. We will find a solution for your remaining issues :)

Conni2461:
If anyone is asking i wasn't here 😆 

@rockerBOO https://github.com/nvim-telescope/telescope.nvim/pull/354/commits/e02c536f8a307f1d1dc3e93b1af839ed1459482f You can find all information in commit message. That is just a test `maker` which allows you to define a array of `regex` and that files will be highlighted with regex rather than tree sitter even though a parser is installed. Should fix the `*.min.js` slowdown issue.
I suggest doing either doing a extension with multiple different makers or putting them inside the wiki.

tami5:
> If anyone is asking i wasn't here 😆 
> 

😂😂😂 I'm sure you couldn't help it. 😝

rockerBOO:
@Conni2461 Yes this works for me :) Thank you!

Conni2461:
> @Conni2461 Yes this works for me :) Thank you!

Nice that's great. I think doing an extension would be the best way to do this.
For example we need a different `file_maker` for windows because we will end up with `^M` see https://github.com/nvim-telescope/telescope.nvim/issues/362#issuecomment-751401635 and have to do `data:gsub([[\r]], '')` so we will end up with multiple different functions and the user can just pick one and maybe even combine them. Lets say, someone wants that `gsub` thing and `binary detection` or want `binary, size and bad-files` what ever 😆  and nobody has to know how everything has to come in place besides us :). That's the most easiest way for the average user.

Also @rockerBOO i looked at the lua `mime-typedetection`. For some parts we are almost checking though the whole file, when calling `ebml_parse` for example but that only happens if the file starts with `\x1AE\xDF\xA3` which shoudn't happen that often 😆 So i think that one looks pretty good. We basically need the whole file not just the mentioned function so we could just bundle it with plenary(MIT License). So more people can make use of it and not just the new extension. Also we can get rid of `file --mime-type` for `termopen cat previewer`

README.md
```diff
@@ -162,6 +162,8 @@ require('telescope').setup{
     file_previewer = require'telescope.previewers'.cat.new, -- For buffer previewer use `require'telescope.previewers'.vim_buffer_cat.new`
     grep_previewer = require'telescope.previewers'.vimgrep.new, -- For buffer previewer use `require'telescope.previewers'.vim_buffer_vimgrep.new`
     qflist_previewer = require'telescope.previewers'.qflist.new, -- For buffer previewer use `require'telescope.previewers'.vim_buffer_qflist.new`
+    buffer_previewer_wrap = false,
+    buffer_previewer_maker = require'telescope.previewers'.buffer_previewer_maker -- Don't touch this one unless you want a bad time. We will do something with this and an extension
```
tjdevries, line 166: maybe we should put a separate section `For Developers` so that we can put something like this here.

README.md
```diff
@@ -162,6 +162,8 @@ require('telescope').setup{
     file_previewer = require'telescope.previewers'.cat.new, -- For buffer previewer use `require'telescope.previewers'.vim_buffer_cat.new`
     grep_previewer = require'telescope.previewers'.vimgrep.new, -- For buffer previewer use `require'telescope.previewers'.vim_buffer_vimgrep.new`
     qflist_previewer = require'telescope.previewers'.qflist.new, -- For buffer previewer use `require'telescope.previewers'.vim_buffer_qflist.new`
+    buffer_previewer_wrap = false,
```
tjdevries, line 165: perhaps instead of this, we can add an autocommand that we fire when we're done loading a previewer.

People can do something like `autocmd User TelescopePreviewerLoaded setlocal nowrap` or similar. I otherwise see maybe having a lot of these in the future. (I think that would work and it only would run in the buffer)

README.md
```diff
@@ -436,22 +440,50 @@ Built-in functions. Ready to be bound to any key you like. :smile:
...
-better performance and are ready for daily usage (`vim_buffer_cat` more than the others), but there might be
-cases where it can't detect a Filetype correctly, thus leading to wrong highlights.
-Also `vimgrep` and `qflist` might be slower due to synchronous file loading.
-
+better performance and are ready for daily usage, but there might be cases where it can't detect a Filetype
+correctly, thus leading to wrong highlights. This is because we can't determine the filetype in the traditional way
+(we don't do `bufload`. We read the file async with `vim.loop.fs_` and attach only a highlighter), because we can't
+execute autocommands, otherwise the speed of the previewer would slow down considerably.
+We determine the filetype by the extension. You can add more extensions to the filetypeguesser:
+Create file `~/.config/nvim/data/plenary/filetypes/builtin.lua` and add following:
```
tjdevries, line 460: This should just link to some place in plenary where we show people how to do this.

This is the wrong repo. Perhaps we can add it in a help doc for plenary and say: `see ":help plenary-filetype"` or similar.

Conni2461: Tami didn't like that here as well. People were just asking over and over and we need a place to document that :)

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -1,60 +1,34 @@
...
 local defaulter = utils.make_default_callable
 
 local previewers = {}
 
-local previewer_ns = vim.api.nvim_create_namespace('telescope.previewers')
+local ns_previewer = vim.api.nvim_create_namespace('telescope.previewers')
 
-local file_maker_async = function(filepath, bufnr, bufname, callback)
-  local ft = pfiletype.detect(filepath)
+previewers.file_maker = function(filepath, bufnr, bufname, use_ft_detect, callback)
```
tami5, line 17: Maybe it would be better to have any function more then 3 params to accept a dict to increase readability

lua/telescope/previewers/utils.lua
```diff
@@ -14,4 +20,48 @@ utils.with_preview_window = function(status, bufnr, callable)
   end
 end
 
+-- Job maker for buffer previewer
+utils.job_maker = function(cmd, env, value, bufnr, bufname, callback)
+  if bufname ~= value then
+    local command = table.remove(cmd, 1)
+    local j = Job:new{
```
tami5, line 27: I know you hate me for this but I don't see any reason to bind j 🤣🤣 @Conni2461

lua/telescope/previewers/utils.lua
```diff
@@ -14,4 +20,48 @@ utils.with_preview_window = function(status, bufnr, callable)
...
+        if callback then callback(bufnr, j:result()) end
+      end)
+    }
+    j:start()
+  else
+    if callback then callback(bufnr) end
+  end
+end
+
+-- API helper functions for buffer previewer
```
tami5, line 42: Couldn't this be grouped under one factor function

lua/telescope/previewers/buffer_previewer.lua
Conni2461: Nope :). All are required and need to be handled and that function is basically only public for an extension

PR #356, opened 2020-12-20
LSP check capabilities

sunjon:
Checks supported capabilities before issuing any unnecessary LSP requests.

### DIFF
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -12,8 +12,10 @@ local function check_capabilities(feature)
   local supported_client = false
   for _, client in pairs(clients) do
     supported_client = client.resolved_capabilities[feature]
+    if supported_client then goto continue end
   end
 
+  ::continue::
   if supported_client then
     return true
   else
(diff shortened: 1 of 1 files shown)


lua/telescope/builtin/lsp.lua
```diff
@@ -6,9 +6,33 @@ local utils = require('telescope.utils')
...
+      print("LSP: server does not support " .. feature)
+    end
+    return false
+  end
+end
+
 local lsp = {}
 
 lsp.references = function(opts)
+  if not check_capabilities("find_references") then return end
```
Conni2461, line 34: There is an apply_checks function at the bottom of the file which will wrap all functions that will be exported. So you can do this only once down there

sunjon: I think that's taken care of. I had to create a new table to map the function names to the actual parameter name used in the capability check, rather than renaming the functions and changing peoples mappings.

Conni2461 (approved):
LGTM
I just realized that you are checking capabilities. Thats really neat Thanks :)

tami5:
😍😍😍😍😍 👌

PR #357, opened 2020-12-21
README.md | fix lua example configuration

SebastianKeller:
Both `find_files()` and `help_tags()` were bound to `<leader>ff`. This resulted in `<leader>ff` not opening the `find_files()` dialog


tami5:
Nice catch, @SebastianKeller thank you

PR #361, opened 2020-12-23
refactor: Move toward more callback centered approach

tjdevries:
As we can move closer to more callbacks, it will be easier to (1) test the code and (2) use threading when we figure it out.


tjdevries:
(Old PR that I may end up doing a few things on)

tjdevries:
did better now. ty ober

PR #364, opened 2020-12-25
fix: builtin.buffers to override entry_maker

tamago324:
This will allow for more customization and make telescope more fun to use!

```lua
require('telescope.builtin').buffers {
  ...
  entry_maker = my_entry_maker.gen_from_buffer_better(),
  ...
}
```

Should it be possible to override it with other pickers?

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -473,14 +473,10 @@ function make_entry.gen_from_buffer(opts)
 
     local icon, hl_group = get_devicons(entry.filename, disable_devicons)
 
-    if hl_group then
-      icon = { icon, hl_group }
-    end
-
     return displayer {
       {entry.bufnr, "TelescopeResultsNumber"},
       {entry.indicator, "TelescopeResultsComment"},
-      icon,
+      { icon, hl_group },
       display_bufname .. ":" .. entry.lnum
       }
   end
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -33,6 +33,7 @@ local lsp_type_highlight = {
 local make_entry = {}
 
 local transform_devicons
+local get_devicons
 if has_devicons then
   if not devicons.has_loaded() then
     devicons.setup()
@@ -52,9 +53,26 @@ if has_devicons then
       return icon_display
     end
   end
+
+  get_devicons = function(filename, disable_devicons)
+    if disable_devicons or not filename then
+      return ''
+    end
+
+    local icon, icon_highlight = devicons.get_icon(filename, string.match(filename, '%a+$'), { default = true })
+    if conf.color_devicons then
+      return icon, icon_highlight
+    else
+      return icon
+    end
+  end
 else
   transform_devicons = function(_, display, _)
     return display
+  end
+
+  get_devicons = function(_, _)
+    return ''
   end
 end
 
@@ -426,11 +444,19 @@ end
 function make_entry.gen_from_buffer(opts)
   opts = opts or {}
 
+  local disable_devicons = opts.disable_devicons
+
+  local icon_width = 0
+  if not disable_devicons then
+    icon_width = vim.fn.strdisplaywidth(get_devicons('fname', disable_devicons))
+  end
+
   local displayer = entry_display.create {
     separator = " ",
     items = {
       { width = opts.bufnr_width },
       { width = 4 },
+      { width = icon_width },
       { remaining = true },
     },
   }
@@ -445,9 +471,16 @@ function make_entry.gen_from_buffer(opts)
       display_bufname = entry.filename
     end
 
+    local icon, hl_group = get_devicons(entry.filename, disable_devicons)
+
+    if hl_group then
+      icon = { icon, hl_group }
+    end
+
     return displayer {
       {entry.bufnr, "TelescopeResultsNumber"},
       {entry.indicator, "TelescopeResultsComment"},
+      icon,
       display_bufname .. ":" .. entry.lnum
       }
   end
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -172,7 +172,7 @@ files.treesitter = function(opts)
     prompt_title = 'Treesitter Symbols',
     finder    = finders.new_table {
       results = results,
-      entry_maker = make_entry.gen_from_treesitter(opts)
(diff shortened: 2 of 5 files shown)


Conni2461:
LGTM, great idea btw, we should probably do this for all :)

Also if you think your `entry_maker` is "better" :) feel free to show it(code/image) maybe you can implement some of the ideas here.

tamago324:
I'm glad you said that!

I used to use [LeaderF](https://github.com/Yggdroot/LeaderF) until I started using Telsceope, and I favorite the buffers display in LeaderF.
So I decided to apply it to Telescope as well.

I wrote this code just for my own use, so it's not the best code, but I'll post it here.

## LeaderF's buffer

This is my favorite display of LeaderF's buffer.

leaderf_buffer


## My Telescope buffer

image

### movie


https://user-images.githubusercontent.com/16581287/103154042-618bec00-47d8-11eb-86a8-10f3c416382e.mp4


### Source code

I wrote this with reference to [make_entry.gen_from_buffer](https://github.com/nvim-lua/telescope.nvim/blob/244584ced901c60e244a5979aa19a0e0b5604439/lua/telescope/make_entry.lua#L426-L479).

```lua
local invalid_regex_list = {'^term://', '^deol%-edit@'}

local is_valid_bufnr = function(bufnr)
  for _, re in ipairs(invalid_regex_list) do
    if vim.api.nvim_buf_get_name(bufnr):match(re) then
      return false
    end
  end
  return true
end

function make_entry.gen_from_buffer_like_leaderf(opts)
  opts = opts or {}
  local default_icons, _ = devicons.get_icon('file', '', {default = true})

  local bufnrs = filter(function(b)
    return 1 == vim.fn.buflisted(b) and is_valid_bufnr(b)
  end, vim.api.nvim_list_bufs())

  local max_bufnr = math.max(unpack(bufnrs))
  local bufnr_width = #tostring(max_bufnr)

  local max_bufname = math.max(
    unpack(
      map(function(bufnr)
        return vim.fn.strdisplaywidth(vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ':p:t'))
      end, bufnrs)
    )
  )

    separator = " ",
    items = {
      { width = bufnr_width },
      { width = 4 },
      { width = vim.fn.strwidth(default_icons) },
      { width = max_bufname },
      { remaining = true },
    },
  }

    return displayer {
      {entry.bufnr, "TelescopeResultsNumber"},
      {entry.indicator, "TelescopeResultsComment"},
      {entry.devicons, entry.devicons_highlight},
      entry.file_name,
      {entry.dir_name, "Comment"}
      }
  end

  return function(entry)
    local bufname = entry.info.name ~= "" and entry.info.name or '[No Name]'
    local hidden = entry.info.hidden == 1 and 'h' or 'a'
    local readonly = vim.api.nvim_buf_get_option(entry.bufnr, 'readonly') and '=' or ' '
    local changed = entry.info.changed == 1 and '+' or ' '
    local indicator = entry.flag .. hidden .. readonly .. changed

    local dir_name = vim.fn.fnamemodify(bufname, ':p:h')
    local file_name = vim.fn.fnamemodify(bufname, ':p:t')

    local icons, highlight = devicons.get_icon(bufname, string.match(bufname, '%a+$'), { default = true })

    return {
      valid = is_valid_bufnr(entry.bufnr),

      value = bufname,
      ordinal = entry.bufnr .. " : " .. file_name,
      display = make_display,

      bufnr = entry.bufnr,

      lnum = entry.info.lnum ~= 0 and entry.info.lnum or 1,
      indicator = indicator,
      devicons = icons,
      devicons_highlight = highlight,

      file_name = file_name,
      dir_name = dir_name,
    }
  end
end
```


I hope this is one of your ideas to improve it.

Conni2461:
Looks great :)
I think icons would be a great addition. I am not sure about having just filename and then whole path.
If you want you can add icons to the current PR or if you don't want to do it here/at all then i can merge this PR :) 

CC @tami5 @sunjon

tami5:
@tamago324 😍 loved your version great work. LGTM I think its a powerful addition to our modularity. cc @sunjon 

@tamago324 could you add you amazing custom buffer as an example to the wiki? config recipes.

Also, @Conni2461 would be better for new/current users if @tamago324 version became default?

sunjon:
I see no issue with allowing make_entry to be overridden. I haven't tried the çustomised buffer config yet - series it do more than add (the same) icon to all entries?

Conni2461:
I am not asking about having more customization. 😆 
I was asking about the different `display` tamago324 has. Right now we have basically just a copy of `:buffers` with `nr indicator name:lnum` he is using: `nr indicator symbol onlyfilename fullpath`

tamago324:
Thanks to everyone for their comments.

@Conni2461 I'm thinking of adding an icon display like in `nr indicator symbol name:lnum`. Is that ok?

@tami5 Thank you! I'll add the configuration to the `Configuration-Recipes` section of the wiki then. Also, can I post a screenshot?

tami5:
@tamago324 sure, I'd prefer the screen to be dark mode, but do what you like.

tamago324:
@tami5 I see, I'll shoot in dark mode to make it consistent. Thank you.

tamago324:
@tami5 https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes

tamago324:
I added the devicons.
I think it's probably working well.

lua/telescope/make_entry.lua
```diff
@@ -445,9 +471,16 @@ function make_entry.gen_from_buffer(opts)
       display_bufname = entry.filename
     end
 
+    local icon, hl_group = get_devicons(entry.filename, disable_devicons)
+
+    if hl_group then
+      icon = { icon, hl_group }
+    end
```
Conni2461, line 478: You don't need to do that nil check here. Just pass in the table to the `displayer`. If hl_group is nil then we will not do anything :) I think

Conni2461 (approved):
LGTM besides that one comment. Thanks :)

Will merge after when that is updated, unless you want to make every entry_maker configurable :) which i would love

tamago324:
@Conni2461 I would like to make every entry_maker configurable.
 Later, you'll modify the code and commit.

lua/telescope/make_entry.lua
tamago324: Thank you! I fixed it.

Conni2461:
Thanks :)

tamago324:
Thank you for merging!

PR #366, opened 2020-12-26
Fix bug in scroller.create.

MenBrial:
Corrects error creation.


Conni2461:
Thanks :)
CI is currently not working because neovim nightly build vanished from github. 😆

PR #371, opened 2020-12-29
Previewer windows fix

Conni2461:
Edit: Should work. Just not in general. We need to determine for each file which line endings are used. I think you can also use linux line endings on windows and these change would break it

Edit 2: Should be better now


raiguard:
image

Tested on Windows 10, can confirm that this works properly!

PR #372, opened 2020-12-29
add "append" mode to previewers.utils.job_maker

pwntester:


Conni2461 (approved):
LGTM but now I have the feeling that we are losing control of the parameters. 🤣

But tbh i also don't really mind.

Conni2461:
Talked with @tami5. Lets make a table and have the table as last parameter. That means you also have to update 4 builtin previewer. Sorry about that :)
We can add:
```lua
{
  env,
  value, -- value can be optional because its only here for the cache
  mode,
  callback
}
```

pwntester:
> LGTM but now I have the feeling that we are losing control of the parameters. 🤣

Yep, maybe it should be better to make it an `opts` dict.

pwntester:
> ```lua
> value can be optional because its only here for the cache
> ```

so is `bufname`, should I make it also part of the dict and optional?

Conni2461:
Yes thanks :). Forgot about that 😆

Conni2461 (approved):
Thats much better. Thanks :)

tami5:
LGTM thanks @pwntester

PR #373, opened 2020-12-30
Add ts_parsers.ft_to_lang to detect buffer's lang

fffed:
Add detection of a proper buffer language by `ts_parsers.ft_to_lang` as `ts_parsers.has_parser` works with `language` but not with `filetype`.
https://github.com/nvim-treesitter/nvim-treesitter/issues/796

### DIFF
--- a/lua/telescope/previewers/utils.lua
+++ b/lua/telescope/previewers/utils.lua
@@ -49,31 +49,40 @@ utils.job_maker = function(cmd, bufnr, opts)
   end
 end
 
+local function has_filetype(ft)
+    return ft and ft ~= ''
+end
+
 --- Attach default highlighter which will choose between regex and ts
 utils.highlighter = function(bufnr, ft)
-  if ft and ft ~= '' then
-    if has_ts and ts_parsers.has_parser(ft) then
-      ts_highlight.attach(bufnr, ft)
-    else
-      vim.cmd(':ownsyntax ' .. ft)
-    end
+  if not(utils.ts_highlighter(bufnr, ft)) then
+    utils.regex_highlighter(bufnr, ft)
   end
 end
 
 --- Attach regex highlighter
 utils.regex_highlighter = function(_, ft)
-  if ft and ft ~= '' then
+  if has_filetype(ft) then
     vim.cmd(':ownsyntax ' .. ft)
+
+    return true
   end
+
+  return false
 end
 
 -- Attach ts highlighter
 utils.ts_highlighter = function(bufnr, ft)
-  if ft and ft ~= '' then
-    if has_ts and ts_parsers.has_parser(ft) then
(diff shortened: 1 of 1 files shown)


fffed:
I've tried to DRY highlighter calls but not sure if it worth it.

Conni2461:
> I've tried to DRY highlighter calls but not sure if it worth it.

What do you mean with that.

lua/telescope/previewers/utils.lua
```diff
@@ -51,29 +51,35 @@ end
...
-      vim.cmd(':ownsyntax ' .. ft)
-    end
+  if not(utils.ts_highlighter(bufnr, ft)) then
+    utils.regex_highlighter(bufnr, ft)
   end
 end
 
 --- Attach regex highlighter
 utils.regex_highlighter = function(_, ft)
+  local isHighlighted = false
```
Conni2461, line 61: Lets not introduce a new var. Do `return true` inside the if and `return false` at the end

fffed:
> > I've tried to DRY highlighter calls but not sure if it worth it.
> 
> What do you mean with that.

I removed conditions from utils.highlighter

Conni2461:
> I removed conditions from utils.highlighter

Yeah but now we run `ft and ft ~= ''` twice when we have a regex highlight. I didn't mind the code duplications for that part. We only use `highlighter` here and the other two functions were meant for extension developers that they don't have to do the `has_ts` stuff

fffed:
> Yeah but now we run `ft and ft ~= ''` twice when we have a regex highlight.

It's true and this is why I also was thinking hard on it ))). But on other side we would have the same part of logic in the two functions(ts lang detection) and if you totally OK with it I could revert it.

Conni2461:
Okay it shouldn't be an issue really :)
Will merge. Thanks for investigating and reaching out to the nvim-tree-sitter devs :)

clason:
Broke telescope for me :]

```
Error executing vim.schedule lua callback: .../start/telescope.nvim/lua/telescope/previewers/utils.lua:77: attempt to call field 'ft_to_lang' (a nil value)
```

(Please remember that not everyone has tree-sitter installed!)

lua/telescope/previewers/utils.lua
```diff
@@ -49,31 +49,40 @@ utils.job_maker = function(cmd, bufnr, opts)
...
+
+  return false
 end
 
 -- Attach ts highlighter
 utils.ts_highlighter = function(bufnr, ft)
-  if ft and ft ~= '' then
-    if has_ts and ts_parsers.has_parser(ft) then
-      ts_highlight.attach(bufnr, ft)
+  if has_filetype(ft) then
```
clason, line 76: This should be `if has_ts and has_filetype(ft)`.

PR #378, opened 2021-01-03
fix: live_grep throws an error if you insert `\`

### LINKED ISSUE #347: live_grep errors if you use `\`
```
telescope.nvim/lua/telescope/make_entry.lua:143: attempt to concatenate field 'filename' (a nil value)
```

2020-12-18-111149_4480x1440_scrot

elianiva:
fixed #347. Also fixed an issue with `(` or `)`.

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -19,6 +19,8 @@ files.live_grep = function(opts)
         return nil
       end
 
+      if string.match(prompt, "[\\|%(%)]") then prompt = "" end
+
       return flatten { conf.vimgrep_arguments, prompt }
     end,
     opts.entry_maker or make_entry.gen_from_vimgrep(opts),


lua/telescope/builtin/files.lua
```diff
@@ -19,6 +19,8 @@ files.live_grep = function(opts)
         return nil
       end
 
+      if string.match(prompt, "[\\|%(%)]") then prompt = "" end
```
elianiva, line 22: This works on `live_grep` and still match the `\` or `()`, but if I do this on `grep_string`, it won't match `\` nor `()`.

tami5:
:+1: it did solve the issue. Not sure if this is normal but there side effect from typing \.

Conni2461:
I think we just have to escape certain chars. I will have a look at it in the next hour. I am not sure if we should "reset" prompt

Conni2461:
Nope. I was wrong. Your fix seems to work. Thanks @elianiva :)

PR #379, opened 2021-01-03
fix: escape chars instead of resetting the prompt

elianiva:
@Conni2461 nope, you were right :laughing:. We should escape it instead. This escapes `\`, `()`, and `[]`, I don't know what other characters that can cause an error. It also fixes `grep_string`. Follow up of #378

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -11,6 +11,17 @@ local flatten = vim.tbl_flatten
 
 local files = {}
 
+local escape_chars = function(string)
+  return string.gsub(string,  "[%(|%)|\\|%[|%]|%-|%{%}|%?|%+|%*]", {
+    ["\\"] = "\\\\", ["-"] = "\\-",
+    ["("] = "\\(", [")"] = "\\)",
+    ["["] = "\\[", ["]"] = "\\]",
+    ["{"] = "\\{", ["}"] = "\\}",
+    ["?"] = "\\?", ["+"] = "\\+",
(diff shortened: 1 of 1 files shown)


Conni2461:
First of all. You can do multiple substituts in one `gsub`? Thats soooo good to know :)
Also i kinda did that and it was still failing for me so, good job for figuring it out :)

Conni2461:
We need to escape `-` as well

elianiva:
ok I've added it

Conni2461:
I will continue to name everything i found 😆 
`:`

elianiva:
also, do you think we should move it to its own function? It's kinda tedious to change it on 2 places.

Conni2461:
Have a local function that takes a string and returns the new string. Really good work btw :)

elianiva:
done :)

Conni2461:
Found more :)
- `?`
- `+`
- `*`

Conni2461 (approved):
LGTM Thank you :)
I couldn't find another char which doesn't mean we are covering all but we will figure that one out in the future :)

elianiva:
Thanks!

errx:
I'm sorry but why do we need to escape chars? this change breaks regexp search with ripgrep

errx:
I've changed this line
return flatten { vimgrep_arguments, prompt, opts.search_dirs or '.' }
to
return flatten { vimgrep_arguments, '--', prompt, opts.search_dirs or '.' }

but we need to handle ripgrep errors (like invalid regexp)

PR #380, opened 2021-01-03
add extension in telescope command

glepnir:
Call the extension from command line when user register extension to telescope
```vim
-- before
lua require"telescope".extensions.dotfiles.dotfiles{path=os.getenv("HOME").."/.dotfiles"}
lua require"telescope".extensions.project.project{}
-- now
Telescope dotfiles path=xxxx
Telescope project
```

### DIFF
--- a/plugin/telescope.vim
+++ b/plugin/telescope.vim
@@ -49,7 +49,9 @@ cnoremap <silent> <Plug>(TelescopeFuzzyCommandSearch) <C-\>e
 
 " Telescope builtin lists
 function! s:telescope_complete(...)
-  return join(luaeval('vim.tbl_keys(require("telescope.builtin"))'), "\n")
+  let l:builtin_list = luaeval('vim.tbl_keys(require("telescope.builtin"))')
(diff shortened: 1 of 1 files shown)


Conni2461 (approved):
LGTM Thanks :) Thats really helpful. Will merge in 5 min. I will try first

Do you need a repo under this org for your dotfiles extension?

Conni2461:
Okay i don't think that is the correct solution yet.
e.g. with dap we have following finders:
- commands
- configurations
- list_breakpoints
- variables

But i am only getting `dap` as finder which technically doesn't exits. I think similar thinks would apply to `telescope-github` but that one is currently cheating to get that behavior so i can't test :(

glepnir:
I don't test telescope-dap:) only test my extension and telescope-project.

tami5:
😍😍 thanks a lot @glepnir

glepnir:
now we can do this `Telescope dap commands`

Conni2461:
Oh nice @glepnir Thanks for that. Everything looks good to me so i will merge

Issue #367, opened 2020-12-26
feat: Command history Picker should add to history


tami5:
Commands ran with command history picker should  be add to command history by default.

PR #383, opened 2021-01-04
add selected command to cmd-history

sunjon:
Add commands to cmd-history. Fixes https://github.com/nvim-telescope/telescope.nvim/issues/367


tami5:
:heart_eyes: :heart_eyes: :heart_eyes: thanks @sunjon

PR #386, opened 2021-01-04
make pickers not block typing when it is processing (using libuv async handle)

oberblastmeister:
I have made the `on_lines` callback for `nvim_buf_attach` async using `uv_async_t`. This has extreme performance benefits even with such a small code change because the cursor can keep moving while telescope is processing. I am using `uv_async_t` instead of `luv_work_ctx_t` or `luv_thread_t` because I don't think that the thread ones can change state. In my tests they crash while the async one is fine. We probably still need to make the actual picker faster but making it async is a good step as it makes the user feel like it is going fast.

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -558,7 +558,7 @@ function Picker:find()
 
   -- Register attach
   vim.api.nvim_buf_attach(prompt_bufnr, false, {
-    on_lines = on_lines,
+    on_lines = utils.async(on_lines),
     on_detach = vim.schedule_wrap(function()
       self:_reset_highlights()
 
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -556,10 +556,13 @@ function Picker:find()
   on_lines(nil, nil, nil, 0, 1)
   update_status()
 
+  local send, run, stop = utils.async_queue(on_lines)
+
   -- Register attach
   vim.api.nvim_buf_attach(prompt_bufnr, false, {
-    on_lines = utils.async(on_lines),
+    on_lines = send,
     on_detach = vim.schedule_wrap(function()
+      stop()
       self:_reset_highlights()
 
       on_lines = nil
@@ -574,6 +577,7 @@ function Picker:find()
       collectgarbage(); collectgarbage()
     end),
   })
+  run()
 
   -- TODO: Use WinLeave as well?
   local on_buf_leave = string.format(
@@ -803,10 +807,10 @@ function Picker:set_selection(row)
   if status.preview_win and self.previewer then
     self:_increment("previewed")
 
-    self.previewer:preview(
+    utils.async(function() self.previewer:preview(
       entry,
       status
-    )
+    ) end)()
   end
 end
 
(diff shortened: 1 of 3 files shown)


Conni2461:
I notified tj. He wanna look at that but if that works that could be huge :) Thank you very much 😄

clason:
Briefly testing it, I noticed that the `grep_string` finder spawns a new `rg` instance every time it is run and then  abandoned -- leaving a bunch of `rg`s hanging around and eating memory until neovim is closed.

oberblastmeister:
@clason did you mean it happens on `live_grep` because I tested it and it only happens on `live_grep`. I also tested it on the main branch and my async branch and the same behavior seems to happen.

tjdevries:
@oberblastmeister Thanks! I will try and test soon.

@clason sounds like I need to track down this issue anyway :) I can try and figure it out -- I thought we were freeing those results, but maybe we are not. Thanks for testing.

clason:
@oberblastmeister no, it's on `builtin.grep_string { search = '' }` (on my home directory, so the result list is _long_)

tami5:
according to https://github.com/nvim-telescope/telescope.nvim/issues/392#issuecomment-754773890 the issue is still present

lua/telescope/pickers.lua
```diff
@@ -803,10 +807,10 @@ function Picker:set_selection(row)
   if status.preview_win and self.previewer then
     self:_increment("previewed")
 
-    self.previewer:preview(
+    utils.async(function() self.previewer:preview(
       entry,
       status
-    )
+    ) end)()
```
Conni2461, line 813: I wound not do that at all. The buffer previewers are already async and termopen previewer does also not need that.

oberblastmeister: I will probably remove that. I might put stuff that doesn't make sense because I am still testing.

oberblastmeister: yeah that was an issue that I was finding, rust and markdown files were very slow

Conni2461:
I tried that PR again. It does absolutely nothing for me :|. Run on home dir

lua/telescope/pickers.lua
oberblastmeister: for me though that actually made markdown files preview faster but still not fast

Conni2461: I also tried to make just ts highlighter async. Doesn't do anything because we are just attaching a highlighter which is already super fast and parsing happens not when attaching.

oberblastmeister:
What is not working for you? Is it the typing experience that is freezing or the fact that grepping on home dir will not open?

Conni2461:
Still freezing and i didn't notice any difference to current version. But i also didn't used a stopwatch just my feeling.
Also i was getting an previewer issue with `1004 is not a valid winid` which is probably an issue with that async previewer

tjdevries:
I have a feeling this doesn't do what we  think it does :) I'm very busy applying for jobs, but we can try some stuff later. Sorry for the delays :)

tjdevries:
done with new async await strat

PR #388, opened 2021-01-04
Add colors to builtin.git_status results

cempassi:
I would like to add colors to builtin.git_status results. 

I made a basic implementation from what is already existing. Is there any improvements i can make?

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -117,16 +117,7 @@ git.status = function(opts)
     prompt_title = 'Git Status',
     finder = finders.new_table {
       results = vim.split(output, '\n'),
-      entry_maker = function(entry)
-        if entry == '' then return nil end
-        local mod, file = string.match(entry, '(..).*%s[->%s]?(.+)')
-        return {
-          value = file,
-          status = mod,
-          ordinal = entry,
-          display = entry,
-        }
-      end
+      entry_maker = make_entry.gen_from_git_status(opts)
     },
     previewer = previewers.git_file_diff.new(opts),
     sorter = conf.file_sorter(opts),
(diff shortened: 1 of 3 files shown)


Conni2461:
We probably want to write a real entry_maker in  `make_entry.lua`

lua/telescope/builtin/git.lua
```diff
@@ -113,6 +114,31 @@ git.status = function(opts)
...
+  local make_display = function(entry)
+    local status_highlight
+
+    if entry.status == "??" then
+      status_highlight = "DiffDelete"
+    elseif string.match(entry.status, " .") then
+      status_highlight = "DiffChange"
+    else
+      status_highlight = "DiffAdd"
+    end
```
Conni2461, line 134: Thats not correct. I would make two 1 width items for status and split it.
Because what about. I have a staged file and do some more changes we will end up with this `MM`.
Now the first one has to be `DiffAdd` and the second one `DiffChange`

lua/telescope/builtin/git.lua
```diff
@@ -113,6 +114,31 @@ git.status = function(opts)
...
+
+    return displayer {
+      { entry.status, status_highlight},
+      { entry.value, "NormalNC"},
```
Conni2461, line 138: I would not add a `hl` for entry.value at all.

Conni2461:
Looks way better now. I still think we should move it `make_entry`

Conni2461 (approved):
LGTM. Anything missing from your side? Or why WIP.
If not i will merge

cempassi:
Everything good to me if there's nothing else!

Conni2461:
Oh one more thing. Could you do Telescope hl groups, so we have more customize options? Like 
```
highlight default TelescopeResultsDiffChange DiffChange
...
```

Conni2461:
Thanks @cempassi :)

PR #391, opened 2021-01-05
allow actions.get_current_line in addition to actions.get_selected_entry

oberblastmeister:
This can be useful for example if the user has typed in something that does not match any entry. Then we want to perform something else. A use case is for [neuron.nvim](https://github.com/oberblastmeister/neuron.nvim) where the user can query for notes. If the user has typed in something that does not match any entry, we want to create a new note, kind of like Notational Velocity. I would need to be able to get the current line for that to work.


Conni2461:
Thanks :) Great addition

PR #395, opened 2021-01-05
fix: directory handle in windows

Mike325:
Use "dir" to list the content of directories in Windows

### DIFF
--- a/lua/telescope/previewers/term_previewer.lua
+++ b/lua/telescope/previewers/term_previewer.lua
@@ -33,9 +33,14 @@ local get_file_stat = function(filename)
   return vim.loop.fs_stat(vim.fn.expand(filename)) or {}
 end
 
+local function list_dir(dirname)
+    local qdir = add_quotes .. vim.fn.expand(dirname) .. add_quotes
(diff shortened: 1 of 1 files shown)


Conni2461:
Thanks :) Good fix

lua/telescope/previewers/term_previewer.lua
```diff
@@ -33,9 +33,14 @@ local get_file_stat = function(filename)
   return vim.loop.fs_stat(vim.fn.expand(filename)) or {}
 end
 
+local function list_dir(dirname)
+    local qdir = add_quotes .. vim.fn.expand(dirname) .. add_quotes
+    return vim.fn.has('win32') == 1 and {'cmd.exe', '/c', 'dir', qdir} or { 'ls', '-la', qdir}
```
tjdevries, line 38: we actually only need to do this check once, rather than every time.

So we could do `local function list_dir` and tehn assign it based on the function result.

Conni2461: Oh each thats a good idea :)

Now that we are talking about it i also need to add a directory viewer to `buffer_previewer`, probably with my WIP `luafind` and then using `vim.loop.fs_stat` to get all that long stuff. That could be cool

PR #397, opened 2021-01-06
Use the same maker as other file pickers

delphinus:
Make devicons visible for `oldfiles`


tami5:
Nice @delphinus Thanks

PR #398, opened 2021-01-06
refactor telescope command

glepnir:
Now support config telescope theme from command line.

telescope

### DIFF
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -1,0 +1,46 @@
+local themes = require('telescope.themes')
+local builtin = require('telescope.builtin')
+local extensions = require('telescope._extensions').manager
+local command = {}
+
+-- receive the viml command args
+-- it should be a table value like
+-- {
+--   cmd = 'find_files',
+--   theme = 'dropdown',
+--   extension_type  = 'command'
+--   opts = {
+--      cwd = '***',
+-- }
+function command.run_command(args)
+  local user_opts = args or {}
+  if next(user_opts) == nil and not user_opts.cmd then
+    print('[Telescope] your command miss args')
+    return
+  end
+
+  local cmd = user_opts.cmd
+  local opts = user_opts.opts or {}
+  local extension_type = user_opts.extension_type or ''
+  local theme = user_opts.theme or ''
+
+  if string.len(theme) > 0 then
+    opts = themes[theme](opts)
+  end
+
+  if string.len(extension_type) > 0 then
+    extensions[cmd][extension_type](opts)
+    return
+  end
+
+  if builtin[cmd] then
+    builtin[cmd](opts)
+    return
+  end
+
+  if rawget(extensions,cmd) then
+    extensions[cmd][cmd](opts)
+  end
+end
+
+return command
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -4,6 +4,8 @@ local extensions = require('telescope._extensions').manager
 local config = require('telescope.config')
 local command = {}
 
+-- convert command line string arguments to
+-- lua number boolean type and nil value
 local function convert_user_opts(user_opts)
   local default_opts = config.values
 
@@ -65,7 +67,6 @@ function command.run_command(args)
   if next(opts) ~= nil then
     convert_user_opts(opts)
   end
-  print(vim.inspect(opts))
 
   if string.len(theme) > 0 then
     opts = themes[theme](opts)
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -1,7 +1,45 @@
 local themes = require('telescope.themes')
 local builtin = require('telescope.builtin')
 local extensions = require('telescope._extensions').manager
+local config = require('telescope.config')
 local command = {}
+
+local function convert_user_opts(user_opts)
+  local default_opts = config.values
+
+  local _switch = {
+    ['boolean'] = function(key,val)
+      if val == 'false' then
+        user_opts[key] = false
+      end
+      user_opts[key] = true
+    end,
+    ['number'] = function(key,val)
+      user_opts[key] = tonumber(val)
+    end,
+    ['string'] = function(key,val)
+      if val == 'nil' then
+        user_opts[key] = nil
+      end
+    end
+  }
+
+  local _switch_metatable = {
+    __index = function(_,k)
+      print(string.format('Type of %s does not match',k))
+    end
+  }
+
+  setmetatable(_switch,_switch_metatable)
+
+  for key,val in pairs(user_opts) do
+    if default_opts[key]  == nil then
+      print('[Telescope] argument not exists in default options')
+      return
+    end
+    _switch[type(default_opts[key])](key,val)
+  end
+end
 
 -- receive the viml command args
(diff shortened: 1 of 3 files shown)


tami5:
:heart_eyes: works really well. Thanks @glepnir . LGTM

If we are creating an interface for telescope through the command line, I'd suggest that we abbreviate stuff, don't you think?

E.g.

```vim
Telescope cmd -t/t= theme -l/l= layout -t/t= path/cwd -p/p= preview/true-false
```

sunjon:
> heart_eyes works really well. Thanks @glepnir . LGTM
> 
> If we are creating an interface for telescope through the command line, I'd suggest that we abbreviate stuff, don't you think?
> 
> E.g.
> 
> ```viml
> Telescope cmd -t/t= theme -l/l= layout -t/t= path/cwd -p/p= preview/true-false
> ```

abbreviated parameter equivalents would be nice, as would completion so there's no need to memorize parameter names.

glepnir:
Now support completion default options in command line.

sunjon:
> Now support completion default options in command line.

Thanks!
There's a few issues that need resolving at the moment:

1) completion for the options is getting mixed with the original results for the available finders - it's not clear at all which are options and which are finders.

2)  there are issues that would be solved with some sort of input validation/sanitzation. A few examples:

- Problem setting an empty string value
`prompt_prefix = " "` will work via lua config; via command like it will result in literal quotes being inserted, `" "` results in an error.

- issue 1 makes it a bit harder to describe this accurately, but some options seem to be missing from the completion. For example prompt_title and  result_title (but work if you manually enter them).

- some options are being passed through as the incorrect type (I assume they need wrapping in a table somewhere. `Telescope vim_options border=false` results in an `Expected table, got string` error. Similarly, `Telescope vim_options winblend=50` results in `Option 'winblend' requires an integer value'. Setting no value at all results in an error where the finder doesn't load at all. Perhaps for each option you can check its type against that of the default config rather than maintaining a manually created map.

glepnir:
The conversion of the type of vim to the type of lua is a very troublesome problem.

Conni2461:
for what its worth: https://github.com/nvim-telescope/telescope.nvim/pull/329 also did some converting for `string` -> `boolean`. I forgot about it even existing until this conversation 😆

glepnir:
Excessive `if else` nesting does not seem to be a good design. This will increase the cyclomatic complexity of the code

Conni2461:
Yes i think your way is better :) He needs to remove it after your are done 😆 Good job

lua/telescope/command.lua
```diff
@@ -0,0 +1,90 @@
...
+    end,
+    ['number'] = function(key,val)
+      user_opts[key] = tonumber(val)
+    end,
+    ['string'] = function(key,val)
+      if val == 'nil' then
+        user_opts[key] = nil
+      end
+    end
+  }
```
Conni2461, line 27: Lets pull that table out of that scope. No need to define it more then once.

lua/telescope/command.lua
```diff
@@ -0,0 +1,90 @@
...
+
+  local _switch_metatable = {
+    __index = function(_,k)
+      print(string.format('Type of %s does not match',k))
+    end
+  }
```
Conni2461, line 33: Should we fail silent? Its not really useful for end user and more just helpful for debugging now

lua/telescope/command.lua
glepnir: hmm create a new file just for this `_switch`

lua/telescope/command.lua
glepnir: yep. just for debug we never know that what arguments that user passed.

lua/telescope/command.lua
Conni2461: No not a new file. Just out of that function scope.

glepnir: hmm value in `_switch` is a closure function use variable `user_opts`. remove out function. It will receive three arguments..

Conni2461: Oh yeah. I am dumb i haven't seen that 😆

Conni2461: I will test your PR later.

glepnir:
any update for this pr?

tami5:
@glepnir can you update readme `themes section`, under 
> to use a theme, simply append it to a built-in function "or command line" ,,,:

To include the new feature.

Thanks

glepnir:
Update the Readme.

Conni2461:
Okay i tested it works almost perfect. Good job

The only issue i have right now and i don't know if you can fix it is that when typing `:Telescope dap c...` i get all completion items for the second part. Not just dap completion items. Is that fixable? If not i think we can merge

glepnir:
> completion for the options is getting mixed with the original results for the available finders - it's not clear at all which are options and which are finders.

I don't know does we can set two completion list in command ? @tamago324

glepnir:
I found a way to do that wait.

glepnir:
Done.

Conni2461:
I will try it out in the next 15 mins. Thanks for all your work :)

glepnir:
I had update my comment with a gif

Conni2461 (approved):
I just tested it. Thats some impressive work. Good job :)

Conni2461:
@sunjon Any comments on the current version or can i merge?

sunjon:
I get some breakage if I try setting strings to whitespace only:
`:Telescope vim_options prompt_prefix="`   - errors
`:Telescope vim_options prompt_prefix=" "` - fails to load the module
`:Telescope vim_options prompt_prefix=""` - places quotes as the prompt

I think the error handling is going to make the experience much less frustrating for the user.
Feedback on what they can and can't do, rather than the raw result of different error paths.

windwp:
@glepnir

Is there anyway to add this support for autocomplete property on extensions 

sample `Telescope gh pull_request author=glepnir`

i think we can add a new function like this on extensions

``` lua
return require('telescope').register_extension {
    autocomplete= function(cmd)
     -- extensions will handle autocomplete for that
    end,
    exports= {
      gist = gh_b.gh_gist ,
      issues = gh_b.gh_issues,
      pull_request = gh_b.gh_pull_request ,
    },
}

```

glepnir:
@windwp  support complete item like  `author`?

glepnir:
@sunjon  fixed.Now

```vim
:Telescope vim_options prompt_prefix=" "  --now work 
:Telescope vim_options prompt_prefix=""   --now work 
:Telescope vim_options prompt_prefix=      --does not support
```

windwp:
@glepnir  yes  "author label .... "

glepnir:
If can passed it into config.values. now it's support.

sunjon:
I'm going to merge this now, as the basic functionality is there.
If there are items missing, we'll take them up as issues.

Thanks again to @glepnir for this

windwp:
@glepnir how can i do that
in my situation it have different autocomplete per function

`telescope gh gist` has public and secret option

`telescope gh issues` has many option

glepnir:
like this way?
```lua
local your_options = { author= ''} 
-- in gh.lua
local default_opts = require('telescope.config').values

then merge your_options into default_opts
```

PR #401, opened 2021-01-06
feat: add icons to git_status finder

sunjon:
This is WIP as there's an issue with _some_ of the double width icons getting cut in half.

giticons

Can anyone test and see if they see the same?


lua/telescope/make_entry.lua
```diff
@@ -899,28 +899,48 @@ function make_entry.gen_from_autocommands(_)
...
+      untracked = giticons.untracked or "?"
+    }
+
+    local str_1 = string.sub(entry.status, 1, 1)
+    local str_2 = string.sub(entry.status, -1)
+
+    local git_abbrev = {
+      ["A"] = icons.add,
+      ["M"] = icons.change,
+      ["D"] = icons.change,
```
cempassi, line 929: Since we can distinguish between modified and deleted now, we could use the deleted icon instead?

sunjon: ah that's a copy/paste oversight. fixed

cempassi (approved):
LGTM!

sunjon:
> LGTM!

alright, I'll assume the icon problems are on my end until more people have tested it.

sunjon:
The default icon choices are a bit random.
I'm also not sure if each X/ Y column should be treated individually, or if their state should be represented by the combination of them both (if that makes sense).

sunjon:
these are the options available if anyone else has time to test;
```
opts.git_icons = {
  added      = "  ",
  changed    = "  ",
  copied     = "  ",
  deleted    = "  ",
  untracked  = "  ",
  unmerged   = "  ",
}
```

Conni2461:
Hmm i don't know how i feel about this. I used `git status --short` on cli for years and i got so used to them now that i need my `M` and `A` and so 😆 I know i can change them now but i think the default should resemble `git status --short` or `git status --porcelain` as close as possible. 
Also having left and right the same default color is kinda weird. I sometimes stage things what i have done but i still have to do more work so i eventually will end up with `MM` and both having a different color. Thats the whole point of splitting them into two parameters.

I think that message is a mess. I am still to tired.

sunjon:
> Hmm i don't know how i feel about this. I used `git status --short` on cli for years and i got so used to them now that i need my `M` and `A` and so laughing I know i can change them now but i think the default should resemble `git status --short` or `git status --porcelain` as close as possible.
> Also having left and right the same default color is kinda weird. I sometimes stage things what i have done but i still have to do more work so i eventually will end up with `MM` and both having a different color. Thats the whole point of splitting them into two parameters.
> 
> I think that message is a mess. I am still to tired.

I agree. I wanted someone else's opinion as to how to make the display coherent.
I'm open to:
- ) leaving everything as it is.
- ) adding some color to the existing displayer
- ) find a way to make icons make sense to people used to the original format.

Conni2461:
Its all good :) Still thanks for putting it up. Its important that we talk about this and find a common ground and ultimately have a better builtin in the end :)

Right now i am thinking having a second `entry_maker` which can be enabled with an `opts` and that one can have icons. You can use only one row and so on. Because i think it will be very hard to realize that all in one `entry_maker`.

elianiva:
are you going to make the `git_icons` option configurable via `setup{}` as well? Everything seems to be working and I didn't get that cut out icons

cempassi:
> > LGTM!
> 
> alright, I'll assume the icon problems are on my end until more people have tested it.
> edit: adding an extra space fixed things a little bit more.. but still quite broken (on kitty).

I’m using alacrity, bu

lua/telescope/make_entry.lua
```diff
@@ -899,28 +899,52 @@ function make_entry.gen_from_autocommands(_)
...
+      { width = col_width},
       { remaining = true },
+      -- { width = 70 },
     }
   }
 
   local make_display = function(entry)
-    local modified = "TelescopeResultsDiffChange"
-    local staged = "TelescopeResultsDiffAdd"
+    local icons = opts.git_icons or git_icon_defaults
```
tjdevries, line 928: We should probably do something where we just create a new icons table that has a merging of the passed in icons and the defaults, if you only want to override one thing.

Also, maybe a list of the names somewhere and/or an example with pretty icons.

tjdevries (changes requested):

tjdevries:
If anyone doesn't like the new setup, let's chat in a new issue / PR. I think this is cool :)

PR #402, opened 2021-01-07
Use valid paths to preview considering cwd

delphinus:
I found `live_grep` picker shows corrupted previews when `cwd` option is used.

```vim
lua require'telescope.builtin'.live_grep{cwd = '/foofoo/barbar'}
```

When this command is executed in `/foo/bar`, and it founds `some-file/found.txt`, preview window will be corrupted because `/foo/bar/some-file/found.txt` does not exist.

### TODO

* [x] I added Unix path separator `/` to join. For Windows, I should use `\` for this? (sry, I do not have Windows)

### DIFF
--- a/lua/telescope/previewers/term_previewer.lua
+++ b/lua/telescope/previewers/term_previewer.lua
@@ -258,14 +258,15 @@ previewers.vimgrep = defaulter(function(opts)
       local win_id = status.preview_win
       local height = vim.api.nvim_win_get_height(win_id)
 
-      local filename = entry.filename
+      local p = from_entry.path(entry, true)
+      if p == nil or p == '' then return end
       local lnum = entry.lnum or 0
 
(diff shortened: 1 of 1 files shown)


delphinus:
I rewrite logic with `path.separator` in path.lua. This can work on Windows, probably.

Conni2461:
Thanks for your work :)

I wish i would have seen that yesterday because there is already a helper function to deal with that problem.
We use it everywhere besides `vimgrep` and `qflist` in `term_previewers`. Take a look how `vimgrep` in `buffer_previewer` solves that issue of `cat` in `term_previewers`

delphinus:
@Conni2461 Thank you for signposts and I rewrote logic with using `from_entry.path()` to get paths.

Conni2461 (approved):
Thanks :) LGTM

PR #405, opened 2021-01-08
WIP: Switchable window attempt

Conni2461:
New:
- Switch to buffer_previewer with `<C-w>l` in normal mode (we can talk about more mappings, for insert mode as well)

~~Fixes:~~
~~- Buffer previewer scrolling. Now we can do something like smooth scrolling @tami5~~ doing this somewhere else

Missing:
- [ ] Switch back from previewer. We have to inject a mappings so that is possible
- [ ] Handle edits with interface so that each `buffer_previewer` has to write a save function so we can realize that stuff @sunjon suggested
- [ ] We have to noop a lot of commands and mappings in previewer so the user doesn't close that window or does other unpredictable things. We can also map `:q` to `actions.close` so we close telescope correctly.

CC @hoob3rt

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -55,6 +55,14 @@ function actions.preview_scrolling_down(prompt_bufnr)
   actions.get_current_picker(prompt_bufnr).previewer:scroll_fn(30)
 end
 
+function actions.preview_switch_window_left(prompt_bufnr)
+  actions.get_current_picker(prompt_bufnr).previewer:switch_window(state.get_status(prompt_bufnr).prompt_win)
+end
+
+function actions.preview_switch_window_right(prompt_bufnr)
+  actions.get_current_picker(prompt_bufnr).previewer:switch_window(state.get_status(prompt_bufnr).preview_win)
+end
+
 -- TODO: It seems sometimes we get bad styling.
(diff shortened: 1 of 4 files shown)


elianiva:
are you planning on making the buffer scroll amount configurable? Because 30 is way too much for me and I want to be able to override it.

Conni2461:
I was thinking about that when i saw the 30 and i think that option can be even part of the `layout_defaults`. Lets say i want `horizontal` have scroll_length (whatever 😆 ) to be 30 and `vertical` to be 10.

tami5:
- is there away to make user mappings works for switch windows?
- an option for mappings should work in insert mode as well?
- no smooth scrolling

tjdevries:
@Conni2461 This is the PR you're looking for that would make this a lot easier:

https://github.com/neovim/neovim/pull/13280

I haven't completed it yet, but if you wanted to do it, it could be fun :laughing: This will make it so you can pass the lua ref directly and call it without triggering autocmds, etc.

Conni2461:
Its probably not a good idea to do this. There are a lot of things to consider because the buffer previewer doesn't have any association with a file and we have to manually handle everything. Like on a `:w` we would need to invoke our own write function that would probably has to invoke autocmds as well. And there are a lot of other things as well, that we need to consider.

Also i don't have the time and motivation to do it so closing for now.

PR #413, opened 2021-01-11
Set valid commands for apropos in macOS

delphinus:
In maacOS apropos needs args and outputs differently.

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -403,12 +403,30 @@ internal.help_tags = function(opts)
 end
 
 internal.man_pages = function(opts)
-  local pages = utils.get_os_command_output(opts.man_cmd or { 'apropos', '--sections=1', '' })
+  opts.sections = utils.get_default(opts.sections, '1')
+  opts.man_cmd = utils.get_default(opts.man_cmd, {'apropos', ''})
+
+  local sections = {}
+  for _, section in ipairs(vim.split(opts.sections, ',', true)) do
+    sections[section] = true
+  end
+  local pages = utils.get_os_command_output(opts.man_cmd)
+  local results = {}
+  for _, page in ipairs(pages) do
+    local cmd, section, desc = page:match'^(.+)%s*%((.+)%)%s*%-%s*(.*)$'
+    if sections[section] then
+      table.insert(results, {
+        cmd = cmd,
+        section = section,
+        desc = desc,
+      })
+    end
+  end
 
   pickers.new(opts, {
     prompt_title = 'Man',
     finder    = finders.new_table {
-      results = pages,
+      results = results,
       entry_maker = opts.entry_maker or make_entry.gen_from_apropos(opts),
     },
     previewer = previewers.man.new(opts),
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -418,14 +418,15 @@ internal.man_pages = function(opts)
     attach_mappings = function(prompt_bufnr)
       actions._goto_file_selection:replace(function(_, cmd)
         local selection = actions.get_selected_entry()
+        local args = selection.section .. ' ' .. selection.value
 
         actions.close(prompt_bufnr)
         if cmd == 'edit' or cmd == 'new' then
-          vim.cmd('Man ' .. selection.value)
+          vim.cmd('Man ' .. args)
         elseif cmd == 'vnew' then
-          vim.cmd('vert bo Man ' .. selection.value)
+          vim.cmd('vert bo Man ' .. args)
         elseif cmd == 'tabedit' then
-          vim.cmd('tab Man ' .. selection.value)
+          vim.cmd('tab Man ' .. args)
         end
       end)
 
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -408,30 +408,11 @@ internal.man_pages = function(opts)
     opts.man_cmd,
     vim.fn.has'mac' and {'apropos', ' '} or {'apropos', ''}
   )
-
-  local sections = {}
-  for _, section in ipairs(vim.split(opts.sections, ',', true)) do
-    sections[section] = true
-  end
-  local pages = utils.get_os_command_output(opts.man_cmd)
-  local results = {}
-  for _, page in ipairs(pages) do
-    local cmd, section, desc = page:match'^(.+)%s*%((.+)%)%s*%-%s*(.*)$'
-    if sections[section] then
-      table.insert(results, {
-        cmd = cmd,
-        section = section,
-        desc = desc,
-      })
-    end
-  end
+  opts.entry_maker = opts.entry_maker or make_entry.gen_from_apropos(opts)
 
   pickers.new(opts, {
     prompt_title = 'Man',
-    finder    = finders.new_table {
-      results = results,
-      entry_maker = opts.entry_maker or make_entry.gen_from_apropos(opts),
-    },
+    finder    = finders.new_oneshot_job(opts.man_cmd, opts),
     previewer = previewers.man.new(opts),
     sorter = conf.generic_sorter(opts),
     attach_mappings = function(prompt_bufnr)
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -404,7 +404,10 @@ end
 
 internal.man_pages = function(opts)
   opts.sections = utils.get_default(opts.sections, '1')
-  opts.man_cmd = utils.get_default(opts.man_cmd, {'apropos', ''})
+  opts.man_cmd = utils.get_default(
+    opts.man_cmd,
+    vim.fn.has'mac' and {'apropos', ' '} or {'apropos', ''}
+  )
 
   local sections = {}
   for _, section in ipairs(vim.split(opts.sections, ',', true)) do
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -403,7 +403,8 @@ internal.help_tags = function(opts)
 end
 
 internal.man_pages = function(opts)
-  opts.sections = utils.get_default(opts.sections, '1')
+  opts.sections = utils.get_default(opts.sections, {'1'})
+  assert(vim.tbl_islist(opts.sections), 'sections should be a list')
   opts.man_cmd = utils.get_lazy_default(opts.man_cmd, function()
     return vim.fn.has'mac' and {'apropos', ' '} or {'apropos', ''}
   end)
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -409,9 +409,6 @@ internal.man_pages = function(opts)
     return vim.fn.has'mac' and {'apropos', ' '} or {'apropos', ''}
   end)
   opts.entry_maker = opts.entry_maker or make_entry.gen_from_apropos(opts)
-  opts.PAGER = utils.get_lazy_default(opts.PAGER, function()
-    return vim.fn.executable('col') and 'col -bx' or ''
-  end)
 
   pickers.new(opts, {
     prompt_title = 'Man',
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -408,6 +408,9 @@ internal.man_pages = function(opts)
     return vim.fn.has'mac' and {'apropos', ' '} or {'apropos', ''}
   end)
   opts.entry_maker = opts.entry_maker or make_entry.gen_from_apropos(opts)
+  opts.PAGER = utils.get_lazy_default(opts.PAGER, function()
+    return vim.fn.executable('col') and 'col -bx' or ''
+  end)
 
   pickers.new(opts, {
     prompt_title = 'Man',
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -404,10 +404,9 @@ end
 
 internal.man_pages = function(opts)
   opts.sections = utils.get_default(opts.sections, '1')
-  opts.man_cmd = utils.get_default(
-    opts.man_cmd,
-    vim.fn.has'mac' and {'apropos', ' '} or {'apropos', ''}
-  )
+  opts.man_cmd = utils.get_lazy_default(opts.man_cmd, function()
+    return vim.fn.has'mac' and {'apropos', ' '} or {'apropos', ''}
+  end)
   opts.entry_maker = opts.entry_maker or make_entry.gen_from_apropos(opts)
 
   pickers.new(opts, {
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -406,7 +406,8 @@ internal.man_pages = function(opts)
   opts.sections = utils.get_default(opts.sections, {'1'})
   assert(vim.tbl_islist(opts.sections), 'sections should be a list')
   opts.man_cmd = utils.get_lazy_default(opts.man_cmd, function()
-    return vim.fn.has'mac' and {'apropos', ' '} or {'apropos', ''}
+    local is_darwin = vim.loop.os_uname().sysname == 'Darwin'
+    return is_darwin and {'apropos', ' '} or {'apropos', ''}
   end)
   opts.entry_maker = opts.entry_maker or make_entry.gen_from_apropos(opts)
 
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -623,28 +623,29 @@ function make_entry.gen_from_apropos(opts)
     items = {
       {},
       {},
-      {},
     },
   }
 
   local make_display = function(entry)
     -- Stopgap measure to set `width` for this field. See #414
-    local formatted = ('%-30s'):format(entry.value)
+    local formatted = ('%-30s'):format(entry.keyword)
     return displayer {
       {formatted, 'TelescopeResultsFunction'},
-      {'('..entry.section..')', 'TelescopeResultsVariable'},
       entry.description
     }
   end
 
   return function(line)
-    local cmd, section, desc = line:match'^(.+)%s*%((.+)%)%s*%-%s*(.*)$'
+    local keyword, desc = line:match'^(.+)%s+%-%s+(.*)$'
+    local section = keyword and keyword:match'%(([^)])%)' or nil
+    local cmd = keyword and keyword:match'(%S+)%(' or nil
     return sections[section] and {
       value = cmd,
       description = desc,
       ordinal = cmd,
       display = make_display,
       section = section,
+      keyword = keyword,
     } or nil
   end
 end
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -637,7 +637,7 @@ function make_entry.gen_from_apropos(opts)
   return function(line)
     local keyword, desc = line:match'^(.+)%s+%-%s+(.*)$'
     local section = keyword and keyword:match'%(([^)]+)%)' or nil
-    local cmd = keyword and keyword:match'(%S+)%(' or nil
+    local cmd = keyword and keyword:match'(%S+)%s*%(' or nil
     return sections[section] and {
       value = cmd,
       description = desc,
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -623,7 +623,7 @@ function make_entry.gen_from_apropos(opts)
     separator = ' ',
     items = {
       { width = 30 },
-      {},
+      { remaining = true },
     },
   }
 
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -612,29 +612,40 @@ function make_entry.gen_from_packages(opts)
   end
 end
 
-function make_entry.gen_from_apropos()
+function make_entry.gen_from_apropos(opts)
+  local sections = {}
+  for _, section in ipairs(vim.split(opts.sections, ',', true)) do
+    sections[section] = true
+  end
+
   local displayer = entry_display.create {
-    separator = "",
+    separator = ' ',
     items = {
-      { width = 30 },
-      { remaining = true },
+      {},
+      {},
+      {},
     },
   }
 
   local make_display = function(entry)
+    -- Stopgap measure to set `width` for this field. See #414
+    local formatted = ('%-30s'):format(entry.value)
     return displayer {
-      entry.value,
+      {formatted, 'TelescopeResultsFunction'},
+      {'('..entry.section..')', 'TelescopeResultsVariable'},
       entry.description
     }
   end
 
-  return function(result)
-    return {
-      value = result.cmd,
-      description = result.desc,
-      ordinal = result.cmd,
+  return function(line)
+    local cmd, section, desc = line:match'^(.+)%s*%((.+)%)%s*%-%s*(.*)$'
+    return sections[section] and {
+      value = cmd,
+      description = desc,
+      ordinal = cmd,
       display = make_display,
-    }
+      section = section,
+    } or nil
   end
 end
 
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -628,13 +628,11 @@ function make_entry.gen_from_apropos()
     }
   end
 
-  return function(line)
-    local cmd, _, desc = line:match("^(.*)%s+%((.*)%)%s+%-%s(.*)$")
-
+  return function(result)
     return {
-      value = cmd,
-      description = desc,
-      ordinal = cmd,
+      value = result.cmd,
+      description = result.desc,
+      ordinal = result.cmd,
       display = make_display,
     }
   end
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -635,10 +635,8 @@ function make_entry.gen_from_apropos(opts)
   end
 
   return function(line)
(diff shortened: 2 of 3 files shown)


Conni2461:
I wanted to change `get_os_command_output` to plenary job for some time now and i just did that in #428 because i needed `get_os_command_output` in that PR so i also did the refactor. Problem is we can't use `|` in plenary job like that.

So i am thinking right now that we should just do the `(1)` in lua. It should be faster then spawning an grep process anyway.
We can also just do this for all platforms and we can do an opt like `included_sections = { 1, 5, 7 }` for example. And all we need to then is construct a regex and do iterate `pages`.

Thoughts?

delphinus:
Nice. It is also good users can select sections for man. I'll implement such logic.

Conni2461:
Great then i will merge my refactor now and you need to rebase. Doesn't change anything on logic side, you are doing this, just updates how `get_os_command_output` is called :)

Thanks for all your work :) i really appreciate it

delphinus:
I used `get_os_command_output()` and successfully got results. But I found macOS `apropos` takes too much time and Neovim stucks until telescope windows appear.

```
time apropos ' ' > /dev/null

________________________________________________________
Executed in  787.30 millis    fish           external
   usr time  201.29 millis  131.00 micros  201.16 millis
   sys time  638.43 millis  783.00 micros  637.65 millis
```

So I decide to use `finders.new_oneshot_job()` to execute `apropos` asynchronously and I finally get desirable results.

スクリーンショット 0003-01-17 12 46 36

...but, I also found the preview is corrupted. This probably have been occurring on macOS the whole time until now. How can I fix this? @Conni2461

Conni2461:
Try that:
```diff
diff --git a/lua/telescope/previewers/buffer_previewer.lua b/lua/telescope/previewers/buffer_previewer.lua
index 5bf9bcd..2351446 100644
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -365,8 +365,8 @@ previewers.man = defaulter(function(_)

     define_preview = function(self, entry, status)
       local win_width = vim.api.nvim_win_get_width(self.state.winid)
-      putils.job_maker({'man', '-P', 'cat', entry.value}, self.state.bufnr, {
-        env = { ["MANWIDTH"] = win_width },
+      putils.job_maker({'man', entry.value}, self.state.bufnr, {
+        env = { ["MANWIDTH"] = win_width, ["PAGER"] = 'cat' },
         value = entry.value,
         bufname = self.state.bufname
       })
```

lua/telescope/builtin/internal.lua
```diff
@@ -403,27 +403,30 @@ internal.help_tags = function(opts)
 end
 
 internal.man_pages = function(opts)
-  local pages = utils.get_os_command_output(opts.man_cmd or { 'apropos', '--sections=1', '' })
+  opts.sections = utils.get_default(opts.sections, '1')
+  opts.man_cmd = utils.get_default(
+    opts.man_cmd,
+    vim.fn.has'mac' and {'apropos', ' '} or {'apropos', ''}
```
Conni2461, line 409: `apropos ' '` and `apropos ''` is not the same. 
So we are probably looking for `apropos -w '*'` for both

lua/telescope/make_entry.lua
```diff
@@ -612,31 +612,41 @@ function make_entry.gen_from_packages(opts)
...
     items = {
-      { width = 30 },
-      { remaining = true },
+      {},
+      {},
     },
   }
 
   local make_display = function(entry)
+    -- Stopgap measure to set `width` for this field. See #414
```
Conni2461, line 630: We should actually fix it rather than doing that hack.
One thing we can do is pass in like `truncate = false` which will disable `truncate`.

lua/telescope/make_entry.lua
```diff
@@ -612,31 +612,41 @@ function make_entry.gen_from_packages(opts)
   end
 end
 
-function make_entry.gen_from_apropos()
+function make_entry.gen_from_apropos(opts)
+  local sections = {}
+  for _, section in ipairs(vim.split(opts.sections, ',', true)) do
+    sections[section] = true
+  end
```
Conni2461, line 619: Hmmm. I thought we can just pass in `sections = { 1, 5 }` but i only thought about the lua side. I am not sure if i like a comma separated string

lua/telescope/make_entry.lua
```diff
@@ -612,31 +612,41 @@ function make_entry.gen_from_packages(opts)
...
       entry.description
     }
   end
 
   return function(line)
-    local cmd, _, desc = line:match("^(.*)%s+%((.*)%)%s+%-%s(.*)$")
-
-    return {
+    local keyword, desc = line:match'^(.+)%s+%-%s+(.*)$'
+    local section = keyword and keyword:match'%(([^)])%)' or nil
```
Conni2461, line 640: rather then `[^)]` why not `%d`

delphinus: Sections in man pages may have non-digit characters such as `1m`, `x`, etc. ([man page - Wikipedia][])

[man page - Wikipedia]: https://en.wikipedia.org/wiki/Man_page

On my Mac, many pages have such sections.

```
apropos ' ' | egrep -v '\(\d+\)' | head

TIFFClose(3tiff)         - close a previously opened TIFF file
TIFFDataWidth(3tiff)     - Get the size of TIFF data types
TIFFError(3tiff), TIFFSetErrorHandler(3tiff) - library error handling interface
TIFFFieldDataType(3tiff) - Get TIFF data type from field information
TIFFFieldName(3tiff)     - Get TIFF field name from field information
TIFFFieldPassCount(3tiff) - Get whether to pass a count to TIFFGet/SetField
TIFFFieldReadCount(3tiff) - Get number of values to be read from field
TIFFFieldTag(3tiff)      - Get TIFF field tag value from field information
TIFFFieldWriteCount(3tiff) - Get number of values to be written to field
TIFFFlush(3tiff), TIFFFlushData(3tiff) - flush pending writes to an open TIFF file
```

lua/telescope/builtin/internal.lua
delphinus: hmm... apropos in macOS is so primitive and has no options such as `-w`.

[apropos Man Page - macOS - SS64.com](https://ss64.com/osx/apropos.html)

I saw here and knew `apropos ' '` is enough for this time.

Conni2461: Oh okay. I just did `apropos ' ' | wc -l` and `apropos '' | wc -l` and got a different count. 😆 Lets keep it that way

lua/telescope/make_entry.lua
Conni2461: You learn something every day :)

lua/telescope/make_entry.lua
delphinus: I wrote this for launching from commands such as `:Telescope man_pages sections=1,5`. Ideally, `convert_user_opts()` should parse these comma-separated values into Lua tables.

I will file an issue for this.

lua/telescope/make_entry.lua
delphinus: I found way to solve #414 fundamentally. I rewrote `strdisplaywidth()` with `ffi` module and enable to use `width` feature in vim.loop.

https://github.com/delphinus/telescope.nvim/commit/f5d6c98293dc25a59d88811c55e29a9934a340ec

I will test more and apple this to this PR.

Conni2461: First of all i love the idea :) Good job

Now three other things:
- I am concerned about windows. I don't know `linetabsize_col`, so is it a vim function we can always call on all platforms or is it a unix function? 
- Do we wanna support not jit at all. Every time i do something with `ffi` i currently just ignore the fact. Telescope without jit is not fast enough anyway. Just thinking out load
- i think you don't need to create the `c_` types you should just be able to pass the lua int and lua string. I think that just works for primitive types. At least thats what i am doing 😆 And thats what the documentation does: https://luajit.org/ext_ffi.html

delphinus: > * I am concerned about windows. I don't know linetabsize_col, so is it a vim function we can always call on all platforms or is it a unix function?

I just converted the logic of `strdisplaywidth()` from Neovim's source. `linetabsize_col` is in it. It should work everywhere Neovim works. ;)

https://github.com/neovim/neovim/blob/a1ed941a7881122fda2fd48e71e890ed55e4d08e/src/nvim/eval/funcs.c#L9845-L9858

For other 2 points, I will reply later. 👍

Conni2461: I thought so. Thanks for clarifying :)

delphinus: > * Do we wanna support not jit at all. Every time i do something with ffi i currently just ignore the fact. Telescope without jit is not fast enough anyway. Just thinking out load. Nvm we also doing if jit in path.shorten. Didn't know that

I see. I'm not familiar with LuaJIT affairs in Neovim, but I think the fallback measure `#str - col` is enough for ordinary latin1 characters. I'll use that.

> * i think you don't need to create the c_ types you should just be able to pass the lua int and lua string. I think that just works for primitive types. At least thats what i am doing 😆 And thats what the documentation does: https://luajit.org/ext_ffi.html

I found `c_col` is unnecessary as you say, but still necessary for `c_str`.

```lua
local ffi = require('ffi')

ffi.cdef[[
  typedef unsigned char char_u;
  int linetabsize_col(int startcol, char_u *s);
]]

local col = 0
local str = 'foobar'
print( ffi.C.linetabsize_col(col, str) - col )
--local c_str = ffi.new('char[?]', #str + 1)
--ffi.copy(c_str, str)
--print( ffi.C.linetabsize_col(col, c_str) - col )
```

```vim
E5113: Error while calling lua chunk: scratch/strdisplaywidth.lua:10: bad argument #2 to 'linetabsize_col' (cannot convert 'string' to 'unsigned char *')
```

----

I'm thinking for posting another PR for this. I found other problems for `strdisplaywidth` 🤔

delphinus: Created #413. I'll try to solve problems for multi-byte strings later in another PR...

Conni2461: > but still necessary for c_str

Oh yeah it works for `const char*` but not for `unsigned char*`

delphinus: This will be fixed in #464.

delphinus:
@Conni2461

> Try that:
>
> ```diff
> diff --git a/lua/telescope/previewers/buffer_previewer.lua b/lua/telescope/previewers/buffer_previewer.lua
> index 5bf9bcd..2351446 100644
> --- a/lua/telescope/previewers/buffer_previewer.lua
> +++ b/lua/telescope/previewers/buffer_previewer.lua
> @@ -365,8 +365,8 @@ previewers.man = defaulter(function(_)
>
>      define_preview = function(self, entry, status)
>        local win_width = vim.api.nvim_win_get_width(self.state.winid)
> -      putils.job_maker({'man', '-P', 'cat', entry.value}, self.state.bufnr, {
> -        env = { ["MANWIDTH"] = win_width },
> +      putils.job_maker({'man', entry.value}, self.state.bufnr, {
> +        env = { ["MANWIDTH"] = win_width, ["PAGER"] = 'cat' },
>          value = entry.value,
>          bufname = self.state.bufname
>       }) 
> ```

This did not solve the problem. At least in macOS, telescope still shows broken previews.

Finally I found this is due to `env` params.

```diff
diff --git a/lua/telescope/previewers/buffer_previewer.lua b/lua/telescope/previewers/buffer_previewer.lua
index 0491640..07cd4c1 100644
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -379,8 +379,8 @@ previewers.man = defaulter(function(_)
 
     define_preview = function(self, entry, status)
       local win_width = vim.api.nvim_win_get_width(self.state.winid)
-      putils.job_maker({'man', '-P', 'cat', entry.value}, self.state.bufnr, {
-        env = { ["MANWIDTH"] = win_width },
+      local cmd = ('MANWIDTH=%d PAGER=cat man %s'):format(win_width, entry.value)
+      putils.job_maker({'bash', '-c', cmd}, self.state.bufnr, {
         value = entry.value,
         bufname = self.state.bufname
       })
```

Without using `env` and including them in bash commands, it shows valid previews. Strange... plenary.nvim has bugs??

Conni2461:
Does plenary has bugs? Of course it does like all software. Does `plenary.job` env has a bug? I highly doubt it :)

Your command is something completely different. Let me explain.
We are using `uv.spawn` [see](https://github.com/luvit/luv/blob/master/docs.md#uvspawnpath-options-on_exit) to spawn external processes. This does not involve bash or any other shell at all. So we are spawing processes at "os level" so if there is a "bug" then i would guess its mac os that does something differently compared to other unix systems. We actually also doesn't even touch env besides converting the table into the correct format. It goes straight into `uv.spawn` as option.

Spawing a bash might also be not the best idea regards to performance and env. While uv always guaranties us that we have a clear env, bash would load in other env vars. Just run that script with `:luafile %`
```lua
local Job = require'plenary.job'

local s1 = os.clock()
Job:new({
  command = 'env',
  args = {},
  env = { ['PAGER'] = 'cat', ['MANWIDTH'] = 100 },
  on_exit = function(j)
    print(vim.inspect(j:result()))
    print(os.clock() - s1)
  end
}):sync()

local s2 = os.clock()
Job:new({
  command = 'bash',
  args = { '-c', 'env' },
  env = { ['PAGER'] = 'cat', ['MANWIDTH'] = 100 },
  on_exit = function(j)
    print(vim.inspect(j:result()))
    print(os.clock() - s2)
  end
}):sync()
```

For me it takes twice as much time to complete the job and i have a much richer env.

You can also try setting: `['PAGER'] = ''` rather than cat. Maybe that fixes it. That still has the same behaivior on linux. At least for me 😆

clason:
One thing one has to keep in mind with macOS is that environment variables differ (wildly!) whether you open nvim from a shell or via a GUI from the Finder. (And it depends on the version how to set the latter...)

delphinus:
After reading your signposts, I tested `man ls` with several conditions.

```lua
local Job = require'plenary.job'

for i, env in ipairs{
  {},
  {PAGER = ''},
  {PAGER = 'cat'},
  {PAGER = 'col -bx'},
} do
  local env_result
  Job:new{
    command = 'env',
    args = {},
    env = env,
    on_exit = function(j) env_result = vim.inspect(j:result()) end,
  }:sync()
  Job:new{
    command = 'man',
    args = {'ls'},
    env = env,
    on_exit = function(j)
      print(('%d: %s, env: %s'):format(i, j:result()[4], env_result))
    end,
  }:sync()
end
```

```bash
# launch bash with minimum envs
$ env -i bash

bash-3.2$ env
PWD=/Users/jinnouchi.yasushi
SHLVL=1
_=/usr/bin/env

# execute the script above
bash-3.2$ nvim +'luafile /tmp/foo.lua' +q
1: N^HNA^HAM^HME^HE, env: {}
2: N^HNA^HAM^HME^HE, env: { "PAGER=cat" }
3: N^HNA^HAM^HME^HE, env: { "PAGER=" }
4: NAME, env: { "PAGER=col -bx" }
```

This script shows the 4th line from `man ls`. The line contains the text `NAME`. The result shows `PAGER=`, `PAGER=cat`, and no envs make `man ls` show formatted texts.

Only the last attempt `PAGER="col -bx"` is the winner. So I want to use this for telescope. How do you think?

Conni2461:
Yeah we can do this. Would just check if col is installed. Just to make sure and if not installed we just do `PAGER=`.

We can do that check once on script level, or before `return previewers.new_buffer_previewer` to do the check once when `:Telescope man_pages` gets called. I prefer latter so people can just install `col` without reloading vim. Even thought `col` is part of `linux-utils` for me so it should be present on all systems. But i wouldn't bet on it 🤣 

Good work figuring that out :)

lua/telescope/make_entry.lua
delphinus: Issued in #447 and will be fixed in #450.

54f68cf I've removed this logic to expect #450 to be merged.

delphinus:
The remained problem to be fixed is https://github.com/nvim-telescope/telescope.nvim/pull/413#discussion_r559087447. I'm waiting for #464 to be merged.

lua/telescope/builtin/internal.lua
```diff
@@ -403,27 +403,33 @@ internal.help_tags = function(opts)
...
-  local pages = utils.get_os_command_output(opts.man_cmd or { 'apropos', '--sections=1', '' })
+  opts.sections = utils.get_default(opts.sections, {'1'})
+  assert(vim.tbl_islist(opts.sections), 'sections should be a list')
+  opts.man_cmd = utils.get_lazy_default(opts.man_cmd, function()
+    return vim.fn.has'mac' and {'apropos', ' '} or {'apropos', ''}
+  end)
+  opts.entry_maker = opts.entry_maker or make_entry.gen_from_apropos(opts)
+  opts.PAGER = utils.get_lazy_default(opts.PAGER, function()
+    return vim.fn.executable('col') and 'col -bx' or ''
+  end)
```
Conni2461, line 414: I wanted that in previewers. No need to pass it in a opts. You can do that.
```lua
previewers.man = defaulter(function()
  local pager = ... -- this will run once. When the previewer gets created
  return previewers.new_buffer_previewer {
```

lua/telescope/builtin/internal.lua
```diff
@@ -403,27 +403,33 @@ internal.help_tags = function(opts)
 end
 
 internal.man_pages = function(opts)
-  local pages = utils.get_os_command_output(opts.man_cmd or { 'apropos', '--sections=1', '' })
+  opts.sections = utils.get_default(opts.sections, {'1'})
+  assert(vim.tbl_islist(opts.sections), 'sections should be a list')
+  opts.man_cmd = utils.get_lazy_default(opts.man_cmd, function()
+    return vim.fn.has'mac' and {'apropos', ' '} or {'apropos', ''}
```
Conni2461, line 409: A vimL free version to check if mac would be: `vim.loop.os_uname().sysname == 'Darwin'`

lua/telescope/builtin/internal.lua
delphinus: Got it. Moved logic but still can select `PAGER` by `opts.PAGER`. a8ee358

lua/telescope/builtin/internal.lua
delphinus: Oh, good stuff. I did not know this. ee5c8f0

lua/telescope/make_entry.lua
delphinus: Fixed on 7de61bc

delphinus:
And fixed https://github.com/nvim-telescope/telescope.nvim/pull/413#discussion_r559087447 by #464.

lua/telescope/make_entry.lua
```diff
@@ -613,31 +613,39 @@ function make_entry.gen_from_packages(opts)
...
+    sections[section] = true
+  end
+
   local displayer = entry_display.create {
-    separator = "",
+    separator = ' ',
     items = {
       { width = 30 },
-      { remaining = true },
+      {},
```
Conni2461, line 626: we should keep the `remaining = true` here after that i think we are done right?

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -372,16 +372,19 @@ previewers.help = defaulter(function(_)
...
   return previewers.new_buffer_previewer {
     get_buffer_by_name = function(_, entry)
       return entry.value
     end,
 
     define_preview = function(self, entry, status)
       local win_width = vim.api.nvim_win_get_width(self.state.winid)
-      putils.job_maker({'man', '-P', 'cat', entry.value}, self.state.bufnr, {
-        env = { ["MANWIDTH"] = win_width },
+      putils.job_maker({'man', entry.value}, self.state.bufnr, {
```
Conni2461, line 386: don't we also need to do here now: `selection.section .. ' ' .. selection.value`?

Conni2461:
I just tested it. It doesn't work on linux. I get 0 results. But count tells me: `23086/25618`. I will do some digging

tami5:
> I just tested it. It doesn't work on linux. I get 0 results. But count tells me: `23086/25618`. I will do some digging

Same empty list

Conni2461:
Okay these changes seems to work best for me:
```diff
diff --git a/lua/telescope/make_entry.lua b/lua/telescope/make_entry.lua
index 7355a73..0409307 100644
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -623,28 +623,26 @@ function make_entry.gen_from_apropos(opts)
     separator = ' ',
     items = {
       { width = 30 },
-      {},
+      { remaining = true },
     },
   }
 
   local make_display = function(entry)
     return displayer {
-      { entry.keyword, 'TelescopeResultsFunction' },
+      { entry.value .. ' (' .. entry.section .. ')', 'TelescopeResultsFunction' },
       entry.description
     }
   end
 
   return function(line)
-    local keyword, desc = line:match'^(.+)%s+%-%s+(.*)$'
-    local section = keyword and keyword:match'%(([^)]+)%)' or nil
-    local cmd = keyword and keyword:match'(%S+)%(' or nil
+    local cmd, section, desc = line:match'^([^%s]*)%s%(([^)]+)%)%s+%-%s+(.*)$'
+    if not cmd then return nil end
     return sections[section] and {
       value = cmd,
       description = desc,
       ordinal = cmd,
       display = make_display,
       section = section,
-      keyword = keyword,
     } or nil
   end
 end
```

The issue with the prev first regex is this example:
`abrev (1)              - long name - desc`
Long name was part of the name now, even tho it should have been part of the description. Also one regex seems like the ideal think to do. But i don't know the format on `macos` apropos :| so we might need to do a even smart regex so both formats are accepted. Maybe with `?`

delphinus:
Sorry, I fixed to show candidates in Linux on 2a69687.

From reading your reports, I invested the result of `apropos ' '` on my box again. It outputs such lines below...

```sh
# most general entry
ls(1)                    - list directory contents

# multi `-`'s appear
php(1)                   - PHP Command Line Interface 'CLI' php-cgi - PHP Common Gateway Interface 'CGI' command

# multi entries in one line
bzcmp(1), bzdiff(1)      - compare bzip2 compressed files

# super long line
builtin(1), !(1), %(1), .(1), :(1), @(1), {(1), }(1), alias(1), alloc(1), bg(1), bind(1), bindkey(1), break(1), breaksw(1), builtins(1), case(1), cd(1), chdir(1), command(1), complete(1), continue(1), default(1), dirs(1), do(1), done(1), echo(1), echotc(1), elif(1), else(1), end(1), endif(1), endsw(1), esac(1), eval(1), exec(1), exit(1), export(1), false(1), fc(1), fg(1), filetest(1), fi(1), for(1), foreach(1), getopts(1), glob(1), goto(1), hash(1), hashstat(1), history(1), hup(1), if(1), jobid(1), jobs(1), kill(1), limit(1), local(1), log(1), login(1), logout(1), ls-F(1), nice(1), nohup(1), notify(1), onintr(1), popd(1), printenv(1), pushd(1), pwd(1), read(1), readonly(1), rehash(1), repeat(1), return(1), sched(1), set(1), setenv(1), settc(1), setty(1), setvar(1), shift(1), source(1), stop(1), suspend(1), switch(1), telltc(1), test(1), then(1), time(1), times(1), trap(1), true(1), type(1), ulimit(1), umask(1), unalias(1), uncomplete(1), unhash(1), unlimit(1), unset(1), unsetenv(1), until(1), wait(1), where(1), which(1), while(1) - shell built-in commands

# cmd containing spaces
Keychain Circle Notification(8) - part of iCloud Keychain syncing

# corrupted lines (should be ignored)
usage: /usr/libexec/makewhatis.local [options] directories ...
```

I tried to use the one regexp to solve all types. With it, I tested with this script below.

```lua
local Job = require'plenary.job'

_G.parsed = {}

Job:new{
  command = 'apropos',
  args = vim.loop.os_uname().sysname == 'Darwin' and {' '} or {''},
  env = {},
  on_exit = function(j)
    local result = j:result()
    local count = 0
    local re = '^(.-)%s*%(([^)]+)%).-%s+%-%s+(.*)$'
    local delim = ' | '
    local parsed = io.open('/tmp/parsed.txt', 'w')
    for _, line in ipairs(result) do
      local cmd, section, desc = line:match(re)
      if cmd then
        count = count + 1
        parsed:write(cmd..delim..section..delim..desc..'\n')
      end
    end
    parsed:close()
    print(('all: %d, parsed: %d'):format(#result, count))
  end,
}:sync()
```

```
all: 6337, parsed: 6336
```

Yes! This one regexp can parse all entries (and ignore one line) on my box at least. I'll rewrite logic with this.

delphinus:
I tested this script above on ubuntu by [this Dockerfile][]. It seems this works good both on Linux and macOS.

[this Dockerfile]: https://gist.github.com/delphinus/321750e40869937294e0821d918d30c5

Conni2461 (approved):
LGTM. There is one more comment left in previewer. I think we should do `entry.selection, entry.value` now.
I give it a try and will merge if you answered that comment.

Thanks for all your work :)

delphinus:
Oops, wait a minute. I'm still writing the logic for "this one regexp" ;)

lua/telescope/previewers/buffer_previewer.lua
delphinus: Good catch! Fixed on 79fa15a

delphinus:
Done! check it.

Conni2461:
Works really good now thanks for all that hard work :) And i think its a cool new addition being able to set the `sections` :D

@tami5 Could you take a second and retest after that i am going to merge

tami5:
LGTM. Thanks @delphinus

PR #423, opened 2021-01-12
Do not remove a part of branch names

delphinus:
Before this, branch names such as `feature/foo` became `foo`. Fix it.


Conni2461:
Thanks :)

PR #430, opened 2021-01-12
Preview translated help

delphinus:
When `helplang` is set, it searches `tags-XX` and takes priority to them over `@en` entries.

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -357,9 +357,9 @@ internal.help_tags = function(opts)
 
   local function iterate_tags(filename, fn)
     local delimiter = string.char(9)
-    local lines = vim.split(path.read_file(filename), '\n')
+    local lines = vim.split(path.read_file(filename), '\n', true)
     for _, line in ipairs(lines) do
-      local fields = vim.split(line, delimiter)
+      local fields = vim.split(line, delimiter, true)
       if #fields == 3 then
         fn{
           name = fields[1],
@@ -383,7 +383,7 @@ internal.help_tags = function(opts)
 
   local helplangs = vim.tbl_filter(
     function(lang) return lang_tags[lang] and true or false end,
-    vim.split(vim.o.helplang, ',')
+    vim.split(vim.o.helplang, ',', true)
   )
 
   local tags = {}
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -358,18 +358,21 @@ internal.help_tags = function(opts)
     end
   end
 
-  local function iterate_files(files, fn)
+  local function iterate_files(lang_files, fn)
     local delimiter = string.char(9)
-    for _, file in ipairs(files) do
-      local lines = vim.split(path.read_file(file), '\n', true)
-      for _, line in ipairs(lines) do
-        local fields = vim.split(line, delimiter, true)
-        if #fields == 3 then
-          fn{
-            name = fields[1],
-            filename = help_files[fields[2]],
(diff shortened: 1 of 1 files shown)


lua/telescope/builtin/internal.lua
```diff
@@ -334,18 +334,51 @@ end
...
+  local all_tag_lang_files = {}
   local all_help_files = {}
-  for _, v in ipairs(vim.split(vim.fn.globpath(vim.o.runtimepath, 'doc/*', 1), '\n')) do
+  local langs = {}
+  for _, v in ipairs(vim.fn.globpath(vim.o.runtimepath, 'doc/*', 1, 1)) do
     local split_path = vim.split(v, path.separator, true)
     local filename = split_path[#split_path]
     if filename == 'tags' then
       table.insert(all_tag_files, v)
+    elseif filename:match('^tags%-..$') then
```
Conni2461, line 345: Does the file always has a 2 char width lang code or are there exceptions?
Because i most definitely see exceptions in `/etc/locale.gen` like some have 3

lua/telescope/builtin/internal.lua
```diff
@@ -357,12 +390,22 @@ internal.help_tags = function(opts)
...
+              break
+            end
+          end
+          if not found then
+            table.insert(tags, {
+              name = matches[1],
+              filename = all_help_files[matches[2]],
+              cmd = matches[3]
+            })
+          end
```
Conni2461, line 408: What about when people have setup `vim.o.helplang` and still want the eng version at some point maybe because the translated version for something is not yet up to date? (Is that an issue) Should we do an opt so people can specify the preferred language

lua/telescope/builtin/internal.lua
delphinus: [Neovim doc][] says the extension of help files has `.XXx` format and tags have `tags-XX` filename. The lang `XX` has two letters. I see the 3rd and later letters are cut in the source.

[Neovim doc]: https://neovim.io/doc/user/helphelp.html#help-translated

https://github.com/neovim/neovim/blob/77a6049e07d070470d3274348ac61ab9b96cac6f/src/nvim/option.c#L874

lua/telescope/builtin/internal.lua
delphinus: Good catch. I'll add it.

Conni2461:
Is it worth looking at it again? Just ping me if you need another review or mark the PR as ready for review :)

delphinus:
Thank you and wait just a moment.

I implemented `ignore_helplang` option, but I'm considering better interface.

```lua
opts = {
  lang = <string>,       -- languages to show. ex. ja,ru
  fallback = <boolean>,  -- if true, fallbacks to en help for ones that are not found
}
```

```vim
" use helplang and show en help when it is not found.
:Telescope help_tags

" use helplang, but do not show en help when it is not found.
:Telescope help_tags fallback=false

" ignore helplang and show help from specified langs.
" when helps cannot be found, show en help.
:Telescope help_tags lang=ja,ru

" ignore helplang and show help from specified langs only.
:Telescope help_tags lang=ja,ru fallback=false
```

When you need the latest help, you can use just `:Telescope help_tags lang=en`.

delphinus:
finished. review again @Conni2461

Conni2461 (approved):
LGTM Thanks :) English still works thats most important for me.
I can't test other languages, no idea how to set it up, i trust you that it also works :)

@sunjon you originally wrote that. Mind looking over it for a sec? If you don't find any time i will merge

Conni2461:
I haven't heard of anyone so i just merge it :)

Issue #437, opened 2021-01-16
Should all args in a command be converted into Lua's syntax?


delphinus:
**Is your feature request related to a problem? Please describe.**

When I call `:Telescope` with builtin config values in args, [convert_user_opts][] parse them and translate vimL values to Lua's ones.

[convert_user_opts]: https://github.com/nvim-telescope/telescope.nvim/blob/00e898a1f919a2330c3b0fdf21429627cbaf2175/lua/telescope/command.lua#L9

```lua
:Telescope find_files color_devicons=false

=> opts = {color_devicons = false}
```

But this does not parse other opts defined in each subcommand. I found this in developing #430.

```lua
:lua require'telescope.builtin'.help_tags{fallback = false}

=> opts = {fallback = false}

:Telescope help_tags fallback=false

=> opts = {fallback = 'false'}
```

**Describe the solution you'd like**

I think such opts should be also converted into Lua values in the same way. How do you think?

**Describe alternatives you've considered**

**Additional context**

Conni2461:
I ran into the same issue and wanted to ask @glepnir if this was possible :)

glepnir:
should be possible. I will improve it later and test.

glepnir:
@delphinus  now in my local this can work, other request?
```
:Telescope help_tags fallback=false
``

delphinus:
Yeah, that's it ;)

glepnir:
closed by #439

PR #439, opened 2021-01-16
support convert custom command arg

glepnir:
support #437

### DIFF
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -1,8 +1,14 @@
+local fn = vim.fn
 local themes = require('telescope.themes')
 local builtin = require('telescope.builtin')
 local extensions = require('telescope._extensions').manager
 local config = require('telescope.config')
 local command = {}
+
+local bool_type = {
+  ['false'] = false,
+  ['true'] = true
+}
 
 -- convert command line string arguments to
 -- lua number boolean type and nil value
@@ -29,6 +35,9 @@ local function convert_user_opts(user_opts)
       if val == '"' then
         user_opts[key] = ''
       end
+      if bool_type[val] ~= nil then
+        user_opts[key] = bool_type[val]
+      end
     end
   }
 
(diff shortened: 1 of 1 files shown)


glepnir:
@delphinus  Does this can work for you?

Conni2461:
Works for me :) Thanks

tami5:
Nice @glepnir , I love how powerful this is becoming

delphinus:
confirmed. It does all good. thx!

glepnir:
Ok Now merged.

MuhammedZakir:
A doubt:

```
    ['boolean'] = function(key,val)
      if val == 'false' then
        user_opts[key] = false
      end
      user_opts[key] = true
    end,
```

From what I understand, the `user_opts[key] = true` statement is executed regardless of `val`'s value? Shouldn't this be a if-else conditional for this to work correctly?

```
    ['string'] = function(key,val)
      if val == 'nil' then
        user_opts[key] = nil
      end
      if val == '""' then
        user_opts[key] = ''
      end
      if val == '"' then
        user_opts[key] = ''
      end
      if bool_type[val] ~= nil then
        user_opts[key] = bool_type[val]
      end
    end
```

Isn't if-elseif better for this? Not that this is wrong, but I feel like it's redundant as each if block is executed at the moment even when it doesn't need to. For example, if `val` is `nil`, there is no need to execute subsequent statements.

glepnir:
I  had rewrire this in my local branch.  make pr later. but don't use if else. too many if else will increase cycle complexity of code

MuhammedZakir:
> I had rewrire this in my local branch. make pr later

Okay! :-)

> but don't use if else. too many if else will increase cycle complexity of code

You're right about that. I assume you're talking about second one, right? But, the code already have those if conditionals. I personally think changing it to `if-elseif` makes it easier to read/debug because if one `if` statement is matched, you can ignore all other `if` branches. That's not the case with current code/style. This is my personal option. So I maybe wrong, and to be honest, for this one, it's probably very easy to know we only need to care about one `if` conditional.

PR #444, opened 2021-01-19
add ignore_current_buffer options to builtin buffers

relnod:
Hello,
this adds a new option `ignore_current_buffer` to the builtin `buffers`.

When `ignore_current_buffer` is set to `true`, the current buffer won't get listed.
The option also disables setting the `default_selection_idx`, since it is not needed in that case.

This change greatly improves buffer switching for me.

I'm also welcome for suggestions on how to handle this differently.
Also please correct me if I made a mistake while converting the boolean statement, or if you feel it's uneccecary.

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -472,10 +472,16 @@ end
 
 internal.buffers = function(opts)
   local bufnrs = filter(function(b)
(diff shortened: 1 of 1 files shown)


Conni2461:
LGTM. Works for me as well.
Thanks :)

PR #450, opened 2021-01-20
support custom table type keyword

### LINKED ISSUE #447: Should Telescope command parse more complex arguments?

I'm refactoring `man_pages` picker in #413. I added an option: `sections` to limit sections to list man pages.

```vim
:Telescope man_pages sections=1,3,5
```

In Lua codes, it transforms `1,3,5` string into `{'1', '3', '5'}` table. Should `convert_user_opts()` have this logic to use this conversion for any other command?

```vim
sections=1,2,3tiff,n,x
```

```lua
{sections = {'1', '2', '3tiff', 'n', 'x'}}`
```

It should convert values into not numbers but strings.

Any thoughts?

glepnir:
try to close #447  cc @Conni2461

### DIFF
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -2,7 +2,6 @@ local themes = require('telescope.themes')
 local builtin = require('telescope.builtin')
 local extensions = require('telescope._extensions').manager
 local config = require('telescope.config')
-local utils = require('telescope.utils')
 local command = {}
 
 local arg_value = {
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -2,6 +2,7 @@ local themes = require('telescope.themes')
 local builtin = require('telescope.builtin')
 local extensions = require('telescope._extensions').manager
 local config = require('telescope.config')
+local utils = require('telescope.utils')
 local command = {}
 
 local arg_value = {
@@ -66,7 +67,7 @@ end
 --   opts = {
 --      cwd = '***',
 -- }
-function command.run_command(args)
+local function run_command(args)
   local user_opts = args or {}
   if next(user_opts) == nil and not user_opts.cmd then
     print('[Telescope] your command miss args')
@@ -99,6 +100,34 @@ function command.run_command(args)
   if rawget(extensions,cmd) then
     extensions[cmd][cmd](opts)
   end
+end
+
+function command.load_command(cmd,...)
+  local args = {...}
+  local user_opts = {}
+  user_opts['cmd'] = cmd
+  user_opts.opts = {}
+
+  for _,arg in ipairs(args) do
+    if arg:find('=',1) == nil then
+      user_opts['extension_type'] = arg
+    else
+      local param = utils.split_by_pattern(arg,'[^=]+')
+      if param[1] == 'theme' then
+        user_opts['theme'] = param[2]
+      elseif param[2]:find(',',1) ~= nil then
+        if param[1] == 'search' then
+          user_opts.opts[param[1]] = param[2]
+        else
+          user_opts.opts[param[1]] = utils.split_by_pattern(param[2],'[^,]+')
+        end
+      else
+        user_opts.opts[param[1]] = param[2]
+      end
+    end
+  end
+
+  run_command(user_opts)
 end
 
 return command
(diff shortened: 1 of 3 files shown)


lua/telescope/command.lua
```diff
@@ -101,4 +102,32 @@ function command.run_command(args)
...
+      user_opts['extension_type'] = arg
+    else
+      local param = utils.split_by_pattern(arg,'[^=]+')
+      if param[1] == 'theme' then
+        user_opts['theme'] = param[2]
+      elseif param[2]:find(',',1) ~= nil then
+        if param[1] == 'search' then
+          user_opts.opts[param[1]] = param[2]
+        else
+          user_opts.opts[param[1]] = utils.split_by_pattern(param[2],'[^,]+')
```
Conni2461, line 122: Why not vim.split?

glepnir: this function useful than `vim.fn.split`  with lua regex

lua/telescope/command.lua
```diff
@@ -101,4 +102,32 @@ function command.run_command(args)
...
+
+  for _,arg in ipairs(args) do
+    if arg:find('=',1) == nil then
+      user_opts['extension_type'] = arg
+    else
+      local param = utils.split_by_pattern(arg,'[^=]+')
+      if param[1] == 'theme' then
+        user_opts['theme'] = param[2]
+      elseif param[2]:find(',',1) ~= nil then
+        if param[1] == 'search' then
```
Conni2461, line 119: What about other parameters which could expect a `,`. I would do the other way around. Define a table with options that need to be converted like `find_command`, `vimgrep_arguments`, `selections` and convert them.

We need other opinions on that
And i gotta go now :)

Still thanks for your work :)

lua/telescope/command.lua
Conni2461: There is a `vim.split` which is in `shared.lua` which is exactly that and there is a `vim.gsplit`. See 
https://github.com/neovim/neovim/blob/f75be5e9d510d5369c572cf98e78d9480df3b0bb/runtime/lua/vim/shared.lua#L53-L113

lua/telescope/command.lua
glepnir: you mean create a table that collection table type keywords then if match we convert the value right? But check issue some extensions developer will create their own keywords .so these keywords we don't know which type it is. so I try to use comma as a deal to convert value to table

Conni2461: As i said earlier these extension dev gotta do something like this to support lua and vim side:
```lua
opts.option = type(opts.option) == 'table' and opts.option or vim.split(opts.option, ',')
```
I kinda have it that way not convert one rather than converting and its not expected :|

lua/telescope/command.lua
glepnir: Got it. remove custom split function

lua/telescope/command.lua
glepnir: If they can do this. everything come to easy :)

glepnir:
@Conni2461 
Is it possible to provide a registration function for extension developers to register custom keywords and types

Conni2461:
Hmm we could do:
```lua
local split_keywords = {
  ['find_command'] = true,
  ['vimgrep_arguments'] = true,
  ['sections'] = true
}

function command.load_command(cmd, ...)
...
  if split_keywords[param[2]] then
    -- split
  end
...
end

function command.register_keyword(keyword)
  split_keywords[keyword] = true
end
```

And then maybe a connection with the current interface

glepnir:
done.

Conni2461:
Great :) thanks. Do you like it or not so much? Because it was just an idea 😆 and you are the viml side expert :)

One thing we can discuss is a interface with extension so people don't have to `require'command``. Any ideas?

glepnir:
LGTM. I thought for a long time. I find it difficult to do it based on the current code. Because the key in opts can be added at will, use closures for each function to capture opts? This seems to require a lot of refactoring. Therefore, if extension developers want to explicitly type completion on the command line, they must manually require command

Conni2461 (approved):
The code part looks good to me. Thanks @glepnir 

But i am still not sure that this is the best way to do it, so if you are okay with the current state merge it.

Maybe we come up with a better idea to handle that even more general later :)

glepnir:
> But i am still not sure that this is the best way to do it

Agree with it. but check some issue and other pr.  they need this pr. so merge it.

PR #453, opened 2021-01-22
fix: escape filename before opening it

elianiva:
I have a weird case where I need to open a file with `$` in its filename so we need to escape it.


elianiva:
Also, I've tried to escape it for the previewer as well but couldn't make it work. It just fails silently, no error message and no preview.

Conni2461:
We can do this.

It fails in `from_entry.lua` because the file is not readable just do `path = vim.fn.fnameescape(path)` before that if validate :)

elianiva:
aha, thanks! That fixed it.

Conni2461 (approved):
LGTM
Thanks :)

Feel free to merge

PR #455, opened 2021-01-23
allow a callback to be called on on_lines update (for filtering tags from query text)

sunjon:


tjdevries:
I don't understand the filter marker?

Can't you just pass a function and manage the filter marker thing as the caller of this? Can you show me how you're using this?

Conni2461:
Here is the related PR https://github.com/nvim-telescope/telescope-frecency.nvim/pull/3

sunjon:
@tjdevries The idea is that if the optional callback is present, it's called on text entry (on_lines) with the current query text sent as a parameter.

That function has the choice/ability to filter out part(s) of the query text and return the remainder.

If the callback decides that whatever it did with the removed text (matched a filter) requires the results to be regenerated, it can do that and return them via a new finder.

This allows the following: 


[Frecency Workspace Filters](https://streamable.com/n85bjd)

Every time a valid 'tag' is matched there, completely new results are being generated.

tjdevries:
I don't like `filter_marker`.

Can't you just make `filter_marker` be part of the callback?

```lua
function my_thing(input, filter)
end

function passed_to_telescope(input)
  return my_thing(input, ':')
end
```

I do not want to add too many things to `picker`s if we can help it. I'm trying to move in the opposite direction actually.

lua/telescope/pickers.lua
```diff
@@ -446,6 +448,16 @@ function Picker:find()
...
 
     local prompt = self:_get_prompt()
+
+    local updated_finder
+    if self._on_input_filter_cb then
+      prompt, updated_finder = self._on_input_filter_cb(prompt, self._filter_marker)
+    end
+
+    if updated_finder then
+      self.finder = updated_finder
```
tjdevries, line 458: I think if we update to a new finder, we should make sure we `close` the old one or something. I can't remember if finders already have a hook for closing them down or not, but we should formalize that if we're going to randomly abandon them at times.

sunjon:
> I don't like `filter_marker`.
> 
> Can't you just make `filter_marker` be part of the callback?
> 
> ```lua
> function my_thing(input, filter)
> end
> 
> function passed_to_telescope(input)
>   return my_thing(input, ':')
> end
> ```
> 
> I do not want to add too many things to `picker`s if we can help it. I'm trying to move in the opposite direction actually.

It wasn't actually being used. I've removed it.

lua/telescope/pickers.lua
sunjon: I don't see any hook present at the moment. What would this involve?
The finder object consists of the results table and its metatable. Is it enough to set the finder/results tables to nil? Out of curiousity - would this have any practical benefit over abandoning them GC wise?

tjdevries: Well some finders start jobs -- they should have the jobs get stopped before we start a new one. Things like that.

We may want to add that before we do this (or I can potentially do it after)

sunjon: Doesn't almost everything other than grep use a `StaticFinder` that doesn't run jobs? It seemed that way when I looked at putting in some job cleanup code.

As it doesn't seem to apply to this context, I'd rather merge sooner rather than later. If the above statement is correct, any issues should only appear in modification of the live-grep(`JobFinder`) finder.  

That way I could leave the jobs cleanup changes for you to do, as you're a lot more familiar with it.

PR #456, opened 2021-01-23
fix: escape filename for previewer and action

elianiva:


Conni2461:
Also we might wanna do something else as well to fix another issue 😆  If you could do that as well :) thanks
```lua
    filepath = vim.fn.expand(filepath)
    local stat = vim.loop.fs_stat(filepath) or {}
    if stat.type == 'directory' then
```

Also does both work again. Opening these files and previewing it?

elianiva:
yes, both opening and previewing works

Conni2461 (approved):
LGTM
Lets hope it doesn't break anything. I am too lazy to test it right now :)
Feel free to merge

PR #460, opened 2021-01-24
Allow previewer window customization

pwntester:


Conni2461 (approved):
Good catch  :) thanks

Issue #403, opened 2021-01-07
[suggestion] use vim movements in results/preview window


hoob3rt:
I don't know if this has been discussed but I think it would be nice to use vim movements in results window. This "normal" mode is really limited...

Say you live grep something and see your result on top of the result window, now you have to hold `k` and wait instead of pressing `gg` or `H`. Range operators would also be nice e.g. `10k`.

This is a problem in all fuzzy finders for vim I know and it doesn't feel vim-like but since telescope is so close to neovim core I think it would be a nice feature to have. 

As of `<C-u>` and `<C-d>` I think scrolling in preview window is more of a gimmick than a feature but I understand it could be useful for someone. Imo there should be an option to switch between results and preview window using `<C-w>hl>` and use all normal mode movements in both.

Let me know what you think :)

Conni2461:
That are some really good suggestions.
1. `G` or `gg` should be super easy to realize. We only have to do a `shift_current_selection` with the amount of lines currently displayed or 0. After that mapping the function is trivial
2. The second suggestion(10k) should be relative easy to do as well and should be a really good addition. I just need to figure out if we can get `v:count`. Have not tried it in lua yet.
3. This one is tricky(previewer). Switching window focus is super easy. Don't get me wrong. We are doing it all the time for example `help_tags` to get to the correct position we are doing a `vim.fn.search`. So let me explain the tricky part.
We have two base_previewers. One based on `termopen` which will start `cat` or `bat`. With this one we are basically pretty limited. If `less` is installed we have a pager so you should be able to do searching and jumping not that interesting. 

With the experimental `buffer_previewer` we can do all most of stuff. Everything that we can do in a real buffer besides actually edit the file. And thats the point that will be hard for people to understand. This file we are looking at is just the content of the file and the buffer has actually no relation with any file. Why? Because if we do something like `bufload` which loads the file into the buffer we will end up running all autocmds. Synchronously. Which means everything blocks when moving though your results. So our solution right now is reading the file async with `uv.fs_open` ... and attaching the highlighter to the buffer. So to make the file editable we have to actually overwrite the whole file on save by ourselves and i am not really happy with that.

But we actually discussed making buffer_previewer editable and doing some other cool stuff. @sunjon suggested this in telescope gitter on December 29 2020:

> I've wanted to know what the fascination with org-mode is for a while. I've not used emacs, and org-mode get mentioned far more as a "thing", rather than "it's great because x,y and z"
[...]

PR #463, opened 2021-01-24
Advanced normal mode

### LINKED ISSUE #466: feat: Normal mode: Walking multiple steps at once (ie 5j or 5k)
I'm always frustrated when I need to type single j and k to walk inside the editor, so I always do 5j,5k etc for navigation.
However, in normal mode I can not use 5j or 5k to walk 5 steps down and up.

I would like to have the same walking behavior like (neo)vim, wher you can use 5j and 5k (after changing to normal mode) for walking 5 steps down and up.

I think the default behavior of neovim should be represented, if feasible, to reduce learning or configuration need.

Conni2461:
Close #466

Ref #403
Works only for the current screen because we can not yet scroll through all results.
Sorry it take me that long

- [x] `<count>j` and `<count>k`
- [x] G and gg (doing that tomorrow)

CC @hoob3rt

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -5,6 +5,7 @@ local a = vim.api
 local log = require('telescope.log')
 local state = require('telescope.state')
 local utils = require('telescope.utils')
+local p_scroller = require('telescope.pickers.scroller')
 
 local action_state = require('telescope.actions.state')
 local action_set = require('telescope.actions.set')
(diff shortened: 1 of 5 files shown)


Conni2461:
Okay that should be good now. I was thinking maybe we could use the `scroller` for `top` and `bottom` but i need to look at that one first. Also maybe `H` and `L` are better mappings rather than `G` and `gg`?! Maybe we can add `G` and `gg` later if we have scrolling?!

Also @tjdevries should we merge that before https://github.com/nvim-telescope/telescope.nvim/pull/472 or after. If before i take a look at the `scroller` much sooner. Also depends on what the people demand 🤣

Conni2461:
Its mapped to `H` and `L` now. It would be done. @tami5 probably does some testing first tho

tami5:
Works very well, is there way to map it in normal mode, maybe \<c-s-l\>/\<c-s-h\>. Also `G` and `g` maybe be more generic and `<c-g>/<c-G>`

Conni2461:
Sure you can map it in insert mode:
```lua
mappings = {
  i = {
    ["<c-h>"] = actions.move_to_top,
    ["<c-l>"] = actions.move_to_bottom,
  },
},
```
But you can't map `<c-g>` and `<c-G>` because ctrl and ctrl_shift is the same.

Will add some tests then ask tj

PR #465, opened 2021-01-25
add actions.select

### LINKED ISSUE #461: Have an actions.select
Have an `actions.select` to replace `_goto_file_selection` as default actions so `actions.select:replace` is more intuitive. 

CC @anott03

anott03:
Close #461

CC @Conni2461

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -144,8 +144,12 @@ function actions.center(_)
   vim.cmd(':normal! zz')
 end
 
-function actions.goto_file_selection_edit(prompt_bufnr)
+function actions.select(prompt_bufnr)
   actions._goto_file_selection(prompt_bufnr, "edit")
+end
+
+function actions.goto_file_selection_edit(prompt_bufnr)
+  actions.select(prompt_bufnr)
 end
 
 function actions.goto_file_selection_split(prompt_bufnr)
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -66,7 +66,7 @@ function actions.preview_scrolling_down(prompt_bufnr)
 end
 
 -- TODO: It seems sometimes we get bad styling.
-function actions._goto_file_selection(prompt_bufnr, command)
+function actions._select(prompt_bufnr, command)
   local entry = actions.get_selected_entry(prompt_bufnr)
 
   if not entry then
@@ -145,22 +145,23 @@ function actions.center(_)
 end
 
 function actions.select(prompt_bufnr)
-  actions._goto_file_selection(prompt_bufnr, "edit")
+  actions._select(prompt_bufnr, "edit")
 end
 
 function actions.hselect(prompt_bufnr)
-  actions._goto_file_selection(prompt_bufnr, "new")
+  actions._select(prompt_bufnr, "new")
 end
 
 function actions.vselect(prompt_bufnr)
-  actions._goto_file_selection(prompt_bufnr, "vnew")
+  actions._select(prompt_bufnr, "vnew")
 end
 
 function actions.tabselect(prompt_bufnr)
-  actions._goto_file_selection(prompt_bufnr, "tabedit")
+  actions._select(prompt_bufnr, "tabedit")
 end
 
 -- aliases
+actions._goto_file_selection = actions._select
 actions.goto_file_selection_edit = actions.select
 actions.goto_file_selection_split = actions.hselect
 actions.goto_file_selection_vsplit = actions.vselect
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -148,21 +148,23 @@ function actions.select(prompt_bufnr)
   actions._goto_file_selection(prompt_bufnr, "edit")
 end
 
-function actions.goto_file_selection_edit(prompt_bufnr)
-  actions.select(prompt_bufnr)
-end
-
-function actions.goto_file_selection_split(prompt_bufnr)
+function actions.hselect(prompt_bufnr)
   actions._goto_file_selection(prompt_bufnr, "new")
 end
 
-function actions.goto_file_selection_vsplit(prompt_bufnr)
+function actions.vselect(prompt_bufnr)
   actions._goto_file_selection(prompt_bufnr, "vnew")
 end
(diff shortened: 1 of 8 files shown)


lua/telescope/actions/init.lua
```diff
@@ -144,10 +144,14 @@ function actions.center(_)
...
 end
 
-function actions.goto_file_selection_edit(prompt_bufnr)
+function actions.select(prompt_bufnr)
   actions._goto_file_selection(prompt_bufnr, "edit")
 end
 
+function actions.goto_file_selection_edit(prompt_bufnr)
+  actions.select(prompt_bufnr)
+end
```
Conni2461, line 153: actions.goto_file_selection_edit = actions.select

Conni2461:
functions actionsCI failure unrelated :)

Maybe we should just rename all and provide alias for the old methods 🤔 Like @tami5 said :)

anott03:
Sure! I'll work on that today :)

lua/telescope/actions/init.lua
anott03: That makes sense. I'll add that change along with aliases for all the other functions.

lua/telescope/mappings.lua
```diff
@@ -30,7 +30,7 @@ mappings.default_mappings = config.values.default_mappings or {
 
     n = {
       ["<esc>"] = actions.close,
-      ["<CR>"] = actions.goto_file_selection_edit + actions.center,
+      ["<CR>"] = actions.select + actions.center,
       ["<C-x>"] = actions.goto_file_selection_split,
```
tami5, line 34: How about those ? will they stay the same?

anott03: If I understand what you're asking, I think the idea is that everything internally changes to the new naming scheme (split, vsplit, hsplit, etc.) and we leave aliases with the old names so that we don't break compatibility.

Conni2461:
We probably also want a `_select` which points to `_goto_file_selection` so we never have to write that function name every again 😆 

Thanks for your work. :)

anott03:
+

> We probably also want a `_select` which points to `_goto_file_selection` so we never have to write that function name every again laughing
> 
> Thanks for your work. :)

All done! :)

Conni2461 (approved):
LGTM. I try it later then i merge
Thanks :)

Conni2461:
Works for me. Did some quick testing :)
Thanks :D

Conni2461:
And we broke backwards compatibility 🤣  😭 😭 😭 
I should step down as maintainer.

Can no longer replace `_goto_file_selection`

anott03:
Should I go back and remove that change? It's easy to do.

Also, why did that break it? Does it have to do with when users/extensions override it?

Conni2461:
I think: `actions._goto_file_selection = actions._select` as i thought maybe `function actions._goto_file_selection(...) actions._select(...) end` is the correct version.

Could you try it out and if so open a new PR. I gotta sleep now
Thanks :)

anott03:
I've made some changes, this is how it looks now.
```lua
function actions._goto_file_selection(...)
...
end
...
actions._select = actions._goto_file_selection
```
I believe this is what you were describing (swapping the alias and the function definition). I'm going to leave it here for tonight and continue tomorrow :)

Conni2461:
I think the aliases don't work. Even doing something like this:
```lua
function actions._goto_file_selection(prompt_bufnr, cmd)
  actions._select(prompt_bufnr, cmd)
end
```
doesn't let me replace it. My replace function just never gets called 😕 Is this just on my end? Can someone confirm that this is an issue? Like do `_goto_file_selection` in `help_tags` rather then `_select` and print like `hello` inside the function. That never gets printed if i confirm my entry.

anott03:
I'm getting the same issue. I went into help_tags, swapped out  `_select` with `_goto_file_selection` and added a print statement like you said. When I tried to open a help tag, nothing was printed, and the help didn't open in a split.

anott03:
I've reverted `_select` to `_goto_file_selection` on my fork. Is there any concern that the other aliases will cause similar issues? If not I can open a new PR.

Conni2461:
Oh yeah would be great if you could provide a fix, so new interface and old one works :)

anott03:
I've been doing some more testing, and it seems that `:replace` failed for all of the aliases (select, vselect, etc.). The solution to this is going to be a bit more involved...

Conni2461:
I just figured out what the issue is 😆

Conni2461:
Seriously why didn't i see that earlier. We can only override mappings that are also bound to keys.
So the thing why our replace isn't called is because the `goto_file_selection_edit` is not mapped only `select`.
So executing so doing `<CR>` would mean we run `actions.select` or the replaced function.

No idea how to fix that

anott03:
I think the best thing to do now is to revert back to entirely so that comparability is good for now (I believe these changes are still on master right now). I'll open a PR doing that. After that maybe on a fork or something we work on determining the right way to do this.

anott03:
I've opened the PR. Here' my messy idea for a solution that might work. We take `mt.replace`, which currently looks like this
```lua
  function mt:replace(v)
    assert(#self == 1, "Cannot replace an already combined action")

    local action_name = self[1]
    mt._replacements[action_name] = v
  end
```
and edit a second entry in `mt._replacements` so that it looks like this:
```lua
  function mt:replace(v)
    assert(#self == 1, "Cannot replace an already combined action")

    local action_name = self[1]
    local alias_name = actions.aliases[action_name] -- we would need to define this table
    mt._replacements[action_name] = v
    mt._replacements[alias_name] = v
  end
```
Any thoughts?

tjdevries:
Sorry, can someone re-summarize the problem? I can try and take a look sometime this week for sure.

anott03:
@tjdevries 
- What we're trying to do: pretty clear, switch to a better naming scheme for actions
- What we did: rename all the functions to the new names and then create "aliases" (to preserve backwards compatibility) that look something like this
    - `actions.goto_file_selection_edit = actions.select` (goto... is the old name, select is the new name)
- the issue we encountered: when we try to use `:replace` on one of the aliases, it doesn't work, because as @Conni2461 said above, `:replace` is only works on the functions that are mapped
- a possible solution (maybe?): what I describe above

PR #469, opened 2021-01-26
Undo #465

anott03:
Undo #465 because of breaking backwards compatibility.
CC @Conni2461

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ require('telescope').setup{
         -- So, to not map "<C-n>", just put
         ["<c-x>"] = false,
         -- Otherwise, just set the mapping to the function that you want it to be.
-        ["<C-i>"] = actions.goto_file_selection_split,
+        ["<C-i>"] = actions.hselect,
         -- Add up multiple actions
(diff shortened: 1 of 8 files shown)


tami5:
:|

Conni2461:
@tami5 I also kinda don't want to revert it but i think we broke all extensions that use some form of `goto_file_selection:replace` and even more.

Described here: https://github.com/nvim-telescope/telescope.nvim/pull/465#issuecomment-767831897
Possible solution: https://github.com/nvim-telescope/telescope.nvim/pull/465#issuecomment-767808213

I would love a solution for that asap but if we don't figure out one today i probably will revert. Thoughts? Or you decide tami.

I messed it up this time :| i should have figured it out before merging 😭

anott03:
@Conni2461 I tested the solution I mentioned, no luck with it yet. The idea may be there and I just implemented something incorrectly if you want to take a look I made a branch on my fork for it: [anott03/telescope.nvim](https://github.com/anott03/telescope.nvim/tree/renameActions).

PR #472, opened 2021-01-28
feat: Action improvements

tjdevries:
We now provide `actions.select_*` as opposed to `actions.goto_file_*` because not everything is a file, but people were using those as the selection thing because that's what we defaulted the map to do.

So, I have deleted (and given an error message on how to migrate) for anyone overriding `actions._goto_file_selection` and introduced `action_set`.

```lua
-- Override just edit requests
require('telescope.actions.set').edit:replace(new_edit_function)

-- Override all selections
require('telescope.actions.set').select:replace(new_edit_function)
```

Related: #465 

Also note that I have moved some actions that weren't really actions OUT of `actions/init.lua` because it was confusing to have them there.

See new docs for more info as well.


tami5:
It would be nice if we have more abstraction.

```
require('telescope.actions.set').edit:replace(new_edit_function)
```

```
pickers.new(opts, {
   mappings = {
     edit = function(..) end
   }
})
```

README.md
```diff
@@ -298,13 +298,14 @@ its `attach_mappings` key to a function, like this
 
 ```lua
+local action_set = require('telescope.actions.set')
 -- Picker specific remapping
 ------------------------------
 require('telescope.builtin').fd({ -- or new custom picker's attach_mappings field:
     -- This will replace select no mather on which key it is mapped by default
     actions.goto_file_selection_edit:replace(function()
```
Conni2461, line 307: We have to make sure that we renamed all forms of `goto_file_selection`

developers.md
```diff
@@ -0,0 +1,52 @@
...
+    - These are most useful for writing your own actions and interacting with telescope at that time
+- `lua/telescope/actions/mt.lua`
+    - You probably don't need to look at this, but it defines the behavior of actions.
+
+### `:replace(function)`
+
+Directly override an action with a new function
+
+```lua
+actions.goto_file_selection_edit:replace(git_checkout_function)
```
Conni2461, line 26: 😆 Typo. There is no `goto_file_selection` anymore :)

lua/telescope/builtin/internal.lua
```diff
@@ -83,7 +85,7 @@ internal.planets = function(opts)
     sorter = conf.generic_sorter(opts),
     attach_mappings = function(prompt_bufnr)
       actions.goto_file_selection_edit:replace(function()
```
Conni2461, line 87: Here are more `goto_file_selection_edit`

Conni2461:
I was thinking something else.
E.g. `'telescope.builtin'.git_branches` we have a `attach_mappings` set to do a checkout on `edit`.

How about users can pass in even more `attach_mappings` like that:
```lua
require'telescope.builtin'.git_branches({ attach_mappings = function(prompt_bufnr, map)
  map(mode, key, function()
    -- do something else like rebase
  end)
  return true
end})
```

Right now we would override the `attach_mappings` set by builtin because we do this:

```lua
pickers.new = function(opts, defaults)
  return Picker:new(extend(opts, defaults))
end
```

tjdevries:
Oh, that's a good point on the overriding the mappings thing.... hmmm. I'll have to think a bit more.

Conni2461:
https://github.com/nvim-telescope/telescope.nvim/issues/480 what he said :)

tjdevries:
@Conni2461 What do you think now?

I still left in the actions themselves, since it makes that you might want to be able to select something, but edit a file in a different way.

Also, the default behavior is now backwards compatible I'm pretty sure, if you check the `action_set` tests in `action_spec.lua`

tjdevries:
> It would be nice if we have more abstraction.
> 
> ```
> require('telescope.actions.set').edit:replace(new_edit_function)
> ```
> 
> ```
> pickers.new(opts, {
>    mappings = {
>      edit = function(..) end
>    }
> })
> ```

Let's do this at a different time (or maybe we won't be able to do it just based on how we're setting up the configuration now. I suppose it could be possible to make it so you can do only `mappings` or `attach_mappings` to make it quite simple on how you would use it.

tami5:
works well no issues LGTM

elianiva:
I just tried this out, works great! I don't have any issues at all.

Conni2461:
Important information post :)

@pwntester(octo), @cjjlaidler(project), @windwp(github, media-files) @Conni2461(dap, snippets), @sunjon(frecency, packer), @delphinus(node-modules, z, ghq), @noahares(bibtex), @tami5(cheat) and @sharksforarms(vimspector)

This PR will have breaking changes for `actions.goto_file_selection_:replace`. You need to update your extension accordingly and use:
- actions.select_default
- actions.select_horizonal
- actions.select_vertical
- actions.select_tab
- actions.set.edit   // for replacing all select functions with keys: `default`, `horizontal`, `vertical`, `tab`

There is a function `actions.state.select_key_to_edit_key(key)` that maps these keys to `edit`, `new`, `vnew`, `tabedit`. (The function name and module position may vary in the end :)

We also moved `actions` that weren't really actions into `actions.state` like `get_selected_entry`, `get_current_line` and `get_current_picker`.

All informations can be found in developers.md afterwards. :)

cjjlaidler:
Hi @Conni2461 any idea when you plan to merge this one, so we can get a rough idea when the plugins need updating? Thanks

tjdevries:
@Conni2461 let's merge today?

Conni2461:
@tjdevries sounds good to me

tjdevries:
cc @ThePrimeagen

tjdevries:
@Conni2461 I will click the button in 25 minutes

tjdevries:
Here goes nothin! :)

Issue #394, opened 2021-01-05
Support symlinks


sharksforarms:
### Description
Trying to use builtin find_files on a directory which is a symlink

**Expected Behavior**
Should show contents of path symlink points to

**Actual Behavior**
Telescope pops up, shows no files, no errors are shown

### Details

```lua
function M.nvim_config(opts)
  local opts = require('telescope.themes').get_dropdown{
    -- cwd = "~/.config/nvim" is a symlink to "~/source/dotfiles/nvim",
    cwd = vim.fn.stdpath("config"),
    file_ignore_patterns = { "plugged/" },
  }

  require('telescope.builtin').find_files(opts)
end
```

Conni2461:
right now you should be able to do `cwd = vim.loop.fs_realpath('~/.config/nvim')` but we should check if something like this is in `plenary.path` and fix it when doing #387

Conni2461:
@sharksforarms Does this still happen? I am trying to reproduce this for @anott03 but i can't :laughing: everything works as expected.

sharksforarms:
The issue is about following symlinks inside folders.

```
mkdir /tmp/test
mkdir /tmp/other
touch /tmp/test/file1
touch /tmp/other/file2
ln -s /tmp/other /tmp/test/other
```

```rust
  local opts = require('telescope.themes').get_dropdown{
    cwd = "/tmp/test/",
  }

  require('telescope.builtin').find_files(opts)
```

Only `file1` will be shown. The symlink to `/tmp/other` is not followed.

Conni2461:
Oh that is completely different then your original post. Is it not?

> By default, fd does not descend into symlinked directories. Using this flag, symbolic links are also traversed.

Just add `-L` to the options. So do a new find_files command. We could do a opt for that as well and add it to the other commands. But i don't know if `rg` supports it.

sharksforarms:
Ahh thanks! That makes sense. I actually made the issue as I had symlinked folders in my nvim config, but then ended up just making nvim the top level directory and didn't have this issue anymore.

Conni2461:
I still think we can do an opt here :) It seems to be `-L` for all commands.

DavidePatria:
could you please share how this option should be added to telescope config?

dgrabla:
@DavidePatria Did you find out how to add the option?

dgrabla:
@DavidePatria Got it

[...]

PR #473, opened 2021-01-29
move from telescope/path to plenary/path

### LINKED ISSUE #387: Move from `telescope/path` to `plenary/path`
I think everything from `telescope.path` can already be found in `plenary.path` so we should use that module instead and remove `telescope.path`

anott03:
Closes #387

### DIFF
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -13,7 +13,7 @@
 local a = vim.api
 
 local log = require('telescope.log')
-local path = require('telescope.path')
+local path = require('plenary.path')
 local state = require('telescope.state')
 
 local action_state = require('telescope.actions.state')
@@ -130,7 +130,7 @@ action_set.edit = function(prompt_bufnr, command)
     -- check if we didn't pick a different buffer
     -- prevents restarting lsp server
     if vim.api.nvim_buf_get_name(0) ~= filename or command ~= "edit" then
-      filename = path.normalize(vim.fn.fnameescape(filename), vim.loop.cwd())
+      filename = path:new(vim.fn.fnameescape(filename)):normalize(vim.loop.cwd())
       vim.cmd(string.format("%s %s", command, filename))
     end
   end
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -13,7 +13,7 @@
 local a = vim.api
 
 local log = require('telescope.log')
-local path = require('plenary.path')
+local Path = require('plenary.path')
 local state = require('telescope.state')
 
 local action_state = require('telescope.actions.state')
@@ -130,7 +130,7 @@ action_set.edit = function(prompt_bufnr, command)
     -- check if we didn't pick a different buffer
     -- prevents restarting lsp server
     if vim.api.nvim_buf_get_name(0) ~= filename or command ~= "edit" then
-      filename = path:new(vim.fn.fnameescape(filename)):normalize(vim.loop.cwd())
+      filename = Path:new(vim.fn.fnameescape(filename)):normalize(vim.loop.cwd())
       vim.cmd(string.format("%s %s", command, filename))
     end
   end
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -46,10 +46,9 @@ files.live_grep = function(opts)
     end, vim.api.nvim_list_bufs())
     if not next(bufnrs) then return end
 
-    local tele_path = require'telescope.path'
     for _, bufnr in ipairs(bufnrs) do
       local file = vim.api.nvim_buf_get_name(bufnr)
-      table.insert(filelist, tele_path.make_relative(file, opts.cwd))
+      table.insert(filelist, Path:new(file):make_relative(opts.cwd))
     end
   elseif search_dirs then
     for i, path in ipairs(search_dirs) do
@@ -249,11 +248,10 @@ files.file_browser = function(opts)
     return finders.new_table {
       results = data,
       entry_maker = (function()
-        local tele_path = require'telescope.path'
         local gen = make_entry.gen_from_file(opts)
         return function(entry)
           local tmp = gen(entry)
-          tmp.ordinal = tele_path.make_relative(entry, opts.cwd)
+          tmp.ordinal = Path:new(entry):make_relative(opts.cwd)
           return tmp
         end
       end)()
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -284,7 +284,7 @@ files.file_browser = function(opts)
           return
         end
 
-        local fpath = current_picker.cwd .. os_sep .. file
+        local fpath = Path:new({current_picker.cwd, file})
         if string.sub(fpath, -1) ~= os_sep then
           actions.close(prompt_bufnr)
           Path:new(fpath):touch({ parents = true })
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -229,6 +229,10 @@ end
 files.file_browser = function(opts)
   opts = opts or {}
 
+  local is_dir = function(value)
+    return value:sub(-1, -1) == os_sep
+  end
+
   opts.depth = opts.depth or 1
   opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or vim.loop.cwd()
   opts.new_finder = opts.new_finder or function(path)
@@ -245,17 +249,53 @@ files.file_browser = function(opts)
     })
     table.insert(data, 1, '..' .. os_sep)
 
-    return finders.new_table {
-      results = data,
-      entry_maker = (function()
-        local gen = make_entry.gen_from_file(opts)
-        return function(entry)
-          local tmp = gen(entry)
-          tmp.ordinal = Path:new(entry):make_relative(opts.cwd)
-          return tmp
+    local maker = function()
+      local mt = {}
+      mt.cwd = opts.cwd
+      mt.display = function(entry)
+        local hl_group
+        local display = utils.transform_path(opts, entry.value)
+        if is_dir(entry.value) then
+          display = display .. os_sep
+          if not opts.disable_devicons then
+            display = (opts.dir_icon or "") .. " " .. display
+            hl_group = "Default"
+          end
+        else
+          display, hl_group = utils.transform_devicons(entry.value, display, opts.disable_devicons)
         end
-      end)()
-    }
+
+        if hl_group then
+          return display, { { {1, 3}, hl_group } }
+        else
+          return display
+        end
+      end
+
+      mt.__index = function(t, k)
+        local raw = rawget(mt, k)
+        if raw then return raw end
+
+        if k == "path" then
+          local retpath = Path:new({t.cwd, t.value}):absolute()
+          if not vim.loop.fs_access(retpath, "R", nil) then
+            retpath = t.value
+          end
+          if is_dir(t.value) then retpath = retpath .. os_sep end
+          return retpath
+        end
+
+        return rawget(t, rawget({ value = 1 }, k))
+      end
+
+      return function(line)
+        local tbl = {line}
+        tbl.ordinal = Path:new(line):make_relative(opts.cwd)
+        return setmetatable(tbl, mt)
+      end
+    end
+
+    return finders.new_table { results = data, entry_maker = maker() }
   end
 
   pickers.new(opts, {
@@ -265,7 +305,7 @@ files.file_browser = function(opts)
     sorter = conf.file_sorter(opts),
     attach_mappings = function(prompt_bufnr, map)
       action_set.select:replace_if(function()
-        return action_state.get_selected_entry().path:sub(-1) == os_sep
+        return is_dir(action_state.get_selected_entry().path)
       end, function()
         local new_cwd = vim.fn.expand(action_state.get_selected_entry().path:sub(1, -2))
         local current_picker = action_state.get_current_picker(prompt_bufnr)
@@ -284,8 +324,8 @@ files.file_browser = function(opts)
           return
         end
 
-        local fpath = Path:new({current_picker.cwd, file})
-        if string.sub(fpath, -1) ~= os_sep then
+        local fpath = current_picker.cwd .. os_sep .. file
+        if not is_dir(fpath) then
           actions.close(prompt_bufnr)
           Path:new(fpath):touch({ parents = true })
           vim.cmd(string.format(':e %s', fpath))
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -3,7 +3,7 @@ local action_set = require('telescope.actions.set')
 local action_state = require('telescope.actions.state')
 local finders = require('telescope.finders')
 local make_entry = require('telescope.make_entry')
-local path = require('telescope.path')
+local Path = require('plenary.path')
 local pickers = require('telescope.pickers')
 local previewers = require('telescope.previewers')
 local sorters = require('telescope.sorters')
@@ -135,7 +135,7 @@ internal.symbols = function(opts)
 
   local results = {}
   for _, source in ipairs(sources) do
-    local data = vim.fn.json_decode(path.read_file(source))
+    local data = vim.fn.json_decode(Path:new(source):read())
     for _, entry in ipairs(data) do
       table.insert(results, entry)
     end
@@ -346,8 +346,8 @@ end
 
 internal.vim_options = function(opts)
   -- Load vim options.
-  local vim_opts = loadfile(utils.data_directory() .. path.separator .. 'options' ..
-                            path.separator .. 'options.lua')().options
+  local vim_opts = loadfile(utils.data_directory() .. Path.path.sep .. 'options' ..
+                            Path.path.sep .. 'options.lua')().options
 
   pickers.new(opts, {
     prompt = 'options',
@@ -449,7 +449,7 @@ internal.help_tags = function(opts)
   local delimiter = string.char(9)
   for _, lang in ipairs(langs) do
     for _, file in ipairs(tag_files[lang] or {}) do
-      local lines = vim.split(path.read_file(file), '\n', true)
+      local lines = vim.split(Path:new(file):read(), '\n', true)
       for _, line in ipairs(lines) do
         -- TODO: also ignore tagComment starting with ';'
         if not line:match'^!_TAG_' then
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -346,8 +346,9 @@ end
 
 internal.vim_options = function(opts)
   -- Load vim options.
-  local vim_opts = loadfile(utils.data_directory() .. Path.path.sep .. 'options' ..
-                            Path.path.sep .. 'options.lua')().options
+  local vim_opts = loadfile(
+    Path:new({utils.data_directory(), 'options', 'options.lua'}):absolute()
+  )().options
 
   pickers.new(opts, {
     prompt = 'options',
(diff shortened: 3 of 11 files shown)


anott03:
Some of the changes in the diffs don't make sense... it's showing a bunch of changes regarding the names actions that shouldn't be there. When comparing my code to what I see on master by hand it seems to match. Can someone take a look?

README.md
```diff
@@ -278,11 +278,11 @@ require('telescope').setup{
         -- So, to not map "<C-n>", just put
         ["<c-x>"] = false,
         -- Otherwise, just set the mapping to the function that you want it to be.
-        ["<C-i>"] = actions.hselect,
+        ["<C-i>"] = actions.goto_file_selection_split,
```
anott03, line 281: Example of what I mentioned in my comment. This isn't a change I'm making (as far as I can tell)... this is already what's on master.

Conni2461:
You probably based this branch on your last branch rather than on master. You could try to rebase it on current master. Not sure if it will work tho 😆 

Note: I will look at your plenary work first. I probably won't look at this PR today :)

anott03:
@Conni2461 I'll try rebasing - thanks :)

> Note: I will look at your plenary work first. I probably won't look at this PR today :)

Of course. We all have things we need to do :)

anott03:
This needs to be worked on some more after changes being made to plenary

lua/telescope/actions/init.lua
```diff
@@ -120,7 +120,7 @@ function actions._goto_file_selection(prompt_bufnr, command)
         vim.cmd(string.format(":tab sb %d", entry_bufnr))
       end
     else
-      filename = path:new(vim.fn.fnameescape(filename)):normalize()
+      filename = path:new(vim.fn.fnameescape(filename)):normalize(vim.fn.getcwd())
```
Conni2461, line 123: Can we check if we still need `vim.fn.fnameescape` when doing path:new? @elianiva 

Also I kinda wanna move all `vim.fn.getcwd()` to `vim.loop.cwd()` to reduce our use of vimL

Conni2461:
@anott03 Could you check if this one will be fixed when using `path:new` https://github.com/nvim-telescope/telescope.nvim/issues/394

Also did you have any luck with rebasing?

lua/telescope/actions/init.lua
elianiva: was about to try this but I got this error.
```
...packer/start/telescope.nvim/lua/telescope/make_entry.lua:122: attempt to call method 'make_relative' (a nil value)
```

Conni2461: Oh yeah requires a plenary PR : https://github.com/nvim-lua/plenary.nvim/pull/60

I am not sure if its done already 😆 but thanks for testing

elianiva: Ah, so that's that I was missing :laughing:
Unfortunately we still need to escape it. You think it's a good idea to escape it ourself instead of relying on `fnameescape`?
My issue isn't that common and the only thing that's causing it is because `$` conflict with bash/zsh variable naming.

Conni2461: Could you try this path for plenary on top of that PR:
```diff
diff --git a/lua/plenary/path.lua b/lua/plenary/path.lua
index 33372f9..c42c97e 100644
--- a/lua/plenary/path.lua
+++ b/lua/plenary/path.lua
@@ -200,14 +200,15 @@ end
 
 function Path:make_relative(cwd)
   cwd = F.if_nil(cwd, self._cwd, cwd)
-  if self.filename:sub(1, #cwd) == cwd  then
+  local filename = self:absolute() or self.filename
+  if filename:sub(1, #cwd) == cwd  then
     local offset =  0
     -- if  cwd does ends in the os separator, we need to take it off
     if cwd:sub(#cwd, #cwd) ~= path.separator then
       offset = 1
     end
 
-    self.filename = self.filename:sub(#cwd + 1 + offset, #self.filename)
+    self.filename = filename:sub(#cwd + 1 + offset, #filename)
   end
 
   return self.filename
@@ -217,7 +218,8 @@ function Path:normalize(cwd)
   cwd = F.if_nil(cwd, self._cwd, cwd)
   self:make_relative(cwd)
   -- Substitute home directory w/ "~"
-  self.filename = self.filename:gsub("^" .. path.home, '~', 1)
+  local filename = self:absolute() or self.filename
+  self.filename = filename:gsub("^" .. path.home, '~', 1)
   -- Remove double path seps, it's annoying
   self.filename = self.filename:gsub(path.sep .. path.sep, path.sep)
   print(self.filename)
@@ -247,7 +249,7 @@ local shorten = (function()
 end)()
 
 function Path:shorten()
-  return shorten(self.filename)
+  return shorten(self:absolute() or self.filename)
 end
 
 function Path:mkdir(opts)
```

elianiva: Ok it's sorta works. I can open the file but it's not a valid file I think. Its content is different from what I have and I can't the file.

Conni2461: Hmm okay. I think `Path:new` should handle the escaping better than vimL 🤣 . I will take a look at it later today. Shouldn't be to hard to create a file with `$` and do some testing 😆 Or maybe @anott03 wants to look at it :)

Still thanks for testing :)

elianiva: no problem :)

anott03:
@Conni2461 sure I'll take a look at #394 today.

As for the rebasing, I took a step back from telescope and was trying to get everything done with plenary first (which I think I have, so I'll get back to this today or tomorrow).

anott03:
@Conni2461 I just rebased against master, and it seems to have worked (all tests pass, it seems to be running fine, etc.). As for #394, I tested it in the following way:
1. Checkout telescope master
2. `cd ~/.config/nvim` (which is a symlink to `~/repos/dotfiles/nvim`
3. run `:Telescope find_files`
I then repeated the above steps on my branch, `plenaryPath`. The results I yielded were the same for both branches, and that is that it successfully listed the files I had in that directory. I'm not sure if that means the issue has been solved, or if I need to take other steps to reproduce it. I'll also perform these tests on a symlink that doesn't point to a git repo and see if that makes a difference.

tjdevries:
(What do you need me to do :laughing: )

anott03:
@tjdevries Unless there are changes that need to be made to this code before it can be merged (stuff to resolve the failed linting test maybe?) I'd say there isn't much to do regarding this issue until the PR to plenary has been merged since this is dependent on that. It's probably worth it to see if someone can recreate #394. I've tried and haven't been able to.

anott03:
@Conni2461 I don't think we should wait on `Path:escape` before merging this... I think it makes more sense just to resolve the conflicts and merge this as is and then create a separate PR for escaping paths if necessary.

Conni2461:
Works for me. I am currently really busy (sorry exams and stuff) so i won't be around that much but i 100% trust @elianiva if he wants to finish it with you :)

anott03:
No worries - good luck with the exams! I'll talk to @elianiva here or on gitter when he's free :)

elianiva:
probably better talk it here, Gitter is a bit janky for me. It deletes DM from other people randomly for some reason 😆

tjdevries:
Do you need anything from me?

anott03:
@tjdevries I believe this is good to go - it just needs to be reviewed by someone. I had talked with Conni about waiting until we have a `Path:escape` in plenary but I think we should just merge this as is and then open a new PR if/when an escape function comes around.

anott03:
@elianiva Can you take a look at the second conflict in `init.lua`? That's the only one I'm not quite sure how to resolve yet...

elianiva:
remove [this part](https://github.com/nvim-telescope/telescope.nvim/blob/e94fdff12359d9dc3f93830c92f15d9fc8121e11/lua/telescope/actions/init.lua#L112-L130) from your branch and edit [this file](https://github.com/nvim-telescope/telescope.nvim/blob/6e941e0ecec1ab6a1b4ce40c693ef1272c505abb/lua/telescope/actions/set.lua) accordingly, specifically [this line](https://github.com/nvim-telescope/telescope.nvim/blob/d4bf1181ea6ef9aec675ff404d65098cbb91d9b1/lua/telescope/actions/set.lua#L4) and [this line](https://github.com/nvim-telescope/telescope.nvim/blob/d4bf1181ea6ef9aec675ff404d65098cbb91d9b1/lua/telescope/actions/set.lua#L111)

 We've moved the action that was previously in `actions/init.lua` to `actions/set.lua`. Hope it helps :)

anott03:
Makes sense - thanks :)

anott03:
I ended up resetting and doing everything over again because it turned out to be easier than rebasing against the latest changes (at least for me). In the processing of doing everything I realized that not only did `telescope.path` have some path related functions, but so did `utils`, so this now gets rid of any path functions in `utils` in favor of plenary as well.

@Conni2461 and @elianiva can one of you check this out?

lua/telescope/builtin/internal.lua
```diff
@@ -305,8 +305,8 @@ end
 
 internal.vim_options = function(opts)
   -- Load vim options.
-  local vim_opts = loadfile(utils.data_directory() .. path.separator .. 'options' ..
-                            path.separator .. 'options.lua')().options
```
Conni2461, line 309: Can we do `Path:new({data_dir, options, options.lua}):absolute()` to get the filepath (or similar)

lua/telescope/make_entry.lua
```diff
@@ -98,7 +98,7 @@ do
       if raw then return raw end
 
       if k == "path" then
-        local retpath = t.cwd .. path.separator .. t.value
```
Conni2461, line 101: Same here we shouldn't do concats manually anymore

Conni2461: Or wait
Maybe we should here
just for performance. This code happens a lot

lua/telescope/path.lua
```diff
@@ -1,89 +1,33 @@
...
-    -- if  cwd does ends in the os separator, we need to take it off
-    if cwd:sub(#cwd, #cwd) ~= path.separator then
-      offset = 1
-    end
-
-    filepath = filepath:sub(#cwd + 1 + offset, #filepath)
-  end
-
-  return filepath
+path.make_relative = function()
```
Conni2461, line 11: maybe we should just do. Similar to actions deprecated message
```lua
path.__index = function()
  print('deprecated message')
end
```

Conni2461: https://github.com/nvim-telescope/telescope.nvim/blob/2c4efc2f8af0575526f7b5dcc35fba9e0f1cc944/lua/telescope/actions/init.lua#L22-L35

Conni2461: Just can just always do a error. You don't need to do the string.finds

lua/telescope/utils.lua
```diff
@@ -229,7 +201,7 @@ function utils.data_directory()
   local sourced_file = require('plenary.debug_utils').sourced_filepath()
   local base_directory = vim.fn.fnamemodify(sourced_file, ":h:h:h")
 
-  return base_directory .. pathlib.separator .. 'data' .. pathlib.separator
```
Conni2461, line 232: We should not doing concats here

elianiva:
is there any reason why they're called `path` *or* `pathlib` ? I personally prefer `pathlib` to avoid `path.path.sep`, but I saw on some places, it's just called as `path` and some places called it as `pathlib`.

anott03:
I'm guessing that's just an artifact of different code being written at different times. I just left the name as whatever it was before, but I agree that `pathlib` is a bit nicer. I've also been considering have `plenary.path` return `sep` since `path.path.sep` seems redundant.

anott03:
@Conni2461 I've finally gotten back to work on this! I reset against master and everything seems to work with `plenary.path` now.

lua/telescope/builtin/internal.lua
```diff
@@ -330,8 +330,8 @@ end
...
+  local vim_opts = loadfile(utils.data_directory() .. Path.path.sep .. 'options' ..
+                            Path.path.sep .. 'options.lua')().options
```
Conni2461, line 334: I think i already commented this :rofl: but: can we do `Path:new({utils.data_directory(), 'options', 'options.lua'}):absolute()` or so

lua/telescope/actions/set.lua
```diff
@@ -13,7 +13,7 @@
 local a = vim.api
 
 local log = require('telescope.log')
-local path = require('telescope.path')
+local path = require('plenary.path')
```
Conni2461, line 16: Maybe we should make it more consistant. Sometimes `path` is written starting with a big letter and sometimes not.

If i have to choose i would take `Path`

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -78,7 +78,7 @@ previewers.file_maker = function(filepath, bufnr, opts)
             if opts.callback then opts.callback(bufnr) end
         end)})
       else
-        path.read_file_async(filepath, vim.schedule_wrap(function(data)
+        path:new(filepath):read(vim.schedule_wrap(function(data)
```
Conni2461, line 81: I actually prefer `_read_async` here i know its kinda cool that it detects it automatically but we can save the `if` check here

lua/telescope/utils.lua
```diff
@@ -1,12 +1,12 @@
 local has_devicons, devicons = pcall(require, 'nvim-web-devicons')
 
-local pathlib = require('telescope.path')
+local pathlib = require('plenary.path')
```
Conni2461, line 3: Yep def some consistency when it comes to naming

Conni2461:
Left some nitpicks

lua/telescope/actions/set.lua
anott03: I agree - I had just tried to leave things as they were, but `Path` seems better to me.

anott03:
Given that path is now from an external library (plenary), maybe `pathlib` is the best name. Either that or `Path`...

anott03:
@Conni2461 As I've been working on this, I realized that `find_files` and `git_files` are both broken. It seems more general though - any of the builtin pickers that sort files seem to be broken. If you have a chance, can you take a look?

Conni2461:
I can look at it. But can you rebase or so. Somehow the history and the diff looks super odd. Its hard to go through the changes with that :rofl: Ping me if its done.

l-kershaw:
@anott03 I think the problem that you are encountering is probably coming from the fact that Telescope's `make_relative` has different parameters to Plenary's `make_relative`.
The first has inputs `filepath` and `cwd` which are both strings. The second has inputs of `self` and `cwd` where `self` is a `Path` object and `cwd` is a string.

This ends up having plenary's make relative being given a string where it expects a `Path` object, so it passes `self.filename` (which is `nil`) to the `clean` function. This then causes errors when using `gsub` in the current way.

---

As mentioned by @Conni2461, it is a lot easier to help with bugs if you rebase your PR branch to the latest `master` rather than merge, as the latter causes there to be lots of irrelevant changes in the history of this branch (as they have already been merged into Telescope `master`).

anott03:
@l-kershaw @Conni2461 apologies for disappearing for a while - I'm fairly busy for the next week but after that I'll be able to spend more time working on this. I've rebased against master, so the history should be cleared up now (and hopefully I didn't break anything in the process🤞).

lua/telescope/make_entry.lua
```diff
@@ -446,7 +444,7 @@ function make_entry.gen_from_buffer(opts)
   return function(entry)
     local bufname = entry.info.name ~= "" and entry.info.name or '[No Name]'
     -- if bufname is inside the cwd, trim that part of the string
-    bufname = path.normalize(bufname, cwd)
+    bufname = Path.normalize(bufname, cwd)
```
Conni2461, line 447: this part looks wrong. `Path:new(bufname):normalize(cwd)`

lua/telescope/make_entry.lua
```diff
@@ -815,7 +813,7 @@ function make_entry.gen_from_ctags(opts)
   opts = opts or {}
 
   local cwd = vim.fn.expand(opts.cwd or vim.fn.getcwd())
-  local current_file = path.normalize(vim.fn.expand('%'), cwd)
+  local current_file = Path.normalize(vim.fn.expand('%'), cwd)
```
Conni2461, line 816: probably also `Path:new(vim.fn.expand('%')):normalize(cwd)`

Conni2461:
Also we wanna remove `telescope.path` so we should replace each function with a error message that says. "Please use plenary.path" or similar. So extensions know that they have to move to the new way and then remove the whole module in a month or so.

Similar to https://github.com/nvim-telescope/telescope.nvim/blob/49b86b4610199d879f539d9193fbee45792ba712/lua/telescope/utils.lua#L420-L434 for example (i think we could remove those functions now btw)

lets finish this pr in the next two days. I am here if you need me :)

anott03:
@Conni2461 how do we typically handle deprecating things like `path.separator` which are variables rather than functions?

anott03:
Could we do something like this?
```lua
local path = setmetatable({}, {
  __index = function()
    error("telescope.path is deprecated. please use plenary.path instead")
  end
})

return path

```

Conni2461:
Yes thats better for our case because we deprecate a whole module, not just a couple of functions :+1:

Conni2461 (approved):
LGTM

I need to test it, so i will run it today and see if something doesnt work. If all works out i'll merge this evening

Thank you very much :) and thank you for your patience :)

lua/telescope/builtin/files.lua
```diff
@@ -249,11 +248,10 @@ files.file_browser = function(opts)
     return finders.new_table {
       results = data,
       entry_maker = (function()
-        local tele_path = require'telescope.path'
         local gen = make_entry.gen_from_file(opts)
         return function(entry)
           local tmp = gen(entry)
-          tmp.ordinal = tele_path.make_relative(entry, opts.cwd)
+          tmp.ordinal = Path:new(entry):make_relative(opts.cwd)
```
Conni2461, line 254: Hmm `Telescope file_browser` is broken because dirs no longer have the os sep at the end of the file. We need to make sure that it stays there.

I dont even this this change is coursing it but changes in `gen_from_file`. Maybe we should write a completly new `entry_maker` for file_browser. What do you think?

lua/telescope/builtin/files.lua
```diff
@@ -286,7 +284,7 @@ files.file_browser = function(opts)
           return
         end
 
-        local fpath = current_picker.cwd .. os_sep .. file
+        local fpath = Path:new({current_picker.cwd, file})
```
Conni2461, line 287: Pretty sure this is wrong. Because we are doing `Path:new(fpath)` later. I haven't tested this but maybe `fpath:` stuff that happens below might just work.

lua/telescope/builtin/files.lua
anott03: Yeah it's not 100% clear to me what you mean, but I'd be happy to take a crack at it later today :)

lua/telescope/builtin/files.lua
Conni2461: Right might be but we also do a string sub in the if. I keep it as string for now :)

lua/telescope/builtin/files.lua
Conni2461: Sorry i just did it myself :)

Conni2461:
I talked with tj for a second. He wants to keep the module for now but throw an error. Just so extensions have a week or so to switch over to plenary.path

I am going to merge this now.

Thanks for all your work and your patience with me :)

clason:
Shouldn't this get an entry in https://github.com/nvim-telescope/telescope.nvim/blob/master/doc/telescope_changelog.txt?

Conni2461:
I forgot about this. Calling the function will yell at you tho :laughing:

clason:
Yeah, but some explanation -- and migration guide! -- would be good since it's not as simple as replacing `require'telescope.path'` with `require'plenary.path'` ;)

Conni2461:
I'll add one i noticed something else super small too too, i shouldn't have merged. Honest mistake. I was just glad right now that it was done and it didn't break anything inside telescope :laughing:

anott03:
@Conni2461 thanks for all your help! Lmk if there's anything else you'd like me to do :)

PR #478, opened 2021-01-29
Fix: error when repo has a lot of branches, forks and merges

Luxed:
This PR simply checks if `cend` evaluates to false, if it doesn't, it will add the highlight to the line.
Unfortunately, I don't have steps to reproduce or any more information like screenshots as I was only able to reproduce this on my current work repo.


elianiva:
LGTM! I can reproduce it with Neovim's repo and this PR fixed it :)

Conni2461:
Thanks :)

PR #479, opened 2021-01-30
fix: no TS highlight if it gets loaded after Telescope

elianiva:
credit goes to @Conni2461


Conni2461 (approved):
I think that should not break anything especially for clason. So you can merge :)

PR #486, opened 2021-02-01
fix: fix an invalid code example

yutakatay:


elianiva:
This has been fixed in #487 I think. I'm sorry, didn't saw your PR earlier so I merged the other one. But thanks for submitting this anyway :)

Issue #485, opened 2021-02-01
Missing a comma in README#mapping


peacock0803sz:
Missing a comma in README, mapping section

I tried the key mapping given in the README, but I got the following error.

```
E5107: Error loading lua [string ":lua"]:20: '}' expected (to close '{' at line 18) near '='
```

After the `["<esc>"] = actions.close`, does it need comma?

```lua
local actions = require('telescope.actions')
-- Global remapping
------------------------------
require('telescope').setup{
  defaults = {
    mappings = {
      i = {
        -- To disable a keymap, put [map] = false
        -- So, to not map "<C-n>", just put
        ["<c-x>"] = false,
        -- Otherwise, just set the mapping to the function that you want it to be.
        ["<C-i>"] = actions.goto_file_selection_split,
        -- Add up multiple actions
        ["<CR>"] = actions.goto_file_selection_edit + actions.center,
        -- You can perform as many actions in a row as you like
        ["<CR>"] = actions.goto_file_selection_edit + actions.center + my_cool_custom_action,
      },
      n = {
        ["<esc>"] = actions.close
        ["<C-i>"] = my_cool_custom_action,
      },
    },
  }
}
```

Should I send a pull request?

Thanks.

elianiva:
Yes, it needs a comma. Feel free to send a PR :)

PR #487, opened 2021-02-01
Correct a missing comma in README, Mapping section

peacock0803sz:
I found a missing comma in README, Mapping section. ref: #485


elianiva:
Good catch. Thank you!

PR #490, opened 2021-02-02
Add user autocmd Telescope 

glepnir:
Add user autocmd `Telescope` that do it before create Telescope window


Conni2461:
For next time. You should just be able to rebase on current master :)

glepnir:
Okay, I am used to creating new branches to avoid other situations

PR #491, opened 2021-02-02
Add user autocmd TelescopeFindPre

glepnir:
Add user autocmd Telescope that do it before create Telescope window

cc @Conni2461

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -326,8 +326,8 @@ function Picker:find()
 
   self.original_win_id = a.nvim_get_current_win()
 
-  -- Use autocmd run it before create Telescope window
-  vim.cmd'do User Telescope'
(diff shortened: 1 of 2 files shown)


lua/telescope/pickers.lua
```diff
@@ -326,6 +326,9 @@ function Picker:find()
 
   self.original_win_id = a.nvim_get_current_win()
 
+  -- Use autocmd run it before create Telescope window
+  vim.cmd'do User Telescope'
```
tjdevries, line 330: Can we call it `TelescopeFindPre`? Since it will be run before every find?

If that's fine, I'd like to see it as `vim.cmd [[doautocmd User TelescopeFindPre]]`

If you do that, then we can merge is fine with me.

glepnir:
@tjdevries  Done.

tjdevries:
Thanks glep 😁

PR #494, opened 2021-02-03
Adding visual selection for grep_string

fdschmidt93:
Hi there!

I humbly present my (very first) PR to add visual selection to grep_string. I noticed the comment suggesting to add the feature when I wrote the functionality for myself.

In particular, this PR supports multi-line forwards and backwards selection from 'visual' and 'visual line' modes. Multi-lines are space-delimited concatenated.

In addition, it might be sensible to update the prompt title with the visual selection or pipe the selection itself into the prompt rather than immediate search for feedback of selection though I remained in line with the word under cursor implementation to search for the visual selection.

Looking forward to feedback!

(Apologies for any possible rookie mistakes, first time 'proper use of git' and hacking lua in neovim :)

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -10,8 +10,6 @@ local flatten = vim.tbl_flatten
 
 local files = {}
 
-local supported_visual_modes = {'v', 'V'}
-
 local escape_chars = function(string)
   return string.gsub(string,  "[%(|%)|\\|%[|%]|%-|%{%}|%?|%+|%*]", {
     ["\\"] = "\\\\", ["-"] = "\\-",
@@ -64,16 +62,10 @@ end
 --  opts.search -- the string to search.
 --  opts.search_dirs -- list of directory to search in
 files.grep_string = function(opts)
+  -- TODO: This should probably check your visual selection as well, if you've got one
 
   local vimgrep_arguments = opts.vimgrep_arguments or conf.vimgrep_arguments
   local search_dirs = opts.search_dirs
-
-  -- set search from visual mode
-  if vim.tbl_contains(supported_visual_modes, vim.fn.mode()) then
-      opts.search = utils.get_visual_selection(opts.delimiter)
-      vim.api.nvim_command('normal :esc<CR>')
-  end
-
   local search = escape_chars(opts.search or vim.fn.expand("<cword>"))
   local word_match = opts.word_match
   opts.entry_maker = opts.entry_maker or make_entry.gen_from_vimgrep(opts)
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -10,6 +10,8 @@ local flatten = vim.tbl_flatten
 
 local files = {}
 
+local supported_visual_modes = {'v', 'V'}
+
 local escape_chars = function(string)
   return string.gsub(string,  "[%(|%)|\\|%[|%]|%-|%{%}|%?|%+|%*]", {
     ["\\"] = "\\\\", ["-"] = "\\-",
@@ -62,10 +64,16 @@ end
 --  opts.search -- the string to search.
 --  opts.search_dirs -- list of directory to search in
 files.grep_string = function(opts)
-  -- TODO: This should probably check your visual selection as well, if you've got one
 
   local vimgrep_arguments = opts.vimgrep_arguments or conf.vimgrep_arguments
   local search_dirs = opts.search_dirs
+
+  -- set search from visual mode
+  if vim.tbl_contains(supported_visual_modes, vim.fn.mode()) then
+      opts.search = utils.get_visual_selection(opts.delimiter)
+      vim.api.nvim_command('normal :esc<CR>')
+  end
+
   local search = escape_chars(opts.search or vim.fn.expand("<cword>"))
   local word_match = opts.word_match
   opts.entry_maker = opts.entry_maker or make_entry.gen_from_vimgrep(opts)
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -295,7 +295,7 @@ function utils.get_visual_selection_range()
 end
 
 function utils.get_visual_selection(delimiter)
-    local delimiter = delimiter or ' '
+    delimiter = delimiter or ' '
     local mode = vim.fn.mode()
     local line_start, column_start, line_end, column_end = utils.get_visual_selection_range()
     local lines = vim.api.nvim_buf_get_lines(0, line_start - 1, line_end, false)
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -284,40 +284,96 @@ function utils.strcharpart(str, nchar, charlen)
   return str:sub(nbyte + 1, nbyte + len)
 end
 
-function utils.get_visual_selection_range()
+
+function utils.get_visual_pos(exit_visual_mode)
+    exit_visual_mode = exit_visual_mode or true
+
+    -- get {start: 'v', end: curpos} of visual selection 0-indexed
     local line_v, column_v = unpack(vim.fn.getpos("v"), 2, 3)
     local line_cur, column_cur = unpack(vim.fn.getcurpos(), 2, 3)
+
+    if exit_visual_mode then
+      vim.cmd [[normal :esc<CR>]]
+    end
+
     -- backwards visual selection
-    if line_v > line_cur or (line_v == line_cur and column_cur < column_v) then
-        return line_cur, column_cur, line_v, column_v
+    if line_v > line_cur then
+        line_cur, line_v = line_v, line_cur
+    end
+    if column_v > column_cur then
+        column_cur, column_v = column_v, column_cur
     end
     return line_v, column_v, line_cur, column_cur
 end
 
-function utils.get_visual_selection(delimiter)
+
+-- Recursively resolve beginning or ending byte column for multi-width characters
+--  Example: multi-width Japanese Zenkaku, see change in columns from left to right char
+--  全角
+-- @param line string: buffer line
+-- @param byte_col integer: initial byte column
+-- @param offset integer: resolver char border towards char beginning (-1) or char end (+1)
+-- return byte_col integer: byte column of char beginning or end
+function utils.resolve_col(line, byte_col, offset)
+    local utf_start, _ = vim.str_utfindex(line, math.min(byte_col, #line))
+    local utf_start_offset, _ = vim.str_utfindex(line, math.min(byte_col + offset, #line))
+    if utf_start == utf_start_offset and byte_col + offset <= #line then
+        return utils.resolve_col(line, byte_col + offset, offset)
+    else
+        return byte_col
+    end
+end
+
+function utils.get_visual_selection(delimiter, trim, exit_visual_mode)
     delimiter = delimiter or ' '
-    local mode = vim.fn.mode()
-    local line_start, column_start, line_end, column_end = utils.get_visual_selection_range()
+    trim = trim or true
+    exit_visual_mode = exit_visual_mode or ' '
+
+    local mode = vim.api.nvim_get_mode().mode
+    local line_start, column_start, line_end, column_end = utils.get_visual_pos(exit_visual_mode)
     local lines = vim.api.nvim_buf_get_lines(0, line_start - 1, line_end, false)
 
     local concat = {}
     local first_line = 1
     local last_line = line_end - (line_start - 1)
-    if first_line == last_line then
+
+    if first_line == last_line or mode == '' then
         -- get difference in columns, inclusive
-        column_end = column_end - column_start + 1
+        column_end =  column_end - column_start + 1
     end
     for row=first_line,last_line do
         local line = lines[row]
         if mode ~= 'V' then
-            if row == first_line then
-                line = line:sub(column_start)
+            if row == first_line or mode == '' then
+                -- Recursively get first byte index of utf char for initial byte column selection
+                --  Example: multi-width Japanese Zenkaku, see change in columns from left to right char
+                --  全角
+                local byte_col = utils.resolve_col(line, column_start, -1)
+                line = line:sub(byte_col)
             end
-            if row == last_line then
-                line = line:sub(1, column_end)
+            if row == last_line or mode == '' then
+                -- Recursively get last byte index of utf char for last byte column selection
+                local byte_col = utils.resolve_col(line, column_end, 1)
+                if mode == '' then
+                    -- if column was resolved and not initial char did not surpass column end, extra character required
+                    local char_begin_col = utils.resolve_col(line, byte_col, -1)
+                    if char_begin_col < column_end and byte_col >= column_end and row ~= last_line then
+                        byte_col = utils.resolve_col(line, byte_col+1, 1)
+                    end
+                end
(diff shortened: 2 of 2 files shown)


elianiva:
LGTM! The only nitpick I have is `:Telescope grep_string` doesn't work when using range (well, none of Telescope commands do). I don't know if we should make them accept range since the only picker that supports range is `grep_string` when this PR gets merged.

Conni2461:
I am busy today. I will take a closed look at it tomorrow :)
The only thing that i am currently thinking is replacing some of `vim.fn.` with `vim.api.nvim_` functions.
- `vim.fn.mode` -> `vim.api.nvim_get_mode`
- `nvim_win_get_cursor()`
- ...

fdschmidt93:
Many thanks for the comments!

> LGTM! The only nitpick I have is `:Telescope grep_string` doesn't work when using range (well, none of Telescope commands do). I don't know if we should make them accept range since the only picker that supports range is `grep_string` when this PR gets merged.

Hmm, I suppose visual selection technically applies just as well to other pickers (most importantly `current_buffer_fuzzy_find`), though the same thing is I guess just as true for word under cursor selection.

I guess there's two ways to think about it -

1. Exclusively adopt visual_selection_range and let users pass `opt.search` themselves
    * Move exiting visual mode to  `utils.get_visual_selection_range()` and expose with bool flag (probably good idea anyways)
    * Clean in implementation, though I guess less ergonomic in 'use' (i.e. always requires definition of mapping or separate function)
2. Switch picker behavior by currently selected mode, either solely for `grep_string`, selected functions, or any picker

Any thoughts?

> The only thing that i am currently thinking is replacing some of vim.fn. with vim.api.nvim_ functions.

Generally agree and locally switched to `vim.api.nvim_get_mode().mode` already!

However `utils.get_visual_selection_range()` for the time being I presume better relies on `vim.fn.` for, to the best of my knowledge, two reasons. On one hand, there is no elegant `vim.api` counterpart for `getpos('v')` (i.e. row and col offsets of start of visual selection). On the other hand, `nvim_win_get_cursor()` returns 0-indexed columns (as opposed to 1) and hence introduces some inconsistency to `getpos('v')`, i.e. it would like this at the expense of clarity.

```
function utils.get_visual_selection_range()
    local line_v, column_v = unpack(vim.fn.getpos("v"), 2, 3)
    local line_cur, column_cur = unpack(vim.api.nvim_win_get_cursor(0))
    column_cur = column_cur + 1
    -- backwards visual selection
    if line_v > line_cur or (line_v == line_cur and column_cur < column_v) then
        return line_cur, column_cur, line_v, column_v
    end
    return line_v, column_v, line_cur, column_cur
end
```
Though should there be a `vim.api`-equivalent of `getpos('v')`, then of course, it makes sense to use that api. Is there any tangible benefit to `nvim_win_get_cursor` over `getcurpos`?

> But it looks like a great addition so thank you :) and welcome to open source and telescope :)

Thanks again :)

elianiva:
I think visual selection input could also useful for other pickers. Maybe we should make Telescope commands accept range?

lua/telescope/builtin/files.lua
```diff
@@ -62,10 +64,16 @@ end
...
 files.grep_string = function(opts)
-  -- TODO: This should probably check your visual selection as well, if you've got one
 
   local vimgrep_arguments = opts.vimgrep_arguments or conf.vimgrep_arguments
   local search_dirs = opts.search_dirs
+
+  -- set search from visual mode
+  if vim.tbl_contains(supported_visual_modes, vim.fn.mode()) then
+      opts.search = utils.get_visual_selection(opts.delimiter)
+      vim.api.nvim_command('normal :esc<CR>')
```
tjdevries, line 74: Usually we write `vim.cmd [[...]]` . In this case I think maybe you want `:stopinsert` though?

lua/telescope/utils.lua
```diff
@@ -284,4 +284,42 @@ function utils.strcharpart(str, nchar, charlen)
   return str:sub(nbyte + 1, nbyte + len)
 end
 
+function utils.get_visual_selection_range()
```
tjdevries, line 287: I think `:help vim.region()` may be of use to you here. I'm not 100% sure though without reading some more code.

tjdevries:
I think a better way to do something like this would be to expose a telescope utility function that could basically be used like this:

```lua
require('telescope.builtin').grep_string {
  search = require('telescope.utils').get_visual_selection()
  ...
}
```

lua/telescope/utils.lua
fdschmidt93: Thanks for the pointer, I'll look into it!

lua/telescope/builtin/files.lua
fdschmidt93: I see! `:stopinsert` throws an error though saying `no range allowed` since its called from visual mode. I'll recheck to see if there is a more idiomatic way to leave visual mode.

Conni2461:
See https://github.com/neovim/neovim/pull/13896

fdschmidt93:
This branch now supports variable-width characters and selection via visual block mode with `utils.get_visual_selection()` and can be passed as proposed above to pickers that support `search`. 

This differs from neovim/neovim#13896 that can't support v-block mode with `vim.fn.strpart` as I think that assumes multi-width chars were rendered single width. This branch recursively resolves multi-width chars and catches v-block mode scenarios in best-effort.

Known issue: 
- Starting byte columns of subsequent lines might be smaller than of the beginning line when character-widths vary across lines leading to cutoff of starting byte column difference
 
Happy to get this merged though PR can also be closed and discussion be taken upstream since core functionality will likely be eventually supported there.

clason:
If this improves on the implementation in https://github.com/neovim/neovim/pull/13896, please consider an upstream patch!

clason:
In particular, `resolve_col` seems like something that could go into core as a replacement/improvement on `vim.str_byteindex`.

fdschmidt93:
Full block mode support for (visual) selection is now in this branch, barring block mode selections that include EOL chars, which pre-/appends an extra char (which I tried to address with `virtualedit` though that introduced other side effects, so leaving this for now since I'm not sure if it can be supported). 

On a high-level, v-block mode is supported as follows:

1. Get byte column of start / end positions (e.g. `getpos("v")`)
2. Get corresponding visual columns (see `:help wincol`)
3. Resolve edge visual columns (for multi-width chars, merely first visual column is returned)
4. Get beginning and end byte offsets by line that are within visual wincol boundaries

Pinging @clason, would you suggest to get this reviewed here or should I refactor `vim.region` right away and submit upstream? Not sure what's easier to review, as the function as is returns a visual selection to possibly simplify any debugging.

clason:
Awesome! (Since you mention `virtualedit`: have you tested this, too? There was an issue about `vim.region` not working correctly with that set, which I fixed at some point.)

> Pinging @clason, would you suggest to get this reviewed here or should I refactor `vim.region` right away and submit upstream? Not sure what's easier to review, as the function as is returns a visual selection to possibly simplify any debugging.

Good question! You are right that it's easier to test, iterate (and get merged) here. On the other hand, you'd get more eyes at the main repo -- and get the tests already in place for `vim.region` and `highlight` for free.

lua/telescope/utils.lua
```diff
@@ -284,4 +284,244 @@ function utils.strcharpart(str, nchar, charlen)
...
+      return l_start, c_start, l_end, c_end
+    end
+end
+
+-- Retrieve wincol from pos
+-- wincol (see :help wincol) denotes the visual column of the current cursor position
+-- Since wincol can only be retrieved for the current position,
+-- the cursor has to be set prior
+-- @param pos table position to retrieve wincol for
+-- @param start_pos table optional position to jump back to
```
clason, line 350: Since `get_wincol` does the jump, shouldn't it also be responsible for restoring cursor position?

lua/telescope/utils.lua
```diff
@@ -284,4 +284,244 @@ function utils.strcharpart(str, nchar, charlen)
...
+
+-- Get maximum, possibly synthetic wincol for position
+-- Consider a multi-width char: 全, wincol solely returns the first visual column
+-- edge_wincol returns the left or right border wincol instead depending on parameters
+-- @param pos table {row, column} with {1, 0} indexed values
+-- @param wincol integer initial wincol prior to finding (synthetic) maximum
+-- @param maximum boolean true resolves towards right, false resolves towards left
+function utils.edge_wincol(pos, wincol, maximum)
+    maximum = maximum or false
+    local max_col = utils.get_wincol({pos[1], 2^31-1})
```
clason, line 371: maybe use `math.huge` here?

lua/telescope/utils.lua
```diff
@@ -284,4 +284,244 @@ function utils.strcharpart(str, nchar, charlen)
...
+    local utf_start, _ = vim.str_utfindex(line, math.min(byte_col, #line))
+    local utf_start_offset, _ = vim.str_utfindex(line, math.min(byte_col + offset, #line))
+    if utf_start == utf_start_offset and byte_col + offset <= #line and byte_col + offset >= 0 then
+        return utils.resolve_bytecol(line, byte_col + offset, offset)
+    else
+        return byte_col
+    end
+end
+
+function utils.get_visual_selection(delimiter, trim, exit_visual_mode)
```
clason, line 457: missing doc string?

lua/telescope/utils.lua
fdschmidt93: Not sure I follow. If `start_pos` is set the cursor jumps back to that position. In the actual function `get_visual_selection`, I also jump back right before returning the visual selection as otherwise I'd have to keep track of the starting position everywhere (though I'm sure it currently jumps more often than needed :) - will check what can be removed.

lua/telescope/utils.lua
fdschmidt93: Good point!

lua/telescope/utils.lua
clason: If I understand correctly (and I might not!), what currently happens is:

1. the calling function (`get_visual_selection`) stores the current cursor position
2. it then calls `get_wincol` with that position as argument
3. `get_wincol` jumps to the location of the mark and then jumps back to the passed position

I'd suggest to have `get_wincol` _always_ jump back (saving the current position before jumping), even if that amounts to duplicating saving the current position. Otherwise it seems a bit of a misnomer: I'd expect a function called `get_wincol` to be side-effect free so it's guaranteed to return the position without changing state (like changing cursor position).

fdschmidt93:
Yes, I, for instance, tried setting `virtualedit=all` before getting the {start,end} positions and resetting it to "" before returning the selection. 

However, I just noticed that "v" (or "'<") register actually seems to store the byte column beyond the line, so it should be straightforward to catch after all. I'll include it in the next iteration.

clason:
Just some drive-by comments, not a full review ;)

lua/telescope/utils.lua
```diff
@@ -284,4 +284,244 @@ function utils.strcharpart(str, nchar, charlen)
...
+-- @param maximum boolean true resolves towards right, false resolves towards left
+function utils.edge_wincol(pos, wincol, maximum)
+    maximum = maximum or false
+    local max_col = utils.get_wincol({pos[1], 2^31-1})
+    if wincol >= max_col then
+        return max_col
+    end
+    if wincol == 1 then
+        return 1
+    end
```
tami5, line 377: not sure if this breaks anything but it maybe better to have one parent if expression, parentheses are not required but make it more readable and accept additional `and/or`
```lua
if wincol then 
  return (wincol >= max_col) and max_col or 1 
end 
```

lua/telescope/utils.lua
```diff
@@ -284,4 +284,244 @@ function utils.strcharpart(str, nchar, charlen)
...
+
+    local newcol = utils.get_wincol(pos)
+    if maximum then
+        if newcol > wincol then
+            return newcol - 1
+        else
+            return utils.edge_wincol({pos[1], pos[2] + 1}, wincol, true)
+        end
```
tami5, line 385: same case with having a single parent if expression

lua/telescope/utils.lua
```diff
@@ -284,4 +284,244 @@ function utils.strcharpart(str, nchar, charlen)
...
+-- vim.str_byteindex is similar though can only return the right-hand border
+-- Note: byte columns are 1-indexed for simplified interop with lua string:sub
+-- @param line string: buffer line
+-- @param byte_col integer: initial byte column
+-- @param offset integer: resolver char border towards char beginning (-1) or char end (+1)
+-- return byte_col integer: byte column of char beginning or end
+function utils.resolve_bytecol(line, byte_col, offset)
+    local utf_start, _ = vim.str_utfindex(line, math.min(byte_col, #line))
+    local utf_start_offset, _ = vim.str_utfindex(line, math.min(byte_col + offset, #line))
+    if utf_start == utf_start_offset and byte_col + offset <= #line and byte_col + offset >= 0 then
```
tami5, line 450: store the condition in a valid_x then do if valid_x to increase readability

lua/telescope/utils.lua
```diff
@@ -284,4 +284,244 @@ function utils.strcharpart(str, nchar, charlen)
...
+        return utils.resolve_bytecol(line, byte_col + offset, offset)
+    else
+        return byte_col
+    end
+end
```
tami5, line 455: or better with var valid_x do

```lua
return not valid_x and byte_col or  utils.resolve_bytecol(..)
```

lua/telescope/utils.lua
```diff
@@ -284,4 +284,244 @@ function utils.strcharpart(str, nchar, charlen)
...
+
+function utils.get_visual_selection(delimiter, trim, exit_visual_mode)
+    delimiter = delimiter or ' '
+    trim = trim or true
```
tami5, line 459: use vim.F.if_nil

fdschmidt93: Thanks for your comments! The `x` or true are a bit depressing :) saw a similar pattern somewhere once and copied that, though thinking about it after your comment.. it's a bit painfully obvious I'm afraid haha

clason: _everybody_ gets bitten by that at some point :)

lua/telescope/utils.lua
fdschmidt93: Thanks for elaborating, makes perfect sense! Your understanding is correct, `vim.fn.wincol` only returns the visual column from the current cursor position. Other than that, speed is a non-issue (I guess thanks to luajit) as e.g. 500 loc are  processed instantaneously.

lua/telescope/utils.lua
tami5: :laughing: actually conni and tj told me about it, once I went crazy debug and that was the reason :laughing: . 

basically vim.F.if_nil is the following

```lua
x = x == nil and true or x

function F.if_nil(a, b)
  return a == nil and b or a
end

```

tjdevries:
Honestly, this PR looks really interesting and seems to solve a lot of problems that people have with multi-byte things.

We should just add this to neovim core directly I think :) Is that OK with you @fdschmidt93 ? I think we can do this in favor of: https://github.com/neovim/neovim/pull/13896 and then after that we can look into re-using around several places within neovim core.

Of course, when merged to neovim core, then people can just use that function to get the visually selected text :)

Does that sound OK?

fdschmidt93:
Sounds great! :) I'll take care of last issues around EOL characters tomorrow and will then open a PR upstream.

PR #497, opened 2021-02-04
Open quickfix list on it's own action

tssm:
Given that Telescope has a picker for quickfix items, opening its window automatically isn't that useful, so this commit removes the call from send_to_qflist and send_selected_to_qflist. Some users may prefer the old behaviour though so a new action open_qflist was created.


tjdevries:
:+1: LGTM.

One thing to note for people coming here later to check, is that you can do this:

```
```

in your mappings dictionary to do both of these actions in one command.

PR #505, opened 2021-02-06
Buffer previewer ls colorized

Conni2461:
Need help with color groups otherwise should be done.

Also probably we should create Telescope hl_groups for that. I am open for default links :)
Also i need to think about the code 😆 Just wrote that in 30 mins so might not be optimal in terms of everything

| Before | After |
|--------|-------|
|  | t0NvTME |

Edit: new idea plenary already has the sections so why not use that 🤣

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -15,6 +15,50 @@ local previewers = {}
 
 local ns_previewer = vim.api.nvim_create_namespace('telescope.previewers')
 
+local color_hash = {
+  ['p'] = 'TelescopePreviewPipe',
+  ['c'] = 'TelescopePreviewCharDev',
+  ['d'] = 'TelescopePreviewDirectory',
+  ['b'] = 'TelescopePreviewBlock',
+  ['l'] = 'TelescopePreviewLink',
+  ['s'] = 'TelescopePreviewSocket',
+  ['.'] = 'TelescopePreviewNormal',
+  ['r'] = 'TelescopePreviewRead',
+  ['w'] = 'TelescopePreviewWrite',
+  ['x'] = 'TelescopePreviewExecute',
+  ['-'] = 'TelescopePreviewHyphen',
+  ['T'] = 'TelescopePreviewSticky',
+  ['S'] = 'TelescopePreviewSticky',
+  [2]   = 'TelescopePreviewSize',
+  [3]   = 'TelescopePreviewUser',
+  [4]   = 'TelescopePreviewGroup',
+  [5]   = 'TelescopePreviewDate',
+}
+color_hash[6]   = function(line)
+  return color_hash[line:sub(1, 1)]
+end
+
+local colorize_ls = function(bufnr, data, sections)
+  local windows_add = path.separator == '\\' and 2 or 0
+  for lnum, line in ipairs(data) do
+    local section = sections[lnum]
+    for i = 1, section[1].end_index - 1 do -- Highlight permissions
+      local c = line:sub(i, i)
+      vim.api.nvim_buf_add_highlight(bufnr, ns_previewer, color_hash[c], lnum - 1, i - 1, i)
+    end
+    for i = 2, #section do -- highlights size, (user, group), date and name
+      local hl_group = color_hash[i + (i ~= 2 and windows_add or 0)]
+      vim.api.nvim_buf_add_highlight(bufnr,
+        ns_previewer,
+        type(hl_group) == "function" and hl_group(line) or hl_group,
+        lnum - 1,
+        section[i].start_index - 1,
+        section[i].end_index - 1
+      )
+    end
+  end
+end
+
 previewers.file_maker = function(filepath, bufnr, opts)
   opts = opts or {}
   if opts.use_ft_detect == nil then opts.use_ft_detect = true end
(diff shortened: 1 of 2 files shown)


lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -15,6 +15,74 @@ local previewers = {}
...
+    local sections = get_sections(line)
+    for i = 1, #sections[1].str do -- Highlight permissions
+      local c = sections[1].str:sub(i, i)
+      vim.api.nvim_buf_add_highlight(bufnr, ns_previewer, color_hash[c], lnum - 1, i - 1, i)
+    end
+    -- TODO(conni2461): Change hardcoded colors
+    for i = 2, #sections do
+      local color = color_hash[i]
+      color = type(color) == 'function' and color(line) or color
+      vim.api.nvim_buf_add_highlight(bufnr,
```
elianiva, line 75: I managed to break this line, let me find a way to reproduce it.

elianiva: ok I found it. If we have a directory with a space in its name, it breaks. I was wildly guessing (without reading the code) that you use a space for splitting the section so I decided to try it out.

Conni2461: Yeah doing that. I have thought about having spaces in filenames and then i forgot about it 🤣 One sec

Conni2461: Btw thanks good catch

Conni2461: Should work now.

Conni2461: Oh windows might be broken 😆 They get a different output. I do that later

Conni2461:
@elianiva Can you retest now requires https://github.com/nvim-lua/plenary.nvim/pull/63

We already kinda already have the section start and end index in scandir.ls so we should just do it there and pass it down.

elianiva:
just tested it out, works great!

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -24,9 +67,13 @@ previewers.file_maker = function(filepath, bufnr, opts)
     filepath = vim.fn.expand(filepath)
     local stat = vim.loop.fs_stat(filepath) or {}
     if stat.type == 'directory' then
-      pscan.ls_async(filepath, { hidden = true, on_exit = vim.schedule_wrap(function(data)
-        vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, data)
-        if opts.callback then opts.callback(bufnr) end
+      pscan.ls_async(filepath, {
+        hidden = true,
+        group_directories_first = true,
```
Conni2461, line 72: Anyone a issue with this option? Because i like it :)

plugin/telescope.vim
```diff
@@ -27,6 +27,18 @@ highlight default link TelescopePromptPrefix Identifier
...
+highlight default link TelescopePreviewNormal Normal
+highlight default link TelescopePreviewRead Constant
+highlight default link TelescopePreviewWrite Statement
+highlight default link TelescopePreviewExecute String
+highlight default link TelescopePreviewNon NonText
+highlight default link TelescopePreviewSticky Keyword
+highlight default link TelescopePreviewSize String
+highlight default link TelescopePreviewUser Constant
+highlight default link TelescopePreviewGroup Constant
+highlight default link TelescopePreviewDate Directory
```
Conni2461, line 40: I am a noob in hl groups. Thats the best i could do :)

elianiva: `TelescopePreviewNon` is a bit ambiguous, I think. What does it do? (reading from the code, I think it highlights `-`)

Conni2461: Yeah. Highlights `-` so if a permission is not set.

But i also realized i forgot to add those filetypes so if we have a `b` (block device) we error:
```
local type_tbl = { [1]  = 'p', [2]  = 'c', [4]  = 'd', [6]  = 'b', [10] = '-', [12] = 'l', [14] = 's' }
```

Gotta fix that tomorrow :)
If you have a better name for `TelescopePreviewNon` i would be very happy and thankful :)

lua/telescope/previewers/buffer_previewer.lua
elianiva: No, I don't have any issue and I like it as well!

plugin/telescope.vim
Conni2461: Btw just putting this here. We kinda never get a symbolic link. `uv` gives me the same octal for a link as for a file. :|

`Hypen` sounds good :) Thanks elianiva

elianiva: > Btw just putting this here. We kinda never get a symbolic link. uv gives me the same octal for a link as for a file. :|

Yeah, it looks a bit different from doing `ls -l`, Telescope shows `drwxr-xr-x` while `ls -l` shows `lrwxrwxrwx`.  But it doesn't really matter to me personally :laughing:

Conni2461: I think whats happening is that `uv.fs_stat` actually resolves it. Even the inode is the same which it isn't. Or do i resolve it? No i don't think i need to investigate :rofl:

Conni2461: But doesn't change anything. Doing this with `a` being file and `b` a symbolic link of `a` gives me the same inode/mode
```lua
local P = function(...) print(vim.inspect(...)) end
P(vim.loop.fs_stat('a'))
P(vim.loop.fs_stat('b'))
```

Conni2461: `lstat` works. So doing `vim.loop.fs_lstat`

Conni2461:
@elianiva Can you retest? Needs a new version of nvim-lua/plenary.nvim#63 as well. Some more bugfixes and links are now shown and resolved like this: `a -> b` and highlight only `a`


CC @timbedard might be interesting for you as well :) Just because we talked a lot about that ls earlier :)

timbedard:
Neat! Yeah, I'll pull it down today or tomorrow and give it a go.

elianiva:
very cool! now it works with symlinks

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -15,6 +15,50 @@ local previewers = {}
...
+  ['c'] = 'TelescopePreviewCharDev',
+  ['d'] = 'TelescopePreviewDirectory',
+  ['b'] = 'TelescopePreviewBlock',
+  ['l'] = 'TelescopePreviewLink',
+  ['s'] = 'TelescopePreviewSocket',
+  ['.'] = 'TelescopePreviewNormal',
+  ['r'] = 'TelescopePreviewRead',
+  ['w'] = 'TelescopePreviewWrite',
+  ['x'] = 'TelescopePreviewExecute',
+  ['-'] = 'TelescopePreviewHypen',
```
elianiva, line 29: shouldn't this be `hyphen` instead of `hypen`?

Conni2461: Yeah typo thanks

Conni2461:
Last small update to plenary. It now aligns `username` and `group` correctly :) I don't have a before picture but i didn't look good :laughing: 
image

Issue #507, opened 2021-02-07
Feature: buitin.windows picker


kyoh86:
With many tabs and windows, I want to focus it with telescope.
(I'll create a pull-request for this one)

kyoh86:
> I've created an extension [telescope-windows.nvim](https://github.com/kyoh86/telescope-windows.nvim).
> [TC72/telescope-tele-tabby.nvim](https://github.com/TC72/telescope-tele-tabby.nvim) and this are meet the requirements of this PR and #507.

PR #508, opened 2021-02-07
create new builtin "windows" picker

kyoh86:
for: #507

I've tried to create it as extension, but there's some private methods in internal.lua.
So I was judged that it should be as builtin method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,6 +387,7 @@ Built-in functions. Ready to be bound to any key you like. :smile:
 | Functions                           | Description                                                                                 |
 |-------------------------------------|---------------------------------------------------------------------------------------------|
 | `builtin.buffers`                   | Lists Open buffers in the current vim instance.                                             |
+| `builtin.windows`                   | Lists windows in the current vim instance and focus it.                                     |
 | `builtin.oldfiles`                  | Lists Previously open files.                                                                |
 | `builtin.commands`                  | Lists Available plugin/user commands and run it.                                            |
 | `builtin.tags`                      | Lists Tags in current directory with preview (ctags -R)                                     |
@@ -400,7 +401,7 @@ Built-in functions. Ready to be bound to any key you like. :smile:
 | `builtin.vim_options`               | Lists vim options and on enter edit the options value.                                      |
 | `builtin.registers`                 | Lists vim registers and edit or paste selection.                                            |
 | `builtin.autocommands`              | Lists vim autocommands and go to their declaration.                                         |
-| `builtin.spell_suggest`             | Lists spelling suggestions for <cword>.
+| `builtin.spell_suggest`             | Lists spelling suggestions for <cword>.                                                     |
 | `builtin.keymaps`                   | Lists normal-mode mappings.                                                                 |
 | `builtin.filetypes`                 | Lists all filetypes.                                                                        |
 | `builtin.highlights`                | Lists all highlights.                                                                       |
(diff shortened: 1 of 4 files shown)


elianiva:
From what I understand, extension needs extra dependency to install so you're right, this feature belongs in builtin.

delphinus:
It is useful to removing unfocuasable windows from candidates. It can be detected from `nvim_win_get_config()`.

```lua
local win_conf = vim.api.nvim_win_get_config(winnr)
local is_focusable = win_conf.relative == '' or win_conf.focusable
```

delphinus:
And one more suggestion. Can I see tabnr for each window?

Conni2461:
Okay i am currently struggling to understand why we need or want this at all. 
Also if you wanna do something like this i would just make `get_devicons` public. Move it somewhere or so.

kyoh86:
@delphinus 

Thank you for your advice!

> It is useful to removing unfocuasable windows from candidates
> Can I see tabnr for each window?

- removed unfocusable windows
- put tabpage on heads of each item

kyoh86:
What is the status of this PR?

Conni2461:
This is the status from my pov.

> Okay i am currently struggling to understand why we need or want this at all.

I am just not sure if we need a feature that is so similar to builtin.buffers and shares a lot of the same code. But i don't make the final decisions here, thats just my opinion :)
But I am happy to do something similar to fzf with putting those kinda pickers into a new section in the wiki with credit obviously :) We can have collapsible sections for each one. Like this:
## A collapsible section with markdown
<details>
  <summary>Click to expand!</summary>
  
```lua
print('Hello World')
```
</details>

elianiva:
I agree with Conni. If it's so similar to `builtin.buffers`, I think it's better to put this in the Wiki with your name on it

jedrzejboczar:
If you're wondering about use case for this kind of picker I'd say I find it really useful in my workflow. I actually have pretty similar implementation in my config.

I often have several tabs open, each with ~2-3 windows (treating them like "views"). When I know that I have a given file already open in some tab but don't remember which one then the window picker solves the problem. The important part is that it uses `vim.api.nvim_set_current_win(winnr)` which jumps to the window (and so switches to the corresponding tab) instead of opening the file in the current window.

I'm not saying it should be a builtin, maybe not, just wanted to mention that there are use cases for it, which make it differ from `builtin.buffers`. If `builtin.buffers` allowed to filter the buffers to show only the ones open in windows, than it should just be a matter of replacing `_goto_file_selection` to have the window picker, so maybe it's enough to just extend `builtin.buffers` with this option?

For reference here is the version I use (mostly copy-paste from `builtin.buffers`):
https://gist.github.com/jedrzejboczar/0df10e82e481ac62fa62d05a5aa6b1a9

kyoh86:
I agree with @jedrzejboczar. Thank you.

TC72:
I've just created an extension which does the same thing, called tele-tabby, I've described is as a tab switcher.
Even though you are technically switching windows it's not really needed if you only have 1 tab.
I modify and highlight the buffer path so you can easily see the project root of the tab and its current working directory.

My workflow is to use separate tabs with their own working directories, set with tcd, this allows me to focus the results I get back from telescope tools like find_files.

Please try it out, any feedback is appreciated.
https://github.com/TC72/telescope-tele-tabby.nvim

jedrzejboczar:
@TC72 Thanks for that extension, I think I'll switch to it now. I agree that calling it a "tab picker" may make more sense, at least in my workflow. At first glance I'm just missing the preview a bit, but I'm not sure if I actually need it :P

TC72:
@jedrzejboczar that's great to hear someone's using it. I think you could use a preview if you have it in a horizontal split. The layout where the preview is to the side makes it difficult to see the full path which is the main feature of my tab switcher. But I may look at an option to shorten the path.
If you have any issues or ideas please let me know.

kyoh86:
I've created an extension [telescope-windows.nvim](https://github.com/kyoh86/telescope-windows.nvim).
[TC72/telescope-tele-tabby.nvim](https://github.com/TC72/telescope-tele-tabby.nvim) and this are meet the requirements of this PR and #507.

Issue #509, opened 2021-02-07
Feature: builtin.git_recents picker


kyoh86:
With many files in a repo, I want to open one of them which is updated recently.
(I'll create a pull-request for this one)

YodaEmbedding:
Some possible heuristics could be:

 - Files recently committed _by the current author_ on the current branch. (Avoids issues with multi-developer projects.)
 - Modified files in current working tree.
 - MRU database for the current git project. (See also: [fzf-filemru](https://github.com/tweekmonster/fzf-filemru) and [fzf-mru.vim](https://github.com/pbogut/fzf-mru.vim).)

Conni2461:
We already have this mru database extension https://github.com/nvim-telescope/telescope-frecency.nvim

It works with one database for all projects but you can limit it to the directory where the lsp is currently attached with `:LSP:` tag. (I think he is also wanted to add a `:CWD` tag, not sure). Tag will be remebered. You can also define your own Projects and use them as tags. You can also enable `show_unindexed` which will show all files in the current tag even if they are never opened.

PR #510, opened 2021-02-07
add builtin.git_recents picker

kyoh86:
for: #509

EDIT: With `git diff HEAD~n --name-only`, git_recents picks files which changed in recent n-commits.

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -137,6 +137,26 @@ git.status = function(opts)
   }):find()
 end
 
+git.recents = function(opts)
+  local depth = utils.get_default(opts.depth, 5)
+
+  if opts.cwd then
+    opts.cwd = vim.fn.expand(opts.cwd)
+  end
+
+  opts.entry_maker = opts.entry_maker or make_entry.gen_from_file(opts)
+
+  pickers.new(opts, {
+    prompt_title = 'Git Recents',
+    finder = finders.new_oneshot_job(
+      {'git', 'diff', '--name-only', depth and 'HEAD~' .. depth or 'HEAD'},
+      opts
+    ),
+    previewer = conf.file_previewer(opts),
+    sorter = conf.file_sorter(opts),
+  }):find()
+end
+
 local set_opts_cwd = function(opts)
   if opts.cwd then
     opts.cwd = vim.fn.expand(opts.cwd)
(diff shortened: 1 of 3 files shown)


Conni2461:
Doesn't `git_status`, `<CR>` also open files? Doesn't work if we have a new dir but i think everything else should just work?

lua/telescope/builtin/git.lua
```diff
@@ -137,6 +137,26 @@ git.status = function(opts)
   }):find()
 end
 
+git.recents = function(opts)
+  local depth = utils.get_default(opts.depth, 5)
+
+  if opts.cwd then
+    opts.cwd = vim.fn.expand(opts.cwd)
+  end
```
Conni2461, line 145: You don't need to handle `cwd`. When exporting the function we call `apply_checks` which will handle `cwd`

tami5:
It's better to provide an option for git files to sort based on recent changes. New pickers that does little difference is pointless to maintain. Just look at git files picker see if you can provide an option to sort results based on last updated. btw it makes sense to have this feature, thanks.

kyoh86:
Thank you for your advices!

@Conni2461
I often use it (in FZF), but I cannot believe that is major needs.
So if you reject this PR, I'll create new extension for just me.

@tami5 
I'm gonna try it.

kyoh86:
@tami5 

I had should say that `--depth` option is the very thing I want.

Using `git-ls-files(1)`, `git_files` picker will be able to just pick files with "current" status.
And maybe `git_files` with that option will behave looks like a `git_status` picker.

Conni2461:
I think as opt doesn't make sense because of the use of two different functions.

> I often use it (in FZF), but I cannot believe that is major needs.

Does fzf have it builtin? Because when i implemented git i thought that we wanted to support basically the same things as fzf does. So people can switch easily. And not provide too much additional stuff only cool things like (lsp, treesitter) :laughing: 

We can also put stuff like this into the wiki. fzf does the same: https://github.com/junegunn/fzf/wiki/Examples-(vim)

elianiva:
Telescope's wiki is lonely and sad, it needs more contents :laughing:

kyoh86:
I found that I just lied.
This is not like fzf feature but my custom function.

I withdraw this PR once.
Finding good interface, I'll create new PR again.

PR #515, opened 2021-02-08
Do not use C binding in Windows

### LINKED ISSUE #513: Displayer broken on Windows
Reported by @sp1n0r on gitter. Reproducible.

```
cannot resolve symbol 'linetabsize_col'
```
And if that one is ignored:
```
cannot resolve symbol 'utf_ptr2len'
```

Possible hotfix: 
```diff
diff --git a/lua/telescope/utils.lua b/lua/telescope/utils.lua
index 9cfc23c..188c1ea 100644
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -203,7 +203,7 @@ function utils.get_os_command_output(cmd)
 end

 utils.strdisplaywidth = (function()
-  if jit then
+  if jit and pathlib.separator ~= '\\' then
     local ffi = require('ffi')
     ffi.cdef[[
       typedef unsigned char char_u;
@@ -224,7 +224,7 @@ utils.strdisplaywidth = (function()
 end)()

 utils.utf_ptr2len = (function()
-  if jit then
+  if jit and pathlib.separator ~= '\\' then
     local ffi = require('ffi')
     ffi.cdef[[
       typedef unsigned char char_u;
```

Other possible solution from @tjdevries:
> We can just expose the functions in the api and mark it as fast

CC @delphinus

delphinus:
Neovim in Windows cannot detect C funcs that are called here.

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -203,7 +203,7 @@ function utils.get_os_command_output(cmd)
 end
 
 utils.strdisplaywidth = (function()
-  if jit then
+  if jit and pathlib.separator ~= '\\' then
     local ffi = require('ffi')
     ffi.cdef[[
       typedef unsigned char char_u;
@@ -224,7 +224,7 @@ utils.strdisplaywidth = (function()
 end)()
 
 utils.utf_ptr2len = (function()
-  if jit then
+  if jit and pathlib.separator ~= '\\' then
     local ffi = require('ffi')
     ffi.cdef[[
       typedef unsigned char char_u;


Conni2461:
I don't like this fix at all. If i would i would have opened a PR myself :rofl: so can we not close the issue? 
I don't like that displayer will not truncate correctly on windows but for now this is good as hotfix.
Just remove the fix part in your post after that you squash and merge. Message like: `fix: ...`

Thanks :)

delphinus:
Oh, I also think that this is only stopgap measure. I'll fix comments.

Conni2461:
> Oh, I also think that this is only stopgap measure. I'll fix comments.

I don't get what you are saying. Is this mergeable or do you have a different idea?

delphinus:
I have the same thought. Now I don't know the final logic to slove this, but I just thought this fix should be merged for Windows users although it is not very good.

Issue #516, opened 2021-02-08
`git_commits`, `git_bcommits`, and `git_status` do not respect the `cwd` option


mroavi:
Note: I'm new to Lua and I couldn't move all the different configuration pieces into one file.

### Description

`git_commits`, `git_bcommits`, and `git_status` do not respect the `cwd` option.

**Expected Behavior**

Calling any of the functions above when passing the current buffer as argument to the `cwd`
option should run against the repo found relative to that path. 

**Actual Behavior**

Calling any of the functions above ignores the passed argument to the `cwd` option and runs
against the repo found relative to vim's current working directory.

NOTE: `git_files` does work as expected.

### Details

Steps to reproduce:

In `.config/nvim/lua/mrv/telescope/init.lua`

```lua
local M = {}

-- Works as expected!
function M.git_files()
  require("telescope.builtin").git_files{
    cwd = vim.fn.expand("%:p:h"),
  }
end

-- TODO: cwd is ignored
function M.git_commits()
  require("telescope.builtin").git_commits{
    cwd = vim.fn.expand("%:p:h"),
  }
end

-- TODO: cwd is ignored
function M.git_bcommits()
  require("telescope.builtin").git_bcommits{
    cwd = vim.fn.expand("%:p:h"),
  }
end

-- TODO: cwd is ignored
function M.git_status()
  require("telescope.builtin").git_status{
    cwd = vim.fn.expand("%:p:h"),
  }
end

return M
```

In `.config/nvim/lua/mrv/telescope/mappings.lua`:

```lua
local vimp = require 'vimp'
vimp.nnoremap('<Leader>fg', require("mrv.telescope").git_files)
vimp.nnoremap('<Leader>co', require("mrv.telescope").git_commits)
vimp.nnoremap('<Leader>bc', require("mrv.telescope").git_bcommits)
vimp.nnoremap('<Leader>st', require("mrv.telescope").git_status)
```

Steps to reproduce:

1. Start nvim and `cd` into a directory with a git repo.
2. Open a file buffer that's inside *another* repo.
3. Type `<Leader>co`, `<Leader>bc`, and `<Leader>st`.

Result:
The git commands take effect in nvim's current working directory and not in
in the directory of the current buffer.

Conni2461:
Thanks for the report :)
Yeah i realized that myself and have fix that in #290 just because i realized that when i was working on it. If you want i can make a separate PR for it :)

I should probably do that. File browser might take some more time.

[...]

PR #517, opened 2021-02-08
fix: all git builtin respect cwd now

Conni2461:
Fix https://github.com/nvim-telescope/telescope.nvim/issues/516

Please don't be mad if i don't merge that today :) I am planing to go to sleep soon. But would be great if you can test it and report back if some things don't work. I already found issues i haven't considered in my previous attempt to fix that so i might have missed to pass in one cwd :) I am currently not sure about it tho :rofl: 

CC @mroavi


mroavi:
It's working like a charm. The three functions mentioned in the issue are all fixed!

PR #521, opened 2021-02-09
Cycle prompt history

### LINKED ISSUE #520: Recent git_files and find_files history like FZF for Vim
On FZF for Vim there is a history with all last inputs easily access via <c-p> <c-n>.

I would like to see something like that here.

No alternative found until now.

**Additional context**
mac os x 11.1 big sur
nvim version v0.5.0-dev+1062-gcc1851c9f
tmux 3.1c
zsh 5.8 (x86_64-apple-darwin20.0)

Conni2461:
Close #520

Showcase: https://streamable.com/w3wpri

Enabled on default you only need to add mappings

```lua
require('telescope').setup {
  defaults = {
    mappings = {
      i = {
        ["<C-Down>"] = require('telescope.actions').cycle_history_next,
        ["<C-Up>"] = require('telescope.actions').cycle_history_prev,
      },
    },
  }
}
```

For more information `:help telescope.defaults.history`

CC @saragiotto

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -47,46 +47,38 @@ telescope.setup({opts})                                    *telescope.setup()*
 
         Default: '  '
 
-                                       *telescope.defaults.history_handler*
-    history_handler: ~
-        This is a developer setting and is meant for an extension.
+                                               *telescope.defaults.history*
+    history: ~
+          This field handles the configuration for prompt history. It can either
+          be a table or false.
+          If its false then history is disabled.
 
-        history_handler is a function that defines the behavior of the history.
-        Telescope core ships a rather simple implementation of it, with a unified
-        history across all pickers. Its not sensitive for cwd or picker.
-
-        We also offer an extension (nvim-telescope/telescope-smart-history.nvim)
-        which uses a sqlite3 database to store prompt into with the corresponding
-        metadata, like picker and cwd. This means cycling to previous entry will
-        return an entry that was actually used in this cwd. This is helpful if you
-        work on multiple projects at the same time.
-
-        It can be disabled the same way, by setting history_location to nil.
-
-        Expects function
-        Default: require('telescope.actions.history').get_simple_history
-
-                                         *telescope.defaults.history_limit*
-    history_limit: ~
-        You can set a limit on how many history elements will be stored.
-        Set to nil to disable it otherwise set to a number.
-
-        Default: nil
-
-                                      *telescope.defaults.history_location*
-    history_location: ~
-        This setting will enable result history. You can scroll though your
-        previous inputs. The mappings to do so are currently not mapped.
-        You can add them like this:
-
-          mappings = {
-            i = {
-              ["<C-Down>"] = actions.cycle_history_next,
-              ["<C-Up>"] = actions.cycle_history_prev,
+          Currently mappings still need to be added, Example:
+            mappings = {
+              i = {
+                ["<C-Down>"] = require('telescope.actions').cycle_history_next,
+                ["<C-Up>"] = require('telescope.actions').cycle_history_prev,
+              },
             },
-          },
 
-        Default: nil
+          Fields:
+            - path:    the path to the telescope history as string
+                       default: stdpath("data")/telescope_history
+            - limit:   the amount of entries that will be written in the history
+                       if limit is set to nil it will grown unbound
+                       default: 100
+            - handler: is a developer setting that is meant for extensions.
+                       You can replace the actual implementation of the history
+                       and provide your own. One example of this is
+                       nvim-telescope/telescope-smart-history.nvim which allows
+                       context sensitive (cwd + picker) history.
+
+                       Default:
+                       require('telescope.actions.history').get_simple_history
+
+                       Is a simple history implementation that is not context
+                       sensitive, it writes all history in one text file.
+
 
                                          *telescope.defaults.layout_config*
     layout_config: ~
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -41,12 +41,52 @@ telescope.setup({opts})                                    *telescope.setup()*
 
         Default: false
 
-
                                           *telescope.defaults.entry_prefix*
     entry_prefix: ~
         Prefix in front of each result entry. Current selection not included.
 
         Default: '  '
+
+                                       *telescope.defaults.history_handler*
+    history_handler: ~
+        This is a developer setting and is meant for an extension.
+
+        history_handler is a function that defines the behavior of the history.
+        Telescope core ships a rather simple implementation of it, with a unified
+        history across all pickers. Its not sensitive for cwd or picker.
+
+        We also offer an extension (nvim-telescope/telescope-smart-history.nvim)
+        which uses a sqlite3 database to store prompt into with the corresponding
+        metadata, like picker and cwd. This means cycling to previous entry will
+        return an entry that was actually used in this cwd. This is helpful if you
+        work on multiple projects at the same time.
+
+        It can be disabled the same way, by setting history_location to nil.
+
+        Expects function
+        Default: require('telescope.actions.history').get_simple_history
+
+                                         *telescope.defaults.history_limit*
+    history_limit: ~
+        You can set a limit on how many history elements will be stored.
+        Set to nil to disable it otherwise set to a number.
+
+        Default: nil
+
+                                      *telescope.defaults.history_location*
+    history_location: ~
+        This setting will enable result history. You can scroll though your
+        previous inputs. The mappings to do so are currently not mapped.
+        You can add them like this:
+
+          mappings = {
+            i = {
+              ["<C-Down>"] = actions.cycle_history_next,
+              ["<C-Up>"] = actions.cycle_history_prev,
+            },
+          },
+
+        Default: nil
 
                                          *telescope.defaults.layout_config*
     layout_config: ~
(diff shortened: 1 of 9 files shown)


saragiotto:
Maybe I'm doing something wrong or it's related with the PR pointed out on the description.

Anyway, after the installation and configs setup, I'm facing the above error when trying to open files:

```
E5108: Error executing lua ...im/plugged/telescope.nvim/lua/telescope/actions/init.lua:44: attempt to call method 'write_text' (a nil value)
```

@Conni2461 did you've any clue on that?

Conni2461:
Oh yeah i have an update for that that i didn't push because i wasn't done yet :laughing: I will just push the fix for that now :)

Conni2461:
Another new version that should work better now. History gets loaded only once on first time. I still need to clean that up in a separate module and add a `history_limit`.

Shatur95:
@Conni2461, what is the status of it?

Conni2461:
Status is that i was super busy for the past few weeks but this will change next Wednesday. Then i will finish it. Shouldn't take that long :)

lua/telescope/pickers.lua
```diff
@@ -516,6 +516,24 @@ function Picker:hide_preview()
   -- 2. Resize prompt & results windows accordingly
 end
 
+function Picker:reset_prompt()
```
elianiva, line 519: this function has a duplicate. https://github.com/nvim-telescope/telescope.nvim/blob/bfd384fe39e6c79c9530cffcaec4d043f488790c/lua/telescope/pickers.lua#L664

Conni2461: Good catch. Thanks :)

This one is obsolete. I basically already put it in master because sunjon needed it for arecibo and forgot about it when rebasing :laughing:

Conni2461: Did you try it @elianiva? Does everything work as expected?

elianiva: yes, everything works as expected. GJ! :)

tami5:
@Conni2461 awesome pr, it's really an important missing feature, @saragiotto great suggestion

>  https://github.com/nvim-telescope/telescope.nvim/issues/520#issuecomment-776187202 It keeps in just one big file, mine gets 1000 lines, I think that is the maximum. But definitely It's not based on `cwd/project`.

I kinda disagree, I mean I can understand why it shouldn't be unique or should be unified for your use-case as I presume that you intend to use it only for the very last input in case you accidentally closed the picker, or when you open several files in order and want to easily switch between them. But what about other use-cases? for example 

- Imagine you opened a command history or highlight or builtins picker, would like to see unrelated input history?

- You're work on multiple project, in multiple tabs, one of them is rust and the other is cargo, would like prev to show say; `cargo.rs` instead of `setup.py`?

Thus, I strongly dismiss the idea of a unified history for ALL pickers, One of the core features of telescope is that every Picker has it's own world (preview, theme, layout, mappings .. etc) and history shouldn't be treated differently. History, I believe, should be unique PER `picker` and when the picker is a variant of  file/grep finder it should include PRE `cwd` or `search_dirs`. Perhaps pickers such as buffer picker might be better skipped and only saved in local variable since it's unique to the vim instance.

Conni2461:
Idea:
1. A simple, unique history here (as it is right now) with the ability to override it (similar to sorters)
2. A complex context sensitive history as extension utilizing `sql.nvim`. Metadata is `picker`, `cwd` (for now) because it would be kinda annoying to do this in plain text and maybe `\t` separators.
And with that extension idea i can make everyone happy :)

Thoughts @Shatur95 @saragiotto @elianiva @tami5 

I would start working on this Saturday. I wanna get this done as soon as possible. :)

elianiva:
Seems like a good idea to me, though I'm not sure what "context sensitive" history is :laughing:

Conni2461:
Basically what tami said. A history for each picker (i just thought context sensitive made sense, maybe not).
So a history for `find_files + cwd` and another one for `grep_string + that_cwd`. and another one for `grep_string + other_cwd`, etc

elianiva:
Right, that would be nice.

lua/telescope/actions/history.lua
```diff
@@ -0,0 +1,218 @@
...
+  if not conf.history_location then
+    obj.enabled = false
+    return setmetatable(obj, self)
+  end
+  if conf.history_limit then
+    obj.limit = conf.history_limit
+  end
+
+  obj.path = vim.fn.expand(conf.history_location)
+  obj.db = sql:open(obj.path)
```
tami5, line 133: I suggest using `sql.new` here better.

lua/telescope/actions/history.lua
```diff
@@ -0,0 +1,218 @@
...
+  -- any chance i can get a table object here?!?
+  -- because using obj.db:table to create one feels weird because i can't pass in
+  -- a schema at the same time. WTF tami am i just stupid.
+  -- I think the readme is confusing LULW
+  obj.db:create("histories", {
```
tami5, line 138:
```
obj.data = db:table("history",{nocache = true, schema = {..., ensure = true, id=true}
```

This a better approach so that later we can do obj.data:insert(), obj.data:get() .. etc and with sql.new we don't need to worry about opening or closing connections.

id=true is as same as how you defined the id above. nocache=true because history will be always updated on every run.

lua/telescope/actions/history.lua
```diff
@@ -0,0 +1,218 @@
...
+
+  if line ~= '' then
+    self:__ensure_current('find_files', vim.loop.cwd())
+    if self.current[#self.current].content ~= line then
+      -- TODO(conni2461): HANDLE LIMIT. REMOVE first n with id
+      self.db:insert("histories", {
+        content = line,
+        picker = "find_files",
+        cwd = vim.loop.cwd()
+      })
```
tami5, line 172: with using table module this becomes:

```lua
self.data:insert { ... } 
```

btw if you need to append json, just define the field type to be json and sql.nvim will serialize and deserialize lua table automatically. might fail though on functions not sure

lua/telescope/actions/history.lua
```diff
@@ -0,0 +1,218 @@
...
+
+-- TODO(conni2461): Serialize metadata
+-- TODO(conni2461): CURRENTLY HARDCODED picker and cwd
+local SmartHistory = {}
+SmartHistory.__index = SmartHistory
+
+function SmartHistory:new()
+  local obj = {
+    enabled = true
+  }
```
tami5, line 123: you can make the obj to = `(sql.new(path)):table('histories', {})`

this way we can use self:insert, self:get ... etc, u can also add additional methods or overwrite existing methods id you like . see https://github.com/nvim-telescope/telescope-cheat.nvim/blob/master/lua/telescope/_extensions/cheat/db.lua

lua/telescope/actions/history.lua
Conni2461: That is exactly waht i wanted to do but when using `db:table("history", { schema = { ... } })` it wasn't added :thinking: Maybe i had a typo somewhere. i try it again

lua/telescope/actions/history.lua
Conni2461: This enable is for the history feature. No path not enabled.

tami5: oh you can add it like
`obj.smart_enabled ` or enabled to the table object itself. The idea here is to reuse sql table object without need to prefix it as data, self = smarthistory. So in other functions you can do self: to access both your object and sql table object methods and variables

Conni2461: I think you don't get the whole architecture. I don't get what you want from me :laughing: You can still access the db with this, but you shouldn't because you don't know whats behind the table. It could be DBHistory or it could be SimpleHistory. So i kinda want to keep it the same. So if someone wants to do something outside of History, he can get it with:

```lua
local history = action_state.get_current_history()
history.enabled -- will now show if we have a history or not, works for both
history.content -- get content, need to rename the object my bad
history:get_next() -- 
```

tami5: Oh I see what you mean. Yah merging the two is bad idea in case of switching things around with something that implement get/insert method. Good call

Conni2461:
If someone wants to try `smart_history` (will be easier to setup with the extension), these are the config values you should set:
```lua
history_handler = require('telescope.actions.history').get_smart_history,
history_location = '~/.local/share/nvim/databases/telescope_history.sqlite3',
```

Conni2461:
Would be good if someone could review the comments which generate the documentation.
Also i would need a good name for the extension. My propose was `telescope-smart-history.nvim` but i am open for suggestions.

Then i can finish it :)

doc/telescope.txt
```diff
@@ -25,6 +25,44 @@ telescope.setup({opts})                                    *telescope.setup()*
 
         Default: '  '
 
+                                       *telescope.defaults.history_handler*
+    history_handler: ~
+        This is a developer setting and is meant for an extension.
+
+        It defines the behavior of the history. Telescope core ships are
+        rather simple implementation of it, with a unified history across
+        all pickers. Its not sensitive for cwd.
```
tami5, line 34: `history_handler` defines ...

The default is simple ...

doc/telescope.txt
```diff
@@ -25,6 +25,44 @@ telescope.setup({opts})                                    *telescope.setup()*
...
+
+        We also offer an extension which uses sqlite3 bindings (tami5/sql.nvim)
```
tami5, line 36: mention as a more advance implementation history_handler, + link to repo

tami5 (approved):
Fot the name maybe just telescope-history.nvim

lua/telescope/actions/set.lua
```diff
@@ -44,9 +44,18 @@ end
...
-action_set.select = function(prompt_bufnr, type)
-  return action_set.edit(prompt_bufnr, action_state.select_key_to_edit_key(type))
-end
+action_set.select = {
+  -- Will not be called if `select_default` is replaced rather than `action_set.select` because we never get here
+  pre = function(prompt_bufnr)
+    action_state.get_current_history():append(
+      action_state.get_current_line(),
+      action_state.get_current_picker(prompt_bufnr)
+    )
```
Conni2461, line 53: Pre will not be called when `action_select` is replaced because `action_set.select` will never happen

sandangel:
hi, may I ask for an update on this PR?

Conni2461:
I just rebased it. Sadly the action api is lacking, so history will not be saved in all pickers. Missing are the once that do `actions.select_*:replace`. Examples are the git once and man_pages/help_tags (and probably more, looking at extensions). I mentioned this here: https://github.com/nvim-telescope/telescope.nvim/pull/521#discussion_r622264342 and tj wanted to fix the action API here https://github.com/nvim-telescope/telescope.nvim/pull/807 but he hasn't gotten around to finish it.

I could implement a workaround for that but i haven't asked tj if that is the best idea. I will do that when he comes back from his vacation.

But there is no reason why you couldn't check out the branch and run it. :) I would appreciate some feedback ;)

3nuc:
I feel as if actions.cycle_history_prev/next should throw an error to :messages if the user hasn't set history_location. If you bind the actions (C-p, C-n) but forget to set history_location, pressing the bound keys does nothing

RRethy:
> I feel as if actions.cycle_history_prev/next should throw an error to :messages if the user hasn't set history_location. If you bind the actions (C-p, C-n) but forget to set history_location, pressing the bound keys does nothing (which is obviously expected behavior)

Why not default to somewhere in `$XDG_DATA_HOME`? The less required configuration the better.

Shatur:
> Why not default to somewhere in $XDG_DATA_HOME

There is `vim.fn.stdpath`. I would use `data` dir for it.

Conni2461:
AFAIK its not enabled on default for fzf. Enabling it on default also requires mapping `cycle_history_prev/next` and this is a nightmare because we can't break backwards compatibility, `<C-p>` or `<C-n>` would. Its a way to opinionated topic that i don't want to make any decision here. We currently already have mappings that are not ideal in insert mode (`<C-u>`) and we should learn for that "mistake".

I checked and it seems that i am currently suggesting `<C-Up>` `<C-Down>` in the documentation which is (IMO) also not ideal but i think tj suggested them.

But i don't see a reason to print a error message if history isn't enabled and the actions are mapped and/or used

doc/telescope.txt
```diff
@@ -47,46 +47,38 @@ telescope.setup({opts})                                    *telescope.setup()*
...
-        You can add them like this:
-
-          mappings = {
-            i = {
-              ["<C-Down>"] = actions.cycle_history_next,
-              ["<C-Up>"] = actions.cycle_history_prev,
+                                               *telescope.defaults.history*
+    history: ~
+          This field handles the configuration for prompt history. It can either
+          be a table or false.
```
clason, line 53: `false` or `nil`? What is the default?

doc/telescope.txt
```diff
@@ -47,46 +47,38 @@ telescope.setup({opts})                                    *telescope.setup()*
...
+          If its false then history is disabled.
```
clason, line 54: typo: `it's`

doc/telescope.txt
```diff
@@ -47,46 +47,38 @@ telescope.setup({opts})                                    *telescope.setup()*
...
+                ["<C-Up>"] = require('telescope.actions').cycle_history_prev,
+              },
             },
-          },
 
-        Default: nil
+          Fields:
+            - path:    the path to the telescope history as string
+                       default: stdpath("data")/telescope_history
+            - limit:   the amount of entries that will be written in the history
```
clason, line 67: add semicolon after history (to indicate that the next line is a new sentence)

doc/telescope.txt
```diff
@@ -47,46 +47,38 @@ telescope.setup({opts})                                    *telescope.setup()*
...
+                       if limit is set to nil it will grown unbound
```
clason, line 68: "warning: if limit is set to nil, then history will grow unbounded"

doc/telescope.txt
Conni2461: What about?

    This field handles the configuration for prompt history.
    On default its a table, with default values (more below).
    To disable history, set it to false.

doc/telescope.txt
```diff
@@ -47,46 +47,38 @@ telescope.setup({opts})                                    *telescope.setup()*
...
+                       default: 100
+            - handler: is a developer setting that is meant for extensions.
+                       You can replace the actual implementation of the history
+                       and provide your own. One example of this is
+                       nvim-telescope/telescope-smart-history.nvim which allows
+                       context sensitive (cwd + picker) history.
```
clason, line 74: "handler": a Lua function that implements the history.
This is meant as a developer setting for extensions to override the history handling,
e.g., https://github.com/nvim-telescope/telescope-smart-history.nvim, which allows
context sensitive (cwd + picker) history.

doc/telescope.txt
```diff
@@ -47,46 +47,38 @@ telescope.setup({opts})                                    *telescope.setup()*
...
+
+                       Default:
+                       require('telescope.actions.history').get_simple_history
+
+                       Is a simple history implementation that is not context
+                       sensitive, it writes all history in one text file.
```
clason, line 80: can be removed?

Conni2461: Might be duplicate

doc/telescope.txt
clason: "By default it is", but otherwise 👍🏻 

(But why `false` and not `nil`? From my Lua experience so far, I would have expected the latter.)

(Not that it makes much difference since `nil` is falsy.)

doc/telescope.txt
Conni2461: removed

doc/telescope.txt
Conni2461: Yeah because `previewer = false`, etc but i can make sure that both works :)

clason: OK, if it's consistent over the telescope config, then that's fine!

PR #531, opened 2021-02-13
Find editing window in goto file selection action

relnod:
In `_goto_file_selection` it now checks if the current window contains a file buffer. If not it tries to find a window, that does and opens the file/buffer there.

This fixes opening the file/buffer in a terminal window or the NerdTree buffer, when telescoped was called from that buffer.

I could also hide this feature behind an option.
Also I'm not sure, if the heuristic behind `is_editing_win` is completely right...

What do you think? Feedback apreciated!

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -66,6 +66,29 @@ function actions.preview_scrolling_down(prompt_bufnr)
   actions.get_current_picker(prompt_bufnr).previewer:scroll_fn(30)
 end
 
+--- Checks if the window contains a normal file buffer.
+---
+--- @param winnr number The window id
+--- @returns boolean True if the window contains a file buffer
+local function is_editing_win(winnr)
+  local bufnr = vim.api.nvim_win_get_buf(winnr)
+
+  local buftype = vim.api.nvim_buf_get_option(bufnr, "buftype")
+  return buftype == ""
+end
+
+--- Tries to find a window, that contains a file buffer.
+---
+--- @returns number The window number, -1 if no widow was found
+local function get_editing_win()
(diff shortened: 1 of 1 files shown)


weilbith:
Would be cool if it could be eventually customized to use other plugins like [vim-choosewin](https://github.com/t9md/vim-choosewin). Maybe just having such a feature in mind while implementing this. :shrug:

Conni2461:
I am not so sure about this one. I kinda find it unintuitive to not always be the current window.

I would like to have more opinions on that :)

relnod:
How about an option `window_selection`, that returns a window number, where the default implementation would be to just return the current window. Then It would be possible to use plugins like vim-choosewin, or a custom implementation.

Conni2461:
We could talk about a window_selection fn that can be overridden. Good idea can we postpone this after tjs big actions improvements PR. I will try to wrap up that PR tomorrow :)

Conni2461:
This can now be achieved. e.g. with this config (written in github so might have a syntax error). Sorry for the long wait

```lua
local telescope = require "telescope"

local function is_editing_win(winnr)
  local bufnr = vim.api.nvim_win_get_buf(winnr)

  local buftype = vim.api.nvim_buf_get_option(bufnr, "buftype")
  return buftype == ""
end

local function get_editing_win()
  for _, win in ipairs(vim.api.nvim_list_wins()) do
    if is_editing_win(win) then
      return win
    end
  end
  return 0
end

telescope.setup {
  defaults = {
    get_selection_window = function(_picker, _entry)
      if not is_editing_win(vim.api.nvim_get_current_win()) then
        return get_editing_win()
      end
      return 0
    end
  },
}
```

PR #539, opened 2021-02-21
Update Extensions README section

bi0ha2ard:
Thanks for the cool plugin!

I've created a picker for ROS and ROS2 workspace packages. Maybe someone finds it useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -537,14 +537,24 @@ Telescope provides the capabilties to create & register extensions, which improv
 Some extensions provide integration with external tools, outside of the scope of `builtins`. Others provide performance
 enhancements by using compiled C and interfacing directly with Lua.
 
-### Community Extensions
+### Accessing pickers from extensions
 
-Extensions can be refenced by doing the following:
+Pickers from extensions are added to the `:Telescope` command under their respective name.
+For example:
+
+```viml
+" Run the `configurations` picker from nvim-dap
+:Telescope dap configurations
+```
+
+They can also be called directly from lua:
 
 ```lua
--- Run the `configurations` picker from nvim-dap (not yet implemented)
+-- Run the `configurations` picker from nvim-dap
 require('telescope').extensions.dap.configurations()
 ```
+
+### Pre-loading
 
 To pre-load an extension (so that it will override default configurations), you can do:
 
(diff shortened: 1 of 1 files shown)


Conni2461:
Thanks for using telescope, the kind words and for developing a extension :)

Now tbh i hate that section and i hate to be that guy that is complaining again :rofl:. It was cool back when we had like 2 extensions but now we are getting a PR for a new extension every other week. Don't get me wrong its super important that we display these extensions in some kind of way but i don't like them in the README anymore. We already have a section in the wiki that is a 1:1 copy see https://github.com/nvim-telescope/telescope.nvim/wiki/Extensions

So here is what i am thinking we could do. You can rewrite this section (don't wanna take away your PR) to something that points to that wiki page. Really obvious so that everybody has to see it :laughing: and we make that page editable for all people.

We can also redesign the wiki page a little bit:
| Extension       | Description                |
|-----------------|----------------------------|
| fzy native link | Some elaborate description |

People can also get a bit more elaborate with what their extension tries to achieve. 

This isn't probably not the correct place to discuss that so sorry about that. Also i am looking for more opinions from members :D

bi0ha2ard:
Thanks for your feedback.

I've added the extension to the list in the wiki (without touching the existing formatting) and replaced the community extension section in the README with a link to the page. I hope it's OK like this, otherwise, feel free to close the PR.

Conni2461:
Thanks :) I'll keep it open for a day and wait for more feedback after that we can merge.

Could you also remove the `(not yet implemented)` and add the viml example of running extensions.
```viml
:Telescope dap configurations
```

bi0ha2ard:
Done!

I've split it up a bit, but I wasn't 100% sure what to do with the "Pre-loading" example, so I just put that in its own subsection.

Conni2461:
Maybe we should reformulate that to something like: "To load an extension do this" 

We still have lazy loading but with that `:Telescope` can't interpret the commands, so people should just always load it

Conni2461:
Thats really really good thanks :)

bi0ha2ard:
Ok that was quick, unfortunately I missed a typo, see #542

Conni2461:
:laughing: Yeah i am currently going through some PR. Most of them are mine :laughing: Thanks

Issue #481, opened 2021-01-31
Telescope find_files resetting indentexpr set by treesitter


yzia2000:
I want to use treesitter indentations. However, when I open a file using `:Telescope find_files` or a grep command, it resets the indentexpr to a vim runtime default. For example when opening a javascript file this way, the `indentexpr=GetJavascriptIndent()` from $runtime/indents/javascript.vim instead of `indentexpr=nvim_treesitter#indent()`. This does not occur when I open the files directly through `nvim *.js` or `:e path/to/*.js`.
Please assist me.

elianiva:
You could try to use this

```viml
au Filetype * set indentexpr=nvim_treesitter#indent()
```

I use that because there are other reasons that resets the `indentexpr` other than Telescope (which I'm not entirely sure, this solution works for me so I couldn't be bothered to debug it :laughing:)

yzia2000:
Haha. I have been trying to do that too but it seems more like a bypass than a fix and my inner OCD gets the best of me. I was hoping that someone could explain to me how indentations work in Telescope to see if I could fix this myself.

elianiva:
I think it's because how `indentexpr` being a local buffer variable. It gets overriden when you open any file using `:edit` or `:buffer`. It's strange that you don't have this issue when you do `:e filename` manually because I do.

yzia2000:
> I think it's because how `indentexpr` being a local buffer variable. It gets overriden when you open any file using `:edit` or `:buffer`. It's strange that you don't have this issue when you do `:e filename` manually because I do.

Do you perhaps have `filetype syntax indent on` in your init.vim?

elianiva:
> > I think it's because how `indentexpr` being a local buffer variable. It gets overriden when you open any file using `:edit` or `:buffer`. It's strange that you don't have this issue when you do `:e filename` manually because I do.
> 
> Do you perhaps have `filetype syntax indent on` in your init.vim?

Nope, I don't have that.

yzia2000:
I see. That is surprising because I only encounter the issue when using telescope haha

Conni2461:
Wild guess

https://github.com/nvim-telescope/telescope.nvim/blob/fc54211868fcca8e53fec7071b3fab3591d250da/lua/telescope/actions/init.lua#L139

Removing that line. Does that fix the issue? We just need that so filetype autocommands run.
So stuff like

```
au Filetype cpp setlocal ...
```

works

elianiva:
[...]

PR #541, opened 2021-02-23
fix: goto_file_selection performance issues, do not reload buffer if it is already loaded

### LINKED ISSUE #476: lsp_document_symbols are not working when buffer is not saved
### Description
lsp_document_symbols are not working when buffer is not saved, when I open lsp_document_symbols preview window is opening, but the when I want to jump to some symbol it doesn't work unless it is a saved buffer.
But when I save the file or buffer it works as expected

**Expected Behavior**
should jump to the symbol

**Actual Behavior**
E5108: Error executing lua ...cker/start/telescope.nvim/lua/telescope/actions/init.lua:127: Vim(edit):E37: No write since last change (add ! to override)
Press ENTER or type command to continue

### Details

1. nvim -nu test.vim
2. :Telescope lsp_document_symbols
3. and here you'll get the error like this
4. E5108: Error executing lua ...cker/start/telescope.nvim/lua/telescope/actions/init.lua:127: Vim(edit):E37: No write since last change (add ! to override)
Press ENTER or type command to continue
</details>

- nvim --version output: NVIM v0.5.0-dev+nightly
Build type: Release
LuaJIT 2.1.0-beta3
- Operating system: Mac OS
- Telescope commit: 9d4670c

</details>

<p>

```viml
" Find files using Telescope command-line sugar.
nnoremap <leader>ff <cmd>Telescope find_files<cr>
nnoremap <leader>fg <cmd>Telescope live_grep<cr>
nnoremap <leader>fb <cmd>Telescope buffers<cr>
nnoremap <leader>fh <cmd>Telescope help_tags<cr>

" only i use this much from telescope plugin
```

oberblastmeister:
Fix #476 

Fixes:
- `goto_file_selection` general performance issues from `doautocmd` and unneeded api calls
- bug when entering the same buffer telescope will reload the enter buffer, making lsp restart and adding to the extreme slowness
- for example, when running `lsp.document_symbols` it will always enter the same buffer as was started in, just set the cursor to a different location. However, telescope reloads the entire buffer restarting lsp when it should just change to that location.

### DIFF
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -50,71 +50,65 @@ set.edit = function(prompt_bufnr, command)
   if not entry then
     print("[telescope] Nothing currently selected")
     return
+  end
+
+  local filename, row, col
+
+  if entry.filename then
+    filename = entry.path or entry.filename
+
+    -- TODO: Check for off-by-one
+    row = entry.row or entry.lnum
+    col = entry.col
+  elseif not entry.bufnr then
+    -- TODO: Might want to remove this and force people
+    -- to put stuff into `filename`
+    local value = entry.value
+    if not value then
+      print("Could not do anything with blank line...")
+      return
+    end
+
+    if type(value) == "table" then
+      value = entry.display
+    end
+
+    local sections = vim.split(value, ":")
+
+    filename = sections[1]
+    row = tonumber(sections[2])
+    col = tonumber(sections[3])
+  end
+
+  local entry_bufnr = entry.bufnr
+
+  require('telescope.actions').close(prompt_bufnr)
+
+  if entry_bufnr then
+    if command == 'edit' then
+      vim.cmd(string.format(":buffer %d", entry_bufnr))
+    elseif command == 'new' then
+      vim.cmd(string.format(":sbuffer %d", entry_bufnr))
+    elseif command == 'vnew' then
+      vim.cmd(string.format(":vert sbuffer %d", entry_bufnr))
+    elseif command == 'tabedit' then
+      vim.cmd(string.format(":tab sb %d", entry_bufnr))
+    end
   else
-    local filename, row, col
-    if entry.filename then
-      filename = entry.path or entry.filename
+    filename = path.normalize(vim.fn.fnameescape(filename), vim.fn.getcwd())
 
-      -- TODO: Check for off-by-one
-      row = entry.row or entry.lnum
-      col = entry.col
-    elseif not entry.bufnr then
-      -- TODO: Might want to remove this and force people
-      -- to put stuff into `filename`
-      local value = entry.value
-      if not value then
-        print("Could not do anything with blank line...")
-        return
-      end
-
-      if type(value) == "table" then
-        value = entry.display
-      end
-
-      local sections = vim.split(value, ":")
-
-      filename = sections[1]
-      row = tonumber(sections[2])
-      col = tonumber(sections[3])
+    -- check if we didn't pick a different buffer
+    -- prevents restarting lsp server
+    if vim.api.nvim_get_current_buf() ~= vim.fn.bufnr(filename) then
+      vim.cmd(string.format("%s %s", command, filename))
     end
 
-    local preview_win = state.get_status(prompt_bufnr).preview_win
-    if preview_win then
-      a.nvim_win_set_config(preview_win, {style = ''})
-    end
-
-    local entry_bufnr = entry.bufnr
-
-    require('telescope.actions').close(prompt_bufnr)
-
-    if entry_bufnr then
-      if command == 'edit' then
-        vim.cmd(string.format(":buffer %d", entry_bufnr))
-      elseif command == 'new' then
-        vim.cmd(string.format(":sbuffer %d", entry_bufnr))
-      elseif command == 'vnew' then
-        vim.cmd(string.format(":vert sbuffer %d", entry_bufnr))
-      elseif command == 'tabedit' then
-        vim.cmd(string.format(":tab sb %d", entry_bufnr))
-      end
-    else
-      filename = path.normalize(vim.fn.fnameescape(filename), vim.fn.getcwd())
-
-      local bufnr = vim.api.nvim_get_current_buf()
-      if filename ~= vim.api.nvim_buf_get_name(bufnr) then
-        vim.cmd(string.format(":%s %s", command, filename))
-        bufnr = vim.api.nvim_get_current_buf()
-        a.nvim_buf_set_option(bufnr, "buflisted", true)
-      end
-
-      if row and col then
-        local ok, err_msg = pcall(a.nvim_win_set_cursor, 0, {row, col})
-        if not ok then
-          log.debug("Failed to move to cursor:", err_msg, row, col)
-        end
+    if row and col then
+      local ok, err_msg = pcall(a.nvim_win_set_cursor, 0, {row, col})
+      if not ok then
+        log.debug("Failed to move to cursor:", err_msg, row, col)
       end
     end
-    vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))
   end
 end
 
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -2,7 +2,6 @@ local a = vim.api
 
 local log = require('telescope.log')
 local path = require('telescope.path')
-local state = require('telescope.state')
 
 local action_state = require('telescope.actions.state')
 
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -1,7 +1,6 @@
 local a = vim.api
 
 local log = require('telescope.log')
-local state = require('telescope.state')
 local path = require('telescope.path')
 
 local action_state = require('telescope.actions.state')
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -87,7 +87,7 @@ set.edit = function(prompt_bufnr, command)
   if entry_bufnr then
     utils.edit_buffer(command, entry_bufnr)
   else
-    filename = path.normalize(vim.fn.fnameescape(filename), vim.fn.getcwd())
+    filename = path.normalize(vim.fn.fnameescape(filename), vim.loop.cwd())
 
     -- check if we didn't pick a different buffer
     -- prevents restarting lsp server
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -1,6 +1,7 @@
 local a = vim.api
 
 local log = require('telescope.log')
+local utils = require('telescope.utils')
 local path = require('telescope.path')
 
 local action_state = require('telescope.actions.state')
@@ -84,15 +85,7 @@ set.edit = function(prompt_bufnr, command)
   require('telescope.actions').close(prompt_bufnr)
 
   if entry_bufnr then
-    if command == 'edit' then
-      vim.cmd(string.format(":buffer %d", entry_bufnr))
-    elseif command == 'new' then
-      vim.cmd(string.format(":sbuffer %d", entry_bufnr))
-    elseif command == 'vnew' then
-      vim.cmd(string.format(":vert sbuffer %d", entry_bufnr))
-    elseif command == 'tabedit' then
-      vim.cmd(string.format(":tab sb %d", entry_bufnr))
-    end
+    utils.edit_buffer(command, entry_bufnr)
   else
     filename = path.normalize(vim.fn.fnameescape(filename), vim.fn.getcwd())
 
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -1,7 +1,7 @@
 local a = vim.api
 
 local log = require('telescope.log')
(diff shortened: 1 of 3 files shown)


lua/telescope/pickers.lua
```diff
@@ -327,6 +327,7 @@ function Picker:find()
   assert(self.finder, "Finder is required to do picking")
 
   self.original_win_id = a.nvim_get_current_win()
+  state.set_global_key('original_buf_id', a.nvim_get_current_buf())
```
Conni2461, line 330: We don't need a `state.set_global_key` just do that the api call when doing the get key access

lua/telescope/actions/init.lua
```diff
@@ -73,71 +73,66 @@ function actions._goto_file_selection(prompt_bufnr, command)
...
-        if not ok then
-          log.debug("Failed to move to cursor:", err_msg, row, col)
-        end
+    if row and col then
+      local ok, err_msg = pcall(a.nvim_win_set_cursor, 0, {row, col})
+      if not ok then
+        log.debug("Failed to move to cursor:", err_msg, row, col)
       end
     end
-    vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))
```
Conni2461, line 140: Not having that breaks some things in the past. See here. I also said that this doesn't happen anymore so yeah. Maybe its fine to remove
https://github.com/nvim-telescope/telescope.nvim/issues/481#issuecomment-771453930

lua/telescope/pickers.lua
oberblastmeister: I am not sure what you mean

Conni2461:
First of all thanks :) You need to rebase actions was completely rewritten in the last commit :)
I think the method is the same you just find the new one in actions/set.lua

Might also fix this https://github.com/nvim-telescope/telescope.nvim/issues/476 if you wanna test it. If not i gonna do it after this is rebased :)

lua/telescope/actions/init.lua
oberblastmeister: That doesn't seem to happen anymore. Also as I said above this pr will make it so that if the buffer is already loaded, it will not reload it again so the buffer local variables should not be reset. It will only load new buffers that haven't already been loaded.

lua/telescope/pickers.lua
Conni2461: Why do we need to to save it as state key? We just can get the current_buffer before doing edit. We are back there because the picker is already closed. Or am i wrong? He also already had that check in place it was just broken because of the normalize.

oberblastmeister:
Yes I have tested and it fixes #476. Conceptually it also makes sense because that issue happens because previously we were using `edit`, which when used on the same file that you are currently in will error when the current file is not saved. Now if the file is the same as the current file it will not `edit`

lua/telescope/pickers.lua
oberblastmeister: That will not work because `get_current_buffer` before enter will get the telescope buffer, not the original one. Thats why we have to get the buffer *before* telescope has opened, hence setting a state key. Unless I am misunderstanding you

Conni2461: Telescope should be closed at this point in time. That works on current master. Printing the `bufnr` before edit gives me the correct one

oberblastmeister: You are right that telescope should be closed. My bad

oberblastmeister:
Should be rebased now

lua/telescope/actions/set.lua
```diff
@@ -50,71 +49,65 @@ set.edit = function(prompt_bufnr, command)
...
+  if entry_bufnr then
+    if command == 'edit' then
+      vim.cmd(string.format(":buffer %d", entry_bufnr))
+    elseif command == 'new' then
+      vim.cmd(string.format(":sbuffer %d", entry_bufnr))
+    elseif command == 'vnew' then
+      vim.cmd(string.format(":vert sbuffer %d", entry_bufnr))
+    elseif command == 'tabedit' then
+      vim.cmd(string.format(":tab sb %d", entry_bufnr))
+    end
```
tami5, line 95: Here's c like switch which is way better then elseifs 

```lua
commands = {
  edit = string.format(":buffer %d", entry_bufnr),
  new = string.format(":sbuffer %d", entry_bufnr),
  vnew = string.format(":vert sbuffer %d", entry_bufnr),
  tabedit = string.format(":tab sb %d", entry_bufnr)
}
  vim.cmd(commands[command])

-- or better

vim.cmd(({
  edit = string.format(":buffer %d", entry_bufnr),
  new = string.format(":sbuffer %d", entry_bufnr),
  vnew = string.format(":vert sbuffer %d", entry_bufnr),
  tabedit = string.format(":tab sb %d", entry_bufnr)
})[command])


```

oberblastmeister: I don't like creating a table just for an if/switch statement. It does look better but it creates 4 strings by default and also creates a table.

Conni2461: The code lies in memory anyway so both ways have there drawback. You could only create that table once which would be in fact the best solution because it doesn't have to create the if else branches. But something like this would be optimal:
```lua
local tbl = {
  edit = 'buffer',
  new = 'sbuffer',
...
}
string.format("%s %d", tbl[command], entry_bufnr)
```

But this is code that runs only when with `builtin.buffers` and `:buffer` etc is basically free. So i wouldn't touch that. But thats up to you guys.

lua/telescope/actions/init.lua
Conni2461: I will test this now. But i think its also no longer necessary.

lua/telescope/actions/set.lua
```diff
@@ -50,71 +49,65 @@ set.edit = function(prompt_bufnr, command)
...
+    -- to put stuff into `filename`
+    local value = entry.value
+    if not value then
+      print("Could not do anything with blank line...")
+      return
     end
 
-    local preview_win = state.get_status(prompt_bufnr).preview_win
-    if preview_win then
-      a.nvim_win_set_config(preview_win, {style = ''})
```
Conni2461, line 83: I think we had this because sometimes preview options were applied to the main window. I am not sure if this would happen again if we remove that part.

lua/telescope/actions/set.lua
```diff
@@ -50,71 +49,65 @@ set.edit = function(prompt_bufnr, command)
...
+      vim.cmd(string.format(":buffer %d", entry_bufnr))
+    elseif command == 'new' then
+      vim.cmd(string.format(":sbuffer %d", entry_bufnr))
+    elseif command == 'vnew' then
+      vim.cmd(string.format(":vert sbuffer %d", entry_bufnr))
+    elseif command == 'tabedit' then
+      vim.cmd(string.format(":tab sb %d", entry_bufnr))
+    end
+  else
+    filename = path.normalize(vim.fn.fnameescape(filename), vim.fn.getcwd())
```
Conni2461, line 97: We kinda wanna want speed now so we should use `vim.loop.cwd` :D no more viml :rofl:

lua/telescope/actions/set.lua
oberblastmeister: thanks for the clarification

oberblastmeister: using your solution now

Conni2461:
I hate that i am still nitpicking :laughing: I wouldn't put any action stuff in `utils.lua` currently. Just do a local table in front of that that function. With a clear name :) Thanks i am testing in 5-10min.
Looks really good to me thanks :)

lua/telescope/actions/set.lua
```diff
@@ -40,6 +39,24 @@ set.select = function(prompt_bufnr, type)
   return set.edit(prompt_bufnr, action_state.select_key_to_edit_key(type))
 end
 
+local edit_buffer
+do
```
Conni2461, line 43: Testing now but another tip :)
```lua
local edit_buffer = (function()
  local map = {}

  return function(command, bufnr)
    ...
  end
end)()
```

Runs first time (and only once) and initializes `edit_buffer`.

oberblastmeister: Thats cool. I think this is the purpose of `do .. end` though and in my opinion `do .. end` is clearer. I think this is small so I hope you are okay with me just leaving it :smile:

Conni2461: Do end just opens a block. Its a little bit different but i won't argue about some syntax sugar :laughing: Both works fine

Conni2461:
`vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))` is still needed. I can't get my folds to work without it. And i know @mhartington will not like it being removed :rofl:. Me too btw

oberblastmeister:
Could you explain how the folds do not work? I have tested and they work for me. `vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))` is *very* slow and the user would expect that a file that is already loaded should open up immediately.

Conni2461:
Its fast without it because we just not execute filetype autocommands. There is probably some neovim bug (or missing some vim patches) that makes them not run. I think i remember tj sending me a PR that tried to fix that.

Example what i have in my config (and probably @mhartington) as well:
```lua
local configs = parsers.get_parser_configs()
local ft_str = table.concat(vim.tbl_map(function(ft) return configs[ft].filetype or ft end, parsers.available_parsers()), ',')
vim.cmd('autocmd Filetype ' .. ft_str .. ' setlocal foldmethod=expr foldexpr=nvim_treesitter#foldexpr()')

-- or you can just do something like this for testing
vim.cmd('autocmd Filetype lua :lua print("hello from ft autocmd")')
```

For me it only prints with `doautocmd`

oberblastmeister:
@Conni2461 I have to go right now. I will come back and test in about an hour.

mhartington:
_You rang...._

Conni2461:
ober wants to remove `vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))` but that breaks folds or all ft autocommands. Just for you information. I am still arguing :laughing:

mhartington:
😄 ah alright.

I guess the better question is why would removing this break folds? 
Is this something that needs to be fixed in neovim?

Conni2461:
Yeah i think that PR https://github.com/neovim/neovim/pull/13247 (tj send that one to me some time ago) tries to fix that. I just build that branch. Doesn't help.

oberblastmeister:
This is getting really wierd. I think that this is a neovim issue. My `Filetype` autocommands are not even working in the first place so that needs to be fixed before I test the other thing. Editing a file regular without `telescope.nvim` will not even trigger filetype autocommands. I made an [issue](https://github.com/neovim/neovim/issues/14000#issue-814857729) on neovim.

lua/telescope/actions/set.lua
tami5: You are welcome, :D

lua/telescope/actions/set.lua
```diff
@@ -40,6 +39,24 @@ set.select = function(prompt_bufnr, type)
...
+  }
+
+  edit_buffer = function(command, bufnr)
+    command = map[command]
+    if command == nil then
+      error('There was no associated buffer command')
+    end
+    vim.cmd(string.format("%s %d", command, bufnr))
+  end
+end
```
tami5, line 58: a version without if else :laughing:, ignore it if you don't like it its okay. I'm just sharing my thoughts

```lua
local errmsg = "There was no associated buffer command"

edit_buffer = function(command, bufnr)
  return map[command] and vim.cmd(map[command]  .. " " .. bufnr) or error(errmsg) 
end

```

oberblastmeister:
@Conni2461 I tested and for me the folds don't even work with the `vim.api.nvim_command("doautocmd filetypedetect BufRead " .. vim.fn.fnameescape(filename))` so the autocommand seems useless and just makes entering a file very slow. For me whenever I open a file with telescope folds don't work, but when I open a file with `edit` it works. The weird thing is that when folds are not working when I do `set foldexpr?` it shows `nvim_treesitter#foldexpr()` but somehow it is not recognizing that. I think that this is some neovim or treesitter buf and that an `doautocmd` seems like it is unnecessary to retrigger everything and make it slow.

Conni2461:
Yeah we ran into that as well. So removing the action.center makes it work. It just soooooo weird. I have that locally in my config. mappings.lua just does `select_default + center` on default. Really weird :laughing:

But you might be right it would be better to remove it. We could add a new action. So i can locally do `select_default + do_auto` or i just add a do auto locally

oberblastmeister:
@Conni2461 changing the actions to only do `select_default` does not fix my folds. I think we should merge this pr because the `doautocmd` does not seem to be doing anything and this pr also fixes #476. For the treesitter folds not working we should move to #559

Conni2461 (approved):
Yeah i agree. I used rust and rust_analyzer yesterday and with `doautocmd` it starts lsp twice. Which isn't bad for clangd and others i usually use but rust_analyzer, which is usually really great has a bad startup time.

Sorry @mhartington you gotta do `:e` for now. I will search more why it just doesn't work because doing `setlocal foldmethod` or `setlocal foldexpr` shows the correct values. You can also create a new action which basically just does `vim.cmd'e'` and do:
`default_select + custom_action` in your config.

Check readme for more info :)

Code LGTM btw :rofl:

Issue #313, opened 2020-12-04
Fuzzy finder for LSP diagnostics


alex-popov-tech:
Hey! Thanks for the plugin!
I could not find some function to search through diagnostic errors for the buffer, which is super useful...
Did i miss that function, or it was not added yet?
Or maybe i can add it myself using telescope api?
Any help would be appreciated!

Conni2461:
Are you talking about `builtin-lsp`?
If so you can do `:lua vim.lsp.diagnostic.set_loclist({open_loclist = false})` to populate location list with diagnostics
After that you can run `:lua require'telescope.builtin'.loclist{}`

alex-popov-tech:
@Conni2461 yes exactly...if you have a complete function, can you please share it? i will try to build it myself meanwhile :)

Conni2461:
```lua
local m = {}

m.show_diagnostics = function(opts)
  opts = opts or {}
  vim.lsp.diagnostic.set_loclist({open_loclist = false})
  require'telescope.builtin'.loclist(opts)
end

return m
```

Would be a lua modul. E.g. `.config/nivm/lua/modules/telescope.lua`.
Then you shoud be able to do `:lua 'require('module/telescope').show_diagnostics()`

alex-popov-tech:
thanks, that works like a charm!

fdschmidt93:
For anyone else coming here and wondering how to highlight the entries resulting from the above function like red for error (something quite worthwhile for my Python coding), etc. , you can easily create a custom `entry_maker` on top of @Conni2461's function (thanks!), similar to the below code which links signs and highlighting to builtin LSP highlight groups. Preview attached at the bottom.

```lua
local telescope_utils = require('telescope.utils')
local entry_display = require('telescope.pickers.entry_display')

local function gen_from_lsp_loclist(opts)
  opts = opts or {}
  opts.tail_path = telescope_utils.get_default(opts.tail_path, true)

  local displayer = entry_display.create {
    separator = "▏",
    items = {
    -- slighltly increased width
      { width = 11 },
      { width = 48 },
      { remaining = true }
    }
  }

  local make_display = function(entry)
    local filename
    if not opts.hide_filename then
      filename = entry.filename
      if opts.tail_path then
        filename = telescope_utils.path_tail(filename)
      elseif opts.shorten_path then
        filename = telescope_utils.path_shorten(filename)
      end
    end

    -- add styling of entries
[...]

PR #543, opened 2021-02-23
Diagnostics & Prefiltering by Entry Attribute

fdschmidt93:
This PR implements a builtin LSP diagnostics picker and optional prefiltering (see also discussion as per #313). Prefiltering subselects entries based on a user-set entry attribute prior to scoring.

Some items to discuss:

- There might be more elegant ways to wrap prefiltering around Sorter - happy to get pointers!  
- Filtering matches items to filter via `:%S[:%s]+` to more leniently allow filter matching though also just as happy to implement it more conservatively.
- Filtering defaults to slightly refactored version of `substr_matcher` that (if my understanding is correct) more easily allows to change filtering to possibly other functions
- Possible additional default prefiltering possibilities (so far only document symbols and diagnostics)

Long-term ideas: completing filter tags like frecency-telescope

**Document Symbols Prefiltered by `symbol_name` (e.g. method):**

document_symbols

**Diagnostics Prefiltered by `type` (i.e. hints):**

diagnostics

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -236,7 +236,8 @@ All layout strategies are functions with the following signature: >
         - columns : number Columns in the vim window
         - lines   : number Lines in the vim window
 
-TODO: I would like to make these link to `telescope.layout_strategies.*`, but it's not yet possible.
+TODO: I would like to make these link to `telescope.layout_strategies.*`,
+but it's not yet possible.
 
 Available layout strategies include:
   horizontal:
@@ -247,6 +248,24 @@ Available layout strategies include:
 
   flex:
   - See |layout_strategies.flex|
+
+Available tweaks to the settings in layout defaults include
+(can be applied to horizontal and vertical layouts):
+  mirror (default is `false`):
+  - Flip the view of the current layout:
+    - If using horizontal: if `true`, swaps the location of the
+      results/prompt window and preview window
+    - If using vertical: if `true`, swaps the location of the results and
+      prompt windows
+
+  width_padding:
+  - How many cells to pad the width of Telescope's layout window
+
+  height_padding:
+  - How many cells to pad the height of Telescope's layout window
+
+  preview_width:
+  - Change the width of Telescope's preview window
 
 
 layout_strategies.center()                        *layout_strategies.center()*
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -252,16 +252,16 @@ previewers.Previewer()                                *previewers.Previewer()*
 
 
 
-previewers.buffer_previewer_maker({bufnr}, {opts}, {filepath})*previewers.buffer_previewer_maker()*
+previewers.buffer_previewer_maker({filepath}, {bufnr}, {opts})*previewers.buffer_previewer_maker()*
     A universal way of reading a file into a buffer previewer. It handles async
     reading, cache, highlighting, displaying directories and provides a
     callback which can be used, to jump to a line in the buffer.
 
 
     Parameters: ~
+        {filepath} (string)  String to the filepath, will be expanded
         {bufnr} (number)  Where the content will be written
         {opts} (table)  keys: `use_ft_detect`, `bufname` and `callback`
-        {filepath} (string)  String to the filepath, will be expanded
 
 
 previewers.cat()                                            *previewers.cat()*
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -13,16 +13,14 @@ https://github.com/nvim-telescope/telescope.nvim
   :h telescope.actions
 
 telescope.extensions()                                *telescope.extensions()*
-
     Use telescope.extensions to reference any extensions within your
     configuration.
-    While the docs currently generate this as a function, it's actually a table.
-    Sorry.
+    While the docs currently generate this as a function, it's actually a
+    table. Sorry.
 
 
 
 telescope.load_extension({name})                  *telescope.load_extension()*
-
     Load an extension.
 
 
@@ -31,7 +29,6 @@ telescope.load_extension({name})                  *telescope.load_extension()*
 
 
 telescope.register_extension({mod})           *telescope.register_extension()*
-
     Register an extension. To be used by plugin authors.
 
 
@@ -40,10 +37,8 @@ telescope.register_extension({mod})           *telescope.register_extension()*
 
 
 telescope.setup({opts})                                    *telescope.setup()*
-
-    Setup function to be run by user. Configures the defaults, extensions
-    and other aspects of telescope.
-
+    Setup function to be run by user. Configures the defaults, extensions and
+    other aspects of telescope.
 
 
     Valid keys for {opts.defaults}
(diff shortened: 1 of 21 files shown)


lua/telescope/builtin/lsp.lua
```diff
@@ -64,7 +64,10 @@ lsp.document_symbols = function(opts)
       entry_maker = opts.entry_maker or make_entry.gen_from_lsp_symbols(opts)
     },
     previewer = conf.qflist_previewer(opts),
-    sorter = conf.generic_sorter(opts),
+    sorter = conf.prefilter_sorter{
+        tag = "symbol_type",
```
tjdevries, line 68: we use two space indententation generally. sorry for the nitpick :)

lua/telescope/builtin/lsp.lua
```diff
@@ -182,6 +185,27 @@ lsp.workspace_symbols = function(opts)
   }):find()
 end
 
+lsp.diagnostics = function(opts)
+    local locations = utils.diagnostics_to_tbl(opts)
+    local filename = vim.api.nvim_buf_get_name(0)
+    for _, value in pairs(locations) do value.filename = filename end
```
tjdevries, line 191: I think best to use `bufnr` over `name` here -- since we already have the buffer.

lua/telescope/builtin/lsp.lua
```diff
@@ -209,6 +233,7 @@ local feature_map = {
   ["document_symbols"]  = "document_symbol",
   ["references"]        = "find_references",
   ["workspace_symbols"] = "workspace_symbol",
+  -- ["diagnostics"] = "diagnostics",
```
tjdevries, line 236: we probably want this either in or out

lua/telescope/make_entry.lua
```diff
@@ -894,6 +901,71 @@ function make_entry.gen_from_ctags(opts)
...
+      filename = entry.filename
+      if opts.tail_path then
+        filename = utils.path_tail(filename)
+      elseif opts.shorten_path then
+        filename = utils.path_shorten(filename)
+      end
+    end
+
+    -- add styling of entries
+    local sign = vim.trim(vim.fn.sign_getdefined("LspDiagnosticsSign" .. lsp_type_diagnostic[entry.type])[1].text)
```
tjdevries, line 930: This should probably only be done once, rather than every time we do the display.

lua/telescope/builtin/lsp.lua
fdschmidt93: You're right! Info was redundant anyways as per corresponding `make_entry`.

lua/telescope/builtin/lsp.lua
```diff
@@ -184,6 +187,25 @@ lsp.workspace_symbols = function(opts)
   }):find()
 end
 
+lsp.diagnostics = function(opts)
+    local locations = utils.diagnostics_to_tbl(opts)
+    if vim.tbl_isempty(locations) then return end
```
Conni2461, line 192: Maybe we also wanna `print('No diagnostics found')` or so. Might be better if we give people some feedback that things are happening

lua/telescope/builtin/lsp.lua
```diff
@@ -184,6 +187,25 @@ lsp.workspace_symbols = function(opts)
...
+        results = locations,
+        entry_maker = opts.entry_maker or make_entry.gen_from_lsp_diagnostics(opts)
+        },
+      previewer = conf.qflist_previewer(opts),
+      sorter = conf.prefilter_sorter{
+        tag = "type",
+        sorter = conf.generic_sorter(opts)
+      }
+    }):find()
+    return locations
```
Conni2461, line 206: Do we need to return locations?

lua/telescope/builtin/lsp.lua
```diff
@@ -184,6 +187,25 @@ lsp.workspace_symbols = function(opts)
...
+lsp.diagnostics = function(opts)
+    local locations = utils.diagnostics_to_tbl(opts)
+    if vim.tbl_isempty(locations) then return end
+
+    pickers.new(opts, {
+      prompt_title = 'LSP Diagnostics',
+      finder = finders.new_table {
+        results = locations,
+        entry_maker = opts.entry_maker or make_entry.gen_from_lsp_diagnostics(opts)
+        },
```
Conni2461, line 199: Shouldn't be indented :)

lua/telescope/make_entry.lua
```diff
@@ -30,6 +30,13 @@ local lsp_type_highlight = {
   ["Variable"] = "TelescopeResultsVariable",
 }
 
+local lsp_type_diagnostic = {
+    ["I"] = "Information",
```
Conni2461, line 34: 2 spaces here

Conni2461:
Indentation is still wrong in sorters and utils, so i guess this is not yet done. You can ping me if i should take another look at this. But right now i really really like it.

fdschmidt93:
Apologies, bad habits in nvim settings from only writing python thus far - will clean up tonight!

Conni2461:
Yeah all good :) I appreciate the work you put into :D

fdschmidt93:
Pinging @Conni2461 triple-checked indentation.. should be fine now (hopefully :)

elianiva:
am I missing something here? The result is only updated when I insert some query and then remove it again

https://user-images.githubusercontent.com/51877647/109780942-0cd27a80-7c3a-11eb-93dc-91541079dffb.mp4

fdschmidt93:
I'm afraid I cannot reproduce your gif :/ it works in `wip.lua` as you'd expect for me.

https://user-images.githubusercontent.com/39233597/109785006-f4e40200-7c0b-11eb-9d52-b77771670b07.mp4

elianiva:
ok gotcha, it was caused by fzy/fzy-native

clason:
> ok gotcha, it was caused by fzy/fzy-native overriding _all_ sorters

Oh, that's probably also why the project functionality in telescope-frecency wasn't working for me! (@sunjon)

That's a big problem... I _cannot_ do without fzy-native...

elianiva:
I fixed it using the default config on fzy-native readme so it won't override *every* sorters. Though it's weird that your frecency doesn't work, mine still does without these lines.

```lua
    fzy_native = {
      override_generic_sorter = false,
      override_file_sorter = true,
    }
```

clason:
Oh, frecency works, just the `:lsp:` prefiltering doesn't. And I _do_ want to override the generic sorter! And prefiltering! I want everything :)

fdschmidt93:
Ok, I'll have to incorporate something more closely resembling frecency & #594 into prefiltering which should hopefully fix the issue. I actually _did not_ originally to incorporate prefiltering more "natively" x).

Conni2461:
I agree with clason. I want fzy everywhere. So i am going to check out this PR this evening (8pm cet) and try to get an understanding whats going on. Maybe i can help :)

@clason about frecency, it ignores your configuration and should always use substr matcher so it shouldn't matter that you are using fzy.

@fdschmidt93 About formatting, first glance everything looks great thanks. I hope the whole formatting situation improves soon. rockerBOO found [StyLua](https://github.com/JohnnyMorganz/StyLua) and tj and i have talked about using it, with a CI job that checks if formatting is respected. That way it will just tell you what lines need to change and how.

In the meantime would be great if you can add your finder to the builtin lsp table in the readme. So people can find because its awesome :). Otherwise if you want you can also start documenting your function header in `lua/telescope/builtin/init.lua` this is not a requirement (yet) because not a single picker is documented but we have to start at some point. This header documentation will then create the corresponding vim documentation in a CI job. See:
- https://github.com/nvim-telescope/telescope.nvim/blob/master/CONTRIBUTING.md
- https://github.com/tjdevries/tree-sitter-lua/blob/85048ad5932232309a2bce50a11c10cbdaddd880/HOWTO.md

We don't have a good way to document the opts table yet (https://github.com/tjdevries/tree-sitter-lua/issues/8) but i also wanna get this done in the next two days as well. And formatting might be a little bit off until https://github.com/tjdevries/tree-sitter-lua/pull/6 is merged but we can just retrigger CI :)

Conni2461:
>  tonight _ideally before _

I kinda don't like the idea of recreating the finder unless everything else doesn't work, because currently we can only recreate `new_table`. I actually love your approach and want to get this to work. 
Also its way cleaner.

clason:
> @clason about frecency, it ignores your configuration and should always use substr matcher so it shouldn't matter that you are using fzy.

Oh, right, that makes sense. Yes, `substr` matcher works well, of course -- telescope doesn't _have_ to be fuzzy, it's not just (anymore) FZF-but-in-Lua

> In the meantime would be great if you can add your finder to the builtin lsp table in the readme. So people can find because its awesome :).

That reminds me: now that we have an awesome gallery in the Wiki, would it make sense to remove the list from the README (which is quite long) and just refer to the Wiki? (I'd also remove the "Your next awesome..." entries; they just take up space.

(Also, while I'm at it, the previewer table is inconsistent: the "new" previewers are listed as experimental while the "cat" previewers are listed as deprecated -- so we currently have _no_ stable previewers ;) The names could also be changed before 1.0 so the buffer ones are called `files_buffer`, `grep_buffer`, etc. and the old ones `files_cat` etc.)

Conni2461:
Oh yeah good catch. `buffer previewers` are default already just forgot to remove the experimental mention.

We should cleanup README before 1.0 but i would wait a little bit longer till all `telescope.builtins` are in `:help`. I will open a tracking issue where we can discuss this better

doc/telescope.txt
```diff
@@ -252,15 +252,15 @@ previewers.Previewer()                                *previewers.Previewer()*
 
 
 
-previewers.buffer_previewer_maker({opts}, {bufnr}, {filepath})*previewers.buffer_previewer_maker()*
+previewers.buffer_previewer_maker({bufnr}, {opts}, {filepath})*previewers.buffer_previewer_maker()*
```
Conni2461, line 255: I am almost done with this fix.

Conni2461 (approved):
LGTM.
Thanks for all your work and patience :)

PR #544, opened 2021-02-24
feat: adds M mapping in normal mode

hoob3rt:
continuing https://github.com/nvim-telescope/telescope.nvim/pull/463

Adds `M` mapping in normal mode for consistency with `H` and `L` mappings

### DIFF
--- a/lua/telescope/pickers/scroller.lua
+++ b/lua/telescope/pickers/scroller.lua
@@ -80,6 +80,10 @@ scroller.top = function(sorting_strategy, max_results, num_results)
   return (num_results > max_results) and 0 or (max_results - num_results)
 end
 
+scroller.middle = function(sorting_strategy, max_results, num_results)
+  return math.floor(max_results/2)
+end
+
 scroller.bottom = function(sorting_strategy, max_results, num_results)
   if sorting_strategy == 'ascending' then
     return math.min(max_results, num_results) - 1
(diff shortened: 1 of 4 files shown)


lua/telescope/pickers/scroller.lua
```diff
@@ -80,6 +80,10 @@ scroller.top = function(sorting_strategy, max_results, num_results)
   return (num_results > max_results) and 0 or (max_results - num_results)
 end
 
+scroller.middle = function(max_results)
```
Conni2461, line 83: Maybe we should have them always take the same params?! Tj wondered :laughing: You can do `(_, max_results, _)`

Otherwise looks good to me :) Thanks that was super fast btw

hoob3rt: Yeah consistency is nice. I will apply that

hoob3rt: as I expected, your luacheck doesn't like the underscore variable, lmk how do you want to proceed with this

Conni2461: `_` is defined for function headers :laughing: otherwise its nil

hoob3rt: so just change it to nil and we should be good to go?

Conni2461: Yeah

Conni2461: Oh updating tests to `nil, value, nil` :)

hoob3rt: Yeah, I noticed

Conni2461: Thanks :)

Issue #545, opened 2021-02-24
actions.move_selection_{next,previous} no longer work after scrolling the previewer.


elianiva:
### Description

Telescope `action.move_selection_next` and `action.move_selection_previous` no longer work after scrolling the previewer buffer.

**Expected Behavior**

It should work as it would normally after I scroll the previewer buffer.

**Actual Behavior**

It can only jump between the topmost or the bottommost item. Here's a screencast that shows which key I'm pressing.

https://user-images.githubusercontent.com/51877647/108946286-b214ae00-7690-11eb-8b5e-cd36a9f76501.mp4

As you can see, I'm holding `<C-n>` and `<C-p>` but the cursor can only move between the topmost and the bottommost item.

### Details

<details><summary>Reproduce</summary>

1. nvim -nu test.vim
2. `:Telescope find_files`
3. Scroll between the result items using `<C-n>` or `<C-p>`
4. Scroll the buffer previewer using `<C-d>` or `<C-u`
5. Try to do step 3 once more.
6. It doesn't scroll like before, it only jumps between the topmost or the bottommost item.
</details>

<details><summary>Environment</summary>

- nvim --version output: NVIM v0.5.0-dev+1109-g6b7cc45c4
- Operating system: Archlinux
- Telescope commit: b5051ee

</details>

<details><summary>Configuration</summary>
<p>

```viml
set nocompatible hidden laststatus=2

if !filereadable('/tmp/plug.vim')
  silent !curl --insecure -fLo /tmp/plug.vim
    \ https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
endif

source /tmp/plug.vim
call plug#begin('/tmp/plugged')
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-telescope/telescope.nvim'
call plug#end()

autocmd VimEnter * PlugClean! | PlugUpdate --sync | close
lua << EOF

require('telescope').setup{
  defaults = {
    vimgrep_arguments = {
      'rg',
      '--color=never',
      '--no-heading',
      '--with-filename',
      '--line-number',
      '--column',
      '--smart-case'
    },
    prompt_position = "bottom",
    prompt_prefix = ">",
    selection_strategy = "reset",
    sorting_strategy = "descending",
    layout_strategy = "horizontal",
    layout_defaults = {},
    file_ignore_patterns = {},
    shorten_path = true,
    winblend = 0,
[...]

PR #547, opened 2021-02-24
fix: scroll misbehaving + fixed jump to middle

elianiva:
#545 https://github.com/nvim-telescope/telescope.nvim/issues/545#issuecomment-784914549
cc @Conni2461 

edit: forgot to test it locally before pushing it :laughing: I'll fix the failing test in a moment

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -81,7 +81,7 @@ end
 function actions.move_to_middle(prompt_bufnr)
   local current_picker = actions.get_current_picker(prompt_bufnr)
   current_picker:set_selection(p_scroller.middle(
-    nil,
+    current_picker.sorting_strategy,
     current_picker.max_results,
     current_picker.manager:num_results()
   ))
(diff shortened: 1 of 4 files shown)


Conni2461:
`sorting_strategy = 'ascending'` was already working. This should break that case. Its only when descending and <

Otherwise good :) So add that case and then tests for both `ascending` and `descending`. I think we should just put the middle tests into     `it('should handle ascending', function()`/ `descending` as well.

lua/telescope/actions/init.lua
```diff
@@ -80,7 +80,11 @@ end
 
 function actions.move_to_middle(prompt_bufnr)
   local current_picker = actions.get_current_picker(prompt_bufnr)
-  current_picker:set_selection(p_scroller.middle(nil, current_picker.max_results, nil))
+  current_picker:set_selection(p_scroller.middle(
+    nil,
```
Conni2461, line 84: Still `nil`

PR #528, opened 2021-02-12
Cycle previewers with git commit and bcommit already using it

Conni2461:
Closes #307
Closes #627

Look at it here: https://streamable.com/sxi8fz

Continuation of #314

Edit: would be cool if we could somehow show the context of the previewer. Idk how we would do this.

PR #558, opened 2021-02-26
Change git -P to --no-pager

jaywonchung:
This pull request changes the `-P` option given to `git` to `--no-pager` in git commit/file/branch previewers because the latter is more backwards compatible.

The `-P` option was added to `git` around three years ago as a shorthand for `--no-pager` (relavant commit: https://github.com/git/git/commit/7213c288187bd9ef1e32e4f86c20c55436f1ae90). My lab server (remarkably) uses `git 2.17.1` and the preview window for git commits displayed nothing. This patch fixed the problem.

Thank you for the great plugin!

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -476,7 +476,7 @@ previewers.git_branch_log = defaulter(function(opts)
       local current_remote = 1
 
       local gen_cmd = function(v)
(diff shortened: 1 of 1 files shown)


Conni2461:
Thats super good to know thanks :)

I gonna make sure i always do that and update https://github.com/nvim-telescope/telescope.nvim/pull/528 which will introduce new git previewers.

Thanks again :D

PR #563, opened 2021-02-26
Allowed git_files to use vim pwd over git root

tapayne88:
Hi, thanks for the amazing work!

I work in a pretty large git repo and spend my time in a nested folder. What I wanted was an option for `git_files` to only show the git files from my cwd - not the whole project.

I _think_ the only other `git_*` command this might be relevant to is `git_commits` (maybe `git_status`?). I'm not too bothered about those but happy to add it to them.

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -154,11 +154,14 @@ local set_opts_cwd = function(opts)
 
   -- Find root of git directory and remove trailing newline characters
   local git_root = vim.fn.systemlist("git -C " .. opts.cwd .. " rev-parse --show-toplevel")[1]
+  local use_git_root = utils.get_default(opts.use_git_root, true)
 
   if vim.v.shell_error ~= 0 then
     error(opts.cwd .. ' is not a git directory')
   else
-    opts.cwd = git_root
+    if use_git_root then
+      opts.cwd = git_root
+    end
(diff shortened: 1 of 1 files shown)


Conni2461:
`set_opts_cwd` will apply cwd to all git_* commands. Its a wrapper that runs in front of all commands. Could you check if that is correct? After that we can merge :)

Thanks for the PR
LGTM

Conni2461:
Okay i just tested this. I think we need to add `'--', '.'` to both `git_commits` and `git_status` would be great if you can do that. `dot` should be enough because we already set cwd over our job interface :)

After that i would merge.

Thanks :)

tapayne88:
> `set_opts_cwd` will apply cwd to all git_* commands. Its a wrapper that runs in front of all commands.

Yeah, that's what my testing shows - all `git_*` commands will run through `set_opts_cwd`.

>  I think we need to add `'--', '.'` to both `git_commits` and `git_status`

Yeah I can get that added, from my testing that looks to scope those commands to the `cwd` too 👍

Conni2461:
Could you make CI happy and split up that line :D Similar to how `bcommits` is formatted :)
Thanks after that i press the button

Conni2461 (approved):
LGTM

Thank you very much :) I know that will make some people happy

PR #571, opened 2021-02-28
feat: open all selected files at once

elianiva:
Added ability to open multiple files at once like `FZF`, someone asked about this feature a long time ago to me when multiselection wasn't available :laughing:

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -413,16 +413,6 @@ actions.send_to_qflist = function(prompt_bufnr)
   vim.fn.setqflist(qf_entries, 'r')
 end
 
-actions.open_selected_files = function(prompt_bufnr)
-  local picker = action_state.get_current_picker(prompt_bufnr)
-
-  actions.close(prompt_bufnr)
-
-  for _, files in ipairs(picker:get_multi_selection()) do
-    vim.cmd(":silent edit "..files[1])
-  end
-end
-
 actions.open_qflist = function(_)
   vim.cmd [[copen]]
 end
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -413,6 +413,16 @@ actions.send_to_qflist = function(prompt_bufnr)
   vim.fn.setqflist(qf_entries, 'r')
 end
 
+actions.open_selected_files = function(prompt_bufnr)
+  local picker = action_state.get_current_picker(prompt_bufnr)
+
+  actions.close(prompt_bufnr)
+
+  for _, files in ipairs(picker:get_multi_selection()) do
+    vim.cmd(":silent edit "..files[1])
+  end
+end
+
 actions.open_qflist = function(_)
   vim.cmd [[copen]]
 end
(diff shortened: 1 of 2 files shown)


Conni2461:
We need something like this 100% but i don't like the idea of having another specific action for that. This should just seamlessly integrate with `actions.select_*` and replacements/enhancements. And that's what this makes so tricky. 

We probably need to make this happen in `mt.lua`. I have some thought on how we could do that and i am sure tj has some too because i briefly talked with him about this feature.

And i also don't like to ship this now, have people setup mappings and then we replace it in a week with the "correct" way.

Feel free to dig into mt.lua or write me in private if you wanna hear my thoughts on how to realize that :)

elianiva:
It feels weird indeed. I was thinking of when you don't have any selected items, it opens up normally. But when you have selections, it will open all of them. I'll DM you on Gitter and hope your message doesn't get swallowed by the clunky matrix<->gitter bridge 😆

weilbith:
I would rather prefer to select all and then perform an action. For example I would use this to put all grep results into the quickfix list.

elianiva:
Yes, that'd be better

Conni2461:
> put results into qf list

is already implemented and mapped to <M-q>. Will put all selected elements into qflist. We are talking here how to handle <cr>, <c-v>, ... and also handle replacements at the same time.

elianiva:
> is already implemented and mapped to .

btw, did your message got cut off or something? I assume you wrote something surrounded by backticks but it's missing.

weilbith:
> > put results into qf list
> 
> is already implemented and mapped to `<M-q>`. Will put all selected elements into qflist. We are talking here how to handle `<cr>`, `<c-v>`, ... and also handle replacements at the same time. 

Yes, but is there an `select all`? :thinking:

Conni2461:
I am still not sure what you mean. There is `send_to_qflist` which will send all entries to qflist. Mapped to `<C-q>`.

If you mean like i select 10 entries with `<TAB>` then press `<CR>` and all 10 entries will be opened for example. Yes this is important that we have that. But elianiva wanted to introduce a new action for that which than had to be mapped to a different key. Thats not that cool. The cool thing would be making it just work with the current select interface but thats a bit tricky because we can compose actions with `+` we can replace actions and we can enhance actions with _pre and _post. All these things need to be taken into consideration so that multiselection will work with everything out of the box.

I send elianiva some info how i would realize this and he is already working on it :D

elianiva:
I think I've managed to get it *kinda* working except the picker doesn't get closed, the backward compatibility test failed, and I'm not sure about the effect for other extensions, they probably also break :rofl:

Conni2461:
You can maybe also just put up the code here. Ping tj :rofl:, tj reads code, tj brain starts thinking about it, tj picks it up and finishes it.

That worked out great for me when i implemented `replace_if`. A day after my implementation he did it :rofl:

elianiva:
right, I'll do that tomorrow since the current code has become more messy and broken because I was *trying* to fix it not knowing I'm just making it worse :rofl:

elianiva:
ok I've pushed my *broken* code, the test is no longer failing though, but it still complains about the picker

elianiva:
superseded by #807

PR #572, opened 2021-02-28
fix: entry_prefix if caret > prefix

Conni2461:
Fix this issue. I just got the length of the wrong string. Bad rebase by me.
image

CC @weilbith

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -707,7 +707,7 @@ function Picker:set_selection(row)
       a.nvim_buf_set_text(
         results_bufnr,
         self._selection_row, 0,
-        self._selection_row, #self.entry_prefix,
+        self._selection_row, #self.selection_caret,
         { self.entry_prefix }
       )
       self.highlighter:hi_multiselect(


weilbith:
I must admit that I have issues to test your fork. I'm using `packer.nvim`, but somehow it always fails if I switch to your fork and this branch. :shrug: 
So I just change into the directory of the `packpath` where Packer puts its plugins, add your remote and switched to the branch. This should now just work. But the issue is still the same. 

This is my setup configuration:

```vim
lua require('telescope').setup({
    \   defaults = {
    \     prompt_prefix = "  ",
    \     selection_caret = " ",
    \     winblend = 10,
    \     file_ignore_patterns = { "%.fls", "%.fdb_latexmk" }
    \   }
    \ })
```

And this is how the bug looks for me:

telecsope

Conni2461:
Don't you need to do packer compile or so. Idk i don't use packer :rofl: not yet.
I will try your config because i wanna ship a fix for that asap

weilbith:
Packer just make use of packages. So this should work as is now. The compile stuff is only for optional packages to load them on conditions and do stuff when certain packages get loaded like source their configuration or so. 

---

Do you need the icons I use? But actually they seem quite independent of the issue.

Conni2461:
I could reproduce this issue with master but not with this branch. So i wait 5 minutes maybe you can try it out but it seems to be fixed on my side

weilbith:
Okay wait a sec. Let me take a sip coffee and double check.

weilbith:
Okay, yes you are right. I'm sorry. :see_no_evil: 
For some reason `git switch` did not pick up your branch but just created a new one from `master`. I should check the man page why. :thinking: 
I'm now on 32372de18977358ae805388bd125227ea67452e2 and it works. Good job! :+1:

Conni2461:
Good thanks for testing. And more like bad job on rebasing :rofl: and i also never used `nvim_buf_set_text` so i wasn't sure if i did the right thing but this seems good now

weilbith (approved):
Fixes my issue. :+1: 
Code wise I think there is nothing to say. :laughing:

PR #573, opened 2021-02-28
Add ability to select in reverse direction

VVKot:
Use Shift+Tab to match fzf's keymappings.

Seems like a reasonable functionality, which will also make it easier for people coming from fzf.vim.


Conni2461:
Thanks :)

PR #578, opened 2021-03-01
feat(git): delete bindings in the action_mappings

ThePrimeagen:
Added the delete bindings.

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -105,6 +105,12 @@ git.branches = function(opts)
 
       map('i', '<c-r>', actions.git_rebase_branch)
       map('n', '<c-r>', actions.git_rebase_branch)
+
+      map('i', '<c-d>', actions.git_delete_branch)
+      map('n', '<c-d>', actions.git_delete_branch)
+
+      map('i', '<c-u>', false)
+      map('n', '<c-u>', false)
       return true
     end
   }):find()


Conni2461:
I have deliberately not mapped this, when i implemented it, because I thought it could be dangerous. Our default `<c-d>` mapping is previewer scrolling even if no previewer is enabled and this mapping override that.

But that was also before i added the confirmation prompt. So maybe i am overthinking things :rofl: 

Someone else has to decide this :)
But thanks for the PR prime :)

tami5:
I believe it should be remap to something else, and the confirmation prompt would make less dangerous indeed.

+ @Conni2461 it would be nice of somehow we can provide an easy way to setup configuration for individual pickers in setup method.

ThePrimeagen:
The setup method was not to hard.  You can provide an `action_mappers`, but it sucks because you have to find all the required actions.

The real question is does it make sense to even keep this ctrl-d for scrolling when it comes to branches?  I mean, its rather surprising to find someone scrolling through commits...  Just seems like the wrong way to do things.  Ctrl-d just feels expected, at least to me, as that is the thing I tried before remapping it in my personal setup, but it sucks to have that setup in my code because it feels so natural.

ThePrimeagen:
One more thing.  Currently telescope doesn't even tell you what your current branch is in the menu.  So technically, if you don't have a status bar telling you, you actually don't even know what the commits diff are betwixt.  At least for me I forget what branch I am on often :)

tjdevries:
If we do this, we should remove `<c-u>` from this picker as well, so that it's not lop-sided.

ThePrimeagen:
so doesthis mean I should override c-u or is there more discussion that is needed for making this change.

I get the fact that it does create new behavior, and I understand that line of reasoning as well.

Conni2461:
Okay i thought about that we can do this. @ThePrimeagen
Could you set `<c-u>` to false after that i merge

Conni2461:
About your that you want to see your current branch that is a really good idea. I can implement that in 15 mins :) unless you want to do it :)

ThePrimeagen:
I am just here for the 200 iq ideas.  I also take payment in github stars, like Telescopic Johnson

ThePrimeagen:
When you say "set to false" do you mean the change I am about to push.

ThePrimeagen:
ok, whoaahhhhh, who has been saying anything about a video.  Is telescopic johnson spilling the beans?

Conni2461:
He ordered that i have to do night shifts so you can make a video and make us famous :rofl: 

Thanks for the PR :)
I am almost done with updating the display. So that will be in master soon as well

ThePrimeagen:
OHH SNAP.  The street cred is going through the roof.

PR #580, opened 2021-03-01
Moved transform_devicons & get_devicons from make_entry.lua to utils.lua

ndavidq0:
...

@Conni2461

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -1,3 +1,5 @@
+local has_devicons, devicons = pcall(require, 'nvim-web-devicons')
+
 local pathlib = require('telescope.path')
 local Job     = require('plenary.job')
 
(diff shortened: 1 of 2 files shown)


lua/telescope/utils.lua
```diff
@@ -295,4 +298,46 @@ utils.align_str = function(string, width, right_justify)
     or string..string.rep(" ", width - str_len)
 end
 
+if has_devicons then
+  if not devicons.has_loaded() then
+   devicons.setup()
+  end
+
+  utils.transform_devicons = function(filename, display, disable_devicons)
```
Conni2461, line 306: I would do this differently and i think that is the issues:
```lua
utils.transform_devicons = (function()
  if has_devicons then
    -- load devicons
    return function(filename, display, disable_devicons)
      -- the content of transform_devicons
    end
  else
    return function(_, display, _)
      return display
    end
  end
end)()
```

And then the same for `get_devicons`

The function will than run the first time and will initialize `transform_devicons` and `get_devicons`

Does this make sense for you?

lua/telescope/utils.lua
```diff
@@ -295,4 +297,50 @@ utils.align_str = function(string, width, right_justify)
...
+      local icon, icon_highlight = devicons.get_icon(filename, string.match(filename, '%a+$'), { default = true })
+      if conf.color_devicons then
+        return icon, icon_highlight
+      else
+        return icon
+      end
+    end
+  else
+    return function(_, display, _)
+      return display
```
Conni2461, line 341: This function was only `(_, _) return ''` it doesn't have a display. It just returns the icon and highlight

ndavidq0: Oh didn't notice it

Conni2461 (approved):
LGTM

Thanks :)

Issue #518, opened 2021-02-09
Can't Use in RaspberryPi (aarch64 Architecture)


mirachan010:
### Description

Can't move in aarch64(RaspberryPi4-8GB) neovim from apt.
This neovim is use Lua5.1 not LuaJIT.

**Expected Behavior**

in use :Telescope <TAB> give some completion.

**Actual Behavior**

echo some error
```viml
:Telescope ...
Error detected while processing function <SNR>36_telescope_complete:
line    1:
E5108: Error executing lua ...k/packer/start/plenary.nvim/lua/plenary/filetype.lua:1: loop or previous error
loading module 'plenary.path'
line    4:
E5108: Error executing lua ...acker/start/telescope.nvim/lua/telescope/command.lua:2: loop or previous error
loading module 'telescope.builtin'
line    6:
E712: Argument of extend() must be a List or Dictionary
line   11:
E714: List required
:Telescope 0
```

### Details

<details><summary>Reproduce</summary>

1. nvim -nu test.vim
2. `:Telescope <Tab>` OR `:Telescope oldfiles`
</details>

<details><summary>Environment</summary>

- nvim --version output: 
```
NVIM v0.5.0-dev
Build type: RelWithDebInfo
Lua 5.1
[...]

PR #583, opened 2021-03-01
remove gotos because we don't need to use them

tjdevries:
Removed gotos


Conni2461:
Yeah thanks wanted to do this for some time. So like it :) Might fix https://github.com/nvim-telescope/telescope.nvim/issues/518

Conni2461 (approved):
LGTM

PR #585, opened 2021-03-01
Add docs for a bunch of actions

tjdevries:
Still need to do some more but I got bored :laughing: Someone else can finish if they want :)


Conni2461:
Its monday and seems like vigoux didn't fix tree sitter

tjdevries:
Ah, they actually did upstream though! We just need to get new version for neovim core and then it will be fixed I think :)

tjdevries:
We have some other PRs doing a better job of this. CLosing

PR #587, opened 2021-03-02
Add documentation for some actions

TC72:
Just a first try at adding some documentation to see if I'm doing this right.


Conni2461:
I am not sure why docgen didn't run for you. I can't see the actions on your fork so. You need open your fork repo, go to actions, and look at docgen output.

You can also just try to retrigger CI. `git commit --amend`, just save and force push

TC72:
When I looked at actions it showed there weren't any workflows being run so I enabled them.
I've pushed again and it looks like the action "Generate docs" ran.

https://github.com/TC72/telescope.nvim/actions/workflows/docgen.yml

Conni2461:
Nice, we still have issues that only the last x queries are exported but this bug was fixed in tree-sitter and we need to wait until neovim works with that version, until then, this will not generate docs for all function. You can still continue to document but and we can retrigger CI when these issues are sorted out

PR #591, opened 2021-03-02
feat: add expand_dir opts for git_status

### LINKED ISSUE #579: Use "-u" option for git status picker to show all files
When I'm working on new features where I add new folders and files inside, git status picker prints only the folders. 

I'd love to append `-u` to the git status command so it prints out all the files inside those new folders, or at least make `git_status` function accept the command through options like `find_files` does.

To create custom picker

**Screenshots**:

Before:
before
after

elianiva:
close #579 

do we want to provide a flag just for this option or do we want to make the entire `cmd` overridable? and I'm not sure if this should be the default or not, I like this better but idk about what other people opinion about it.

also, is there any reason why we use `table.getn(tbl)` instead of just `#tbl`? [This thread](https://stackoverflow.com/questions/31452871/table-getn-is-deprecated-how-can-i-get-the-length-of-an-array) said it's deprecated so I tested them both and they have the same result.

cc @Conni2461 @kristijanhusak

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -127,7 +127,10 @@ end
 
 git.status = function(opts)
   local gen_new_finder = function()
-    local output = utils.get_os_command_output({ 'git', 'status', '-s', '--', '.' }, opts.cwd)
+    local git_cmd = {'git', 'status', '-s', '--', '.'}
+    if opts.expand_dir then table.insert(git_cmd, #git_cmd - 1, '-u') end
+
+    local output = utils.get_os_command_output(git_cmd, opts.cwd)
 
     if table.getn(output) == 0 then
       print('No changes found')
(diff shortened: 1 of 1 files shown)


kristijanhusak:
I agree that it should be the default. I don't see anyone complaining that they see too many files in the git status :D If it cannot be the default, i'm ok with any option to override it, either via full command or just the flag.

elianiva:
lol what happened with the action, just got an email saying it's failed so I panicked and ran to my laptop to check the PR but all seems good

Conni2461:
I think we can just do this every time. I didn't even knew about `-u` when i wrote this :rofl: 

I didn't know it was deprecated :rofl: I just think it makes the code clearer overall. If i read `#` i know its string and `table.getn` that its a table. It was removed in lua 5.2 but i think luajit will never remove it. so we are fine (i hope)

Conni2461 (approved):
LGTM 
thanks :)

Edit: you can merge if this is done

PR #593, opened 2021-03-03
feat: use IncSearch to highlight TelescopeMatching

jceb:
In the two colorschemes that I tried, papercolor and nord, `Special` highlighting doesn't look good - it either doesn't highlight the search term (nord) or it's impossible to read (papercolor). IncSearch feels like the natural match for TelescopeMatching highlighting.

### DIFF
--- a/plugin/telescope.vim
+++ b/plugin/telescope.vim
@@ -18,7 +18,7 @@ highlight default link TelescopeResultsBorder TelescopeBorder
 highlight default link TelescopePreviewBorder TelescopeBorder
 
 " Used for highlighting characters that you match.
-highlight default link TelescopeMatching Special
+highlight default link TelescopeMatching IncSearch
 
 " Used for the prompt prefix
 highlight default link TelescopePromptPrefix Identifier


weilbith:
Why don't you simply extend your Telescope configuration to define a different highlight link? Why does a single colorscheme require to change the default of a plugin? :thinking:

jceb:
That's what I did. I created the pull request because I think that the default could be improved.

Conni2461:
I am not sure about this. I am a `hl_group` noob so i can't make a decision here. :)

elianiva:
I can't decide either, I think `IncSearch` makes more sense but I don't even know what `Special` is :laughing: 
I've tried to search for `:h hl-Special` but there's only `hl-SpecialKey`

tami5:
@jceb I looked into it and seem a bad default to have. but I agree the current one sucks and we need a better one. I'm going to close this for now but feel free to reopen if you come up with something some of us can agree upon

Thanks

jceb:
@tami5 what do you suggest instead? Did `IncSearch` not look good in the themes you tried?

tami5:
> @tami5 what do you suggest instead? Did `IncSearch` not look good in the themes you tried?

Yah it did not. Maybe because IncSearch has background color, which is something I did not get used to. I use red font almost in all my themes, off course with different variants.

jceb:
Interesting, which color schemes do you use?

tami5:
> Interesting, which color schemes do you use?

I'm a bit lazy when it cones to color schemes 😂 so I use base16 in zsh, nvim, st, dwm and pretty much anything with a single switch command. So the colors are depended on base16 classes

jceb:
Looking forward to the proposal of the right highlighting group

PR #594, opened 2021-03-03
[WIP] feat: tag_filter for lsp pickers

### LINKED ISSUE #471: feat: implement tag-filters like frecency for some builtin finders

It's kinda hard to find a variable or stuff like that from lsp finder (or similar, like treesitter finder) if the result is not filtered using its label.

Searching through `lsp` finders would be a better experience if we can filter out `functions`, `variables` and such using something like `:filter:`, similar to what frecency has for workspace filter.

Nothing, I love how frecency do this :D 

**Additional context**

image

Filter out based on these labels, so we would do something like `:field: field_name` or `:variable: variable_name`. It doesn't have to be persistent like frecency, as long as it can filter stuff out then it's great. I believe this depends on #455.

Opened this issue based on Sunjon's suggestion on Gitter.

elianiva:
close #471 

I'll add the completion and customisable delimiter later.

### DIFF
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -9,6 +9,36 @@ local conf = require('telescope.config').values
 
 local lsp = {}
 
+local lsp_tag_filter = function (locations, query_text, opts)
+  local new_locations, new_finder
+
+  local matched = query_text:match('^(:(%S+):)')
+  if matched then
+    query_text = query_text:sub(#matched + 1)
+    new_locations = vim.tbl_filter(function(item)
+      return item.kind:lower() == matched:sub(2, #matched - 1):lower()
+    end, locations)
+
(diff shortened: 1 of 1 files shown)


Conni2461:
I hate to do this. I haven't tried this or the other pr but https://github.com/nvim-telescope/telescope.nvim/pull/543 has filtering with a wrapper around sorter. Maybe you want to look at this first before you spend more time here

elianiva:
oh, I didn't see that PR :rofl: 
probably gonnna close this and wait for that instead

Conni2461:
I haven't tried it and haven't read the last two files but it has this thing:
```lua
      sorter = conf.prefilter_sorter{
        tag = "type",
        sorter = conf.generic_sorter(opts)
      }
```

So i still want to take a closer look how he does it, and how it works

PR #599, opened 2021-03-04
Hotfix: jump to diagnostics ++ Feat: workspace diagnostics and improved styling

fdschmidt93:
I noticed a regression from an intermediate commit of the earlier PR that prevented telescope from being able to jump to the diagnostic line as it seems previewer doesn't like `bufnr` and `filename` to be passed simultaneously.

Moreover, I implemented `get_all` which searches through diagnostics of all available buffers.

all_diag

To that end, I revised styling a bit to meet saner defaults with commands as recommended below:

- Current buffer only: `lsp.diagnostics()`
- All buffers incl. filenames: `lsp_diagnostics{ hide_filename = false, get_all = true}` 

as `hide_filename = true` is now the default, such that messages in current buffer get added space.

As a fun side effect entries are now sorted by type _and_ line as opposed to merely by line; at this late hour I have no idea why that is but I guess it's a welcome change :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,7 +409,8 @@ Built-in functions. Ready to be bound to any key you like. :smile:
 | `builtin.lsp_workspace_symbols`     | Searches in LSP all workspace symbols.                                                      |
 | `builtin.lsp_code_actions`          | Lists LSP action to be trigged on enter.                                                    |
 | `builtin.lsp_range_code_actions`    | Lists LSP range code action to be trigged on enter.                                         |
-| `builtin.lsp_diagnostics`           | Lists LSP Diagnostics in the current document.                                              |
+| `builtin.lsp_document_diagnostics`  | Lists LSP Diagnostics in the current document.                                              |
+| `builtin.lsp_workspace_diagnostics` | Lists LSP Diagnostics in the workspace if supported and otherwise open buffers.             |
 | ..................................  | Your next awesome picker function here :D                                                   |
 
 ### Git Pickers
--- a/README.md
+++ b/README.md
@@ -409,6 +409,7 @@ Built-in functions. Ready to be bound to any key you like. :smile:
 | `builtin.lsp_workspace_symbols`     | Searches in LSP all workspace symbols.                                                      |
 | `builtin.lsp_code_actions`          | Lists LSP action to be trigged on enter.                                                    |
 | `builtin.lsp_range_code_actions`    | Lists LSP range code action to be trigged on enter.                                         |
+| `builtin.lsp_diagnostics`           | Lists LSP Diagnostics in the current document.                                              |
 | ..................................  | Your next awesome picker function here :D                                                   |
 
 ### Git Pickers
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -64,6 +64,7 @@ builtin.spell_suggest = require('telescope.builtin.internal').spell_suggest
 builtin.lsp_references = require('telescope.builtin.lsp').references
 builtin.lsp_document_symbols = require('telescope.builtin.lsp').document_symbols
 builtin.lsp_code_actions = require('telescope.builtin.lsp').code_actions
+builtin.lsp_diagnostics = require('telescope.builtin.lsp').diagnostics
 builtin.lsp_range_code_actions = require('telescope.builtin.lsp').range_code_actions
 builtin.lsp_workspace_symbols = require('telescope.builtin.lsp').workspace_symbols
 
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -64,7 +64,8 @@ builtin.spell_suggest = require('telescope.builtin.internal').spell_suggest
 builtin.lsp_references = require('telescope.builtin.lsp').references
 builtin.lsp_document_symbols = require('telescope.builtin.lsp').document_symbols
 builtin.lsp_code_actions = require('telescope.builtin.lsp').code_actions
-builtin.lsp_diagnostics = require('telescope.builtin.lsp').diagnostics
+builtin.lsp_document_diagnostics = require('telescope.builtin.lsp').diagnostics
+builtin.lsp_workspace_diagnostics = require('telescope.builtin.lsp').workspace_diagnostics
 builtin.lsp_range_code_actions = require('telescope.builtin.lsp').range_code_actions
 builtin.lsp_workspace_symbols = require('telescope.builtin.lsp').workspace_symbols
 
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -64,7 +64,10 @@ lsp.document_symbols = function(opts)
       entry_maker = opts.entry_maker or make_entry.gen_from_lsp_symbols(opts)
     },
     previewer = conf.qflist_previewer(opts),
-    sorter = conf.generic_sorter(opts),
+    sorter = conf.prefilter_sorter{
+      tag = "symbol_type",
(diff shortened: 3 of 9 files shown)


lua/telescope/utils.lua
```diff
@@ -84,22 +84,20 @@ end
...
-  local insert_diag = function(diag)
+
+  local preprocess_diag = function(diag, bufnr)
+    local filename = vim.api.nvim_buf_get_name(bufnr)
     local start = diag.range['start']
     local finish = diag.range['end']
     local row = start.line
     local col = start.character
 
     local line = (vim.api.nvim_buf_get_lines(bufnr, row, row + 1, false) or {""})[1]
```
elianiva, line 98: This line throws a `nil` because `{nil}` is considered as true when I tested it.

```suggestion
    local line = vim.api.nvim_buf_get_lines(bufnr, row, row + 1, false)[1] or ""
```

Conni2461:
I wanted to say lgtm but something isn't 100% right yet. See 
image
image

Tested in telescope with sumneko. Also should `lsp_diagnostics` just always have `get_all=true` on default and we just do a alias with `get_all=false` and name it `lsp_bdiagnostics`?

elianiva:
I think `lsp_document_diagnostics` and `lsp_workspace_diagnostics` is better (similar to how we name lsp_*_symbols) but that might be too long for some people.

I kinda don't like the idea of `lsp_bdiagnostics` and `lsp_document_symbols` being too differrent even though they're similar (they both operate on the buffer)

I'd pick one of them for consistency, either of them is fine as long as it's consistent.

fdschmidt93:
Hopefully outstanding issues are resolved :) 

One thing to note (as reflected in the README) is that what `vim.lsp.diagnostic.get_all` returns appears LSP dependent, since pyright e.g. only shows diagnostics for open buffers whereas sumneko actually returns full diagnostics of workspace.

README.md
```diff
@@ -409,7 +409,8 @@ Built-in functions. Ready to be bound to any key you like. :smile:
 | `builtin.lsp_workspace_symbols`     | Searches in LSP all workspace symbols.                                                      |
 | `builtin.lsp_code_actions`          | Lists LSP action to be trigged on enter.                                                    |
 | `builtin.lsp_range_code_actions`    | Lists LSP range code action to be trigged on enter.                                         |
-| `builtin.lsp_diagnostics`           | Lists LSP Diagnostics in the current document.                                              |
+| `builtin.lsp_buffer_diagnostics`    | Lists LSP Diagnostics in the current document.                                              |
```
elianiva, line 412: Maybe it's just me being a little bit too nitpicky, but I think this is better because it matches with the description :)

```suggestion
| `builtin.lsp_document_diagnostics`    | Lists LSP Diagnostics in the current document.                                            |
```

lua/telescope/builtin/init.lua
```diff
@@ -64,7 +64,13 @@ builtin.spell_suggest = require('telescope.builtin.internal').spell_suggest
...
 builtin.lsp_document_symbols = require('telescope.builtin.lsp').document_symbols
 builtin.lsp_code_actions = require('telescope.builtin.lsp').code_actions
-builtin.lsp_diagnostics = require('telescope.builtin.lsp').diagnostics
+builtin.lsp_buffer_diagnostics = require('telescope.builtin.lsp').diagnostics
+builtin.lsp_workspace_diagnostics = function(opts)
+  opts = vim.F.if_nil(opts, {})
+  opts.hide_filename = false
+  opts.get_all = true
+  require('telescope.builtin.lsp').diagnostics(opts)
+end
```
Conni2461, line 73: Not the biggest fan of code in this file. Move it to lsp and if you have problems with the capabilities check you can use sunjons `feature_map` to go from `document_diagnostics` -> `diagnostics` and from `workspace_diagnostics` -> `diagnostics`
Thanks :)

lua/telescope/builtin/init.lua
```diff
@@ -64,7 +64,13 @@ builtin.spell_suggest = require('telescope.builtin.internal').spell_suggest
 builtin.lsp_references = require('telescope.builtin.lsp').references
 builtin.lsp_document_symbols = require('telescope.builtin.lsp').document_symbols
 builtin.lsp_code_actions = require('telescope.builtin.lsp').code_actions
-builtin.lsp_diagnostics = require('telescope.builtin.lsp').diagnostics
+builtin.lsp_buffer_diagnostics = require('telescope.builtin.lsp').diagnostics
+builtin.lsp_workspace_diagnostics = function(opts)
+  opts = vim.F.if_nil(opts, {})
+  opts.hide_filename = false
```
Conni2461, line 70: I would allow an user override here.

lua/telescope/make_entry.lua
```diff
@@ -890,12 +891,10 @@ function make_entry.gen_from_lsp_diagnostics(opts)
       string.format("%s %s", signs[entry.type], pos),
       string.format("LspDiagnosticsDefault%s", entry.type)
     }
-    -- remove line break to avoid display issues
-    local text = string.format("%-" .. line_width .."s", entry.text:gsub(".* | ", ""):gsub("[\n]", ""))
 
     return displayer {
       line_info,
-      text,
+      entry.text,
```
Conni2461, line 897: Is there still a chance where we get `\n` in display? Because i also ran into this a couple of weeks ago and this breaks things. So we can just do `gsub` just in case.

README.md
fdschmidt93: Fully agree! Too hastily on my end :)

lua/telescope/builtin/init.lua
fdschmidt93: Agreed!

lua/telescope/builtin/init.lua
fdschmidt93: On second thought (first being this commit x) I fully agree!

`capabilities_check` is omitted since diagnostics are not there captured (cf. screenshot):

sumneko_capabilities

As per [upstream implementation](https://github.com/neovim/neovim/blob/df4440024bb1f1ce368f5e5844d8af925e264b63/runtime/lua/vim/lsp/diagnostic.lua#L330) an empty table should be returned which ultimately triggers `print('No diagnostics found')`.

Would you want this message to be more explicit in such a case, i.e. in case the server doesn't support diagnostics at all?

Conni2461: Oh i just assumed it would be in capabilities :laughing: . I think message is good enough.

fdschmidt93:
This _should_ be fine now take :)

lua/telescope/builtin/lsp.lua
```diff
@@ -198,8 +198,9 @@ lsp.diagnostics = function(opts)
     return
   end
 
+  opts.hide_filename = utils.get_default(opts.hide_filename, true)
   pickers.new(opts, {
-    prompt_title = 'LSP Diagnostics',
+    prompt_title = utils.get_default(opts.prompt_title, 'LSP Document Diagnostics'),
```
Conni2461, line 203: This `utils.get_` shouldn't be necessary. We are always preferring the prompt title that is in opts.

Example: `Telescope find_files prompt_title=asdf

fdschmidt93: ...

fixed :D

Conni2461 (approved):
LGTM.
Thanks :)

Will try first after that i'll merge

nfjdu:
Is there a way to setup this to work with 1 vim tab only? like i have 3 tabs, there are 3 dif git projects opened, and this deagnistic show errors of all 3 projects when i open it. Sorry for maybe dumb question, not vim ninja yet unfortunately. Thanks

PR #606, opened 2021-03-04
feat: smart send to qflist

Conni2461:
Is this `smart_send` useful to anyone otherwise i remove it from this PR. Basically what it does it sends all to qflist if nothing multi-selected otherwise send multi selected. I am using it since 3 days and thought i can just upstream it. :)

Also:
- scrolling is half window height because its bothering me for quite some time
- Remove some stuff from readme and fix that experimental default previewer mistake

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,7 +398,6 @@ Built-in functions. Ready to be bound to any key you like. :smile:
 | `builtin.highlights`                | Lists all highlights.                                                                       |
 | `builtin.current_buffer_fuzzy_find` | Searches in current buffer lines.                                                           |
 | `builtin.current_buffer_tags`       | Lists Tags in current buffer.                                                               |
-| ..................................  | Your next awesome picker function here :D                                                   |
 
 ### LSP Pickers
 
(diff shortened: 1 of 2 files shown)


elianiva:
what a coincidence, I was about to submit a PR about the scrolling speed because I can't stand it anymore :rofl:
do we want it to be customisable though? mine has an option for that on `layout_defaults` and uses `math.floor` because apparently it freaks out when it receives a float value (like, it inserts some random characters to the previewer for some reason)
if not, we could use yours

Conni2461:
Feel free to checkout this branch and finish it, i currently don't have time to add config value as well. Just open a new PR and close this one, i don't need the commit :) 

`math.floor(lines / 2)` I also had a floor.

elianiva:
> math.floor(lines / 2) I also had a floor.

ok how did I miss that haha, I guess I should've gone to sleep instead of fixing this annoying thing 😆 
will continue to work on that tomorrow

elianiva:
#610

PR #607, opened 2021-03-04
fix: error from unset LspDiagnosticSign & introduce opts.no_sign

### LINKED ISSUE #603: Error on LSP workspace diagnostics when unsetting LSP sign
### Description

Using lsp_workspace_diagnostics give error when unsetting LSP sign
**Expected Behavior**

It works fine
**Actual Behavior**

Error like : 
image

### Details

1. Start nvim and LSP on a project
2. :lua vim.fn.sign_define("LspDiagnosticsSignError", {text = "", numhl = "LspDiagnosticsDefaultError"})
3. :Telescope lsp_workspace_diagnostics
</details>

- nvim --version output: NVIM v0.5.0-dev+1117-g6995fad26
- Operating system: Linux
- Telescope commit: 908752fc67f0aff43e35ffb175b5e6226874bad5

</details>

<p>

```lua
local vim = vim
vim.cmd [[packadd packer.nvim]]

require('packer').startup(function()

  use {'neovim/nvim-lspconfig'}
  use {
    requires = {'nvim-lua/popup.nvim','nvim-lua/plenary.nvim'}
  }
end)

require('telescope').setup{}
require('lspconfig').rust_analyzer.setup {}

```

fdschmidt93:
This should fix #603  as per discussion in the issue, specifically

- The signs if empty default to {E, W, I, H}
- `opts.no_sign` can be used to turn off the sign entirely
- Additionally: line header width is automatically set and now also respects 1'000s of lines adequately

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -862,7 +862,13 @@ function make_entry.gen_from_lsp_diagnostics(opts)
   if not opts.no_sign then
     signs = {}
     for _, v in pairs(lsp_type_diagnostic) do
-      signs[v] = vim.trim(utils.get_default(vim.fn.sign_getdefined("LspDiagnosticsSign" .. v)[1].text, v:sub(1,1)))
+      -- pcall to catch entirely unbound or cleared out sign hl group
+      local status, sign = pcall(
+        function() return vim.trim(vim.fn.sign_getdefined("LspDiagnosticsSign" .. v)[1].text) end)
+      if not status then
+        sign = v:sub(1,1)
+      end
+      signs[v] = sign
     end
   end
 
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -858,16 +858,19 @@ function make_entry.gen_from_lsp_diagnostics(opts)
   opts = opts or {}
   opts.tail_path = utils.get_default(opts.tail_path, true)
 
-  local signs = {}
-  for _, v in pairs(lsp_type_diagnostic) do
-    signs[v] = vim.trim(vim.fn.sign_getdefined("LspDiagnosticsSign" .. v)[1].text)
+  local signs
+  if not opts.no_sign then
+    signs = {}
+    for _, v in pairs(lsp_type_diagnostic) do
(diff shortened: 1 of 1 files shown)


Conni2461:
Still happens. Because `vim.fn.sign_getdefined` will still return nil and a table access `[1]` will throw that error. 

Steps to reproduce.
- Start nvim
- Open File
- Wait for lsp
- Telescope lsp_document_symbols everything works
- `call sign_undefine("LspDiagnosticsSignWarning")`
- Telescope lsp_document_symbols error

lua/telescope/make_entry.lua
```diff
@@ -858,16 +858,24 @@ function make_entry.gen_from_lsp_diagnostics(opts)
...
 
-  local signs = {}
-  for _, v in pairs(lsp_type_diagnostic) do
-    signs[v] = vim.trim(vim.fn.sign_getdefined("LspDiagnosticsSign" .. v)[1].text)
+  local signs
+  if not opts.no_sign then
+    signs = {}
+    for _, v in pairs(lsp_type_diagnostic) do
+      local status, sign = pcall(
+        function() return vim.trim(vim.fn.sign_getdefined("LspDiagnosticsSign" .. v)[1].text) end)
```
Conni2461, line 866: Wasn't an argument for `pcall` you can do this for example.
```lua
local get_table = function()
  return nil
  -- return { 'a' }
end

print((get_table() or {})[1])
```

Tell me what you like more, after that i can either merge or you update

Conni2461: The thing is we basically just need a table so `x or {}` should just work

fdschmidt93: I'll have to admit `pcall` is more simple here as otherwise there are two issues to fix:
1. The original issue empties out the text of the highlighting group with "" resulting in `SignTable[1].text` being nil - that was what I originally fixed
2. Your issue unbinds the diagnostic making`SignTable` nil, which is just as valid (maybe more) though probably less common than original issue

I'd quickly recommit though with comment as to why pcall is required. Sounds good?

Conni2461: Sounds good

Conni2461 (approved):
LGTM

Thanks :)

PR #610, opened 2021-03-05
feat: define scroll speed + smart_send to qflist

elianiva:
continuing @Conni2461's work #606

you can now do something like this to customise the scrolling speed, the default is `preview_height / 2`

```lua
layout_defaults = {
  horizontal = {
    scroll_speed = 5
  },
  vertical = {
    scroll_speed = 2
  },
}
```

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -143,16 +143,21 @@ function actions.toggle_selection(prompt_bufnr)
   current_picker:toggle_selection(current_picker:get_selection_row())
 end
 
+local get_scroll_speed = function(prompt_bufnr)
+  local status = state.get_status(prompt_bufnr)
+  local default_speed = vim.api.nvim_win_get_height(status.preview_win) / 2
+
+  return status.picker.layout_config.scroll_speed or default_speed
+end
+
 function actions.preview_scrolling_up(prompt_bufnr)
-  -- TODO: Make number configurable.
-  local lines = vim.api.nvim_win_get_height(state.get_status(prompt_bufnr).preview_win)
-  action_state.get_current_picker(prompt_bufnr).previewer:scroll_fn(math.floor(-(lines / 2)))
+  local speed = get_scroll_speed(prompt_bufnr)
+  action_state.get_current_picker(prompt_bufnr).previewer:scroll_fn(-math.floor(speed))
 end
 
 function actions.preview_scrolling_down(prompt_bufnr)
-  -- TODO: Make number configurable.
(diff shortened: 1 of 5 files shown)


lua/telescope/actions/init.lua
```diff
@@ -143,14 +143,21 @@ function actions.toggle_selection(prompt_bufnr)
   current_picker:toggle_selection(current_picker:get_selection_row())
 end
 
+local get_scroll_speed = function(prompt_bufnr)
```
Conni2461, line 146: I am thinking maybe we should move this to `action_set`. Similar to `set.shift_selection = function(prompt_bufnr, change)` and all we need to pass in is direction (-1 or 1)

Conni2461: Otherwise LGTM. Thanks for picking it up :)

lua/telescope/actions/set.lua
```diff
@@ -124,6 +125,18 @@ set.edit = function(prompt_bufnr, command)
   end
 end
 
+--- Scrolls the previewer up or down
+---@param prompt_bufnr number: The prompt bufnr
+---@param direction number: The direction of the scrolling
+--      Valid directions include: 1 or -1
```
Conni2461, line 131: FYI Multiline param currently not possible with `tree-sitter-lua` See https://github.com/tjdevries/tree-sitter-lua/pull/14#discussion_r587908659

Conni2461:
2 things about action docs:
1. There is this: https://github.com/nvim-telescope/telescope.nvim/pull/587 I just don't wanna take away his first PR. Sorry i know you spend time documenting `actions/init.lua`. We can keep the `set.lua` documentation :)
2. We have the max query issue until neovim is using tree sitter 0.19.0 and tj regenerated grammar See https://github.com/nvim-treesitter/nvim-treesitter/issues/963

elianiva:
sure, I can remove them. I didn't spend that much time, it's just a lot of copypaste because I was thinking *why not?* :laughing:

lua/telescope/actions/set.lua
elianiva: I should try that grammar as well, been wanting to try that but I always forgot :D

Conni2461: Wouldn't recommend it quite yet. I am using a quite out of date `nvim-treesitter` to make it work.
I would wait until overriding a query is easier in `nvim-treesitter` and the queries for that grammar need some work too.

The only thing that you would gain from it is the emmy documentation part and you can play with `docgen` which is super cool :)

Conni2461 (approved):
LGTM

Thanks for picking it up :)

PR #613, opened 2021-03-05
[WIP] feat: add rename + copy for file_browser

### LINKED ISSUE #584: Renaming files and folder on Telescope file_browser
It would be really nice to have a feature that pressing <C-r> on a folder or file, it redirect to the prompt where you can type the new name

elianiva:
close #584

this *kinda* works. Still need to deal with the `cwd` being weird like `path/to/file/..`
the default binding is `<C-r>` for rename
`<A-y>` to `mark` the selected files
`<A-p>` to `copy` the `marked` files (using `<C-y>`)
`<A-m>` to `move` the `marked` files

I picked these keybinds randomly, feel free to give a better suggestions.
requires https://github.com/nvim-lua/plenary.nvim/pull/90

cc @Conni2461 @FabioRamatis @sharksforarms @timsofteng


timsofteng:
Does <A> means alt? If so then I guess it's not best idea because many people use alt as leader key for window-manager.
I don't know about insert mode but for normal mode y, p and m single keys looks good.
Btw, kybindindg is a difficult architectural process.

sharksforarms:
Do they require a default binding? Since this could be considered "extra" it may be ok to let users configure it if they want it?

(I would probably change it from the defaults)

timsofteng:
> Do they require a default binding? Since this could be considered "extra" it may be ok to let users configure it if they want it?
> 
> (I would probably change it from the defaults)

Of course we can map and remap everything manually. But in my opinion default mapping in any piece of software is a important part of architecture. 
Look at the editor which you use now:) Default keys and shortcuts are one of a main reasons why you choose (neo)vim, aren't it? :)

elianiva:
> So if tab is using for marking everywhere why we just can't using it to mark files for actions in file browser?

Right, I shouldn't make yet another action for that. 😆 
Will fix that in a moment

lua/telescope/builtin/files.lua
```diff
@@ -267,8 +267,105 @@ files.file_browser = function(opts)
...
+        end
+
+        return selected
+      end
+
+      local move_file = function()
+        local current_picker = action_state.get_current_picker(prompt_bufnr)
+
+        for _, file in ipairs(get_marked_files()) do
+          local filename = file.filename:sub(#file:parents() + 2)
```
elianiva, line 301: this is what I meant by 
> this kinda works. Still need to deal with the cwd being weird like path/to/file/..

This trick works but I sometimes get `path/to/file/../../`. It only happens once and I couldn't find a way to reproduce it though.

Conni2461: cwd is my bad. I stumbled across this a couple of days ago but didn't want to make a PR just for this.
You need to swap out some `vim.fn.expand` to `vim.loop.fs_realpath` because `expand` will not reduce the path.
So example `/home/conni/repos/..` will always be the same and each move will just put the new dir at the end so you might end up with `/home/conni/repos/../repos/neovim/src/..` rather than just `/home/conni/repos/neovim` let me send you a diff :)

```diff
diff --git a/lua/telescope/builtin/files.lua b/lua/telescope/builtin/files.lua
index 9749635..76d6bd8 100644
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -238,7 +238,7 @@ files.file_browser = function(opts)
       action_set.select:replace_if(function()
         return action_state.get_selected_entry().path:sub(-1) == os_sep
       end, function()
-        local new_cwd = vim.fn.expand(action_state.get_selected_entry().path:sub(1, -2))
+        local new_cwd = vim.loop.fs_realpath(action_state.get_selected_entry().path:sub(1, -2))
         local current_picker = action_state.get_current_picker(prompt_bufnr)
         current_picker.cwd = new_cwd
         current_picker:refresh(gen_new_finder(new_cwd), { reset_prompt = true })
@@ -262,7 +262,7 @@ files.file_browser = function(opts)
           vim.cmd(string.format(':e %s', fpath))
         else
           Path:new(fpath:sub(1, -2)):mkdir({ parents = true })
-          local new_cwd = vim.fn.expand(fpath)
+          local new_cwd = vim.loop.fs_realpath(fpath)
           current_picker.cwd = new_cwd
           current_picker:refresh(gen_new_finder(new_cwd), { reset_prompt = true })
         end
```

Conni2461: I also ran into some other issues. For example going into a symbolic link if the resolved link is a directory. Because `vim.loop.fs_scandir_next` (don't remember the function) will return either file or directory. And a link is from typ file. So we don't insert a `/` at the end so `replace_if` will return false and we run `action_set.select` :| So we might wanna change that check to, `vim.loop.fs_stat` (not `lstat` because we are not interested in the stats of the link but of the resolved one) and then check `type`

Conni2461: If you want i can commit these changes to this branch, otherwise you can try to fix it yourself :)

elianiva: oh please :)
I still got some stuff to do this evening

elianiva:
ohno, I messed up the rebase again :rofl:

elianiva:
@timsofteng I've implemented `h` to go up 1 dir, `<BS>` to go up 1 dir without any limit, `l` same as `<CR>` to make navigating through directory nicer, `<S-h>` to toggle hidden files.

lua/telescope/builtin/files.lua
```diff
@@ -201,11 +201,17 @@ files.file_browser = function(opts)
 
   opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or vim.loop.cwd()
 
-  local gen_new_finder = function(path)
-    opts.cwd = path
+  local gen_new_finder = function(new_opts)
+    opts.cwd = new_opts.path
+    opts.hidden = new_opts.hidden
```
Conni2461, line 206: Question. First run i call `Telescope file_browser hidden=true` does hidden still gets applied becuase `new_opts` looks like this. `{ path = cwd }` so `opts.hidden` will be nil afterwards. I would do `opts.hidden = opts.hidden or new_opts.hidden` here.

Haven't tried it tho

Conni2461: Btw everything in this PR looks really promising :) Great work.
I will try this PR in the tomorrow and give you more feedback. How much and what is still missing?

elianiva: I think everything is done, just need some more feedbacks :)

lua/telescope/builtin/files.lua
```diff
@@ -230,17 +237,27 @@ files.file_browser = function(opts)
 
   pickers.new(opts, {
     prompt_title = 'Find Files',
-    finder = gen_new_finder(opts.cwd),
+    finder = gen_new_finder({ path = opts.cwd, hidden = opts.hidden }),
```
elianiva, line 240: we should pass user's hidden opt here because if I do `opts.hidden or new_opts.hidden` then the hidden toggler stopped working

tami5 (approved):

lua/telescope/builtin/files.lua
```diff
@@ -261,15 +277,155 @@ files.file_browser = function(opts)
...
+          local filename = file.filename:sub(#file:parents() + 2)
+
+          local success = file:copy({
+            destination = Path:new({
+              current_picker.cwd, filename
+            }).filename
+          })
+
+          if not success then
+            print("The file is already exists!")
```
elianiva, line 360: This can cause a *false alarm*, eg: it's copied succesfully but it prints this warning
removing this fixed it so I'm gonna remove it. Not sure if removing it is a good idea or not.

elianiva:
wait, did I mess up or something?

Conni2461:
Just docgen update. See https://github.com/tjdevries/tree-sitter-lua/commit/ca38705ab4551db996428ebce6bed9bdc9b77780

https://github.com/nvim-telescope/telescope.nvim/pull/659 should take care of it. CI is mad because of one line being to long and multiline `@param`/`@field` isn't done yet. See https://github.com/tjdevries/tree-sitter-lua/pull/15

oberblastmeister:
I'm not sure that I like the fact that we have to refresh the pickers every time we delete a file or rename a file. This seems a bit wasteful for performance. We should have some api where we can delete an entry from the picker or add an entry. This would make other actions much easier to implement like deleting buffers from the buffers picker. This would also allow file system operations to be async as we can delete or rename a file and then apply that change to the picker without having to wait for the operation to finish.

We should also have more options for the `map` function in `attach_mappings`. Currently the mappings don't work very well because they don't have `<nowait>`. Also, I don't like that we are mapping `d` for a destructive operation, even if there is confirmation.  Maybe we should map it do `dd`.

Maybe we should wait for plenary async await to be merged? That way we don't have to go through a transition. What do you think?

elianiva:
yes, I also don't like the refresh thingy.

Conni2461:
I already planed to implement `remove` (for buffer this week see https://github.com/nvim-telescope/telescope.nvim/issues/689#issuecomment-808075081) i can add `append`, `change` to the list.

What i don't understand is why we would need more options for map to enable async. I think something like this should be possible, assuming i am done with `remove`, etc ...

```lua
local rename = function()
  local remove = action_state.remove_current_selected()
  local picker = action_state.get_current_picker()
  Path:new(remove):rename({ new_name = calculate, callback = function(new)
    picker:change:start()
    picker:change:remove(remove)
    picker:change:append(new)
    -- picker:change:sort() -- maybe we want to sort alphabetically because the new name can 
    -- be at any place in the list we just don't know where (in this function)
    -- maybe we can also do a `picker:change:insert_sorted(new, optional_sort_function)` or so
    picker:change:submit() -- we would only need to schedule this part
  end})
end
```

Currently i can't imagine a map option for async which would provide more flexibility. Do you only want to start your async env when this option is set?  Thats why i don't see why we would need to wait for `async await` PR.

The only thing i could imagine being async (and `async await` could help with that) would be `replace_if`. Because now we have a `uv.fs_stat.type == 'directory'` check for each file we are not sure if its a dir. Mainly symbolic links. `uv.fs_scandir` just doesn't provide this information. So a `async_replace_if` could evaluate the condition async and based on that result make the decision if the function will be replaced or not.

I think we can also make refresh part async. I have a scratch using the rebased `incremental_finder` in #288 https://cdn.discordapp.com/attachments/785532784927703070/819644125585735700/2021-03-11_19-50-57.mkv
First run `plenary.scandir`, second one `fd`. I probably should just finish it for him. I currently just have no idea what is happening with all that `async wait` and the resulting changes to finders :laughing: Break was just too long and everyone else did so much work :) Good job

oberblastmeister:
We also need to do buffer handling in plenary. Currently renaming a path will not check if the buffer is opened in nvim and reload a new buffer. We should check for buffers by default and have an option to disable it, like `Path:rename { buf = false }`

elianiva:
I messed up again :laughing:

danymat:
Hello, I would love to know the current state of this issue, it seems very interesting and could be a full replacement to Nerdtree (ATM I use nerd tree for editing in directory and telescope file browser to navigate)

FabioRamatis:
It would be great to add full bindings on doc/telescope.txt

Like for rename, copy, move, delete, etc...

timsofteng:
> It would be great to add full bindings on doc/telescope.txt
> 
> Like for rename, copy, move, delete, etc...

The PR still opened...

elianiva:
I'm still quite busy atm, maybe will continue it in a few weeks, but if anyone want to takeover then feel free :)

m-bal:
What work is left to be done for this issue besides adding full bindings on doc/telescope.txt?

AllenDang:
I just think about same idea today. Waiting for this PR now...

elianiva:
I think I'm done, not sure if I want the action to be performed async because I don't want the user to interact with a file that's currently being processed. 

My main concern is people complaining their file is gone because Telescope LOL

lua/telescope/builtin/files.lua
```diff
@@ -339,17 +345,22 @@ files.file_browser = function(opts)
...
-    finder = opts.new_finder(opts.cwd),
+    finder = opts.new_finder { path = opts.cwd },
     previewer = conf.file_previewer(opts),
     sorter = conf.file_sorter(opts),
     attach_mappings = function(prompt_bufnr, map)
       action_set.select:replace_if(function()
         return is_dir(action_state.get_selected_entry().path)
       end, function()
-        local new_cwd = vim.fn.expand(action_state.get_selected_entry().path:sub(1, -2))
+        local new_cwd = vim.loop.fs_realpath(action_state.get_selected_entry().path)
```
Conni2461, line 355: I think i proposed this fix because i thought its a good idea to normalize the path with it but i actually think that might be wrong. `plenary.path:normalize` should exist now.

I am not sure. how we should handle symbolic links if i am in `/home/conni/a` and open file `link` which goes to `/home/conni/b` then i expect `../` to take me back to `/home/conni/a` right and not to `/home/conni`

Conni2461 (commented):
Maybe we should move some of the actions in a separate module in telescope.actions so they are remappable.

Also we are missing docs in `telescope/builtin/init.lua` describing these default mappinsg

krishenm94:
very cool! can't wait for this to get merged!

fdschmidt93:
@elianiva if still ok with you, I've started picking this up in a local branch (hopefully pushed as WIP later today) and incorporate also some ideas / requests that came up on reddit/gitter.

elianiva:
@fdschmidt93 I'd be more than happy if you do so, thanks. 
Sorry for not much activity here, been busy doing other stuff :laughing:

Kadrian:
Is there some way to get the 

> `l` same as `<CR>` ? Has that been shipped?

fdschmidt93:
First, https://github.com/nvim-telescope/telescope-file-browser.nvim
Then, https://github.com/nvim-telescope/telescope-file-browser.nvim#remappings

Maybe eventually I reorganize mappings then make `l` in normal mode to select, for now, users need to remap it themselves.

PR #622, opened 2021-03-06
Bugfix: `Error executing lua Vim:E731: using Dictionary as a String`

oberblastmeister:


lua/telescope/actions/init.lua
```diff
@@ -413,7 +413,7 @@ end
 
 actions.smart_send_to_qflist = function(prompt_bufnr)
   local picker = action_state.get_current_picker(prompt_bufnr)
-  if table.getn(picker:get_multi_selection()) > 0 then
+  if #(picker:get_multi_selection()) > 0 then
```
Conni2461, line 416: I don't like this. For me `#` is for string length and `table.getn` for table length. Makes the code cleaner in my opinion.

I know its deprecated and removed in lua 5.2 but that won't happen to luajit

Conni2461:
Out of curiosity, when does this error occur?

oberblastmeister:
This error occurs because sometimes the `value` key of an entry can be a table instead of a string. The `entry_to_qf` function used to put the text as `entry.value`, which could be a table. Now if checks if it is a table.

One `make_entry` function that can cause this is `make_entry.gen_from_file`. Calling `entry_to_qf` will generate an entry of the structure:
```lua
{
  bufnr = ...
  filename = "a/path"
  lnum = ...
  col = ...
  text = {
    text = "some text here"
    filename = "a/path/
  }
}
```
this is because text was directly set to value, which was a table. This brings into question whether the value key of an entry should be a table or whether you should just add more keys directly to the entry and not in a new table inside value.

elianiva:
nice, I also got this error when sending lsp_*_diagnostics to qflist but haven't got time to check what's wrong

Conni2461:
I got the code and the error, i was only interested with which picker it happens, so you answer overshot my question by miles :laughing: 

LGTM, thanks :)

Issue #261, opened 2020-11-18
prompt: can not delete typed text in insert mode (after previously in normal mode)


doubleloop:
# Steps to reproduce:

minimal vimrc:
```vim
call plug#begin('~/.local/share/nvim/plugged')
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-lua/telescope.nvim'
call plug#end()
```
```sh
$ vim -u testrc.vim +"Telescope find_files"
```
type something (ex "foo") in prompt field, press `<esc>` for normal mode  then `a` to insert again. Then try to delete typed text (`backspace` or `c-w`). It does not work. Type further (ex "bar"). Then remove with backspace (without switching to normal mode). You can remove only second part ("foobar" -> "foo") but first part can not be removed (unless you go to normal mode and use some normal mode deletion like `S` or `dbw`)

tjdevries:
Yeah, I believe this is an issue with prompt buffers. I think the bug report should go to neovim/neovim if there isn't one there already. My guess is that we're missing some patches :/ I have experienced the same thing.

mah-shee:
I suppose this issue is related to https://github.com/neovim/neovim/issues/13715  FYI

Conni2461:
This issue is fixed in the latest neovim master: neovim/neovim#16421. Its not part of the 0.6 release

PR #626, opened 2021-03-07
feat: add completion to pre-filtering

fdschmidt93:
This PR adds simple insert mode completion to `tag` prefiltering to both (a) cycle through (in e.g. empty prompt or prompt merely comprises a tag) or (b) incrementally complete `tags` by

1. Collecting `tags` in a set table during scoring
2. Trigger `action` to set insert completion for prompt with `complete` **

https://user-images.githubusercontent.com/39233597/110256294-57385c00-7f98-11eb-86e9-9f647bb2a659.mp4

Accepting a match currently is done via (a)`<C-y>` (standard vim binding), (b) continue typing with open menu or (c) `ESC` as `<CR>` is of course bound to jumping to line of entry. 

Known issue:

- Backspace freezes after going into normal mode (e: i.e. accepting completion with `ESC`): one can still cycle through but only delete prompt from normal mode (i.e. `cc`) which I guess is related to #261 

** Unlike `:help complete` suggests, it works without `<C-R>` as evident by mapping set to `<C-l>` :)

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -420,29 +420,46 @@ actions.smart_send_to_qflist = function(prompt_bufnr)
   end
 end
 
-actions.complete = function(prompt_bufnr)
-  local tags = action_state.get_current_picker(prompt_bufnr).sorter.tags
-  local delimiter = action_state.get_current_picker(prompt_bufnr).sorter._delimiter
+actions.complete_tag = function(prompt_bufnr)
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local tags = utils.get_default(current_picker.sorter.tags, {})
+  local delimiter = current_picker.sorter._delimiter
 
-  if not tags then return end
-
-  local symbols = {}
-  local line = action_state.get_current_line()
-  for t, _ in pairs(tags) do
-    table.insert(symbols, string.format('%s%s%s ', delimiter, t:lower(), delimiter))
+  if vim.tbl_isempty(tags) then
+    print('No tags found or associated with picker.')
+    return
   end
 
-  local filtered_symbols
+  -- format tags to match filter_function
+  local prefilter_tags = {}
+  for tag, _ in pairs(tags) do
+    table.insert(prefilter_tags, string.format('%s%s%s ', delimiter, tag:lower(), delimiter))
+  end
+
+  local line = action_state.get_current_line()
+  local filtered_tags = {}
   -- retrigger completion with already selected tag anew
   -- trim and add space since we can match [[:pattern: ]]  with or without space at the end
-  if vim.tbl_contains(symbols, vim.trim(line) .. " ") then
-    filtered_symbols = symbols
+  if vim.tbl_contains(prefilter_tags, vim.trim(line) .. " ") then
+    filtered_tags = prefilter_tags
   else
-    filtered_symbols = utils.filter_completion(line, symbols)
+    -- match tag by substring
+    for _, tag in pairs(prefilter_tags) do
+      local start, _ = tag:find(line)
+      if start then
+        table.insert(filtered_tags, tag)
+      end
+    end
   end
-  local col = vim.api.nvim_win_get_cursor(0)[2] + 1
 
-  vim.fn.complete(col - #line, filtered_symbols)
+  if vim.tbl_isempty(filtered_tags) then
+    print('No matches found.')
+    return
+  end
+
+  -- col - #line bytes are replaced by accepted match
+  local col = vim.api.nvim_win_get_cursor(0)[2] + 1
+  vim.fn.complete(col - #line, filtered_tags)
 
 end
 
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -457,7 +457,7 @@ actions.complete_tag = function(prompt_bufnr)
     return
   end
 
-  -- col - #line bytes are replaced by accepted match
+  -- incremental completion by substituting string starting from col - #line byte offset
   local col = vim.api.nvim_win_get_cursor(0)[2] + 1
   vim.fn.complete(col - #line, filtered_tags)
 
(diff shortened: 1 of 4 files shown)


lua/telescope/actions/init.lua
```diff
@@ -420,6 +420,33 @@ actions.smart_send_to_qflist = function(prompt_bufnr)
...
 
+actions.complete = function(prompt_bufnr)
+  local tags = action_state.get_current_picker(prompt_bufnr).sorter.tags
+  local delimiter = action_state.get_current_picker(prompt_bufnr).sorter._delimiter
+
+  if not tags then return end
+
+  local symbols = {}
+  -- :sub(2, -1): remove preceding ">"
+  local line = vim.api.nvim_buf_get_lines(prompt_bufnr, 0, 1, false)[1]:sub(2,-1) or ""
```
Conni2461, line 431: Cool feature. I will look at it closer later.
But thats one thing i immediately noticed. This will not work if the prompt is different or not existing.
There is `action_state.get_current_line()` or so that will do give you the line without `prompt_prefix`

fdschmidt93: Thanks! Changed accordingly :)

lua/telescope/actions/init.lua
```diff
@@ -420,6 +420,49 @@ actions.smart_send_to_qflist = function(prompt_bufnr)
   end
 end
 
+actions.complete_tag = function(prompt_bufnr)
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local tags = utils.get_default(current_picker.sorter.tags, {})
+  local delimiter = current_picker.sorter._delimiter
+
+  if vim.tbl_isempty(tags) then
+    print('No tags found or associated with picker.')
```
Conni2461, line 429: I am not sure if we should not just exit silently when someone presses `<c-l>`, for example on accident in `find_files`.

lua/telescope/utils.lua
```diff
@@ -392,4 +392,17 @@ utils.get_devicons = (function()
   end
 end)()
 
+function utils.create_tag_set(tag)
+  tag = vim.F.if_nil(tag, 'ordinal')
+  local set = {}
+  return setmetatable(set, {
+    __index = {
```
Conni2461, line 399: Just thinking out load. There is also `__newindex` do we need to handle that as well?

Conni2461:
> Backspace freezes after going into normal mode (e: i.e. accepting completion with ESC): one can still cycle through but only delete prompt from normal mode (i.e. cc) which I guess is related to #261

Yeah its a neovim upstream issue with prompt buffers. So not setting a `prompt_prefix` fixes that. The vim patch that addressed this issue was `v8.2.1976` but that one cant port until this was ported first:
```
[conni@home neovim]$ ./scripts/vim-patch.sh -L --grep prompt -- src
v8.1.0035
v8.1.0042
v8.1.0064
v8.1.1813
v8.2.0052
v8.2.0198
v8.2.0301
v8.2.0911
v8.2.0946
v8.2.0948
v8.2.0950
v8.2.1147
v8.2.1588
v8.2.1781
v8.2.1844
v8.2.1976 <- This patch fixes your issue
v8.2.2014
```

And the first one `v8.1.0035` is not that easy to port. See problems here: https://github.com/neovim/neovim/pull/13027

lua/telescope/actions/init.lua
fdschmidt93: I initially followed your line of thought though then personally preferred verbosity plus possibly accounting for future scenarios I can't think of yet.

Nevertheless, your point of view is totally fair since, at least for the time being, there should not be scenarios in which `actions.complete_tag` can be intentionally called _and_ `tags` can be an empty table. An empty table should also eventually trigger `print(No matches found.)`.

So, let's settle for the below?

```lua
  local tags = current_picker.sorter.tags
  local delimiter = current_picker.sorter._delimiter

  -- silently exit for pickers that do not prefilter
  if not tags then return end
```

Conni2461: Yeah i know i said hey lets give user feedback if he calls `Telescope lsp_*_diagnostics` but i am not sure how we should handle it here. Like someone hits it accidentally in `find_files` and just sees a random message about `tags`. 

But yeah i see what you saying, i also never know how to handle these cases

lua/telescope/utils.lua
fdschmidt93: No, from reading documentation my understanding is that we don't need it here. Haven't used it and I quickly tried reeimplementing a set with `__newindex`. I might be clumsy but that use case is circular, since `__newindex` defines behavior for keys that are absent from the table. When trying to set a key for a table that's not there in `__newindex` it gets reinvoked. :sweat_smile: 

Sure thing :)

lua/telescope/actions/init.lua
fdschmidt93: No diagnostics for `lsp_*_diagnostics` is a much more common case though. Let's not confuse the 99% of telescope users still unaware of prefiltering with a half-hearted message as per the current commit they lack context to understand and be explicit instead

```lua
  local tags = current_picker.sorter.tags
  local delimiter = current_picker.sorter._delimiter

  if not tags then
    print('No tag pre-filtering set for this picker.')
    return
  end
```

since empty tables on second thought rightfully go to `print('No matches found.')`

If you'd agree we could additionally point to pickers for which tag pre-filtering is set out-of-the-box as a gimmick to increase discoverability of the feature like
`print('No tag pre-filtering set for this picker. Try, for instance, <C-l> for lsp_*_diagnostics or lsp_*_symbols!)` or something similar, since accidentally hitting `<C-l>` is also not that common I suppose, but the above is fine of course, too.

Conni2461: Yeah works now, thanks :)
I was also thinking maybe we should just map that action for functions that actually support pre-filtering.

I am going to test it now :)

Conni2461 (approved):
LGTM i like it :)

Thanks :)

I will wait for an answer to my last two comments after that i would merge

lua/telescope/utils.lua
```diff
@@ -392,4 +392,17 @@ utils.get_devicons = (function()
   end
 end)()
 
+function utils.create_tag_set(tag)
```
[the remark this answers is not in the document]
fdschmidt93: I fully understand - required but not sure where to put it :laughing: , let's do this local in `sorters.lua` then :)

Issue #629, opened 2021-03-08
[Feature request] :Telescope lsp_definition


jaywonchung:
`:Telescope lsp_references` exists because there may (and probably will) be several referenecs to a symbol. As a matter of fact, this is also the case for LSP definition. For instance, the Microsoft `pyright` language server detects overloaded variants of a function and populates the quickfix list.

LSP definitions is a bit different from LSP references because there being only one definition is the absolute common case. In that case, I think it's better to directly jump to the location instead of showing a telescope window that contains only one entry. Thus I think `:Telescope lsp_definition` should first call `vim.lsp.buf_request_sync`, immediately jump if the language server returned only one location, and otherwise show the picker.

Thank you for the great plugin!

Conni2461:
Implemented in https://github.com/nvim-telescope/telescope.nvim/commit/ff428c91879d7d5ffa22d950cd53b5eeff3553dc

Thanks to everyone involved. Closing this

PR #631, opened 2021-03-09
feat: lsp definitions

oberblastmeister:
checkout [this](https://github.com/nvim-telescope/telescope.nvim/issues/629) issue for more info

### DIFF
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -276,6 +276,7 @@ local feature_map = {
   ["code_actions"]      = "code_action",
   ["document_symbols"]  = "document_symbol",
   ["references"]        = "find_references",
+  ["definitions"]        = "definition",
   ["workspace_symbols"] = "workspace_symbol",
 }
 
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -276,7 +276,7 @@ local feature_map = {
   ["code_actions"]      = "code_action",
   ["document_symbols"]  = "document_symbol",
   ["references"]        = "find_references",
-  ["definitions"]        = "definition",
+  ["definitions"]        = "goto_definition",
   ["workspace_symbols"] = "workspace_symbol",
 }
 
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -50,15 +50,12 @@ lsp.definitions = function(opts)
     end
   end
 
-  local locations = vim.lsp.util.locations_to_items(flattened_results)
-
-  if vim.tbl_isempty(locations) then
+  if #flattened_results == 0 then
     return
-  end
-
-  if #locations == 1 then
+  elseif #flattened_results == 1 then
     vim.lsp.util.jump_to_location(flattened_results[1])
   else
+    local locations = vim.lsp.util.locations_to_items(flattened_results)
     pickers.new(opts, {
       prompt_title = 'LSP Definitions',
       finder = finders.new_table {
@@ -69,7 +66,6 @@ lsp.definitions = function(opts)
       sorter = conf.generic_sorter(opts),
     }):find()
   end
-end
 
 lsp.document_symbols = function(opts)
   local params = vim.lsp.util.make_position_params()
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -66,6 +66,7 @@ lsp.definitions = function(opts)
       sorter = conf.generic_sorter(opts),
     }):find()
   end
+end
 
 lsp.document_symbols = function(opts)
   local params = vim.lsp.util.make_position_params()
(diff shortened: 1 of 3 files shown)


lua/telescope/builtin/lsp.lua
```diff
@@ -38,6 +38,39 @@ lsp.references = function(opts)
...
+      prompt_title = 'LSP Definitions',
+      finder = finders.new_table {
+        results = locations,
+        entry_maker = opts.entry_maker or make_entry.gen_from_quickfix(opts),
+      },
+      previewer = conf.qflist_previewer(opts),
+      sorter = conf.generic_sorter(opts),
+    }):find()
+  end
+end
```
jaywonchung, line 72:
```suggestion
  if #flattened_results == 0 then
    return
  elseif #flattened_results == 1 then
    vim.lsp.util.jump_to_location(flattened_results[1])
  else
    local locations = vim.lsp.util.locations_to_items(flattened_results)
    pickers.new(opts, {
      prompt_title = 'LSP Definitions',
      finder = finders.new_table {
        results = locations,
        entry_maker = opts.entry_maker or make_entry.gen_from_quickfix(opts),
      },
      previewer = conf.qflist_previewer(opts),
      sorter = conf.generic_sorter(opts),
    }):find()
  end
```

jaywonchung: Just a small suggestion to avoid calling `vim.lsp.util.locations_to_items` in the common case :)

jaywonchung (approved):
Thank you for your great work!

I've tested the PR as is in my local machine and it works wonderfully, hence the approve. However I did leave one suggestion.

In addition, I think we should add the following item to `feature_map`:
```lua
["definitions"] = "goto_definition",
```

jaywonchung:
Oh, and we should update `README.md`. Is this usually done in a separate PR?

elianiva:
>  Is this usually done in a separate PR?

nope, we should do it in this PR :)
(and probably add a demo for the wiki, too)

oberblastmeister:
I think for the demo someone else can add it for consistency. My vim doesn't look as nice and has a different colorscheme

oberblastmeister:
also what is the feature map

jaywonchung:
https://github.com/nvim-telescope/telescope.nvim/blob/1c80a92d4027c03b65cce1fe27ef664cff77186d/lua/telescope/builtin/lsp.lua#L245-L250

Here! It translates builtin function names to the corresponding LSP capability name.

lua/telescope/builtin/lsp.lua
```diff
@@ -246,6 +276,7 @@ local feature_map = {
   ["code_actions"]      = "code_action",
   ["document_symbols"]  = "document_symbol",
   ["references"]        = "find_references",
+  ["definitions"]        = "definition",
```
fdschmidt93, line 279:
```suggestion
  ["definitions"]        = "goto_definition",
```
I think the corresponding LSP capability is slightly off. You can check them, for instance, with
```lua
print(vim.inspect(vim.lsp.buf_get_clients(0)[1].resolved_capabilities))
```
in which one of the table keys is `goto_definition`

Currently, it should not be able to go to a definition as the feature_map check always returns `false`.

jaywonchung: Yup! :+1::+1:

lua/telescope/builtin/lsp.lua
```diff
@@ -246,6 +276,7 @@ local feature_map = {
   ["code_actions"]      = "code_action",
   ["document_symbols"]  = "document_symbol",
   ["references"]        = "find_references",
+  ["definitions"]        = "goto_definition",
```
jaywonchung, line 279:
```suggestion
  ["definitions"]       = "goto_definition",
```
Aesthetic purposes XD

Conni2461 (approved):
LGTM, haven't tested it tho (been pretty busy lately sorry :)

So if it works, feel free to squash and merge @oberblastmeister

elianiva:
could someone tell me how to get multiple definition results? I was going to make the demo but I failed to get multiple results so it doesn't show the Telescope window

jaywonchung:
```python
from typing import overload, Union

class KVStore:
    def __init__(self):
        self.store = dict()

    @overload
    def set(self, key: bytes, value: int) -> None:
        ...

    @overload
    def set(self, key: bytes, value: str) -> None:
        ...

    def set(self, key: bytes, value: Union[int, str]) -> None:
        self.store[key] = value


if __name__ == "__main__":
    store = KVStore()
    store.set(key=b"KEY", value=None)
```

with the cursor on `set` in the last line.

I use the `pyright` language server: https://github.com/microsoft/pyright

Relevant config:
```lua
if vim.fn.executable('pyright') == 1 then
  require'lspconfig'.pyright.setup{
    on_attach = require'completion'.on_attach,
    settings = {
      python = {
        analysis = {
          typeCheckingMode = "off",
          autoSearchPaths = true,
          useLibraryCodeForTypes = true,
        }
      }
    }
  }
  vim.cmd('autocmd FileType python setlocal omnifunc=v:lua.vim.lsp.omnifunc')
  vim.cmd('autocmd FileType python setlocal signcolumn=yes')
end
```

Screenshot of my editor:
스크린샷 2021-03-10 오후 3 14 23(2)

Hope this helps!

PR #639, opened 2021-03-11
Update actions.open_qflist to use cwindow

rnevius:
`:cwindow` will only open the quickfix window if it has been populated.


Conni2461:
I don't see a reason why we should change the default. I would suggest you write a own custom action for this but that api is currently broken. See https://github.com/nvim-telescope/telescope.nvim/pull/690 (tj is already working on a fix for this)

But i am closing this sorry because the action is called `open_qflist` and i think it should open every time. Also i can't think of a case where `send_to_qflist` doesn't send anything to qflist

PR #644, opened 2021-03-12
Add change directory functionality to file browser & find files

ranjithshegde:
Just a small piece of lua code to change nvim's pwd while using the builtin file_browser
Navigate to the directory and type (capitol) C in normal mode


tami5:
Hey @ranjithshegde, could you please fix indentation to using two space per-tab and force push. Thanks

lua/telescope/builtin/files.lua
```diff
@@ -244,6 +244,12 @@ files.file_browser = function(opts)
         current_picker:refresh(gen_new_finder(new_cwd), { reset_prompt = true })
       end)
 
+      local change_working_directory = function(prompt_bufnr)
+	local dir = actions.get_selected_entry(prompt_bufnr).value
+	vim.fn.execute("cd " .. dir, "silent")
```
tami5, line 249: idk but it might be better to use lchdir or lcd

lua/telescope/builtin/files.lua
```diff
@@ -244,6 +244,12 @@ files.file_browser = function(opts)
         current_picker:refresh(gen_new_finder(new_cwd), { reset_prompt = true })
       end)
 
+      local change_working_directory = function(prompt_bufnr)
```
elianiva, line 247: this `prompt_bufnr` shouldn't be needed because we already have one up above. That's the reason why the CI is failing (you could do `make lint` locally to make sure it passes)

ranjithshegde: You are right! my bad. Let me try this again

oberblastmeister:
I would prefer the mapping `cd` because it reflects what you are doing and also doesn't override a default vim mapping.

ranjithshegde:
> I would prefer the mapping `cd` because it reflects what you are doing and also doesn't override a default vim mapping.

Done

oberblastmeister:
Nice. Can you add this into an action so it can be used by other finders such as `find_files`?

ranjithshegde:
> Nice. Can you add this into an action so it can be used by other finders such as `find_files`?

I will try tomorrow. Will be a bit tricky for find_files..etc as this function gets the value when the selection is a directory, not a file.

ranjithshegde:
Tried to rebase to master, had forgotten about this PR. Will get to it this weekend

ranjithshegde:
> Nice. Can you add this into an action so it can be used by other finders such as `find_files`?

Just did. Please let me know how that looks to you. Sorry this took a while, busy with my thesis

lua/telescope/builtin/files.lua
```diff
@@ -209,6 +209,17 @@ files.find_files = function(opts)
     ),
     previewer = conf.file_previewer(opts),
     sorter = conf.file_sorter(opts),
+    attach_mappings = function(prompt_bufnr, map)
+
+			local change_working_directory = function ()
```
ranjithshegde, line 214: I dont understand why its doing this weird indenting. it looks correct on my vim locally. `ts`, `sw` and `sts` are all set to 2

Conni2461: Because you are using tabs and tabs will always have a width of 8 in github, it doesnt matter what you set in your local vim config.
We are using 2 spaces for indentation so `set expandtab`

lua/telescope/actions/init.lua
```diff
@@ -137,6 +137,13 @@ function actions.remove_selection(prompt_bufnr)
   current_picker:remove_selection(current_picker:get_selection_row())
 end
 
+-- Set working directory to the current selection's dir
+-- For file_browser this corresponds directly to the selection
+-- For find_files this sets it to the directory containing current selection
+function actions.set_cwd_to_selection(path)
```
Conni2461, line 143: no matter what the other people said, this is not an action, this is a mere util function.

A action should ALWAYS be mapable to a picker, so it needs to get the signature `(prompt_bufnr)`, like all other functions in this file.

lua/telescope/builtin/files.lua
```diff
@@ -209,6 +209,17 @@ files.find_files = function(opts)
...
+				local entry = actions.get_selected_entry(prompt_bufnr).value
```
Conni2461, line 215: this is a deprecated function. Please read `action_state.get_selected_entry`

lua/telescope/actions/init.lua
ranjithshegde: I will just remove it from actions and place that inside individual functions. It is a single line anyway. Would that work?

Conni2461: works for me

lua/telescope/builtin/files.lua
```diff
@@ -209,6 +209,17 @@ files.find_files = function(opts)
     ),
     previewer = conf.file_previewer(opts),
     sorter = conf.file_sorter(opts),
+    attach_mappings = function(prompt_bufnr, map)
+
+      local change_working_directory = function ()
+        local entry = action_state.get_selected_entry(prompt_bufnr).value
+        local dir = vim.fn.fnamemodify(entry, ":p:h")
+        vim.fn.execute("lcd " .. dir, "silent")
```
ranjithshegde, line 217: I changed this to lcd as suggested, but I just realized lcd only changes it inside the telescopeprompt buffer and not the active buffer. Will change it back to cd

ranjithshegde:
@Conni2461 
Any comments? do you approve?

Conni2461 (commented):
Technically the code looks good, documentation would be nice, like [so](https://github.com/nvim-telescope/telescope.nvim/blob/999fad2ce514d869b736aa5d948d355fad89be8d/lua/telescope/builtin/init.lua#L96-L97)

But i personally have a couple of issues with this. This is too opinionated and too easy that people could just pass in `opts.attach_mappings` function to their function call. Couple of issue:

Its mapped, and people cant change this mapping because both are local functions.

`cd` vs `lcd` vs `tcd` whats correct for what workflow. `lcd` is not much harder to implement. You just need the previous bufnr and then you can do `vim.api.nvim_buf_call(bufnr, function() lcd_here end)` but i personally only do `tcd` and a lot of other people will feel like that.

Idk what to say, i know you spent time on it. 
And also this is just my opinion, maybe other people like tami or ober see it differently. They at least started reviewing this PR :laughing:

Conni2461:
I added it as configuration Recipes and created you: https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes#change-directory

I think this works better. Still thanks for you work tho :)

ranjithshegde:
It is indeed better as a config recipe than a PR. Besides I see that the file browser itself might be removed from core! 
Thanks @Conni2461.

KodyVB:
@Conni2461 I couldn't get it to work how the recipe showed, as the mappings were under pickers --> find_files --> mappings, while mappings [are usually](https://github.com/nvim-telescope/telescope.nvim#telescope-setup-structure) under defaults --> mappings. Moving them to match what's shown below made it work for me, so I'm wondering if the recipe should be updated to match that. Sorry if bringing this up in a comment on a closed request isn't the proper way to do it or whatever - I don't have much experience with this kind of stuff.

```lua
require("telescope").setup({
  defaults = {
    -- ....
    mappings = {
      n = {
        ["cd"] = function(prompt_bufnr)
          local selection = require("telescope.actions.state").get_selected_entry()
          local dir = vim.fn.fnamemodify(selection.path, ":p:h")
          require("telescope.actions").close(prompt_bufnr)
          -- Depending on what you want put `cd`, `lcd`, `tcd`
          vim.cmd(string.format("silent lcd %s", dir))
        end,
      },
    },
  },
})
```

PR #650, opened 2021-03-13
Prefer vim.api over vim.fn

VVKot:
See https://github.com/neovim/neovim/pull/13875 for details.

Yes, this is commented code - but person uncommenting it will surely be sad for itt not to work.


tami5:
Thanks @VVKot

PR #651, opened 2021-03-13
Reflect cwd in prefix of file browser

oberblastmeister:


lua/telescope/builtin/files.lua
```diff
@@ -229,8 +230,19 @@ files.file_browser = function(opts)
...
   end
 
+  local function get_prefix(path)
+    path = Path:new(vim.fn.fnamemodify(path, ':p'))
+
+    if path.filename:match('%.%.') then
+      path = Path:new(Path:new(path:parents()):parents())
+    end
+
+    return path.filename .. config.values.prompt_prefix
```
Conni2461, line 240: You don't need a `config` require in this file. Use `conf.prompt_prefix`

lua/telescope/builtin/files.lua
```diff
@@ -229,8 +230,19 @@ files.file_browser = function(opts)
     }
   end
 
+  local function get_prefix(path)
+    path = Path:new(vim.fn.fnamemodify(path, ':p'))
+
+    if path.filename:match('%.%.') then
+      path = Path:new(Path:new(path:parents()):parents())
```
Conni2461, line 237: This triple path definition throws me off :laughing: (Which means i have no idea what happens here and i wrote `parents` :laughing: 
Is there an oversight in `plenary.path:parents`?

oberblastmeister: Yeah the `plenary.path` api is a bit weird because the functions do not return paths, so we have to create new paths. It would be good if I could update it to return paths, but that might break other stuff that depends on it.

oberblastmeister: Also I don't know why it is called `parents`. I feel like it should be called `parent` and there should be another function called `parents` or `ancestors` that will return an iterator of `parents`.

Conni2461: You are right, thats because i am stupid and didn't finish reading the [pathlib spec](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parents) (which it is based on) correctly :laughing:

Conni2461: Fix in https://github.com/nvim-lua/plenary.nvim/pull/107

lua/telescope/builtin/files.lua
```diff
@@ -229,8 +230,19 @@ files.file_browser = function(opts)
     }
   end
 
+  local function get_prefix(path)
+    path = Path:new(vim.fn.fnamemodify(path, ':p'))
```
Conni2461, line 234: I just realize we no longer need the `vim.fn.fnamemodify` after https://github.com/nvim-telescope/telescope.nvim/pull/613 is merged. I forget to normalize the path before but #613 has the fix for it.
With normalize i mean:
/home/conni/repos/../ -> /home/conni and i think thats what you want/need

Conni2461:
closing because file_browser was moved to https://github.com/nvim-telescope/telescope-file-browser.nvim

Issue #433, opened 2021-01-13
History/MRU picker


kristijanhusak:
**Is your feature request related to a problem? Please describe.**
I switched from fzf, and one source that I miss the most is `:History`, which lists oldfiles + buffers (sorted by MRU).

**Describe the solution you'd like**
I'd like to have a separate picker that holds these.

**Describe alternatives you've considered**
I didn't find a solution for this so I implemented it myself in a same way fzf does it. Here are part of codes:
* Saving buffers: https://github.com/kristijanhusak/neovim-config/blob/bleeding-edge/nvim/lua/partials/telescope.lua#L12-L27
* Picker: https://github.com/kristijanhusak/neovim-config/blob/bleeding-edge/nvim/lua/partials/telescope.lua#L47-L83

I wanted to check if you would be interested in adding this here, and if yes, I could make a PR with the implementation.

Conni2461:
Thanks for the request.

@sunjon was working on a "smart" MRU based on that [article](https://slack.engineering/a-faster-smarter-quick-switcher/) which should be almost done.
He can tell you more :)

kristijanhusak:
Oh, nice :)
Is there a related issue/PR that I can track? I wasn't able to find anything related to MRU.
I guess I'll just stick with my version until smarter one is finished :)

sunjon:
Hi. It's WIP (as in not usable yet), but I've put the repo up [here](https://github.com/sunjon/telescope-frecency).
The todo's are listed in the readme, and I'll move the project to the 'nvim-lua' namespace when it's complete.

Hope to have it in a testable order in the next couple of days.

sunjon:
Closing this as the frecency MRU finder has [been released](https://github.com/nvim-telescope/telescope-frecency.nvim).
Hope it satisfies your requirements!

kristijanhusak:
Thanks, I'll give it a try.

kristijanhusak:
I gave this a test, it is working fine, but it doesn't work exactly how I'm used to it. If I open a file that I never opened before, I expect it to be first in the list when I open up the picker, which is not the case with this one. Of course, that's how it works by design, so no complains there :) I'll stick with my implementation. If someone shows interest in using my implementation I'll make a separate package for that.
Thanks anyway.

jesseleite:
Hey @kristijanhusak, seconded, and thank you for sharing this!  @Conni2461 @sunjon, I think the frecency thing is really cool, but it's different than MRU...

[...]

PR #657, opened 2021-03-14
Improve oldfiles finder

jesseleite:
This adds options to the `oldfiles` finder to make it optionally more useful as a MRU (most recently used) finder.

- `include_current_session`
    - Normally `v:oldfiles` only lists files from past vim sessions. This option will include recently accessed files in the current vim session using `:buffers! t` (the `!` bang shows unlisted buffers after they are closed, and the `t` modifier sorts the list by MRU).
- `cwd_only`
    - This option filters the results to only show files within the cwd. This makes for a nice project-specific-MRU finder.
    
At the moment both options don't change default `oldfiles` finder behaviour, as they are both opt-in. I might argue making `include_current_session=true` a default, having to restart vim to see new entries populated into `v:oldfiles` is kind of weird. Either way though, it'd just be nice to have the option!

---

PS. I know there was an issue/discussion at #433 around MRU pickers that was closed because of the (very cool) [frecency extension](https://github.com/nvim-telescope/telescope-frecency.nvim), but I'd argue that MRU and frecency are different things:

- **Frecency** = Frequency + Recency
- **MRU** = Recency only

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -231,6 +231,8 @@ internal.loclist = function(opts)
 end
 
 internal.oldfiles = function(opts)
+  opts.include_current_session = utils.get_default(opts.include_current_session, true)
+
   local results = {}
 
   if opts.include_current_session then
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -234,7 +234,7 @@ internal.oldfiles = function(opts)
   local results = {}
 
   if opts.include_current_session then
-    for k,buffer in ipairs(vim.split(vim.fn.execute(':buffers! t'), "\n")) do
+    for _, buffer in ipairs(vim.split(vim.fn.execute(':buffers! t'), "\n")) do
       local match = string.match(buffer, '%s*(%d+)')
       if match then
         local file = vim.api.nvim_buf_get_name(match)
@@ -245,7 +245,7 @@ internal.oldfiles = function(opts)
     end
   end
 
-  for k,file in ipairs(vim.v.oldfiles) do
+  for _, file in ipairs(vim.v.oldfiles) do
     if vim.loop.fs_lstat(file) and vim.fn.index(results, file) == -1 then
       table.insert(results, file)
     end
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -231,9 +231,26 @@ internal.loclist = function(opts)
 end
 
 internal.oldfiles = function(opts)
-  local results = vim.tbl_filter(function(val)
-    return 0 ~= vim.fn.filereadable(val)
-  end, vim.v.oldfiles)
+  local results = {}
+
+  if opts.include_current_session then
+    for k,buffer in ipairs(vim.split(vim.fn.execute(':buffers! t'), "\n")) do
+      local match = string.match(buffer, '%s*(%d+)')
+      if match then
+        local file = vim.api.nvim_buf_get_name(match)
+        if vim.fn.filereadable(file) == 1 then
+          table.insert(results, file)
+        end
+      end
+    end
+  end
+
+  for k,file in ipairs(vim.v.oldfiles) do
+    if vim.fn.filereadable(file) == 1 and vim.fn.index(results, file) == -1 then
+      table.insert(results, file)
+    end
+  end
+
   pickers.new(opts, {
     prompt_title = 'Oldfiles',
     finder = finders.new_table{
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -251,6 +251,12 @@ internal.oldfiles = function(opts)
     end
   end
 
+  if opts.cwd_only then
+    results = vim.tbl_filter(function(file)
+      return vim.fn.matchstrpos(file, vim.fn.getcwd())[2] ~= -1
+    end, results)
+  end
+
   pickers.new(opts, {
     prompt_title = 'Oldfiles',
     finder = finders.new_table{
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -248,7 +248,7 @@ internal.oldfiles = function(opts)
   end
 
   for _, file in ipairs(vim.v.oldfiles) do
-    if vim.loop.fs_stat(file) and vim.fn.index(results, file) == -1 then
+    if vim.loop.fs_stat(file) and not vim.tbl_contains(results, file) then
       table.insert(results, file)
     end
   end
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -237,10 +237,10 @@ internal.oldfiles = function(opts)
 
   if opts.include_current_session then
     for _, buffer in ipairs(vim.split(vim.fn.execute(':buffers! t'), "\n")) do
-      local match = string.match(buffer, '%s*(%d+)')
+      local match = tonumber(string.match(buffer, '%s*(%d+)'))
       if match then
         local file = vim.api.nvim_buf_get_name(match)
-        if vim.loop.fs_stat(file) then
+        if vim.loop.fs_stat(file) and match ~= vim.api.nvim_get_current_buf() then
           table.insert(results, file)
         end
       end
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -238,7 +238,7 @@ internal.oldfiles = function(opts)
       local match = string.match(buffer, '%s*(%d+)')
       if match then
         local file = vim.api.nvim_buf_get_name(match)
-        if vim.fn.filereadable(file) == 1 then
+        if vim.loop.fs_lstat(file) then
           table.insert(results, file)
         end
       end
@@ -246,14 +246,15 @@ internal.oldfiles = function(opts)
   end
 
   for k,file in ipairs(vim.v.oldfiles) do
-    if vim.fn.filereadable(file) == 1 and vim.fn.index(results, file) == -1 then
+    if vim.loop.fs_lstat(file) and vim.fn.index(results, file) == -1 then
       table.insert(results, file)
     end
   end
 
   if opts.cwd_only then
+    local cwd = vim.loop.cwd()
     results = vim.tbl_filter(function(file)
-      return vim.fn.matchstrpos(file, vim.fn.getcwd())[2] ~= -1
+      return vim.fn.matchstrpos(file, cwd)[2] ~= -1
     end, results)
   end
 
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -233,10 +233,11 @@ end
 internal.oldfiles = function(opts)
   opts.include_current_session = utils.get_default(opts.include_current_session, true)
 
+  local current_buffer = vim.api.nvim_get_current_buf()
+  local current_file = vim.api.nvim_buf_get_name(current_buffer)
   local results = {}
 
   if opts.include_current_session then
-    local current_buffer = vim.api.nvim_get_current_buf()
     for _, buffer in ipairs(vim.split(vim.fn.execute(':buffers! t'), "\n")) do
       local match = tonumber(string.match(buffer, '%s*(%d+)'))
       if match then
@@ -249,7 +250,7 @@ internal.oldfiles = function(opts)
   end
 
   for _, file in ipairs(vim.v.oldfiles) do
-    if vim.loop.fs_stat(file) and not vim.tbl_contains(results, file) then
+    if vim.loop.fs_stat(file) and not vim.tbl_contains(results, file) and file ~= current_file then
       table.insert(results, file)
     end
   end
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -236,11 +236,12 @@ internal.oldfiles = function(opts)
   local results = {}
 
   if opts.include_current_session then
+    local current_buffer = vim.api.nvim_get_current_buf()
     for _, buffer in ipairs(vim.split(vim.fn.execute(':buffers! t'), "\n")) do
       local match = tonumber(string.match(buffer, '%s*(%d+)'))
       if match then
         local file = vim.api.nvim_buf_get_name(match)
-        if vim.loop.fs_stat(file) and match ~= vim.api.nvim_get_current_buf() then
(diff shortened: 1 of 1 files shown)


tami5:
Instead of filereadable, vim.loop.fs_lstat is faster. 

```lua
if not vim.loop.fs_lstat(file) then 
  print("file doesn't exist")
end
```

Additionally, there is a vim.loop function to get current working directory, I believe it's `vim.loop.cwd()`

Other then that the pr LGTM, very smart and useful feature, well done 👍🏻

jesseleite:
Haha, pardon my VimL, still learning the Neovim ropes!

I understand the difference between `vim.fn` and `vim.api`, but a bit confused about the difference between `vim.api` and `vim.loop`.

Anyway, thank you for the kind words ❤️

elianiva:
> still a bit confused about the difference between vim.api and vim.loop

`vim.api` is Neovim's specific APIs and `vim.loop` is basically libuv binding in lua using `luv`. If you want the docs for `vim.loop`, they're [here](https://github.com/luvit/luv/blob/master/docs.md), just substitute `uv` with `vim.loop` or create a local variable like `local uv = vim.loop`

tami5:
I agree current session should br included by default but we should have an option to disable it if for some reason others would like to only browse old sessions files

jesseleite:
Cool. One more thought...

As soon as you open a file, it gets added to the finder. I'm just realizing that fzf's `:History` finder doesn't inject the current/active file as the first result. This way you can open the history finder and just hit enter to go to the last opened file. Maybe I should do the same here?

tami5:
Yah it does not make sense to me as well.

One last thing before we merge, conni told that fs_lstat might break for symlinks so we need to use `fs_stat` instead

tami5:
@jesseleite I'm testing it our right now and for some reason I see files I never opened with the same directory. Can you check please

jesseleite:
@tami5 `cwd_only=true` is opt-in. By default, it's not scoped to cwd.

jesseleite:
@tami5 `cwd_only=true` is opt-in. By default, it's not scoped to cwd.  The idea is `oldfiles` would be all old files, regardless of cwd:

- `:Telescope oldfiles`
    - Show all oldfiles history, including oldfiles in current session.
- `:Telescope oldfiles cwd_only=true`
    - Same as above, but scoped to cwd.

tami5:
Oh sorry my bad, I did open those files, and I forgot.

Okay, so the issue now is when I run `:Telescope oldfiles` in, say my nvim config, I don't get the recent files I opened at the top, instead the recent files in that directory gets promoted to the top and  that  defeats the point of mru, I could just use find_files. Thus, we need a way to; yes keep the most recent in current directory around, but not at the top, because the mess up the order.

Does it make sense?

jesseleite:
Sorry, I don't understand.

The whole finder orders the exact same was as `v:oldfiles` does. It doesn't matter which directory those files come from, it just sorts by most recently accessed. This is how `v:oldfiles` works, and it's how fzf's `:History` finder works as well. If you open it and hit enter on the first result, it'll open the last file you accessed.

If you use the `cwd_only=true` option, the whole list is just filtered down to only show files that exist within that `cwd`, but the order does not change at all. If you open it and hit enter on the first result while using the `cwd_only=true` option, it'll open the last file you accessed within that cwd.

Maybe we can do a screen call or something for me to better understand your issue? I'm on the Telescope Gitter as well if you want to DM me there!

lua/telescope/builtin/internal.lua
```diff
@@ -231,9 +231,35 @@ internal.loclist = function(opts)
...
 internal.oldfiles = function(opts)
-  local results = vim.tbl_filter(function(val)
-    return 0 ~= vim.fn.filereadable(val)
-  end, vim.v.oldfiles)
+  opts.include_current_session = utils.get_default(opts.include_current_session, true)
+
+  local results = {}
+
+  if opts.include_current_session then
+    for _, buffer in ipairs(vim.split(vim.fn.execute(':buffers! t'), "\n")) do
```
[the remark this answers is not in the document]
jesseleite: Because if you close a buffer, it'll get removed from the list, which doesn't make sense. It's the same as my `remove_buf()` argument here https://github.com/nvim-telescope/telescope.nvim/issues/433#issuecomment-798831113.

...It does a file existance check though, so it never lists scratch buffers, etc. It actually lists the exact same files as `v:oldfiles` does after you restart Vim. I made sure of that!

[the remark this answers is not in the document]
jesseleite: @kristijanhusak Because if you manually open a file buffer, then `:bdelete` to close that buffer, it'll get removed from the finder until you restart Vim and see it in `v:oldfiles`, which doesn't make any sense. It's the same as my `remove_buf()` argument from our other thread here https://github.com/nvim-telescope/telescope.nvim/issues/433#issuecomment-798831113...

>Whether or not I close a buffer, I would still expect it to show in `:History` or `:Telescope oldfiles`, since I did in fact open it. If I restart Vim, it would be in the oldfiles list, so why not [show that closed/unlisted buffer] in the list?

That said, it does a file existence filter check though, so it never lists temporary scratch buffers, terminal buffers, etc. It actually lists the exact same results as `v:oldfiles`, just that you don't have to restart Vim to see the old files in your current session. I made sure of that!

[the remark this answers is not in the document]
kristijanhusak: Ok, makes sense, thanks!

tami5:
> Whats missing here? Does anyone need something from me? Code looks good to me (first glance) :)

Sorry, couldn't test today and I'm about to sleep.

It looks good but I'm quite unsure if it will introduce undesirable experience. 

Last time I tested it, I had an issue with the order in which files get sorted. I noticed it doesn't get sorted based on recently opened files, rather somehow prompting current directory files.

Conni can you please test, and check is this still the case? 

Basically, checkout the pr -> open few files in specific order , close them (maybe restart neovim) and launch oldfiles picker. 

Thanks

jesseleite:
>Last time I tested it, I had an issue with the order in which files get sorted. I noticed it doesn't get sorted based on recently opened files, rather somehow prompting current directory files.

@tami5 I'm still not sure what you mean here, but I'm all for another set of eyes and thorough testing 🙂

@Conni2461 I'd be glad to answer any questions, screenshare, DM on gitter, whatever is helpful 👍

Conni2461 (approved):

lua/telescope/builtin/internal.lua
```diff
@@ -231,9 +231,35 @@ internal.loclist = function(opts)
...
+        local file = vim.api.nvim_buf_get_name(match)
+        if vim.loop.fs_stat(file) and match ~= vim.api.nvim_get_current_buf() then
+          table.insert(results, file)
+        end
+      end
+    end
+  end
+
+  for _, file in ipairs(vim.v.oldfiles) do
+    if vim.loop.fs_stat(file) and vim.fn.index(results, file) == -1 then
```
Conni2461, line 251: Chant we use `vim.tbl_contains(t, value)` for this. So a lua func over a viml one. Should return true if contains and false if not. (`:help tbl_contains`)

Otherwise LGTM 
Thanks :)

lua/telescope/builtin/internal.lua
```diff
@@ -231,9 +231,35 @@ internal.loclist = function(opts)
...
+  opts.include_current_session = utils.get_default(opts.include_current_session, true)
+
+  local results = {}
+
+  if opts.include_current_session then
+    for _, buffer in ipairs(vim.split(vim.fn.execute(':buffers! t'), "\n")) do
+      local match = tonumber(string.match(buffer, '%s*(%d+)'))
+      if match then
+        local file = vim.api.nvim_buf_get_name(match)
+        if vim.loop.fs_stat(file) and match ~= vim.api.nvim_get_current_buf() then
```
kristijanhusak, line 243: you could store current buf in variable outside the loop instead of calling function each time. It's super tiny micro optimization :)

lua/telescope/builtin/internal.lua
jesseleite: Good one, pardon my VimL 😂

lua/telescope/builtin/internal.lua
jesseleite: Also good call, thanks!

kristijanhusak (approved):

tami5:
Well done @jesseleite

jesseleite:
Thank you for help and encouragement 🙌

kristijanhusak:
@jesseleite I was testing this today, and I still don't like the idea of showing buffers that are not listed. For example I'm using https://github.com/tpope/vim-dadbod and it creates some temporary files that prints out the results, or some scratch buffers that I don't really need.

Could we add an option to hide the buffers that are not listed?

jesseleite:
@kristijanhusak I do have a `vim.loop.fs_stat()` (essentially `filereadable()`) check around each unlisted buffer though, which means dadbod should be creating actual files on your filesystem.

Try this...

- Open a vim-dadbod buffer
- Close nvim
- Re-open nvim
- `echo v:oldfiles`
- You should see that dadbod file at the top of your `v:oldfiles` list, right?

If we hide unlisted files, then it presents a new bug, which is the same bug that fzf's `:History` finder has. If you open a file and then immediately close it, you won't see it in your `:Telescope oldfiles` or fzf `:History` **_UNTIL_** you restart nvim and get the new `v:oldfiles` list. That's awkward, because I know I just had that project file open, and I know it'll be in my `v:oldfiles` for the next vim session.

That said, I'd rather show dadbod files, than hide the user's actual project files that **_SHOULD_** be in the list when the user has closed those buffers. Nawmean?

kristijanhusak:
@jesseleite it does create files, in tmp folder, but buffers created for those files are scratch buffers. Once neovim is restarted, those files are gone, since they were created in another neovim session, and tmp folders are created per session. `v:oldfiles` returns them, but they are no longer readable so they are not shown.
I'm ok to have this as an opt-in, and leave the default functionality as it is now.

jesseleite:
@kristijanhusak I think the fix may then be to ignore scratch files in the tmp folder, rather than ignore all unlisted buffers (which again includes valid user project files that they recently edited and closed). Looking into this! 👍

jesseleite:
@kristijanhusak I think the fix may then be to ignore scratch files in the tmp folder, rather than ignore all unlisted buffers (which again includes valid user project files that they recently edited and closed). Looking into this! 👍 

Maybe we ignore scratch/tmp files by default, and add an option like `include_tmp_files=true` if people actually want access to all that in the finder.

kristijanhusak:
I think that setting is too specific. I believe most of the people who comes from FZF will expect this to work as it worked there. I never had a problem from missing buffers after I delete them, probably because I got used to it in FZF.
I don't see why there wouldn't be an option like `only_buflisted` that would be disabled, and if someone wants same functionality like FZF, they just set it to true.

jesseleite:
I don't see why it would ever be desirable to open `Telescope oldfiles` and **_NOT_** see a file in there that I knew that I just edited, only to have it show after restarting vim 🤷‍♂️

kristijanhusak:
Habit. I'm used to see only things that I'm working with. You could say I'm using this as a buffer switcher, but I need MRU sort.

jesseleite:
>You could say I'm using this as a buffer switcher, but I need MRU sort.

Why not then use `:Telescope buffers sort_lastused=true`? 😁

kristijanhusak:
Because it's not really sorting the whole list by MRU. It just preselects alternate buffer differently.

jesseleite:
What you are describing then sounds like a possible PR to the `buffers` finder, to make `sort_lastused` actually sort by MRU. I could get behind that!

- `:Telescope buffers` - lists open buffers, with whatever sorting options you set on it
- `:Telescope oldfiles` - lists recently opened and/or closed files, which does as advertised

kristijanhusak:
Yeah, i guess it needs to be something like that. It's up to telescope collaborators to decide.

tami5:
I really don't like this new changes, I'm trying to open recent file in the current directory and I can't see it. I just see ALL files in current directory.

Is it possible that files get add oldfiles when previewing them. Is it possible that files is merged into oldfiles?

jesseleite:
@tami5 What do you mean you see ALL files in current directory?

tami5:
I mean like what you would expect find_files to return. 

Okay it's seems to be an issue with my confing. `:oldfiles` sometimes output files I never opened + ALL of the directory files.

hmmmmmmmmm wired issue. maybe I should delete shada file

jesseleite:
@tami5 I did mention to you in gitter DM the other day about the LSP related issue I found, which lists files I never opened when certain LSPs crawl my files. Not sure if this is a Neovim bug. That said, I think I have a solution I'm going to PR shortly.

Could you try something for me though? If you disable all your LSP stuff and delete your shada file, does `oldfiles` still list files you never intentionally opened?

tami5:
@jesseleite oh yes indeed, I wasn't expecting that bug to be related to this issue. Yah it's a lsp thing for sure, lua lsp seems to causing this from time to time.

That would be awesome thanks @jesseleite

YodaEmbedding:
Is there a way to also include other entries from other file pickers, e.g. `builtin.git_files`?

`:Telescope oldfiles cwd_only=true alternate=git_files`

jesseleite:
@YodaEmbedding That might get weird IMO, as many finders aren't file based (ie. command finder, etc). This oldfiles finder is pretty coupled to the concept of `v:oldfiles`. That said, there's a prompt history PR here which you might be interested in: https://github.com/nvim-telescope/telescope.nvim/pull/521

thraizz:
> @tami5 I did mention to you in gitter DM the other day about the LSP related issue I found, which lists files I never opened when certain LSPs crawl my files. Not sure if this is a Neovim bug. That said, I think I have a solution I'm going to PR shortly.
> 
> Could you try something for me though? If you disable all your LSP stuff and delete your shada file, does `oldfiles` still list files you never intentionally opened?

Did you manage to PR your solution for the LSP crawling issue? I suffer from this as well as pyright indexes all 200+ py files and completely overwrites my dear oldfiles picker :(

jesseleite:
@thraizz, I keep meaning to PR a bugfix for this. Curious for you though, which LSP(s) is/are doing it for you?

thraizz:
> @thraizz, I keep meaning to PR a bugfix for this. Curious for you though, which LSP(s) is/are doing it for you?

pyright, i fixed it by changing `:buffers! t` in line 347 of internal.lua to `:buffers t` so unlisted-buffers (the buffers with a number + 'u' shown in `:buffers! t`) are ignored.

jesseleite:
@thraizz Gotcha! Anyway, I see you have that bugfix PR over at https://github.com/nvim-telescope/telescope.nvim/pull/1219 👍  Check out the discussion over there if you haven't already seen ❤️

PR #663, opened 2021-03-16
Preview colorscheme

teto:
I am trying to add live preview of colorscheme since it's a longstanding petpeeve of mine. I expect it to fail miserably but one has to try. My previewer seems to do nothing, any hint why ? is it preview_fn  or scroll_fn that will be called when hovering on a new entry ?

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -560,6 +560,12 @@ internal.buffers = function(opts)
   }):find()
 end
 
+function colorscheme_previewer (opts)
+	get_command = opts.get_command or function (entry)
+		vim.cmd ([[ colorscheme ]]..entry)
+	end
+end
+
 internal.colorscheme = function(opts)
   local colors = vim.list_extend(opts.colors or {}, vim.fn.getcompletion('', 'color'))
 
@@ -569,6 +575,7 @@ internal.colorscheme = function(opts)
       results = colors
     },
     -- TODO: better preview?
+    previewer = colorscheme_previewer(opts),
     sorter = conf.generic_sorter(opts),
     attach_mappings = function(prompt_bufnr)
       actions.select_default:replace(function()
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -560,14 +560,28 @@ internal.buffers = function(opts)
   }):find()
 end
 
-function colorscheme_previewer (opts)
-	get_command = opts.get_command or function (entry)
-		vim.cmd ([[ colorscheme ]]..entry)
+local colorscheme_previewer = previewers.new({
+	setup = function (self)
+		-- save original colorscheme
+		vim.api.nvim_err_writeln("Ln")
+	end,
+	preview_fn  = function(self, entry, status)
+		print("chosen entry:", entry)
+		vim.cmd("colorscheme " .. entry)
+		vim.api.nvim_err_writeln("preview_fn is not defined for this previewer")
+	end,
+	scroll_fn = function(self, dir)
+		vim.api.nvim_err_writeln("scroll_fn is not defined for this previewer")
+	end,
+	teardown = function (self)
+	-- restore original colorscheme
+		print ("TEARDOWN")
 	end
-end
+
+})
 
 internal.colorscheme = function(opts)
-  local colors = vim.list_extend(opts.colors or {}, vim.fn.getcompletion('', 'color'))
+  -- local colors = vim.list_extend(opts.colors or {}, vim.fn.getcompletion('', 'color'))
 
   pickers.new(opts,{
     prompt = 'Change Colorscheme',
@@ -575,7 +589,7 @@ internal.colorscheme = function(opts)
       results = colors
     },
     -- TODO: better preview?
-    previewer = colorscheme_previewer(opts),
+    previewer = colorscheme_previewer,
     sorter = conf.generic_sorter(opts),
     attach_mappings = function(prompt_bufnr)
       actions.select_default:replace(function()
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -562,26 +562,38 @@ end
 
 local colorscheme_previewer = previewers.new({
 	setup = function (self)
+		local state = {}
+		-- print("initial colorscheme:", state.original)
+		state.original = vim.cmd "colorscheme"
+		-- print(state.original)
 		-- save original colorscheme
-		vim.api.nvim_err_writeln("Ln")
+		-- vim.api.nvim_err_writeln("Ln")
+		return state
 	end,
 	preview_fn  = function(self, entry, status)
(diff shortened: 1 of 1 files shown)


lua/telescope/builtin/internal.lua
```diff
@@ -560,15 +560,36 @@ internal.buffers = function(opts)
   }):find()
 end
 
+local colorscheme_previewer = previewers.new({
+	setup = function (self)
+		-- save original colorscheme
```
Conni2461, line 565: setup needs to return a table, basically a state table. (Setup will always run unless it finds a table)

lua/telescope/builtin/internal.lua
```diff
@@ -560,15 +560,36 @@ internal.buffers = function(opts)
...
+		vim.api.nvim_err_writeln("Ln")
+	end,
+	preview_fn  = function(self, entry, status)
```
Conni2461, line 568: `preview_fn` will run on selecting a entry. That is correct. You have `colors` commented out which means a entry will never be selected

lua/telescope/builtin/internal.lua
```diff
@@ -560,15 +560,36 @@ internal.buffers = function(opts)
...
+		print("chosen entry:", entry)
+		vim.cmd("colorscheme " .. entry)
+		vim.api.nvim_err_writeln("preview_fn is not defined for this previewer")
+	end,
+	scroll_fn = function(self, dir)
```
Conni2461, line 573: Will run on `actions.preview_scrolling_up`/`actions.preview_scrolling_down` is mapped to `<c-d>`/`<c-u>`

lua/telescope/builtin/internal.lua
```diff
@@ -560,15 +560,36 @@ internal.buffers = function(opts)
...
+	teardown = function (self)
+	-- restore original colorscheme
+		print ("TEARDOWN")
+	end
+
+})
+
 internal.colorscheme = function(opts)
-  local colors = vim.list_extend(opts.colors or {}, vim.fn.getcompletion('', 'color'))
+  -- local colors = vim.list_extend(opts.colors or {}, vim.fn.getcompletion('', 'color'))
```
Conni2461, line 584: Should work after commenting this in

lua/telescope/builtin/internal.lua
```diff
@@ -560,15 +560,36 @@ internal.buffers = function(opts)
...
 end
 
+local colorscheme_previewer = previewers.new({
+	setup = function (self)
+		-- save original colorscheme
+		vim.api.nvim_err_writeln("Ln")
+	end,
+	preview_fn  = function(self, entry, status)
+		print("chosen entry:", entry)
+		vim.cmd("colorscheme " .. entry)
```
Conni2461, line 570: Entry is a metatable so you would need to do `entry[1]` if i recall correctly.

elianiva:
would be cool if it can preview some code on the previewer and user can cycle through them. 

like this, for example
http://chriskempson.com/projects/base16/

tami5:
> would be cool if it can preview some code on the previewer and user can cycle through them.
> 
> like this, for example
> http://chriskempson.com/projects/base16/

yah, and the content of the previewer is current file.

teto:
so it now changes the current colorscheme dynamically but it can't restore the initial colorscheme because for some reason vim.cmd "colorscheme" returns nil in script while it works ok in ex mode.

Conni2461:
```lua
local tmp = vim.fn.execute("colorscheme")
print("output:", tmp)
```

teto:
thanks. How would I know in teardown that the user has enabled a specific entry rather than just escaped/exited telescope ? (in which case I want to restore the initial colorscheme).

Conni2461:
Currently not possible but i was lately thinking about this feature. We can do a workaround with actions and replace and enhance. I think
```lua
-- keep og not in previewer state but in internal colorscheme

      local success = false      
        if current_picker.previewer == false then -- when someone has disabled previewer
          vim.cmd("colorscheme " .. selection.value)
          success = nil
        else
          success = true
        end

        actions.close(prompt_bufnr)
      end)

      actions.close:enhance {
        post = function()
          if success == false then
             -- restore og
          end
        end
      }

      return true
    end
```

Something like this might work. Haven't tested it tho

teto:
If I define `actions.close:enhance` then I can't select anything anymore plus it looks overly complex. Looks like the picker doesn't have a teardown function but it would be nice to have one with the selected entry/ status (exit/selected).

Conni2461:
Might be never tried it. I don't have the time to design a API for that right now, so the other workaround i thought about (after posting this) is setting a value inside `current_picker.previewer.state.success = true` and then you should have access to that value in `teardown`.

Otherwise, if that is too hacky for you, you can design a interface and i take a look at it at some point :)

andymass:
Using an "inner nvim" seems to work pretty well, as an alternative to changing the editor's main colorscheme.

```diff
diff --git a/lua/telescope/builtin/internal.lua b/lua/telescope/builtin/internal.lua
index 18820f8..c6308f4 100644
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -591,12 +591,22 @@ end
 internal.colorscheme = function(opts)
   local colors = vim.list_extend(opts.colors or {}, vim.fn.getcompletion('', 'color'))

+  local sourced_file = require('plenary.debug_utils').sourced_filepath()
+  local base_directory = vim.fn.fnamemodify(sourced_file, ":h:h:h:h")
+  local preview_file = base_directory .. '/lua/telescope/builtin/internal.lua'
+
   pickers.new(opts,{
     prompt = 'Change Colorscheme',
     finder = finders.new_table {
       results = colors
     },
-    -- TODO: better preview?
+
+    previewer = previewers.new_termopen_previewer {
+      get_command = function(entry)
+        return {'nvim', '-R', "+colorscheme " .. entry.value, preview_file, '+se sol', '+233'}
+      end
+    },
+
     sorter = conf.generic_sorter(opts),
     attach_mappings = function(prompt_bufnr)
       actions.select_default:replace(function()
```

teto:
@andymass thanks for the trick. It's slower here (~1sec ) but it can probably be speed up via -i NONE / disabling swap etc.

I was of the opinion that changing the whole editor was a better preview but in the end I quite like your approach. Feel free to open a PR, I may close this one.

andymass:
> I was of the opinion that changing the whole editor was a better preview but in the end I quite like your approach.

It is surprisingly good, but I am not sure either to be honest.  I actually think the "live preview" aspect of changing the whole editor would be in some ways nicer as long as it is restored if the user cancels the selection.  But it seems tricky to me as I don't really understand the telescope API.

teto:
with my PR I noticed I had really different colorschemes and the transitions made the screen flash in unpleasnt ways. Tthe approach of chaning the whole editor ended up being rather annoying which is why I now favor your approach.

Conni2461:
https://github.com/nvim-telescope/telescope.nvim/pull/980 Implemented this Feature

Closing

PR #664, opened 2021-03-17
fix: no longer leaking one buffer previewer in some occasions

### LINKED ISSUE #247: file_ignore_patterns not showing the correct number of results
I like to hide `node_modules` and `.git` from the telescope view, which can usually be done with `git_files` instead of `find_files`, but sometimes, I want to search just down a specific directory tree. Because of this, I added `node_modules` and `.git` to the `file_ignore_patterns` list. However, this is showing the incorrect `matching/total` indicator when searching.

What I expect to see:
image

What I actually see:
image

As you can see, in the former, it displays `1/62`. Whereas in the second it displays `20435/20508`, despite there only being one item in the list.

### LINKED ISSUE #640: Errors while executing basic commands in latest neovim v0.5.0-dev+1140-g097ec71bd
Same error for all these commands

:Telescope find_files<cr>
:Telescope live_grep<cr>
:Telescope buffers<cr>
:Telescope help_tags<cr>
```bash
E5108: Error executing lua vim.lua:269: Tried to call API function with vim.fn: 
use vim.api.nvim_create_buf instead
```

I know things break while using nightly release.
Still it might be helpful for developers to know

### LINKED ISSUE #658: Telescope keymaps broken for comma
### Description

When searching through `:Telescope keymaps` for mappings which use a comma `,` as the `mapleader`, the fuzzy finder seems to have trouble recognizing the comma character.

**Expected Behavior**

Using `,` in the fuzzy finder returns results that contain a comma and sorts these results properly.

**Actual Behavior**

The comma is not recognized when used with or without an escape character (i.e. searching `,` or `\,`).

### Details

1. `:Telescope keymaps`
2. `,` ... nothing happens

Or...

1. `:Telescope keymaps`
2. `\,` ... The number of results does not change, but the order (sorting) of the results changes
</details>

- `nvim --version` output: NVIM v0.5.0-dev+nightly-93-gc1fbc2ddf
- Operating system: macOS Big Sur
- Telescope commit: 284f38c57539967b25c7d32700acffd46599c49f

</details>

```viml
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-telescope/telescope-fzy-native.nvim'

let mapleader = ","
lua require('telescope').setup {}
```
</details>

Conni2461:
And some other small things i noticed in the past week:
- fix docgen formatting
- fix in_fast_even when overriding file_maker
- fix filtering for space in keymaps with fzy (close #658)
- remove count of files ignored by `file_ignore_patterns` (fix #247)
- don't write to invalid prompt (fix #640)

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -952,6 +952,7 @@ function Picker:get_result_processor(prompt, status_updater)
       if file then
         if string.find(file, v) then
           log.debug("SKPIPING", entry.value, "because", v)
+          self:_decrement("processed")
           return
         end
       end
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -904,7 +904,7 @@ end
 function Picker:get_status_updater(prompt_win, prompt_bufnr)
   return function()
     local text = self:get_status_text()
-    if not vim.api.nvim_buf_is_valid(prompt_bufnr) then return end
+    if self.closed == true or not vim.api.nvim_buf_is_valid(prompt_bufnr) then return end
     local current_prompt = vim.api.nvim_buf_get_lines(prompt_bufnr, 0, 1, false)[1]
     if not current_prompt then
       return
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -450,7 +450,7 @@ function Picker:find()
   -- Register attach
   vim.api.nvim_buf_attach(prompt_bufnr, false, {
     on_lines = on_lines,
-    on_detach = vim.schedule_wrap(function()
+    on_detach = function()
       on_lines = nil
 
       -- TODO: Can we add a "cleanup" / "teardown" function that completely removes these.
@@ -459,9 +459,11 @@ function Picker:find()
       self.sorter = nil
       self.manager = nil
 
+      self.closed = true
+
       -- TODO: Should we actually do this?
       collectgarbage(); collectgarbage()
-    end),
+    end,
   })
 
   -- TODO: Use WinLeave as well?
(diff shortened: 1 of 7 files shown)


Conni2461:
Docgen CI deleted everything because `tree-sitter-lua` wasn't bumped yet :| https://github.com/Conni2461/telescope.nvim/runs/2133421164?check_suite_focus=true

rockerBOO:
I believe this is due to neovim needing to be built with treesitter 19.x. They just released an update for nvim-treesitter. I dont think `tree-sitter-lua` needs to be bumped in this case. Not sure the environment that is building the nightly. 

https://github.com/nvim-treesitter/nvim-treesitter/pull/1000

Conni2461:
The last nightly was build with `0.19.x` so only `ABI 13` is valid. `docgen` and the used grammar for `docgen` `tree-sitter-lua` is still build with `0.18.x`. TJ does the bump to `0.19.x` here: https://github.com/tjdevries/tree-sitter-lua/pull/15

This is unrelated to `nvim-treesitter` because `nvim-treesitter` ships a different grammar.

Conni2461:
Okay lot of fixes here :) Its done so someone should take a quick look at it and after that can merge it.

Docgen also works again :tada: 

CC @elianiva

lua/telescope/pickers.lua
```diff
@@ -952,6 +955,7 @@ function Picker:get_result_processor(prompt, status_updater)
       if file then
         if string.find(file, v) then
           log.debug("SKPIPING", entry.value, "because", v)
```
elianiva, line 957: SKPIPING -> SKIPPING :)

lua/telescope/pickers.lua
```diff
@@ -986,7 +990,7 @@ end
 
 function Picker:get_result_completor(results_bufnr, prompt, status_updater)
   return function()
-    if self:is_done() then return end
+    if self.closed == true or self:is_done() then return end
```
elianiva, line 993: do we need `true` here? I think it could be shortened to `self.closed or self:is_done()`.

elianiva:
except the two comments I left, looks good to me. GJ!

PR #659, opened 2021-03-15
docs: regenerates with new docgen changes

Conni2461:
and 1 builtin example.

Someone can review and if its good either merge it or finish it :) But it has some changes in in the documentation that needs to be included and i already did this when developing the docgen changes so here it is.

PR #666, opened 2021-03-17
picker(live_grep): add option to grep only over open files

jameswalmsley:
No idea if I've implemented this correctly at all.. however it works.

I wanted to be able to grep just the currently open files, and then put
the results into a quick-list.

All feedback appreciated, and happy to refactor as required.

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -13,6 +13,7 @@ local Path = require('plenary.path')
 local os_sep = Path.path.sep
 
 local flatten = vim.tbl_flatten
+local filter = vim.tbl_filter
 
 local files = {}
 
@@ -32,7 +33,24 @@ end
 files.live_grep = function(opts)
   local vimgrep_arguments = opts.vimgrep_arguments or conf.vimgrep_arguments
   local search_dirs = opts.search_dirs
-  opts.cwd = opts.cwd and vim.fn.expand(opts.cwd)
+  local grep_open_files = opts.grep_open_files
+  opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or vim.loop.cwd()
+
+  local filelist = {}
+
+  if grep_open_files then
+   local bufnrs = filter(function(b)
+      if 1 ~= vim.fn.buflisted(b) then return false end
+      return true
+    end, vim.api.nvim_list_bufs())
+    if not next(bufnrs) then return end
+
+    local tele_path = require'telescope.path'
+    for _, bufnr in ipairs(bufnrs) do
+      local file = vim.api.nvim_buf_get_name(bufnr)
+      table.insert(filelist, tele_path.make_relative(file, opts.cwd))
+    end
+  end
 
   if search_dirs then
     for i, path in ipairs(search_dirs) do
@@ -49,15 +67,19 @@ files.live_grep = function(opts)
 
       prompt = escape_chars(prompt)
 
-      local args = flatten { vimgrep_arguments, prompt }
+      local search_list = {}
 
       if search_dirs then
-        table.insert(args, search_dirs)
+        table.insert(search_list, search_dirs)
       elseif os_sep == '\\' then
-        table.insert(args, '.')
+        table.insert(search_list, '.')
       end
 
-      return args
+      if grep_open_files then
+        search_list = filelist
+      end
+
+      return flatten { vimgrep_arguments, prompt, search_list }
     end,
     opts.entry_maker or make_entry.gen_from_vimgrep(opts),
     opts.max_results,
@@ -71,6 +93,7 @@ files.live_grep = function(opts)
     sorter = conf.generic_sorter(opts),
   }):find()
 end
+
 
 -- Special keys:
 --  opts.search -- the string to search.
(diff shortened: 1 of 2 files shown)


elianiva:
thanks, but I'm pretty sure we already have this feature. It's called `current_buffer_fuzzy`

see https://github.com/nvim-telescope/telescope.nvim/wiki/Showcase#current-buffer-fuzzy-find

jameswalmsley:
@elianiva Thanks for the response. current_buffer_fuzzy only does a fuzzy on the current buffer.
live_grep_open_files allows you to grep all files that are currently open in nvim.

current_buffer_fuzzy also doesn't seem to support sending a list of items to the qflist.

I think the advantages are:

  * Matches across all files currently open in nvim
  * grep i.e. full regular expressions matching
  * Send a selection or all results to qflist.

This scenario can be really powerful when you are rebasing a large code base, and need to make a large number of identical changes to all files in a patch.

elianiva:
ah, I see. I thought current open file is the *current* file you're editing. I misread the *files* as *file*.

elianiva:
on second thought, I think it would be better to make the current live_grep have an option to do this instead of creating a new picker. Maybe an option like `current_open_files=true` then it will only grep on the current open files.

jameswalmsley:
@elianiva I like that idea. I'll update my PR.

jameswalmsley:
@elianiva I've updated it to an option (grep_open_files=true) 
Can you think of a better name, (restrict_open=true) ?

Linting seems to be failing on some code I didn't change.

elianiva:
don't worry about the lint, https://github.com/nvim-telescope/telescope.nvim/pull/676 takes care of that. Just need to wait for docgen to be fixed, otherwise if I merge that PR now, our doc will be gone 😆 

`grep_open_files` sounds good to me

elianiva (approved):

lua/telescope/builtin/init.lua
```diff
@@ -23,6 +23,7 @@ local builtin = {}
 
 --- Live grep means grep as you type.
 builtin.live_grep = require('telescope.builtin.files').live_grep
+builtin.live_grep_open_files = require('telescope.builtin.files').live_grep_open_files
```
elianiva, line 26: oh, you forgot to remove this :)

lua/telescope/builtin/files.lua
```diff
@@ -32,9 +33,26 @@ end
...
+    if not next(bufnrs) then return end
+
+    local tele_path = require'telescope.path'
+    for _, bufnr in ipairs(bufnrs) do
+      local file = vim.uri_to_fname(vim.uri_from_bufnr(bufnr))
+      table.insert(filelist, tele_path.make_relative(file, opts.cwd))
+    end
+  end
+
+  if search_dirs and not grep_open_files then
```
elianiva, line 55: ..and I'm pretty sure this `not grep_open_files` isn't needed.
also you can do `vim.api.nvim_buf_get_name(bufnr)` instead of this URI stuff. 

I pressed approve to soon apparently :laughing: Could you fix this as well? :)

jameswalmsley: Nice, I've updated it to use the nvim api.
Also removed the extra `not grep_open_files`.. I was kind of thinking we don't want to iterate the search dirs.

But a user shouldn't provide both options anyway.

jameswalmsley:
@elianiva I've rebased this branch onto the latest master.
Do you know if I can get this merged?

elianiva:
I like it so far, will have to test it again tomorrow. We should ask someone else before merging it, I don't feel comfortable merging stuff on my own :laughing:

@tami5 thoughts?

tami5:
It looks great to me, I'm going to test out in few hours.

elianiva:
LGTM! Just tested it out and it works, Tami can merge if he wants.

tami5:
@jameswalmsley LGTM great work. It would be awesome to do document this in the source code and readme as you see fit so others can use it if the like to.

jameswalmsley:
@elianiva @tami5 Many thanks for the feedback, I'll add some documentation today. :)

jameswalmsley:
@tami5 @elianiva I've added some option documentation for live_grep..
Wasn't sure exactly where options for pickers are documented in the README, so I've added a small table for live_grep under the picker documentation.

Conni2461:
@tami5 fyi we actually wanted to trim down the documentation in readme. We NEVER included options in the readme because if he had we would have ended up with a > 1000 lines and nobody finds anything. Here is an example https://github.com/nvim-telescope/telescope.nvim/pull/659#discussion_r596265905 on how `opts` should be documented post docgen. Nobody has just started work on it.

I will still be here for questions so feel free to ping me :)

tami5:
> @tami5 fyi we actually wanted to trim down the documentation in readme. We NEVER included options in the readme because if he had we would have ended up with a > 1000 lines and nobody finds anything. Here is an example https://github.com/nvim-telescope/telescope.nvim/pull/659#discussion_r596265905 on how `opts` should be documented post docgen. Nobody has just started work on it.
> 
> 
> 
> I will still be here for questions so feel free to ping me :)

Oh my bad conni.

PR #668, opened 2021-03-18
Add option to display in external window

akiyosi:
This PR is the first step in the features to display the windows that make up the telescope in neovim external window.

telescope-with-ext-windows

In this first PR implementation, external windows are not given any valid positional information, so many GUIs will display the prompt, result, and preview windows in strange positions. This is because, in the current neovim specification, external windows generally do not have positional information (See [Multigrid Events](https://neovim.io/doc/user/ui.html#ui-multigrid)), and their position depends on the implementation of the GUI front-end.
In [Goneovim](https://github.com/akiyosi/goneovim), which I am developing, I have implemented a process that layout these external windows appropriately according to their size.

One idea to give positional information to the external windows is to display them as normal float windows beforehand, and then externalize them. However, I am not familiar with the project architecture and have not been able to implement this workaround.

I would appreciate any feedback or advice on this PR.

### DIFF
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -88,7 +88,14 @@ function config.set_defaults(defaults)
   set("layout_strategy", "horizontal")
   set("layout_defaults", {})
 
-  set("externalize", false)
+  set("externalize", false, [[
+    Displays the prompt, result, and preview windows as external windows if set to true.
+    Only available when using an UI with ui-multigrid support.
+
+    Available options are:
+    - "false" (default)
+    - "true"]])
+
   set("width", 0.75)
   set("winblend", 0)
   set("prompt_position", "bottom")
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -88,6 +88,7 @@ function config.set_defaults(defaults)
   set("layout_strategy", "horizontal")
   set("layout_defaults", {})
 
+  set("externalize", false)
   set("width", 0.75)
   set("winblend", 0)
   set("prompt_position", "bottom")
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -99,6 +99,7 @@ function Picker:new(opts)
       -- TODO: If it's a list, of length 2, then it's a range of min to max?
       height = get_default(opts.height, 0.8),
       width = get_default(opts.width, config.values.width),
+      externalize = get_default(opts.externalize, config.values.externalize),
 
       get_preview_width = get_default(opts.preview_width, config.values.get_preview_width),
 
@@ -133,30 +134,41 @@ function Picker:new(opts)
 end
 
 function Picker:_get_initial_window_options()
-  local popup_border = resolve.win_option(self.window.border)
-  local popup_borderchars = resolve.win_option(self.window.borderchars)
+  local popup_border, popup_borderchars
+  if not self.window.externalize then
+    popup_border = resolve.win_option(self.window.border)
+    popup_borderchars = resolve.win_option(self.window.borderchars)
+  end
+
+  -- local preview, results, prompt
 
   local preview = {
     title = self.preview_title,
-    border = popup_border.preview,
-    borderchars = popup_borderchars.preview,
     enter = false,
     highlight = false
   }
+  if not self.window.externalize then
+    preview.border = popup_border.preview
+    preview.borderchars = popup_borderchars.preview
+  end
 
   local results = {
     title = self.results_title,
-    border = popup_border.results,
-    borderchars = popup_borderchars.results,
     enter = false,
   }
+  if not self.window.externalize then
+    results.border = popup_border.preview
+    results.borderchars = popup_borderchars.preview
+  end
 
   local prompt = {
     title = self.prompt_title,
-    border = popup_border.prompt,
-    borderchars = popup_borderchars.prompt,
     enter = true
   }
+  if not self.window.externalize then
+    prompt.border = popup_border.preview
+    prompt.borderchars = popup_borderchars.preview
+  end
 
   return {
     preview = preview,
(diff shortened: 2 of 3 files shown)


lua/telescope/config.lua
```diff
@@ -92,6 +92,7 @@ function config.set_defaults(defaults)
   set("layout_strategy", "horizontal")
   set("layout_defaults", {})
 
+  set("externalize", false)
```
Conni2461, line 95: Can you add documentation for this? Similar how the other config options are documented. e.g. `scroll_strategy`

Also you probably should rebase and pop the docgen PR. (docgen should work again)
Also sorry that i am so late with this review.

lua/telescope/pickers.lua
```diff
@@ -383,6 +388,26 @@ function Picker:find()
...
+  a.nvim_win_set_option(results_win, 'winhl', 'Normal:TelescopeNormal')
+  a.nvim_win_set_option(results_win, 'winblend', self.window.winblend)
+
+  local results_border_win
+  if not self.window.externalize then
+    results_border_win = results_opts.border and results_opts.border.win_id
+    if results_border_win then
+      vim.api.nvim_win_set_option(results_border_win, 'winhl', 'Normal:TelescopeResultsBorder')
+    end
+  end
```
Conni2461, line 408: Why was this whole part moved down. Previously on 340

lua/telescope/pickers.lua
```diff
@@ -507,22 +538,32 @@ function Picker:find()
...
+
+  if self.window.externalize then
+    a.nvim_win_set_config(results_win, {external = true, width=width, height=height})
+    a.nvim_win_set_config(prompt_win, {external = true, width=width, height=height})
+    a.nvim_win_set_config(preview_win, {external = true, width=width, height=height})
+  end
 end
 
+
+
```
Conni2461, line 550: please remove these again thanks :)

lua/telescope/pickers.lua
```diff
@@ -507,22 +538,32 @@ function Picker:find()
...
 function Picker:hide_preview()
   -- 1. Hide the window (and border)
   -- 2. Resize prompt & results windows accordingly
 end
 
-
-function Picker.close_windows(status)
+function Picker.close_windows(self, status)
```
Conni2461, line 556: We could just name say `Picker:close_windows` then we don't need to say self.

lua/telescope/pickers.lua
```diff
@@ -1085,7 +1128,7 @@ function pickers.on_close_prompt(prompt_bufnr)
   -- TODO: This is an attempt to clear all the memory stuff we may have left.
   -- vim.api.nvim_buf_detach(prompt_bufnr)
 
-  picker.close_windows(status)
+  picker.close_windows(picker, status)
```
Conni2461, line 1131: and here we can do `picker:close_windows(status)`

lua/telescope/config.lua
akiyosi: Thank you for your review.
I have added the documentation.
I have also rebased and successfully generated the documentation.

lua/telescope/pickers.lua
akiyosi: This is so that the prompt, result, and preview windows will be placed in the default display position in telescope.nvim when the external window layout process is done in the GUI. 
If this part is not moved, prompt will be displayed on top. However, since this is a small minor issue, and is strongly based on the layout process of goneovim, it may be better to revert it back.

lua/telescope/pickers.lua
akiyosi: This has been fixed.

lua/telescope/pickers.lua
akiyosi: This was also fixed :)

lua/telescope/pickers.lua
Conni2461: Oh interesting i didn't know that (but i also have never used a GUI). We can keep it :)

But does that mean that option `prompt_position = 'top'` doesn't work?

akiyosi: Yes, `prompt_position`, etc. do not work. This is because, as far as I know, there is no way to directly specify the position of the external window in the current neovim implementation. 
One way, as mentioned at the beginning, is to display them as a normal float window once before externalizing them, and then externalize them instantly. However, I have not been able to investigate how to get this to work.

lua/telescope/config.lua
```diff
@@ -88,6 +88,14 @@ function config.set_defaults(defaults)
   set("layout_strategy", "horizontal")
   set("layout_defaults", {})
 
+  set("externalize", false, [[
+    Displays the prompt, result, and preview windows as external windows if set to true.
+    Only available when using an UI with ui-multigrid support.
```
elianiva, line 93: `an UI` -> `a UI`

lua/telescope/pickers.lua
```diff
@@ -509,22 +540,30 @@ function Picker:find()
   elseif self.initial_mode ~= "normal" then
     error("Invalid setting for initial_mode: " .. self.initial_mode)
   end
+
+  if self.window.externalize then
+    a.nvim_win_set_config(results_win, {external = true, width=width, height=height})
```
elianiva, line 545: probably add space to these lines for consistency. `height=height` -> `height = height`

elianiva:
Just some minor issues, everything else looks good to me. Will need to test it out maybe in an hour or two to make sure TUI version works fine :)

elianiva:
huh, I can't see the result items but I *can* open the file.

Screenshot_20210401_072326

akiyosi:
@elianiva 
Have you enabled externalize in the TUI?

elianiva:
> Have you enabled externalize in the TUI?

`false` should disable it right? as in, it works as usual. I tried setting it to true, but the border is gone (and I think that's expected), but if I set it to false, the border is there, it's just that the result window doesn't contain anything like my screenshot above.

Seems like moving [these lines](https://github.com/nvim-telescope/telescope.nvim/blob/cd149356a9f2efe8c427949e50e7b10e52c4b3dc/lua/telescope/pickers.lua#L391-L410) to where it was fixed the issue, but that's required for GUIs right?

Maybe the border window being placed on top of the actual result window; hence I couldn't see the results.

akiyosi:
Yes, false disables the feature.

Hmm, perhaps your guess is correct, it is hidden in the border window.
However, the corresponding lines are moving in order, including the results border window, so I wonder why it appears this way...

Also, this problem is not reproduced in my environment.



I will do some research.

Conni2461:
Is something like this still required for goneovim or can we close this PR?

Conni2461:
Closing because of inactivity.

PR #670, opened 2021-03-19
live_grep_raw

jesseleite:
I proposed this idea on the Telescope Gitter, and there seemed to be interest from fearless leader @tjdevries, among others. The idea is to port [vim-agriculture](https://github.com/jesseleite/vim-agriculture)'s functionality into a ` live_grep_raw` Telescope finder.

### Behaviour

- It would behave like the `live_grep` finder, but would not escape user input, allowing regex search by default:
    - `func.*someFunc`
- It would also allow the user to pass CLI flags and/or a path argument, right into the Telescope prompt:
    - `-u "func.*someFunc" vendor`
- If the command returns without errors, show a typical result count (ie. `2 / 6` or `0 / 0`) at the right side of the prompt.
- We should gracefully handle errors (due to incomplete regex, etc):
    - Replace the `0 / 0` result count with an `Error` string
    - Return all of rg's error output in the preview window
- We should set default max result count to 1000 or something, to minimize chaos.

### Example

CleanShot 2021-03-19 at 01 59 43

### Todo

- [x] Support both `rg` and `ag` by default.
    - Telescope already uses `rg` by default.
    - Configure `vimgrep_arguments = {'ag', '--nogroup', '--column'}` to use `ag`.
- [x] Accept params before a quoted query string.
- [x] Accept params after a quoted query string.
- [x] Accept path(s) after a quoted query string.
- [ ] Highlight/colour matching query output on each result line.
    - Build custom sorter that highlights regex?
    - Or find a way to render highlights from rg directly?
- [ ] Gracefully handle errors.
- [ ] Handle max result chaos.
    - @tjdevries said he might already have something for limiting max results?

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -96,13 +96,19 @@ files.live_grep_raw = function(opts)
 
     if single_quoted or double_quoted then
       local before_args = prompt:match("(.-)['\"]")
-      -- local after_args = prompt:match("['\"](.-)")
-      for arg in before_args:gmatch("%S+") do
-        table.insert(args, arg)
+      local after_args = prompt:match(".*['\"](.*)")
+      local all_args = before_args .. ' ' .. after_args
+      for arg in all_args:gmatch("%S+") do
+        if arg:match('^-') then
+          table.insert(args, arg)
+        else
+          -- Show graceful grep error
+          -- Path cannot come before query
+        end
       end
     end
 
-    return vim.tbl_flatten { args, query, path }
+    return vim.tbl_flatten { args, '--', query, path }
   end
 
   pickers.new(opts, {
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -5,6 +5,7 @@ local finders = require('telescope.finders')
 local make_entry = require('telescope.make_entry')
 local pickers = require('telescope.pickers')
 local previewers = require('telescope.previewers')
+local sorters = require('telescope.sorters')
 local utils = require('telescope.utils')
 local conf = require('telescope.config').values
 
@@ -128,6 +129,7 @@ files.live_grep_raw = function(opts)
     prompt_title = 'Live Grep Raw',
     finder = finders.new_job(cmd_generator, opts.entry_maker, opts.max_results, opts.cwd),
     previewer = conf.grep_previewer(opts),
+    sorter = sorters.grep_highlighter_only(opts),
   }):find()
 end
 
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -64,6 +64,54 @@ files.live_grep = function(opts)
   }):find()
 end
 
+files.live_grep_raw = function(opts)
+  opts.vimgrep_arguments = opts.vimgrep_arguments or conf.vimgrep_arguments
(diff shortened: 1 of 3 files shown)


Conni2461:
@jesseleite do you need any help here?

jesseleite:
Not ignoring Conni here! We chatted in gitter. I just need to find the time to dig in again, will do soon 😅

muniter:
> Highlight/colour matching query output on each result line. 

Ripgrep has the '--json' option which returns the matches to be highlighted. Might be very useful If you haven't consider it. @jesseleite 

https://www.mankier.com/1/rg#--json

jesseleite:
@muniter Oh interesting, will check it out thanks!

Shatur:
@jesseleite, are you planning to continue working on it?

jesseleite:
@Shatur I am, but we just had a baby, and @tjdevries told me it might be good to wait for all the async stuff he's working on to be merged in.

jesseleite:
PS. @tjdevries and I dadn't have a baby; My wife and I had a baby.

jesseleite:
PS. I got asyncing feeling that some people will misinterpret that last comment. @tjdevries and I dadn't have a baby; My wife and I had a baby.

TC72:
Hey Jesse, I've had some time to look at this and it's great.
Just wanted to check what async features you're waiting for from @tjdevries?

Your code is using `finders.new_job` which has now been changed internally to use the new `async_job_finder`.
So I think your code is good to go on that front.

weeman1337:
The new extension exists: https://github.com/nvim-telescope/telescope-live-grep-raw.nvim :partying_face:

tjdevries:
So can we close this PR then?

jesseleite:
Just trying the extension by @weeman1337 now, very cool!

I apologize for letting this go so long, but as far as closing this PR, the above extension doesn't quite work to the spec I was going for. A couple things I noticed off the bat...

- I can't search a string like `function create` without wrapping it in quotes first (ie. `'function create'`).
    - I personally think spaces should be treated as part of the search query, _unless_ command line options are needed.
- The highlighting only works for literal single-word queries, but not when quoting, or when performing a regex query.
    - My idea was to either parse rg's `--json` output, or parse the ansi colour codes in rg's output, to determine how to highlight the telescope results list.
    - I can't remember @TC72, but I believe you might have worked some of this out?

tjdevries:
I'd rather see that work done in the other repo, rather than here. I think it makes more sense to focus these cooler and more powerful pickers in dedicated locations, instead of trying to maintain support in the core repo. I am trying to push more things out into extensions because there are already too many things in the core repo for me to maintain

TC72:
@jesseleite I got the highlighting using code to convert color codes to highlighting but it was a complete mess.
I'd parse it to get the highlights and strip the codes from the text.
but then the text could be adjusted based on path shorting and devicons which meant I then had to adjust every highlight again.

The better option would probably be to take a 1st run which strips the codes and you'd be able to make any adjustments. Then the highlights would have the correct location.

But that means parsing the output twice.

What I'd much prefer is if there was something built into telescope which attached highlights to a part of the final string so the highlight code could take care of any adjustments made to shorten paths for example.

JSON would be great in terms of finding the data, but I'm worried it would make a huge difference to the size of data we get back from `rg`.

jesseleite:
Sounds good Teej 👍

tjdevries:
> Sounds good Teej +1

<3

PR #676, opened 2021-03-21
chore: fix lint error

elianiva:
I did a mistake by not scrolling up the test log, so I missed one of the reason why the test was failing (long lines) and thought it was caused by trailing whitespace which is unrelated to the previous PR.


elianiva:
and the doc gets removed for some reason :))
no idea about that

Conni2461:
docgen weirdness will be over soon. Promise. Just different ABI version. I explained it here: https://github.com/nvim-telescope/telescope.nvim/pull/664#issuecomment-801451285. Tj is aware of it (i told him) and i also saw this: https://twitter.com/TeejDeVries/status/1372881108979871749?s=20

elianiva:
moved the fix to #681

PR #680, opened 2021-03-21
fix: properly set default_text if prompt_prefix exists

fdschmidt93:
Currently, `default_text` will only be properly set if it starts with the `prompt_prefix`.

For example, `lua require'telescope.builtin'.lsp_document_symbols{default_text='>:class:'`} would have to be passed to set the line to ':class' (which in turn triggers class prefiltering) to properly set the line, assuming the prompt_prefix is >.

See also: https://www.reddit.com/r/neovim/comments/m9vwy2/search_for_class_using_telescope_and_native_lsp/

e: why does this PR of a freshly rebased fork drag on some unrelated commits? :disappointed:

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -401,9 +401,8 @@ files.tags = function(opts)
 end
 
 files.current_buffer_tags = function(opts)
-  return files.tags(vim.tbl_extend("force", {only_current_file = true, hide_filename = true}, opts))
+  return files.tags(vim.tbl_extend("force", {prompt_title = 'Current Buffer Tags', only_current_file = true, hide_filename = true}, opts))
 end
-
 
 local function apply_checks(mod)
   for k, v in pairs(mod) do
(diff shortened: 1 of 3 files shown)


Conni2461:
Just prompt buffers being prompt buffer :laughing: (Upstream bug) i am aware of this. But this workaround causes different problems and will be broken as soon as these issues are resolved. Problems with this solution:
- prompt prefix highlighting is broken
- default_text not delete able 

I was playing around with a more robust workaround here https://github.com/nvim-telescope/telescope.nvim/pull/521/files#diff-9ac2063a221b713185ddb6df3157cf01edf3b12894c2ae984e14250bf16ebd89R547-R552


Also no idea why this PR has some unrelated commits :laughing:

fdschmidt93:
> default_text not delete able

Ah yes, saw that  - I guess that stems yet again from the _other_ upstream bug though can also be seen as unintended 'intended behaviour' :thinking:  :laughing:.

> I was playing around with a more robust workaround here

Alrighty, then let's close this :)

PR #681, opened 2021-03-22
Support for preprocessing of line in finders

psiska:
Rationale: Different command line tools outputs single line of json as
result of operation (hoogle - https://github.com/ndmitchell/hoogle) or
(neuron - https://github.com/srid/neuron). Currently there is no way to
create multiple results from one line

Solution: If fn_preprocess function from line to {item} is provided (is
not nil) then run it and call entry_maker and process_result for each
item in this result.
If not provided behave as before.

Example:
Given the program which outputs single json line for single invocation we can use following preprocess function
```lua
local function do_preprocess(line)
  local result = {}
  local t = string.sub(line, 1, 1)
  -- Filter out non json lines
  if t ~= '[' and t ~= '{' then
    return {}
  end
  local pJson = json.decode(line)
  for _, v in ipairs(pJson) do
    table.insert(result, v)
  end
  return result
end
```
In this case entry maker and process result will be called for each result.

### DIFF
--- a/lua/telescope/finders.lua
+++ b/lua/telescope/finders.lua
@@ -37,6 +37,8 @@ local JobFinder = _callable_obj()
 ---
 ---@param opts table Keys:
 --     fn_command function The function to call
+--     fn_preprocess function The optional function to be called for each line before calling entry_maker.
+--       Allows for creating multiple entries (table) from single line.
 function JobFinder:new(opts)
   opts = opts or {}
 
@@ -70,7 +72,8 @@ function JobFinder:_find(prompt, process_result, process_complete)
     if not line or line == "" then
       return
     end
-    -- If fn_preprocess is available use on the line, afterwards run entry_maker on each of the result.
+    -- If fn_preprocess is defined, call it for each line producing multiple results.
+    -- Call entry_maker on each result.
     if self.fn_preprocess ~= nil then
       local lines = self.fn_preprocess(line)
       for _, item in ipairs (lines) do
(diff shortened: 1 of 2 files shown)


tami5:
Okay the linting is complaining about 

-    lua/telescope/pickers/highlights.lua:46:39: (W612) line contains trailing whitespace
-    lua/telescope/builtin/files.lua:404:121: (W631) line is too long (138 > 120)

Also, it would be nice if you can expose the above snippet if it's required to process json through utils.lua.

elianiva:
you could either wait for https://github.com/nvim-telescope/telescope.nvim/pull/676 to fix lint error or do the same thing I did there and I'll close that PR, either way is fine for me. 

The reason I didn't merge that PR is because the docgen is *kinda* broken so it removes the docs and no one will like that :laughing:

lua/telescope/finders.lua
```diff
@@ -115,6 +125,7 @@ function OneshotJobFinder:new(opts)
   local obj = setmetatable({
     fn_command = opts.fn_command,
     entry_maker = opts.entry_maker or make_entry.from_string,
+    fn_preprocess = opts.fn_preprocess,
```
tami5, line 128: @psiska It would be extremely useful for future us if a comment here explain what fn_preporcess does and when to use it. Other than that we should merge

psiska: Of course I will put there a description. Thanks

psiska: I've just pushed the documentation. Since I am new to lua, not sure if it is enough. 
Also I've included the section in Readme.md with usage example.

Conni2461:
We should probably ping @oberblastmeister for this PR because of the upcoming async await changes

tjdevries:
Yes, let's wait til after asyncify pr lands. This will change this quite a bit.

I also have to think about this a bit more as well.

psiska:
Currently I am updating tools I need to support the JSONL support in order not to need to land this PR.
However until those changes are merged is there an option to run tools via pipe and process with jq to format suitable for telescope?

Something along the lines cmd = "tool | jq" in such case I will get one line == one entry. In other words mode which telescope is used to.
If there is no such possibility I would rework the PR and rename preprocess to line_processor to better explain the purpose.

wookayin:
I find this still useful but the PR has been closed unfortunately. Can we consider adding this feature or have we decided to not accept preprocessors?

PR #683, opened 2021-03-23
Better relative path output in `live_grep` finder

jesseleite:
Default to nil path, for better relative path output.  It's subtle, but notice the `./` in the path results, which is unnecessary.

## Before

CleanShot 2021-03-22 at 23 31 02

## After

CleanShot 2021-03-22 at 23 31 44

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -49,7 +49,7 @@ files.live_grep = function(opts)
 
       prompt = escape_chars(prompt)
 
-      return flatten { vimgrep_arguments, prompt, opts.search_dirs or '.' }
+      return flatten { vimgrep_arguments, prompt, opts.search_dirs or nil }
     end,
     opts.entry_maker or make_entry.gen_from_vimgrep(opts),
     opts.max_results,


Conni2461:
Removing `'.'` breaks windows. But we can only add it when we are on windows

jesseleite:
Done!

lua/telescope/builtin/files.lua
```diff
@@ -49,7 +49,13 @@ files.live_grep = function(opts)
 
       prompt = escape_chars(prompt)
 
-      return flatten { vimgrep_arguments, prompt, opts.search_dirs or '.' }
+      local default_search_dir = nil
+
+      if vim.fn.has('win32') == 1 then
```
Conni2461, line 54: I kinda like `os_sep == '\\'` more and it should be faster :laughing: 

Otherwise LGTM

Conni2461 (approved):
LGTM.

Do you also want to add the changes to `grep_string` because it also suffers from this issue

jesseleite:
Good call, done.

Also, I refactored a bit to avoid passing a flattened `nil` to `new_oneshot_job()` (which errored).  Realistically, we don't want/need to pass the search_dir arg **_AT ALL_**, unless in a windows environment. I think this is cleaner, what you think?

elianiva (approved):

Conni2461 (approved):
Yep makes sense to me.
Thanks :)

PR #690, opened 2021-03-26
Use plenary.strings instead of logic here

delphinus:
This needs https://github.com/nvim-lua/plenary.nvim/pull/96 to be merged.


lua/telescope/utils.lua
```diff
@@ -249,95 +249,6 @@ function utils.get_os_command_output(cmd, cwd)
   return stdout, ret, stderr
 end
 
-utils.strdisplaywidth = (function()
```
Conni2461, line 252: Can we not just remove these functions but do something like this
```lua
utils.x = function()
  error("x deprecated. please use plenary.strings.x") 
end
```

Conni2461:
There is more usage of these utils now. (somewhere i can't remember)

lua/telescope/utils.lua
delphinus: Added on fd1a736

Conni2461:
We have still `utils.strdisplaywidth` in `git.branches` and `lsp.code_action`, which breaks both function in this pr.

Conni2461:
Also you can remove `entry_display.truncate` in telescope afterwards and call `strings.truncate`

Conni2461:
@delphinus I haven't heard from you in quite some time, so i finished it myself. (I hope everything works)

Thanks for all the work on this :)

PR #698, opened 2021-03-29
layout(flex): fix not being able to read user config

tami5:


tjdevries:
Is this that it doesn't respect user config for these values? Cause I think this prevents passing in an override when using flex

tjdevries:
I think I've implemented the correct fix here: https://github.com/nvim-telescope/telescope.nvim/pull/773 As well as some cool doc updates

PR #695, opened 2021-03-28
Show committer, upstream and date in branch finder

kyoh86:
Show short informations for all branches.

telescope-git

for: #569

PR #700, opened 2021-03-29
Fix regressions with #695

kyoh86:
Thank @Conni2461 !

for: https://github.com/nvim-telescope/telescope.nvim/pull/695#discussion_r603514718
for: https://github.com/nvim-telescope/telescope.nvim/pull/695#issuecomment-809608089

It may be fixed by this.

:memo: 
> Also with the date information we can also think of initial sorting based on the date.

I can't fix it right away, so it is not included in this one.

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -79,7 +79,7 @@ end
 
 git.branches = function(opts)
   local format = '{'
-              ..   '"head":%(if:equals=*)%(HEAD)%(then)true%(else)false%(end)'
+              ..   '"head":"%(HEAD)"'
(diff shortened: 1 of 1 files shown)


lua/telescope/builtin/git.lua
```diff
@@ -79,7 +79,7 @@ end
 
 git.branches = function(opts)
   local format = '{'
-              ..   '"head":%(if:equals=*)%(HEAD)%(then)true%(else)false%(end)'
+              ..   '"head":"%(HEAD)"'
```
kyoh86, line 82: conversion is not required.

Conni2461:
Close because will happen in #704

Issue #702, opened 2021-03-30
git_branch throws an error for specific commit authors due to invalid JSON 


Strayer:
### Description

Since 2e03f67de9eab023098bd05916bf26556520a467, `:Telescope git_branch` tries to show author information. It does this by telling Git to output the ref with an output string that tries to resemble JSON. This can fail when the author name contains escape sequences like this:

```
❯ git for-each-ref --format '{"head":%(if:equals=*)%(HEAD)%(then)true%(else)false%(end),"refname":"%(refname)","authorname":"%(authorname)"%(if)%(upstream)%(then),"upstream":"%(upstream:lstrip=2)"%(else),"upstream":""%(end),"committerdate":"%(committerdate:format-local:%Y/%m/%d %H:%M:%S)"}' | grep Haj
{"head":false,"refname":"refs/tags/0.5.129","authorname":"U-SALTY-WINTENDO\Haj Okuda","upstream":"","committerdate":"2017/09/21 06:16:43"}
```

Relevant commit: https://github.com/qmk/qmk_firmware/commit/ea7792b6c6a09d98598f25989e08b262641b9b26

```
❯ git show ea7792b6c6a09d98598f25989e08b262641b9b26 | head
commit ea7792b6c6a09d98598f25989e08b262641b9b26
Author: U-SALTY-WINTENDO\Haj Okuda <hajimeokuda@gmail.com>
Date:   Mon Sep 18 23:38:31 2017 -0400

    Adding my keymap with Colemak Mod-DH

diff --git a/keyboards/kinesis/keymaps/salty/Makefile b/keyboards/kinesis/keymaps/salty/Makefile
new file mode 100644
index 000000000..4346cf009
--- /dev/null
```

**Expected Behavior**

git_branch should not throw an error and show the author info correctly.

**Actual Behavior**

`:Telescope git_branch` outputs an error message:

`E5108: Error executing lua Vim:E474: Unknown escape sequence: \Haj Okuda","upstream":"","committerdate":"2017/09/21 06:16:43"}`

### Details

I didn't add a test configuration since this seems configuration-independent. It is very easy to reproduce by cloning the qmk_firmware repository.

<details><summary>Reproduce</summary>

1. git clone https://github.com/qmk/qmk_firmware.git
2. cd qmk_firmware
3. nvim
4. :Telescope git_branch

</details>

<details><summary>Environment</summary>

[...]

PR #704, opened 2021-03-30
use the quoted fields instead of json-formatting

kyoh86:
fix: #702

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -78,18 +78,12 @@ git.bcommits = function(opts)
 end
 
 git.branches = function(opts)
-  local format = '{'
-              ..   '"head":%(if:equals=*)%(HEAD)%(then)true%(else)false%(end)'
-              ..   ',"refname":"%(refname)"'
-              ..   ',"authorname":"%(authorname)"'
-              ..   '%(if)%(upstream)%(then)'
-              ..     ',"upstream":"%(upstream:lstrip=2)"'
-              ..   '%(else)'
-              ..     ',"upstream":""'
-              ..   '%(end)'
-              ..   ',"committerdate":"%(committerdate:format-local:%Y/%m/%d %H:%M:%S)"'
-              .. '}'
-  local output = utils.get_os_command_output({ 'git', 'for-each-ref', '--format', format }, opts.cwd)
+  local format = '%(HEAD)'
+              .. '%(refname)'
+              .. '%(authorname)'
+              .. '%(upstream:lstrip=2)'
+              .. '%(committerdate:format-local:%Y/%m/%d%H:%M:%S)'
+  local output = utils.get_os_command_output({ 'git', 'for-each-ref', '--perl', '--format', format }, opts.cwd)
 
   local results = {}
   local widths = {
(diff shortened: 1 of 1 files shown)


Conni2461:
Can we integrate #700 into this?

kyoh86:
@Conni2461 thanks, I tried it. How's this?

lua/telescope/builtin/git.lua
```diff
@@ -98,26 +92,42 @@ git.branches = function(opts)
     upstream = 0,
     committerdate = 0,
   }
-  local register_entry = function(entry, trim_refname_prefix)
-    entry.name = string.sub(entry.refname, string.len(trim_refname_prefix)+1)
+  local unescape_single_quote = function(v)
+      return string.gsub(v, "\\([\\'])", "%1")
+  end
+  local parse_line = function(line)
+    local fields = vim.fn.split(string.sub(line, 2, -2), "''", true)
```
elianiva, line 99: probably better to use `vim.split` instead? I'd avoid `vim.fn.*` when possible :)

kyoh86: thanks! I fixed it.

lua/telescope/builtin/git.lua
```diff
@@ -98,26 +92,42 @@ git.branches = function(opts)
...
+  local parse_line = function(line)
+    local fields = vim.split(string.sub(line, 2, -2), "''", true)
+    local entry = {
+      head = fields[1],
+      refname = unescape_single_quote(fields[2]),
+      authorname = unescape_single_quote(fields[3]),
+      upstream = unescape_single_quote(fields[4]),
+      committerdate = fields[5],
+    }
+    local prefix
```
elianiva, line 107: is this `prefix` only used for line 120? If so, I think it'd be better to use number directly so we don't need to calculate the length using `string.len`.

lua/telescope/builtin/git.lua
```diff
@@ -98,26 +92,42 @@ git.branches = function(opts)
...
+      return
+    end
+    local index = 1
+    if entry.head ~= '*' then
+      index = #results + 1
+    end
+
+    entry.name = string.sub(entry.refname, string.len(prefix)+1)
     for key, value in pairs(widths) do
       widths[key] = math.max(value, vim.fn.strdisplaywidth(entry[key]))
```
elianiva, line 122: we also have `strdisplaywidth` in `utils.strdisplaywidth` which uses `ffi` instead of `vim.fn.*`. Would be nice if you can change this as well :)

elianiva:
except those 2 comments, LGTM. Thank you!

lua/telescope/builtin/git.lua
kyoh86: fixed.

Conni2461:
I still don't see my current branch. Is this intended behavior? I think its kinda helpful to just see that branch with an indicator that this is my branch, because not everybody has the current branch in the statusbar and sometimes i forget how my current branch is actually named :rofl: 

Also i am waiting for an answer to elianivas last unresolved comment :)

kyoh86:
@Conni2461 

I can see my current branch.

Conni2461:
That kinda embarrassing, i checkout the wrong PR :rofl: so i see my current branch :+1: 

Indicator would be good :)

kyoh86:
I'll show an indicator in other time.

kyoh86:
@Conni2461 I will implement the indicator on another issue (https://github.com/nvim-telescope/telescope.nvim/issues/726).
So could you merge this PR?

Conni2461:
Hmm idk i feel like this should be part of this PR, because we had a `indicator` before https://github.com/nvim-telescope/telescope.nvim/commit/2e03f67de9eab023098bd05916bf26556520a467

with this part:
```lua
local addition = vim.startswith(entry, '*') and '* ' or ' '
...
ordinal = addition .. entry,
display = addition .. entry,
```

kyoh86:
@Conni2461 I see and have done.

Conni2461 (approved):
LGTM thanks :)

PR #705, opened 2021-03-30
added a new DynamicFinder (which can be used with rust_analyzer)

oberblastmeister:
**note** requires async await https://github.com/nvim-lua/plenary.nvim/pull/83

If you want to test this new finder you can run `:Telescope lsp_dynamic_workspace_symbols` with rust_analyzer or another lsp server. The picker is completely asynchronous as it uses `buf_request`

The implementation of this is not good because there is no way for a finder to check if that specific find has been canceled. This can cause many issues with asynchronous finds because if the user has typed an additional character it will still be working on the previous find, causing ui distortions because the old find will still try to change the buffer. This can be fixed by giving the finder a cancellation channel to check. Furthermore, finders will set the screen on every entry instead of processing all the entries first and then setting them all together. This is another issue as it can cause extreme flickering. This can be seen with this finder and also the `JobFinder` (which live grep uses). #691 solves these issues so we will probably have to wait for that. It also allows stuff to be asynchronously processed


lua/telescope/finders.lua
```diff
@@ -185,4 +215,8 @@ finders.new_table = function(t)
   return async_static_finder(t)
 end
 
+finders.new_live = function(t)
```
tjdevries, line 218: let's call it `new_dynamic` I think, right?

lua/telescope/builtin/lsp.lua
```diff
@@ -218,6 +221,35 @@ lsp.workspace_symbols = function(opts)
   }):find()
 end
 
+local function get_workspace_symbols_requester(bufnr)
+  local cancel = function() end
```
tjdevries, line 225: we don't actually use this at all.

tom-anders:
When I try to use this, I always get this error:

`Error executing vim.schedule lua callback: ....vim/plugged/plenary.nvim/lua/plenary/async_lib/util.lua:170: Oneshot channel can only send once`

Is there some special branch of plenary that I need to be on for this to work?

tom-anders:
Ah turns out this happens when having multiple language servers (I use both clangd and ccs). Maybe there could be an option to only use a specific server?

oberblastmeister:
Yeah that is a known issue, it is because I didn't update it to use the new `buf_request_all` that was recently merged into neovim.

kidonng:
`lsp_dynamic_workspace_symbols` should be mentioned in the docs :) I spent several days to find this PR

kidonng:
> Currently there is an attempt to document all builtins (readme + vim doc (includes options of builtins)), here: https://github.com/nvim-telescope/telescope.nvim/pull/783
> 
> Its almost done

That's good to know!

PR #718, opened 2021-04-03
Fix #707: lsp_workspace_symbols add support for multiple clients.

### LINKED ISSUE #707: lsp_workspace_symbols does not work with multiple lsp servers
### Description

If an LSP client receives an id other than 1, then `lsp_workspace_symbols` will fail. This may happen if you use more
than one LSP client, and you are switching between buffers of different file types.

The guard in `telescope/builtin/lsp.lua` will stop if it does not find result from client with id 1:
```lua
-- Clangd returns { { result = {} } } for query=''
  if not results_lsp or vim.tbl_isempty(results_lsp) or
     vim.tbl_isempty(results_lsp[1]) or vim.tbl_isempty(results_lsp[1].result) then
    print("No results from workspace/symbol. Maybe try a different query: " ..
      "Telescope lsp_workspace_symbols query=example")
    return
  end
```

**Expected Behavior**
`lsp_workspace_symbols` should work with even though there is no result for id 1, as long as there is result for
another client.

**Actual Behavior**
`lsp_workspace_symbols` does not work if there is no result for LSP client with id 1. Even though there is result for
another client.

ovikk13:
Fixes issue #707

Fixes issue where lsp_workspace_symbols stops if there are no results
from LSP client with id 1. This is an issue if you have multiple
clients, and only one of them returns results.

### DIFF
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -184,22 +184,20 @@ lsp.workspace_symbols = function(opts)
   local params = {query = opts.query or ''}
   local results_lsp = vim.lsp.buf_request_sync(0, "workspace/symbol", params, opts.timeout or 10000)
 
-  -- Clangd returns { { result = {} } } for query=''
-  if not results_lsp or vim.tbl_isempty(results_lsp) or
-     vim.tbl_isempty(results_lsp[1]) or vim.tbl_isempty(results_lsp[1].result) then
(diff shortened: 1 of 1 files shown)


fdschmidt93:
Works well, I've also tested a scenario similar as per the issue. Thanks for the fix and sorry again for reading over your issue.

cc @Conni2461 this can be merged

ovikk13:
Hey, no problem:) Thanks for the quick review;)

PR #720, opened 2021-04-03
Fix git rebase input message

medeirosthiago:
Very tiny PR that fixes input message when git rebasing branches. 😅 

I was wondering if I should try to also add current branch info like`... rebase master into feat-branch` but was afraid to make it too large for an input message.


tjdevries:
We can add more info later if people want. Thanks for doing this 😁

PR #732, opened 2021-04-06
feat: buf highlights for current buffer fuzzy find

tjdevries:
Don't know why the picture is weird, but gives you the right idea at least.

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -685,7 +685,7 @@ function make_entry.gen_from_buffer_lines(opts)
     separator = ' │ ',
     items = {
       { width = 5 },
-      { dynamic_highlight = true, },
+      { remaining = true, },
     },
   }
 
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -693,7 +693,7 @@ function make_entry.gen_from_buffer_lines(opts)
 
     return displayer {
       { entry.lnum, opts.lnum_highlight_group or 'TelescopeResultsSpecialComment' },
-      { 
+      {
         entry.line, function()
           if not opts.line_highlights then return {} end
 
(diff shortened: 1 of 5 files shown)


lua/telescope/make_entry.lua
```diff
@@ -685,14 +685,30 @@ function make_entry.gen_from_buffer_lines(opts)
     separator = ' │ ',
     items = {
       { width = 5 },
-      { remaining = true },
+      { dynamic_highlight = true, },
```
Conni2461, line 688: Is that actually a thing? I think you solved this differently so we should change that line back to not confuse people reading that code :)

Conni2461 (approved):
Tested it. Works like a charm. Lets ship it :)

Thanks tj :) also treesitter is amazing

lua/telescope/make_entry.lua
tjdevries: well technically, I don't think `remaining = true` does anything either haha

tami5:
Amazing work @tjdevries works well

lua/telescope/make_entry.lua
Conni2461: Good call.
Just thinking out load: We should probably make it explicit and error on wrong key input, so people aren't up being confused if they have a typo, when using `width` or `right_justify` or something sunjon wanted some time ago, was right side align, not column but result side (currently not possible because we need to lazy evaluate the result window width).

For me it makes sense to validate the items tbl.

jesseleite:
Hey @tjdevries, absolutely love this one, BUT...

If I have treesitter highlighting disabled for a specific syntax, ie)

```lua
require'nvim-treesitter.configs'.setup {
  ensure_installed = "maintained",
  highlight = {
    enable = true,
    disable = {'php'},
  },
}
```

Opening this finder will not only use treesitter in the results view, but will switch to treesitter highlighting in my actual file buffer as well. I can flip back to polyglot highlighting after closing telescope by running `:e`, but it's a bit annoying when I've intentionally ignored the syntax in my treesitter config, nawmean?

PS. Don't judge me. I want to fully switch to treesitter, but the php implementation is a bit buggy in my experience compared to polyglot. I'll get there though!

Issue #222, opened 2020-11-05
lsp_code_actions don't execute selected action


ChrisAmelia:
```
NVIM v0.5.0-812-gd17e50879
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

**init.vim**
```vim
set runtimepath^=~/.local/share/nvim/plugged/nvim-lspconfig
set runtimepath^=~/.local/share/nvim/plugged/plenary.nvim
set runtimepath^=~/.local/share/nvim/plugged/telescope.nvim
set runtimepath^=~/.local/share/nvim/plugged/popup.nvim

lua << EOF
	local nvim_lsp = require 'nvim_lsp'
	local root_pattern = nvim_lsp.util.root_pattern

	require'nvim_lsp'.jdtls.setup{
		root_dir = root_pattern(".git", "pom.xml"),
	}
EOF
```

**Main.java**
```java
public class Main {

    public static void main(String[] args) {
        ArrayList list = new ArrayList<>(); // ArrayList cannot be resolved to a type
    }

}
```
 
![image](https://user-images.githubusercontent.com/16366403/98224283-96579180-1f53-11eb-97bc-6d07be5b2830.png)

When calling `lsp_code_actions()` with `:lua require'telescope.bulitin'.lsp_code_actions{}`, the popup with the different actions appears as expected. Though upon selecting one action, nothing happens, I would expect the import to be done.

Edit: Seems the bare function `:lua vim.lsp.buf.code_action()` doesn't, for this case, work anymore.

tjdevries:
Hmmm, is this still happening? Does this happen no matter how you call the code action stuff?

ChrisAmelia:
Yes, the issue is still happening (tested with the last appimage), I've opened a thread there https://github.com/neovim/nvim-lspconfig/issues/397

I simply open the java file then execute `:lua vim.lsp.buf.code_action()` then select any available option.

ChrisAmelia:
With https://github.com/neovim/nvim-lspconfig/pull/429/files

`:lua vim.lsp.buf.code_action()` works but not `:lua require'telescope.builtin'.lsp_code_actions()`.

tjdevries:
Hmmm.... when I just tried it with Go, it worked perfectly.

ChrisAmelia:
Effectively, tested with go and it works.

ChrisAmelia:
Doesn't telescope's `code_actions()` rely on nvim's built-in one?
[...]

PR #735, opened 2021-04-07
feat: Allow overriding transform and execution for code actions

tjdevries:

PR #738, opened 2021-04-08
feat: Remove version field if zero from codeaction calls

Matsuuu:
Fix implemented based on the discussion at https://github.com/neovim/nvim-lspconfig/pull/429#discussion_r537110059

And the code snippets in the [jdtls implementation](https://github.com/neovim/nvim-lspconfig/blob/486f72a25ea2ee7f81648fdfd8999a155049e466/lua/lspconfig/jdtls.lua#L62) 

If merged with the [code_actions_fixup](https://github.com/nvim-telescope/telescope.nvim/pull/735) branch, would fix the problem of Java users not being able to use codeactions

### DIFF
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -217,6 +217,24 @@ lsp.code_actions = function(opts)
     }
   end
 
+  -- If the text document version is 0, set it to nil instead so that Neovim
+  -- won't refuse to update a buffer that it believes is newer than edits.
+  -- See: https://github.com/eclipse/eclipse.jdt.ls/issues/1695
+  -- Source: https://github.com/neovim/nvim-lspconfig/blob/486f72a25ea2ee7f81648fdfd8999a155049e466/lua/lspconfig/jdtls.lua#L62
+  local function fix_zero_version(workspace_edit)
+    if workspace_edit and workspace_edit.documentChanges then
+      for _, change in pairs(workspace_edit.documentChanges) do
+        local text_document = change.textDocument
+        if  text_document 
+            and text_document.version 
+            and text_document.version == 0 then
+          text_document.version = nil
+        end
+      end
+    end
+    return workspace_edit
+  end
+
   --[[
   -- actions is (Command | CodeAction)[] | null
   -- CodeAction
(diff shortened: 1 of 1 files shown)


Matsuuu:
I've tested this code with Javascript, Golang, Java and Clojure, and all seemed to be working.

@tjdevries Do you see any flaws in this? Not a Lua programmer so don't know if I've missed something.

Could maybe move the code from transform_action to it's own function?

Builds upon your PR which implemented the transform hook

Conni2461:
Do i see this correctly that this is a continuation of #735 and fixes #222

Can you rebase it after that i would test it? I know its been a while since you opened this PR. Sorry that you had to wait for that long to receive an answer. I currently clean up some PRs. 

If i understand this correctly this is still required to for java? Can we just merge this into master rather than into tjs pr?

Matsuuu:
Heya,

Oh yeah this still hasn't been merged :D 

I'll try to get around checking this out this week

Matsuuu:
I..

- Verified that the problem still exists
- Applied the fix, rebased the branch and pointed it to master
- Verified that it works after that.


The jdtls seems a bit of a mess since there's issues been open for years, without closure.

This howevery fixes the problem. There is a new small thing with jdtls complaining about `-32601: No delegateCommandHandler for java.apply.workspaceEdit` after codeaction, but I think that's a topic for another issue.

This is good to go imo

Conni2461:
okay cool thanks for the fix :)

I tested it (not with java, i dont write java) with clangd and have not seen any regressions. I took the liberty to fix stylua formatting because the ci didnt ran for some reason, maybe because the pr is to idk.

Matsuuu:
Yeah I'm pretty sure I forgot to run linters and such. Cheers on running em for me and merging!

jose-elias-alvarez:
Is it okay to attempt to access `action.command.arguments[1]` unconditionally? If I understand correctly, [the specification](https://microsoft.github.io/language-server-protocol/specification#command) says that `arguments` is optional, so it seems like we should guard against the case where it's `nil`, but let me know if I'm misunderstanding something.

Conni2461:
Yeah it seems like arguments is optional. You mind opening a PR to fix it?

jose-elias-alvarez:
@Conni2461 Sure, I just put in a simple fix at #1274.

Issue #697, opened 2021-03-29
Files with same names doesn't open. 


hiddenruins:
### Description
Hello, I am having this weird issue when opening files. Telescope won't switch to another file with the same name. I was playing around to see how it occurs and these are few things I know:-

- File opened first should be the nested file
- File should not be on buffers list.
- After selecting other files it works as normal. 
- If I open the root file first and try to open the nested file it works.

**Recording**
https://user-images.githubusercontent.com/5569621/112799860-a270f680-908e-11eb-82a4-fae2bd17d12d.mp4

**Expected Behavior**
Files should be opened by telescope.

**Actual Behavior**
Telescope shows the same file instead of opening the other file.

### Details

<details><summary>Reproduce</summary>

![image](https://user-images.githubusercontent.com/5569621/112798431-9a17bc00-908c-11eb-881d-e07fc9b1d6f1.png)

1. `nvim`.
2. Telescope find_files and open `folder/file.log`
3. Now try to open `file.log`.

</details>

<details><summary>Environment</summary>

- NVIM v0.5.0-dev+1186-g8665a96b9
- Build type: Release
- LuaJIT 2.1.0-beta3
- Telescope a7fa604

</details>

<details><summary>Configuration</summary>
<p>

```lua
telescope.setup{
  defaults = {
    vimgrep_arguments = {
      'rg',
      '--color=never',
      '--no-heading',
      '--with-filename',
      '--line-number',
      '--column',
      '--smart-case'
    },
    file_sorter =  sorters.get_fzy_sorter, -- better sorter than default
    file_ignore_patterns = {"node_modules", "*.zip", "*.rar"},
    generic_sorter =  sorters.get_fzy_sorter, -- better sorter than default
    shorten_path = true,
    winblend = 0,
    width = 0.75,
    preview_cutoff = 120,
    results_height = 1,
    results_width = 0.8,
    border = {},
    borderchars = { '─', '│', '─', '│', '╭', '╮', '╯', '╰'},
    color_devicons = true,
    use_less = true,
    set_env = { ['COLORTERM'] = 'truecolor' }, -- default = nil,
  }
}

```

elianiva:
removing this line fixes the issue

https://github.com/nvim-telescope/telescope.nvim/blob/a7fa60411efbfddc0674d65346a5311e33f6deb1/lua/telescope/actions/set.lua#L111

though I'm not sure why we need this line

Conni2461:
[...]

PR #741, opened 2021-04-09
Fix: open files with the same name (#697)

LarsHaalck:
Fixes a bug where files with the same name could not be opened. This PR implements the change as discussed in the related issue #697.


oberblastmeister:
Thanks! Please also implement the change outlined #760

LarsHaalck:
I just added it. If I understood #760 correctly, it has to be "or" and not "and", am I right?

oberblastmeister:
Yes it should be `or`

PR #745, opened 2021-04-09
feat: Add ability to choose window to edit

tjdevries:
TODO:
- Add utility function or something that can make this easy to pass in "editting_win" only.

When we do that, we can probably close: https://github.com/nvim-telescope/telescope.nvim/pull/531/

And then users can configure how they desire :)

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -658,6 +658,15 @@ telescope.setup({opts})                                    *telescope.setup()*
 
         Default: nil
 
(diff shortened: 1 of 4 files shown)


Conni2461:
continuation of #531 that we could consider adding.

weilbith:
Any updates for this? Can someone help?

Conni2461:
In my opinion we dont need util function. I am going to merge this and then we can see what people do with it and if we there is then any need for util functions

Issue #582, opened 2021-03-01
Enable users to setup configuration for individual pickers


tami5:
**Is your feature request related to a problem? Please describe.**

It is quite hard for new users to configure there telescopes. Configuration recipes in the wiki prove that setting up telescope to user liking is difficult and a bit complex, thus requiring a lot of investment/

**Describe the solution you'd like**

In setup method, the user shall be able to: 

- Setup configuration (mappings, layout, sorters ...) for individual pickers
- Reference functions by string rather than requiring the actions/preview module

```lua
require'telescope'.setup({
  defaults = {...};
  pickers = {
      git_status = { -- overrides git_status
        prompt_prefix = "git_status>",
        mappings = {}, 
        theme = dropdown,
      },
      dotfiles = { -- new custom picker
        func = "fd" -- that requires func to be defined
        prompt_prefix = "dotfiles>",
        cwd = 'path/to/dir',
        theme = full, -- gets merged to picker opts
        
      };
    }
  }

```
**Describe alternatives you've considered**

Currently the user need to define a custom, yet complex alternative to get individual pickers behaving or looking a certain way. For example, this is how my custom setup looks like https://gist.github.com/tami5/d49ac7f5fb6ce7d06788582446652760 

While this setup serve my needs, new commers will need to copy and paste similar setups or be limited by design.

**Additional context**
https://github.com/nvim-telescope/telescope.nvim/pull/578

elianiva:
I agree with this, I have some simple wrapper functions on my config just to pass the different opts.

tjdevries:
We could do this for the builtin ones I think -- it seems a good idea.

I'm not as confident about doing it for custom ones. We can chat about that more though.

PR #755, opened 2021-04-13
git action for creating and checking out a new branch

smithbm2316:
- Added a new action called `actions.git_create_branch` that will allow the user to create and checkout a new branch if it doesn't already exist
- Added some basic docstrings for the other git actions

I set the default binding to `<C-a>` to trigger this action when using the `git_branches` picker, but am happy to change that if people don't like it. The best mnemonic keymapping I could think of was `<C-c>` as in **checkout** or **create**, but that obviously will not work as a binding lol. I'm very much open to suggestion on the binding, my thought process was `<C-a>` could be thought of as **add new branch**, even if that doesn't really match the git command itself (since I used `git checkout -b [new-branch]`).

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -297,6 +297,31 @@ actions.insert_value = function(prompt_bufnr)
   return entry.value
 end
 
+--- Create and checkout a new git branch if it doesn't already exist
+---@param prompt_bufnr number: The prompt bufnr
+actions.git_create_branch = function(prompt_bufnr)
+  local cwd = action_state.get_current_picker(prompt_bufnr).cwd
+  local new_branch = action_state.get_current_line()
+
+  if new_branch == "" then
+    print('Please enter the name of the new branch to create')
+  else
+    actions.close(prompt_bufnr)
+    local _, ret, stderr = utils.get_os_command_output({ 'git', 'checkout', '-b', new_branch }, cwd)
+    if ret == 0 then
+      print(string.format('Switched to a new branch: "%s"', new_branch))
+    else
+      print(string.format(
+        'Error when creating new branch: %s Git returned "%s"',
+        new_branch,
+        table.concat(stderr, '  ')
+      ))
+    end
+  end
+end
+
+--- Checkout an existing git branch
+---@param prompt_bufnr number: The prompt bufnr
 actions.git_checkout = function(prompt_bufnr)
   local cwd = action_state.get_current_picker(prompt_bufnr).cwd
   local selection = action_state.get_selected_entry()
@@ -313,6 +338,8 @@ actions.git_checkout = function(prompt_bufnr)
   end
 end
 
+--- Tell git to track the currently selected remote branch in Telescope
+---@param prompt_bufnr number: The prompt bufnr
 actions.git_track_branch = function(prompt_bufnr)
   local cwd = action_state.get_current_picker(prompt_bufnr).cwd
   local selection = action_state.get_selected_entry()
@@ -329,6 +356,8 @@ actions.git_track_branch = function(prompt_bufnr)
   end
 end
 
+--- Delete the currently selected branch
+---@param prompt_bufnr number: The prompt bufnr
 actions.git_delete_branch = function(prompt_bufnr)
   local cwd = action_state.get_current_picker(prompt_bufnr).cwd
   local selection = action_state.get_selected_entry()
@@ -349,6 +378,8 @@ actions.git_delete_branch = function(prompt_bufnr)
   end
 end
 
+--- Rebase to selected git branch
+---@param prompt_bufnr number: The prompt bufnr
 actions.git_rebase_branch = function(prompt_bufnr)
   local cwd = action_state.get_current_picker(prompt_bufnr).cwd
   local selection = action_state.get_selected_entry()
@@ -369,6 +400,8 @@ actions.git_rebase_branch = function(prompt_bufnr)
   end
 end
 
+--- Stage/unstage selected file
+---@param prompt_bufnr number: The prompt bufnr
 actions.git_staging_toggle = function(prompt_bufnr)
   local cwd = action_state.get_current_picker(prompt_bufnr).cwd
   local selection = action_state.get_selected_entry()
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -306,10 +306,17 @@ actions.git_create_branch = function(prompt_bufnr)
   if new_branch == "" then
     print('Please enter the name of the new branch to create')
   else
+    local confirmation = vim.fn.input(string.format('Create new branch "%s"? [y/n]: ', new_branch))
+    if string.len(confirmation) == 0 or string.sub(string.lower(confirmation), 0, 1) ~= 'y' then
+      print(string.format('Didn\'t create branch "%s"', new_branch))
+      return
+    end
+
     actions.close(prompt_bufnr)
+
     local _, ret, stderr = utils.get_os_command_output({ 'git', 'checkout', '-b', new_branch }, cwd)
     if ret == 0 then
-      print(string.format('Switched to a new branch: "%s"', new_branch))
+      print(string.format('Switched to a new branch: %s', new_branch))
     else
       print(string.format(
         'Error when creating new branch: %s Git returned "%s"',
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -177,8 +177,8 @@ git.branches = function(opts)
       map('i', '<c-r>', actions.git_rebase_branch)
       map('n', '<c-r>', actions.git_rebase_branch)
 
-      map('i', '<c-u>', actions.git_create_branch)
-      map('n', '<c-u>', actions.git_create_branch)
+      map('i', '<c-a>', actions.git_create_branch)
+      map('n', '<c-a>', actions.git_create_branch)
 
       map('i', '<c-d>', actions.git_delete_branch)
       map('n', '<c-d>', actions.git_delete_branch)
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -177,14 +177,12 @@ git.branches = function(opts)
       map('i', '<c-r>', actions.git_rebase_branch)
       map('n', '<c-r>', actions.git_rebase_branch)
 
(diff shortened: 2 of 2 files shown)


tami5:
Although c-a is beginning of the line for me, but I hardly see any need for it in git picker. LGTM 

cc @elianiva

smithbm2316:
Ooh yes good old readline bindings, I forgot about that. I think it would be good to use something else perhaps, that's something I know a lot of people have ingrained in their fingers! Let me think about an alternate binding and check back in later with some options.

smithbm2316:
Alright I have two ideas for the mapping:
1. Map to `<C-u>` by default, since at the moment we are just disabling `<C-u>` altogether in the `git_branches` picker. I couldn't come up with any bindings honestly that had a good mnemonic for remembering the mapping easily, but if people don't like it they can always override it in their config. 
2. Just add the action and remove the default mappings, add some docs for this picker that if someone wants to use it to create branches, they can setup a custom `attach_mappings()` for it in their config. I would not prefer this as I think that this would lower the ability for people to discover this functionality, but it is a fine option too I think.

Either way, let me know what your thoughts are. I also added a confirmation prompt to make sure a user doesn't accidentally hit the mapping and create a branch by accident, similar to how the `git_delete_branch` action works.

I think I will work on some general docs next for all the default mappings for the different builtin pickers, as afaik you really can't find out about those mappings unless you dig into the source code (unless I am mistaken and just have not seen where that is easily able to be found).

smithbm2316:
cc @elianiva @tami5

tami5:
> Alright I have two ideas for the mapping:
> 
> 1. Map to `<C-u>` by default, since at the moment we are just disabling `<C-u>` altogether in the `git_branches` picker. I couldn't come up with any bindings honestly that had a good mnemonic for remembering the mapping easily, but if people don't like it they can always override it in their config. 


> 2. Just add the action and remove the default mappings, add some docs for this picker that if someone wants to use it to create branches, they can setup a custom `attach_mappings()` for it in their config. I would not prefer this as I think that this would lower the ability for people to discover this functionality, but it is a fine option too I think.

I'm kinda leaning towards the second option as it is what telescope about "customizability" however a default as you stated might be a important. I agree customizability sucks and its kinda hard. With #582 I attempt to solve this issue, hopefully by the end of the week I'll start working on it.

> Either way, let me know what your thoughts are. I also added a confirmation prompt to make sure a user doesn't accidentally hit the mapping and create a branch by accident, similar to how the `git_delete_branch` action works.

Hmm since we're adding not deleting perhaps is a bit robust to confirm, no? adding != deleting. So perhaps it would be better without confirmation.

I think you had it right from the first time. `c-a` makes sense and can be easily remembered. I highly doubt￼￼ that any one would use `c-a` to go to the beginning of the line.

> I think I will work on some general docs next for all the default mappings for the different builtin pickers, as afaik you really can't find out about those mappings unless you dig into the source code (unless I am mistaken and just have not seen where that is easily able to be found).

☺️ yah one of the current shortcomings. last time I worked on the readme I forgot to to create a table with all the available action then. It would be exterminated helpful if you managed to document those cluttered lists of actions 🤣.

lua/telescope/actions/init.lua
```diff
@@ -297,6 +297,38 @@ actions.insert_value = function(prompt_bufnr)
...
+  if new_branch == "" then
+    print('Please enter the name of the new branch to create')
+  else
+    local confirmation = vim.fn.input(string.format('Create new branch "%s"? [y/n]: ', new_branch))
+    if string.len(confirmation) == 0 or string.sub(string.lower(confirmation), 0, 1) ~= 'y' then
+      print(string.format('Didn\'t create branch "%s"', new_branch))
+      return
+    end
+
+    actions.close(prompt_bufnr)
```
elianiva, line 315: maybe close the prompt only if the branch is successfully created? I haven't try it though, will do that in a few hours.

smithbm2316:
> I'm kinda leaning towards the second option as it is what telescope about "customizability" however a default as you stated might be a important. I agree customizability sucks and its kinda hard. With #582 I attempt to solve this issue, hopefully by the end of the week I'll start working on it.

Yeah, I think that's sensible. My thought process for adding a default was that since there is _already_ a default for _deleting_ a branch, it would seem a bit odd to me as a new Telescope user to _not_ have a default mapping for _adding_ a new branch. But I am perfectly happy with leaving the keymapping up to the user, I can easily add a custom function to my own config for it.

> Hmm since we're adding not deleting perhaps is a bit robust to confirm, no? adding != deleting. So perhaps it would be better without confirmation.
> 
> I think you had it right from the first time. `c-a` makes sense and can be easily remembered. I highly doubt￼￼ that any one would use `c-a` to go to the beginning of the line.

I can certainly revert it back to `<c-a>`, as I do think that is more intuitive.  I just thought you had a valid point that `<c-a>` is a common keybinding that people use to jump to the beginning of a line when editing text, so I'd rather not make people wonder why their mental mapping for jumping to the beginning of a line no longer works, only when they are using the `git_branches` picker. I can easily see a world where a new Telescope user would be confused by that :sweat_smile:, which if possible I'd always want to avoid if I can. But there really just aren't any other available Ctrl-key-based keymappings that really make any kind of sense lol so it may be fine for this instance. Though, now that I added in the confirmation prompt for creating a new branch, at least if somebody accidentally pressed `<c-a>` mistakenly or to try and jump to the beginning of the line, they wouldn't accidentally create a branch immediately, they'd have to confirm their choice with 'y<cr>' to make that happen. So there is a safeguard there now for that potential issue.

> :relaxed: yah one of the current shortcomings. last time I worked on the readme I forgot to to create a table with all the available action then. It would be exterminated helpful if you managed to document those cluttered lists of actions :rofl:.

Well I am still pretty new to Lua, and still kinda getting my bearings inside of the Telescope codebase, so this may be a perfect place for me to help! I consider myself to be a pretty good writer, and am probably still new enough to Telescope that I can explain some of those in decent enough detail that a newbie could understand what I write! I'll look into working on that over the coming days when I have the chance. I really was only able to make this PR by just refactoring the `git_delete_branch` action, I would have had no idea how to write this from scratch :rofl:

lua/telescope/actions/init.lua
smithbm2316: Glad it works well for you! It should only close the prompt afaik if the branch is either successfully created _or_ if git throws an error. If the user tries to create a branch without giving a name or if they press the binding and then cancel the action, it will still keep the prompt open (just prints an error message in both cases). Here's a video I took of all of those different cases in action:

[asciicast](https://asciinema.org/a/rWVaRxc4gZO7miLHDjgJmBksr)

elianiva:
LGTM! We can probably merge after you add the docs for those actions that Tami mentioned, if you want.

smithbm2316:
@elianiva @tami5 I just switched the default mapping back to `<c-a>` for now. If you're happy with this, I don't have any other changes to make. However, I'm happy to remove the default binding if we don't want to give one for now, and then I can just add some info in some docs about how to create the binding or something of the sort so people know about it before we merge. 

Sidenote: if you see the asciicast video I responded to elianiva with on the review, do either of you have any idea how to fix the weird spacing that happens with the error messages I `print()`? I had no idea how to try and either clear the previous message before printing a new one, or just fix the spacing. It was driving me crazy :rofl:!

elianiva:
> do either of you have any idea how to fix the weird spacing that happens with the error messages I print()?

It only happens when we print the message as soon as Telescope gets closed. Wrapping it with `vim.defer_fn(print_stuff_here, 1)` or `vim.schedule(print_stuff_here)` works but they seem hacky to me, I'm not sure.

> Do you want me to add docs I mentioned in this PR or open a new one with that?

I think opening a new one is better so we keep this PR small and we can merge it sooner.

tami5:
Yes I agree this PR should be merged for now and later on we can collaborate more on documening available action. Thanks @smithbm2316 tag me whenever you need help with something

smithbm2316:
> It only happens when we print the message as soon as Telescope gets closed. Wrapping it with `vim.defer_fn(print_stuff_here, 1)` or `vim.schedule(print_stuff_here)` works but they seem hacky to me, I'm not sure.

Gotcha. After looking at the help docs for both of those I can see why you'd feel those are a bit hacky. I think at least wrapping the case where we have a git error with one of those might be a decent idea because of how awful the formatting is, but it's your call if you want me to try that or not. The other cases the formatting is not so bad, just a bit weird. But a git error (see the asciinema vid) really gets formatted in a hard to read way imo.

> I think opening a new one is better so we keep this PR small and we can merge it sooner.

I agree!

tami5:
> Sidenote: if you see the asciicast video I responded to elianiva with on the review, do either of you have any idea how to fix the weird spacing that happens with the error messages I print()? I had no idea how to try and either clear the previous message before printing a new one, or just fix the spacing. It was driving me crazy 🤣!

Damn I just noticed it, not sure what's the issue. 

I think this action can later be improved to close the prompt after clicking `c-a` -> create new popup in center of the screen with a message and a field to inter branch name and enter to confirm or escape to cancel. If i'm not mistaken @windwp accomplished similar thing in the pass. let me check

smithbm2316:
> Damn I just noticed it, not sure what's the issue. 
> 
> I think this action can later be improved to close the prompt after clicking `c-a` -> create new popup in center of the screen with a message and a field to input branch name and enter to confirm or escape to cancel. If i'm not mistaken @windwp accomplished similar thing in the pass. let me check 
> 
> edit: https://github.com/nvim-telescope/telescope-github.nvim/commit/255dd41f0761a7e8010e76e9122fb7c2f01e2856

I definitely like this idea. That implementation may be a bit over my head atm with my current Lua knowledge, but I'm sure with some more practice and tinkering I could make something like that happen. I'll keep that on my radar for a future improvement to this.

Conni2461:
uhhhhh @smithbm2316 is writing docs :) There is this: https://github.com/nvim-telescope/telescope.nvim/pull/587 also max result query issue still exist but vigoux was looking into that.

Also why did docgen not run for you? Do you have gh actions disabled for that fork?

smithbm2316:
> uhhhhh @smithbm2316 is writing docs :) There is this: https://github.com/nvim-telescope/telescope.nvim/pull/587 also max result query issue still exist but vigoux was looking into that.
> 
> Also why did docgen not run for you? Do you have gh actions disabled for that fork?

I was wondering why this didn't work. I'll take a look into this later and see if I can figure out if it's an issue in my fork.

PR #759, opened 2021-04-15
Git create branch action

smithbm2316:


smithbm2316:
I went and enabled the github actions in my fork, seems like they were not enabled before. This should now include the docs for the PR we just merged @tami5 @elianiva in #755. @Conni2461 thanks for letting me know how to fix the docgen. Sorry for all the git commit history mess, I actually didn't realize I was opening this PR on the actual telescope repo, thought I was opening it on my fork :sweat_smile: my bad about that.

doc/telescope.txt
```diff
@@ -116,22 +116,6 @@ builtin.live_grep()                                      *builtin.live_grep()*
 
 Actions functions that are useful for people creating their own mappings.
 
-actions.move_selection_next({prompt_bufnr})    *actions.move_selection_next()*
```
elianiva, line 119: hmm, I wonder why this part got deleted as well, was about to merge but I noticed this. I don't want to _accidentally_ delete the existing docs :laughing:

Conni2461: This is the max result for a query bug, that is in treesitter core. See https://github.com/nvim-treesitter/nvim-treesitter/issues/963 for all more information. Its super annoying.

Don't know how to proceed right now here.

smithbm2316: Ooh good catch. I didn't even notice that. @Conni2461 is there any way to *manually* add to the docs for the time being? I'm not sure if that's been tried yet with other PRs, or if it'll still break the docgen action.

smithbm2316:
@elianiva @Conni2461 after one unsuccessful try (forgot to turn off github actions on my fork first), I have now manually put back the docs that were deleted. Not sure if this is a real solution for now, but #773 looks like it was also deleting the same docs that my PR here deleted because of that max-query bug.

Conni2461:
No please no manual changing in the help file. FYI you could do `skip-checks: true` as commit body (not message, so second line) to skip CI i think.

Are we handcuffing us here, with that dilemma?! Because if yes we might want to just remove the last 2 function headers and you add them back if the bug is fixed. Because i kind wanna finish "prompt cycle", can't remember PR number :laughing:, and as long as this issue exist it doesn't make any sense for me to finish it. 

(Bug is file local, so you still write documentation for a different file :smile:)

smithbm2316:
@elianiva @Conni2461 ok I think I was able to do as you suggested now! Sorry about the git commit history mess. I just reverted to the commit that added this feature in with my previous PR, and then removed 3 of the function headers I added in this PR originally causing the bug. Seems like it didn't cause that issue this time! Lmk if there's anything else I need to do to fix that bug we've been having. The last thing I want to do is hold up other PRs lol

Conni2461:
this diff looks completely wrong. Isn't the feature already merged? Shouldn't here only be the removed headers + the generated docs?

smithbm2316:
> this diff looks completely wrong. Isn't the feature already merged? Shouldn't here only be the removed headers + the generated docs?

It is already merged, the diff kindof confused me too. I thought I reverted this back to the right commit, but maybe I didn't pick the right one. @Conni2461 I think I'll close this PR and open a new one and just start over with what I had before I think. This is turning into a bit of a mess on my part, sorry. I'm familiar with the basics of git, but still learning how to navigate a big repository like this, revert to previous commits, rebase properly, etc. I may have just made a mistake in what I reverted to.

PR #767, opened 2021-04-17
Add use_regex option to grep_string

khalid151:
Add `use_regex` option to disable `escape_chars` function so that regex would work when using `grep_string`.
It's disabled by default to preserve old behavior.


Conni2461:
Sure we can do this :) Thanks

PR #768, opened 2021-04-17
fix: live_grep prepending cwd to paths that are already absolute

levouh:
I noticed that as part of #666 (coincidence???) that sometimes we end up with entries that include the current working directory concatenated with an already absolute path. This would break passing `search_dirs` to `live_grep` as of that commit due to the change in how we were calling `flatten`, which would now mean the search directories weren't a table of their own, but included in the returned arguments. Regardless, this broke something downstream of that picker where the entries would have a path like:

```
/home/amasquelier/.local/share/nvim/site/pack/packer/start/telescope.nvim/lua/telescope/state.lua
```

and we'd prepend them with the `cwd`, resulting in:

```
/home/amasquelier/.local/share/nvim/site/pack/packer/start/telescope.nvim/lua/telescope//home/amasquelier/.local/share/nvim/site/pack/packer/start/telescope.nvim/lua/telescope/state.lua
```

which breaks a number of things.

This PR fixes three things:

1. The case mentioned above, where if a path is already absolute, we do not try to prepend the current working directory to it.
2. The previewer for `live_grep` when passing `grep_open_files` as an option now works as well.
3. Make `grep_open_files` and `search_dirs` mutually exclusive, as we are either doing one of the other. It doesn't make a ton of sense to do both.

I'm not sure how kosher it is to use `plenary.path`, as it is already a dependency, but there seems to be a general trend of using `telescope.path` instead. Let me know if this is an issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,10 +376,10 @@ Built-in functions. Ready to be bound to any key you like. :smile:
 
 #### Options for builtin.live_grep
 
-| Keys                   | Description                                           | Options      |
-|------------------------|-------------------------------------------------------|--------------|
-| `grep_open_files`      | Restrict live_grep to currently open files.           | boolean      |
-| `search_dirs`          | List of directories to search in.                     | list         |
+| Keys                   | Description                                                                        | Options      |
+|------------------------|------------------------------------------------------------------------------------|--------------|
+| `grep_open_files`      | Restrict live_grep to currently open files, mutually exclusive with `search_dirs`  | boolean      |
+| `search_dirs`          | List of directories to search in, mutually exclusive with `grep_open_files`        | list         |
 
 
 ### Vim Pickers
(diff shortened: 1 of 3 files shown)


tami5:
> how kosher it is

🤩 I see you cousin 🤣, It's strictly kosher, your good. In all seriousness, plenary is the std of many of plug-ins. And for `telescope.path` I believe is just temporary or specific for functions to specific to telescope .

I haven't tested out yet but from reading the code it looks good to me. 

Is it ready for merge? or there are additional things you want to further test out?

levouh:
I've tested it with the `search_dirs` and `grep_open_files` arguments to `live_grep`, as well as without them, etc.  so it is good to go as far as I'm aware. If you'd like me to write explicit tests I'm happy to do that as well.

Conni2461:
`plenary.path` is good, `telescope.path` was just a temporary solution, we needed to make `plenary.path` was not really mature enough back then :laughing: 
There is a PR that removes `telescope.path` and replaces it with `plenary.path` #473 Maybe @tami5 or you can help finish it :)

PR #769, opened 2021-04-18
feat: add search history picker

VVKot:
Adding a history picker analogous to fzf.vim's `History/`. There is no Lua api like `vim.search` as far as I can tell, so I just send the keys a la `/foo<CR>` when selecting a row. Is that the best/acceptable way to do that? I also have not found any tests for command history so haven't wrote any of my own, I am missing something here?

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -234,6 +234,19 @@ actions.set_command_line = function(prompt_bufnr)
   vim.cmd(entry.value)
 end
 
+actions.edit_search_line = function(prompt_bufnr)
+  local entry = action_state.get_selected_entry(prompt_bufnr)
+  actions.close(prompt_bufnr)
+  a.nvim_feedkeys(a.nvim_replace_termcodes("/" .. entry.value , true, false, true), "t", true)
+end
+
+actions.set_search_line = function(prompt_bufnr)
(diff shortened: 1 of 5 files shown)


tami5:
Interesting. Let me give it a quick test

lua/telescope/actions/init.lua
```diff
@@ -234,6 +234,19 @@ actions.set_command_line = function(prompt_bufnr)
   vim.cmd(entry.value)
 end
 
+actions.edit_search_line = function(prompt_bufnr)
```
tami5, line 237: Inline documentation

lua/telescope/builtin/internal.lua
```diff
@@ -297,6 +297,36 @@ internal.command_history = function(opts)
   }):find()
 end
 
+internal.search_history = function(opts)
```
tami5, line 300: There are some work going on to add inline function documentation. It would be awesome if you can add an inline documentation for the search picker in internal.lua and for the new action added.

tami5 (approved):
@VVKot Works really good, and it's very useful picker indeed.  other than the comments above I believe it should be merged. Thanks

lua/telescope/builtin/internal.lua
VVKot: @tami5 I am happy to do so. Do you have an example of such docs somewhere already?

PR #771, opened 2021-04-19
feat: add new layout strategy

tjdevries:
image

image


vladsolokha:
Seems like the height option is not working, meaning any height I pick for the bottom_pane is not affecting its actual height. 
Screenshot 2024-04-18 at 9 41 49 PM

vladsolokha:
I did find a work around where I have to specify in each key mapping that I'm using "get_ivy" and in options a "layout_config = { height = 10 }" or some other value like that, to make the bottom pane smaller and less intrusive. 
Screenshot 2024-04-19 at 8 41 36 AM

PR #773, opened 2021-04-19
feat: Add much better documentation for layout_strategies

tjdevries:
Still need to clean up a few other bits in the thing here, but once we do that, it will all make sense and items can be placed in the right spot.

This should really clear up a lot of confusion.

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -116,22 +116,6 @@ builtin.live_grep()                                      *builtin.live_grep()*
 
 Actions functions that are useful for people creating their own mappings.
 
-actions.move_selection_next({prompt_bufnr})    *actions.move_selection_next()*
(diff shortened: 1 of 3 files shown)


doc/telescope.txt
```diff
@@ -116,22 +116,6 @@ builtin.live_grep()                                      *builtin.live_grep()*
 
 Actions functions that are useful for people creating their own mappings.
 
-actions.move_selection_next({prompt_bufnr})    *actions.move_selection_next()*
```
smithbm2316, line 119: I had this same issue in #759 @tjdevries, because of the max query bug with treesitter this section of docs is getting deleted by the docgen

Conni2461:
Closing l-kershaw and tj just done this in #922 thanks :)

PR #781, opened 2021-04-21
[RFC] attempt to fix fuzzy_with_index_bias sorter

levouh:
I use this sorter to do things like bias terminal history towards those that are newer when using _Telescope_ to query it, etc. as you'd expect. I noticed that this finder isn't broken, but throws a handful of warnings due to the `:sort` not being called with the new `cb_add`, etc. functions.

If we use the locally-defined sorter, we get double the results back so I just ended up calling `:sorting_function` directly.

Not sure if this is the best way to do this, but looking for feedback.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -862,14 +862,15 @@ actions.which_key = function(prompt_bufnr, opts)
   opts.mode_width = utils.get_default(opts.mode_width, 1)
   opts.keybind_width = utils.get_default(opts.keybind_width, 7)
   opts.name_width = utils.get_default(opts.name_width, 30)
-  opts.column_padding = utils.get_default(opts.column_padding, "  ")
-  opts.column_indent = table.concat(utils.repeated_table(utils.get_default(opts.column_indent, 4), " "))
   opts.line_padding = utils.get_default(opts.line_padding, 1)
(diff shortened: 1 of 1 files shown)


Conni2461:
Are you sure this is the correct PR? :laughing: 

I think you force pushed over the previous PR. Always make sure to open a new branch for each PR ;)

I think i looked at this in the past and it looked good. Do you have a way to get back the previous diff? I think its this commit https://github.com/nvim-telescope/telescope.nvim/commit/bc410b87397612069fd89ae96bb6c4f9565de2ab

levouh:
No branch only `main`! But you are absolutely correct; I forgot this _PR_ even existed and overwrote my local changes/force pushed to my fork.

I'll have to see if this is somewhere in the `reflog`, thanks for the heads up @Conni2461

Conni2461:
The commit is the commit i linked. Here again: https://github.com/nvim-telescope/telescope.nvim/commit/bc410b87397612069fd89ae96bb6c4f9565de2ab

But i am closing this, please open a new PR :) This situation is confusing. You should adept 1 PR == 1 branch

levouh:
>This situation is confusing. You should adept 1 PR == 1 branch

Yup, apologies. I had this on `master` of my fork (which I normally never do) and force-pushed over it when updating with the upstream here. I'll open up another one.

PR #785, opened 2021-04-21
extensions - search current buffer with regex

JoseConseco:
It will let you search current buffer with regex support.

https://user-images.githubusercontent.com/13521338/115562527-dcf93800-a2b6-11eb-8385-27258417a9f7.mp4

User is expected to escape with \  (then in lua script \ is converted to % so that string.find works). I'm new to this so there may be bugs.
Rather than writing this extension, I would rather use:
 - live_grep on current file, but I did not find way to make it work.
 - or use current_buffer_fuzzy_find but without fuzzy, but also I also did not found way to make it work...
 
 In the end me think above extension could be replaced in future with new sorter: non_fzf_regex  or something  like that.  Plus new easy finder:  grep(%)  +  grep(cwd_files)

### DIFF
--- a/lua/telescope/_extensions/buffer_search.lua
+++ b/lua/telescope/_extensions/buffer_search.lua
@@ -47,7 +47,7 @@ local get_str_matcher = function()
   }
 end
 -- copy pasted from tele/lua/builtin/files.lua
-local current_buffer_fuzzy_find = function(opts)
+local current_buffer_regex_find = function(opts)
   -- All actions are on the current buffer
   local bufnr = vim.api.nvim_get_current_buf()
   local filename = vim.fn.expand(vim.api.nvim_buf_get_name(bufnr))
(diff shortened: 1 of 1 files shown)


Conni2461:
Why do you want to merge a extension?
The only new thing here seems to be the sorter. the other code is just `current_buffer_fuzzy_find` isn't it?

tami5:
It would be better to add it as optional feature to current fuzzy finder and enable regex when that option is passed

JoseConseco:
It is based on live_grep (if I  remember correctly) but  it is using regex and it is using only current file.  Preferably instead of merging this extension, I would rather have functionality in Telescope where userI can use any type of search: 
- fuzzy,  regex, string_match  etc   
and  it could work on any target:  current file, opened buffers,  or cwd files etc. But there is no way to achieve these combinations it seems right now (I'm now to nvim and tele so I may be missing something) . I know its is simple script so feel free to close it if it does not match telescope vision.

Conni2461:
Its definitely not based on `live_grep` its based on `current_buffer_fuzzy_find` and the only new thing here is the sorter.

I am closing this here because we are not merging `extensions` into core. `extensions` are meant for external repos. Also there is an attempt to make live_grep work with regex see https://github.com/nvim-telescope/telescope.nvim/pull/670

PR #786, opened 2021-04-21
feat: file browser navigate to parent directory

al3xfischer:
File Browser: Instead of adding the parent directory as `../` we could add it as `..\\`.
This tiny change would allow us to navigate to the parent directory.
Otherwise, it would open the parent direcotry in netrw.

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -247,7 +247,7 @@ files.file_browser = function(opts)
         table.insert(data, typ == 'directory' and (entry .. os_sep) or entry)
       end
     })
-    table.insert(data, 1, '..\\')
+    table.insert(data, 1, '..' .. os_sep)
 
     return finders.new_table {
       results = data,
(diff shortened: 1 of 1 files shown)


elianiva:
isn't it the current behaviour already?

https://user-images.githubusercontent.com/51877647/116062592-3e057f00-a6ae-11eb-8a02-688ba42d4300.mp4

lua/telescope/builtin/files.lua
```diff
@@ -247,7 +247,7 @@ files.file_browser = function(opts)
         table.insert(data, typ == 'directory' and (entry .. os_sep) or entry)
       end
     })
-    table.insert(data, 1, '../')
+    table.insert(data, 1, '..\\')
```
Conni2461, line 250: this should be `'..' .. os_sep` so it works for both Linux and Windows

al3xfischer: I changed it to your suggested solution. (Thx for the hint.)

PR #787, opened 2021-04-21
Git actions docs fix

smithbm2316:
This fixes the max-query treesitter issue in the actions documentation that was caused by #755's docs additions, and is replacing #759 as that PR got messy quickly. @Conni2461 I think the diff looks correct now.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -345,6 +345,8 @@ actions.git_checkout = function(prompt_bufnr)
   end
 end
 
+--- Tell git to track the currently selected remote branch in Telescope
+---@param prompt_bufnr number: The prompt bufnr
 actions.git_track_branch = function(prompt_bufnr)
   local cwd = action_state.get_current_picker(prompt_bufnr).cwd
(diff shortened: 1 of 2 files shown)


smithbm2316:
It looks like some changes were merged recently into master that ended up deleting the docs that my original PR with the feature I created, so I had to fix a merge conflict here with those new updates. That's why the diff now shows that we are adding *back* the `actions.move_selection_next` and `actions.move_selection_previous` docs, whatever recent PR was merged into master seems to have deleted them by accident. I think the diff still should be ok, but that's why it looks weird as far as I can tell.

lua/telescope/actions/init.lua
```diff
@@ -345,8 +345,6 @@ actions.git_checkout = function(prompt_bufnr)
   end
 end
 
---- Tell git to track the currently selected remote branch in Telescope
```
tjdevries, line 348: can you just change these to be `-- comment here` rather than `---` with a todo that says we can add these back in when the upstream issue is fixed?

smithbm2316: Yup! Happy to do that!

PR #797, opened 2021-04-25
prevent fouc on the first preview loading

mnowotnik:
There's a slight lag on the first preview loading (during preview buffer creation).
It is not visible the next time a user chooses a file for preview because a scratch buffer
for the file already exists. This lag *and* setting preview window to display the newly
created buffer before its fully initialized causes a brief flash of a blank terminal background.

This change delays setting preview window to display the new preview buffer and consequently
eliminates the flash. It should improve user experience since flickering can be distracting.

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -184,8 +184,10 @@ previewers.new_buffer_previewer = function(opts)
     else
       local bufnr = vim.api.nvim_create_buf(false, true)
       set_bufnr(self, bufnr)
-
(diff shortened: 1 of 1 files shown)


Conni2461:
I can't reproduce the issue you are describing. So i am not sure what this tries to achieve

mnowotnik:
> I can't reproduce the issue you are describing. So i am not sure what this tries to achieve

Prevent flash of blank terminal background on initial preview buffer loading. If you can't reproduce then maybe your computer calculates the preview too fast for a human to notice.

Conni2461:
Okay i needed to read your post again. I see. We should do this. Can you fix the linting after that i would merge.

Thanks :)

PR #798, opened 2021-04-25
create new action: git switch

kyoh86:
If the branch already exists in local, switch to that.
If the branch is only in remote, create new branch tracking remote and switch to new one.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -358,6 +358,31 @@ actions.git_checkout = function(prompt_bufnr)
   end
 end
 
+--- Switch to git branch
+--- If the branch already exists in local, switch to that.
+--- If the branch is only in remote, create new branch tracking remote and switch to new one.
+---@param prompt_bufnr number: The prompt bufnr
+actions.git_switch = function(prompt_bufnr)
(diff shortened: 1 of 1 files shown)


lua/telescope/actions/init.lua
```diff
@@ -358,6 +358,31 @@ actions.git_checkout = function(prompt_bufnr)
   end
 end
 
+--- Switch to git branch
+--- If the branch already exists in local, switch to that.
+--- If the branch is only in remote, create new branch tracking remote and switch to new one.
+---@param prompt_bufnr number: The prompt bufnr
```
Conni2461, line 364: Make sure github actions for your fork are enabled so docgen is running. We are also running into a maximum amount of docs we can generate for that file (some bug in treesitter) so we are currently documenting new functions like this: 
```lua
-- TODO: add this function header back once the treesitter max-query bug is resolved
--
--
```
With two rather than 3 `-`. See 379

Conni2461:
Oh sorry i think you misunderstood me or i just explained it bad :rofl: 

Keep the documentation in just reduce it too two `-` rather than `---` (because two will not be picked up by the generator). That way we can just add back in the third `-` and remove the todo and the documentation will be generated :)

Issue #795, opened 2021-04-23
errors when using current buffer fuzzy find


Mechap:
I'm getting errors when I want to  open the current buffer fuzzy find.
![image](https://user-images.githubusercontent.com/54742369/115934967-7af52a00-a492-11eb-840d-01e1c4e49c9d.png)

EthanJWright:
Same error here, I'm on Ubuntu 20.04, 0.5 Neovim, and using Kitty.

thenewvu:
Not the same error, but I got another error when opening the current buffer fuzzy find:

![Screen Shot 2021-04-24 at 13 23 38](https://user-images.githubusercontent.com/6092585/115949623-8f700b80-a500-11eb-9fec-a75d5f8f0cc9.png)

I believe this commit causes the error: https://github.com/nvim-telescope/telescope.nvim/commit/0b2c801978e9cfc9b6351fc9f037ff2ba93805bd

dhruvmanila:
I got the same error as @thenewvu got. 
```
E5108: Error executing lua ...ker/start/telescope.nvim/lua/telescope/builtin/files.lua:387: attempt to index local 'query' (a nil value)
```

It's because telescope uses `filetype` to get the treesitter query and the function expects the language name. In my case I was opening the current buffer fuzzy find for a bash file and the filetype for that is `sh` but the language name in treesitter is `bash` so we get a `nil` value and thus the error.

marcelbeumer:
Can confirm what @dhruvmanila describes. Have the same issue with a `typescriptreact` file

Conni2461:
There is a `ft_to_lang` in `nvim-treesitter.parsers`. Someone can add that (don't forget the pcall).

But i am not sure if people report different issues here. would be good if op could have filled out the bug template (next time)

Mechap:
I updated neovim and everything works so far now

Conni2461:
Does anyone here still have issues? Otherwise i will close this issue.

EthanJWright:
I'm still getting the error:

```
Error executing vim.schedule lua callback: ...gged/telescope.nvim/lua/telescope/pickers/highlights.lua:32: Expected lua string                                                                                                                                                                                 
Error executing vim.schedule lua callback: ...gged/telescope.nvim/lua/telescope/pickers/highlights.lua:32: Expected lua string
```

When calling `:Telescope current_buffer_fuzzy_find` on a python file -- I can open up a bug ticket with the proper template if you like.

[...]

PR #801, opened 2021-04-27
fix: use treesitter language name instead of ft if available

dhruvmanila:
This will fix the problem where the filetype is different than the treesitter lang name. Eg., filetype -> "sh", language name -> "bash"

Ref:
- https://github.com/nvim-telescope/telescope.nvim/issues/795#issuecomment-826043348
- https://github.com/nvim-telescope/telescope.nvim/issues/795#issuecomment-826086757

Fix suggestion by @Conni2461: https://github.com/nvim-telescope/telescope.nvim/issues/795#issuecomment-826769060

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -368,6 +368,11 @@ files.current_buffer_fuzzy_find = function(opts)
     })
   end
 
+  local ts_ok, ts_parsers = pcall(require, 'nvim-treesitter.parsers')
+  if ts_ok then
+    filetype = ts_parsers.ft_to_lang(filetype)
+  end
+
   local ok, parser = pcall(vim.treesitter.get_parser, bufnr, filetype)
   if ok then
     local query = vim.treesitter.get_query(filetype, "highlights")
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -373,9 +373,9 @@ files.current_buffer_fuzzy_find = function(opts)
     filetype = ts_parsers.ft_to_lang(filetype)
   end
 
-  local ok, parser = pcall(vim.treesitter.get_parser, bufnr, filetype)
+  local parser_ok, parser = pcall(vim.treesitter.get_parser, bufnr, filetype)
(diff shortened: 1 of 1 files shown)


lua/telescope/builtin/files.lua
```diff
@@ -368,6 +368,11 @@ files.current_buffer_fuzzy_find = function(opts)
...
   end
 
+  local ts_ok, ts_parsers = pcall(require, 'nvim-treesitter.parsers')
+  if ts_ok then
+    filetype = ts_parsers.ft_to_lang(filetype)
+  end
+
   local ok, parser = pcall(vim.treesitter.get_parser, bufnr, filetype)
   if ok then
     local query = vim.treesitter.get_query(filetype, "highlights")
```
tjdevries, line 378: I suppose the other thing is that we should pcall here as well -- I didn't realize this would fail if it didn't find anything.

If we do `local query_ok, query = pcall(...)` here, then that would be double safe.

If you add that in I'll merge.

Thanks for the PR :)

dhruvmanila: Yeah, that sounds good. Got it done :)

lua/telescope/builtin/files.lua
```diff
@@ -368,10 +368,14 @@ files.current_buffer_fuzzy_find = function(opts)
...
 
-  local ok, parser = pcall(vim.treesitter.get_parser, bufnr, filetype)
-  if ok then
-    local query = vim.treesitter.get_query(filetype, "highlights")
+  local ts_ok, ts_parsers = pcall(require, 'nvim-treesitter.parsers')
+  if ts_ok then
+    filetype = ts_parsers.ft_to_lang(filetype)
+  end
 
+  local ok, parser = pcall(vim.treesitter.get_parser, bufnr, filetype)
```
tjdevries, line 376: Sorry, this is small nit, but can you change this to parser_ok? After reading it now, this ok looks lonely 😂

dhruvmanila: Haha, no problem. I get it ;)

Issue #689, opened 2021-03-26
Feature: Possibility to remove buffers in the buffer list


timjackleus:
**Is your feature request related to a problem? Please describe.**
When I've been moving around in a project for long enough my buffer list is filling up and I find it very convenient to simply remove the files I'm no longer working in. Before Telescope I used Denite where this feature is available.

**Describe the solution you'd like**
When browsing the buffer list `:Telescope buffers`, I would like to remove the buffer that is focused by a short command like `<c-d>`.

**Describe alternatives you've considered**
Please let me know if you already have alternative / preferred ways of achieving this.

**Additional context**
Nope, but I will take the opportunity to say this plugin truly is amazing and beautiful 🤩

elianiva:
I think we have a couple of these issues, I'm in the middle of something at the moment (internship :laughing:) so I won't be able to implement it. But feel free to make a PR :) 

see these issues:
https://github.com/nvim-telescope/telescope.nvim/issues/621 (Conni has written a solution for this)
https://github.com/nvim-telescope/telescope.nvim/issues/672
https://github.com/nvim-telescope/telescope.nvim/issues/655 (this has more ideas than just closing a buffer)

Conni2461:
I am busy till next Wednesday after that i can finish it. I also kinda want a remove selected entry (that's the tricky part). But i would also accept a PR :) just don't expect a review before Wednesday :laughing:

timjackleus:
Thank you for info, I'm completely new to lua but I'll have a check on the related issues and see if I can figure something out :)

Conni2461:
I am sorry but i wont do this. Someone else should pick it up :)

dhruvmanila:
I can complete this as I already implemented it in my own config. A question: Should this be included in the buffers picker or should it be implemented as an action with a default mapping?

I will open the PR with the implementation and we can discuss there :)

PR #803, opened 2021-04-28
feat: Add delete_buffer action and mappings

dhruvmanila:
This can delete the currently selected entry as well as the selections done using multi-select.

Question: Should this be included in the buffers picker or should it be implemented as an action with a default mapping? Currently, I have implemented in the picker itself but that would hard code the mappings for the action. Although this action is only useful in the buffers picker so it might make sense :)

fixes: #689 

Other mappings which sounds good for this action: `<C-x>` (taken by `select_horizontal`), `<C-d>` (taken by `preview_scrolling_down`)

Maybe provide an option (`delete_buf_mapping`) for the user to choose the mappings? That way they can override other mappings which they don't use.

### DIFF
--- a/lua/telescope/mappings.lua
+++ b/lua/telescope/mappings.lua
@@ -28,9 +28,7 @@ mappings.default_mappings = config.values.default_mappings or {
       ["<S-Tab>"] = actions.toggle_selection + actions.move_selection_better,
       ["<C-q>"] = actions.send_to_qflist + actions.open_qflist,
       ["<M-q>"] = actions.send_selected_to_qflist + actions.open_qflist,
-      ["<C-l>"] = actions.complete_tag,
-
-      ["<M-w>"] = actions.delete_buffer,
+      ["<C-l>"] = actions.complete_tag
     },
 
     n = {
@@ -57,8 +55,6 @@ mappings.default_mappings = config.values.default_mappings or {
 
       ["<C-u>"] = actions.preview_scrolling_up,
       ["<C-d>"] = actions.preview_scrolling_down,
-
-      ["<M-w>"] = actions.delete_buffer,
     },
   }
 
--- a/lua/telescope/mappings.lua
+++ b/lua/telescope/mappings.lua
@@ -28,7 +28,9 @@ mappings.default_mappings = config.values.default_mappings or {
       ["<S-Tab>"] = actions.toggle_selection + actions.move_selection_better,
       ["<C-q>"] = actions.send_to_qflist + actions.open_qflist,
       ["<M-q>"] = actions.send_selected_to_qflist + actions.open_qflist,
-      ["<C-l>"] = actions.complete_tag
+      ["<C-l>"] = actions.complete_tag,
+
+      ["<M-w>"] = actions.delete_buffer,
     },
 
     n = {
@@ -55,6 +57,8 @@ mappings.default_mappings = config.values.default_mappings or {
 
       ["<C-u>"] = actions.preview_scrolling_up,
       ["<C-d>"] = actions.preview_scrolling_down,
+
+      ["<M-w>"] = actions.delete_buffer,
     },
   }
 
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -560,13 +560,12 @@ end
 actions.delete_buffer = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   local multi_selection = current_picker:get_multi_selection()
+  actions.close(prompt_bufnr)
 
   if vim.tbl_isempty(multi_selection) then
     local selection = action_state.get_selected_entry()
-    actions.close(prompt_bufnr)
     vim.api.nvim_buf_delete(selection.bufnr, {force = true})
   else
-    actions.close(prompt_bufnr)
     for _, selection in ipairs(multi_selection) do
       vim.api.nvim_buf_delete(selection.bufnr, {force = true})
     end
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -554,8 +554,9 @@ actions.open_qflist = function(_)
   vim.cmd [[copen]]
 end
 
---- Delete the selected buffer or all the buffers selected using multi selection.
----@param prompt_bufnr number: The prompt bufnr
+-- TODO: add this function header back once the treesitter max-query bug is resolved
+-- Delete the selected buffer or all the buffers selected using multi selection.
+-- @param prompt_bufnr number: The prompt bufnr
 actions.delete_buffer = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   local multi_selection = current_picker:get_multi_selection()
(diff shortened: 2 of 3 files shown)


tami5:
mappings are tricky business. I think it should me implemented as action for future reuse and decrease the complexity of the main function.

So when the user select multiple buffers it can be deleted with the same mapping that of a single one?

For the mapping, maybe M-w or C-S-w makes sense, since c-w delete the pervious word.

dhruvmanila:
> mappings are tricky business. I think it should me implemented as action for future reuse and decrease the complexity of the main function.

Got it. 

> So when the user select multiple buffers it can be deleted with the same mapping that of a single one?

Yes, I think this type of behavior would be preferred than having two different mappings for similar cases:
1. Delete the currently selected buffer
2. Delete all the buffers selected using multi-select

dhruvmanila:
I have made the necessary changes. Also, if possible, can you briefly explain how would I go about generating the docs? I have installed the `tree-sitter-lua` plugin.

lua/telescope/mappings.lua
```diff
@@ -28,7 +28,9 @@ mappings.default_mappings = config.values.default_mappings or {
       ["<S-Tab>"] = actions.toggle_selection + actions.move_selection_better,
       ["<C-q>"] = actions.send_to_qflist + actions.open_qflist,
       ["<M-q>"] = actions.send_selected_to_qflist + actions.open_qflist,
-      ["<C-l>"] = actions.complete_tag
+      ["<C-l>"] = actions.complete_tag,
+
+      ["<M-w>"] = actions.delete_buffer,
```
Conni2461, line 33: Why is this mapping global? You should do a `map(mode, key, actions.delete_buffer)` in `attach_mappings`.

doc/telescope.txt
```diff
@@ -257,6 +257,15 @@ actions.open_qflist()                                  *actions.open_qflist()*
 
 
 
+actions.delete_buffer({prompt_bufnr})                *actions.delete_buffer()*
```
Conni2461, line 260: Is there a chance your fork has workflows disabled? Because the CI should handle the generation. We are also hitting a max query result issue for this file (which would remove another documentation for a function see https://github.com/tjdevries/tree-sitter-lua/issues/5 and all linked threads).

So other actions only use `--` spaces (that documentation will not be exported) (example `actions.git_staging_toggle`) to avoid this issue

lua/telescope/mappings.lua
dhruvmanila: Do you mean something like this: a38e3ebfe04ffde68bdf9328ff9c229ee667db56? 

> mappings are tricky business. I think it should me implemented as action for future reuse and decrease the complexity of the main function.

That was my original plan but @tami5 suggested to make it global or are you suggesting something else? I am sorry if I am misunderstanding :)

doc/telescope.txt
dhruvmanila: I do have workflows enabled but it did not kick in. It might be something related to GitHub's new feature where a user with privileged rights will have to approve the workflow runs for a first-time contributor: https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/

> So other actions only use -- spaces (that documentation will not be exported) (example actions.git_staging_toggle) to avoid this issue

Got it, I will update the function doc but do we want this action as global or local to just `Telescope  buffers` as done in my first commit?

lua/telescope/mappings.lua
dhruvmanila: Oh, I am sorry, I got it now what you mean. You mean to keep the actions global but mappings local.

tami5: Global in the sense that it applies to all the pickers. mappings should applied to the picker it self. I guess you can refer git pickers as they have local mappings

doc/telescope.txt
dhruvmanila: Nevermind, the workflows were enabled in the repo but I had to give my explicit permission. Sorry for all the trouble :)

lua/telescope/mappings.lua
dhruvmanila: Got it, thanks for clearing it up!

dhruvmanila:
Any updates here? Should there be any more changes which needs to be made? @Conni2461

Conni2461 (approved):
LGTM I am not sure about the mapping thought. Also i would have preferred to delete a buffer and remove that entry in the result list but this is fine for now.

I can merge it if you want

dhruvmanila:
Yes, keeping telescope open would be even better. I have seen arecibo doing similar to that, although that is completely changing the finder which does not seem to be possible in this case.

I looked through the picker code and thinking about doing something similar to this:
1. Get the selection row
2. Either shift the selection up or down based on where we are
3. Remove the selected entry from the finders results table
4. Remove the line from the results buffer
5. Delete the buffer

After the fourth step, the entry is removed from the results buffer as well as the results table. The problem is that I need to update the results buffer as when I move up and down, the lines are restored, but when I type a character then the buffers are updated with the information about the removed entry.

Here's the code I tried:
```lua
  local row = current_picker:get_selection_row()
  local index = current_picker:get_index(row)

  if index == #current_picker.finder.results then
    actions.move_selection_previous(prompt_bufnr)
  else
    actions.move_selection_next(prompt_bufnr)
  end

  table.remove(current_picker.finder.results, index)
  vim.api.nvim_buf_set_lines(current_picker.results_bufnr
end
```
It seems that there is some other object holding on to this information, is it the entry manager? If it is, there does not seem to be any method of removing an entry.

In the above recording, I am deleting the entries but they're not actually being deleted from the buffer, but when I refresh it by typing in the prompt, the results take effect.

https://user-images.githubusercontent.com/67177269/117549746-b62e5600-b059-11eb-8cf1-cbd8786df011.mov

Conni2461:
If you wanna do it that would be awesome. I can help you :) (A interface in picker for this would be awesome). 
Like `remove_current_selected`.

I think you need to `table.remove` the entry and trigger: `current_picker.__on_lines(nil, nil, nil, 0, 1)`

Also i think index is wrong if `sorting_strategy` is `descending` because then the index for the highest matching element is 50 (for example)

dhruvmanila:
That sounds awesome! Yes, a second eye would be amazing. Should we continue with that in this PR or open another one for that?

So, using your suggestion `current_picker.__on_lines(nil, nil, nil, 0, 1)` gives the perfect results although it resets the cursor to the top of the result. And also, by doing that we do not need to shift the cursor up or down. It also gives a weird effect about removing the first two letters of the entry from all the pickers, but this does not happen when we are removing the `default_selection_index` entry.

Thanks for noticing the index part. I will fix that at the end, although I doubt we need to worry about that now.

Idea: Should we provide a callback parameter which will be called with the removed entry, so that the user or the specific picker can do exactly what needs to be done outside of telescope? Eg., delete the buffer, delete the file, etc. 

Also, the status (`14/20`) on the prompt is not being updated for the `find_files` picker. I suspect it is because of the one shot job finder, what do you think?

https://user-images.githubusercontent.com/67177269/117560139-cb7ea100-b0a8-11eb-86e2-331c1a311cec.mov

https://user-images.githubusercontent.com/67177269/117560144-d3d6dc00-b0a8-11eb-99fd-888c49300c35.mov

dhruvmanila:
Closed in favor of #828 

The action will be implemented in the above PR using the new interface.

Issue #684, opened 2021-03-23
custom action example in README produces error


AndrewECarlson:
### Description
When I try to use the example config in the README for my_cool_custom_action, it produces an error. When I don't use the transform_mod function to wrap the custom function and just create the new action function directly, it works. 

It looks like the transform_mod way should add some nice stuff like conflict detection that the workaround would miss out on. Also, when I remove the transform_mod call, the workaround breaks with an error message from mt saying it expected a table.

**Expected Behavior**
Printing "This function ran after another action. Prompt_bufnr: <number>" when I hit <c-i> in normal mode from Telescope
**Actual Behavior**
E5108: Error executing lua ...k/packer/start/telescope.nvim/lua/telescope/mappings.lua:231: attempt to call local 'key_func' (a table value)

### Details

<details><summary> try to use the custom function in Telescope </summary>

1. nvim -nu test.vim
2. :Telescope <any>
3. .... bang here is the issue

<details><summary>Environment</summary>

- nvim --version output: 
NVIM v0.5.0-dev+1157-g0ab88c2ea
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/cc -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/runner/work/neovim/neovim/build/config -I/home/runner/work/neovim/neovim/src -I/home/runner/work/neovim/neovim/.deps/usr/include -I/usr/include -I/home/runner/work/neovim/neovim/build/src/nvim/auto -I/home/runner/work/neovim/neovim/build/include
Compiled by runner@fv-az204-778

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/share/nvim"

Run :checkhealth for more info
- Operating system: Arch Linux
[...]

PR #807, opened 2021-04-28
feat: Multi-select style actions

tjdevries:
WIP: Will update this later :)

Also will try and fix:
- https://github.com/nvim-telescope/telescope.nvim/issues/684#issuecomment-828517987
- https://github.com/nvim-telescope/telescope.nvim/pull/521/commits/0db1c19c624b60bee3d5174d00421b206ed35fa5#

And I'll add docs when I'm done as well. Hopefully this is backwards compat, so when testing, please let me know if you experience breaking changes.


fdschmidt93:
With functional wrappers around selections & entries (see #931 and `telescope.actions.utils`) and [telescope-hop.nvim](https://github.com/nvim-telescope/telescope-hop.nvim) in place, I'd be happy to take this over  @tjdevries :) (after my vacation next week tho) If you prefer to work on that; no worries, I'd still love to see neogit or maybe neorg having full built-in telescope integration :sweat_smile: 

I'd have in mind that, for any `action` applicable, you should have full flexibility in running them on (i) the currently selected entry, (ii) all entries, (iii) a particular multi selection (if I understand `context` correctly), (iv) all multi selections. My design would be to ultimately wrap the `actions` into something like this:

```lua
local actions = require 'telescope.actions'
local action_utils = require 'telescope.actions.utils'
telescope.setup {
  defaults = {
    mappings = {
      i = {
        ["<C-v>"] = action_utils.with_selections(actions.select_vertical),
        -- obviously only for builtin.git_status :)
        ["<C-s>"] = action_utils.with_entries(actions.git_staging_toggle)
      },
    },
  },
```
Let me know what you think :)

Integralist:
Looks like this PR might have long gone stale. 

Does anyone know if there's a workaround for selecting multiple files in telescope yet as this is quite a key missing feature compared to FZF. 

Thanks.

Conni2461:
we currently dont have the time to finish this in a somewhat good interface, so it also works for extensions. Its not planned for telescope 0.1 and is scheduled for post 0.1. See #1938 

There is an issue with workarounds for it https://github.com/nvim-telescope/telescope.nvim/issues/1048

rkulla:
+1 very much looking forward to this being official. Being able to select multiples (with TAB) but not being able to open them still is a bummer. Especially for a feature I use multiple times a day every day but I will try the workarounds.

max397574:
I think for now sending to quickflist is the best alternative when you can't open the files directly

rkulla:
> I think for now sending to quickflist is the best alternative when you can't open the files directly

Yeah, I can stand that with vim-bqf's features but it's still a different feature all together.  Luckily I have both now thanks to vim-bqf and [this](https://github.com/nvim-telescope/telescope.nvim/issues/1048#issuecomment-1679797700) workaround

PR #809, opened 2021-04-29
feat: actions to create a new qflist

rudotriton:
Little something to accompany the current actions for quickfix lists that support the `"r"` and `"a"` `{action}` on `setqflist({list} [, {action}[, {what}]])`. These two new actions call `setqflist` with the `" "` empty `{action}` in order to create a new list instead of acting on the current one.


rudotriton:
I slept on this idea, and I'm doing it with a custom action instead. For my ideal case I also needed to adjust Telescope's local `send_all_to_qf` to append the new list to the end of everything instead of putting it after the current list and possibly overriding multiple list.

PR #812, opened 2021-05-01
Add option to configure width of items in gen_from_lsp_symbols()

tom-anders:
Symbol names can get very long, e.g. when namespaces are involved in C++. I thought it would be a nice addition to make the default values customizable here.


smithbm2316 (commented):
This LGTM, we should be able to merge this once #821 is fixed (that's what is causing the failed tests atm)

Issue #794, opened 2021-04-23
Quickfix picker doesn't jump to the location inside the file


weeman1337:
### Description

I am storing `live_grep`s result into the quickfix list. Then I am using Telescope to go through the quickfix list. Unfotunately on selecting an item it doesn't jump to the line and column.

**Expected Behavior**

Selecting a quickfix item jumps to the line and column.

**Actual Behavior**

Selecting a quickfix item jumps to the start of the file.

### Details

<details><summary>Reproduce</summary>

1. `nvim -nu test.vim`
2. `:Telescope live_grep`
3. Grep anything so that it doesn't point to the start of the file
4. Press C-q to send the results to the quickfix list
5. Close the search
6. `:Telescope quickfix`
7. Select any item
8. → it opens at start of the file

This information is available:

![image](https://user-images.githubusercontent.com/6216686/115913749-49b83200-a471-11eb-9123-08d40ce4f1b4.png)

</details>

<details><summary>Environment</summary>

- nvim --version output: 

```
NVIM v0.5.0-dev+1233-g82ac44d01
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/cc -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/runner/work/neovim/neovim/build/config -I/home/runner/work/neovim/neovim/src -I/home/runner/work/neovim/neovim/.deps/usr/include -I/usr/include -I/home/runner/work/neovim/neovim/build/src/nvim/auto -I/home/runner/work/neovim/neovim/build/include
Compiled by runner@fv-az69-718

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "
/home/runner/work/neovim/neovim/build/nvim.AppDir/usr/share/nvim"
```

- Operating system: Ubuntu 20.04.2
[...]

Issue #715, opened 2021-04-03
undolist builtin (maybe even a session builtin)


zenfailure:
I think the tilte suffices, a jump list builtin with file preview and undolist builtin with vimdiff preview could be an interesting additional default feature. Perhaps also a session builtin for vim sessions.

weilbith:
The undolist is kinda hard as it is a tree not a list. But I like the jumplist.

Conni2461:
Jumplist is in master. I am not sure about the undolist tho. 

You might want to use this for undo https://github.com/mbbill/undotree

Conni2461:
For session you can use this session manager: https://github.com/Shatur/neovim-session-manager
For undotree: https://github.com/mbbill/undotree

Closing for now

PR #813, opened 2021-05-02
Jumplist picker and jump to row/col on existing buffers.

caojoshua:
First time developer for lua, neovim plugin, and open source. Please help me in this learning experience.

Partially address #715 

1. Create a `jumplist` picker
2. Jump to row/col on existing buffers (helps for jumplist and also some existing pickers ie. quickfix)

The major limitation right now is that vim `getjumplist()` does not include file text. `vim.api.nvim_buf_get_lines()` does not work because some jump targets are in unloaded buffers. I decided to only include the filename and row/col in the entries, and rely on the previewer to view contents. Open to any suggestions here.

### DIFF
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -113,19 +113,18 @@ action_set.edit = function(prompt_bufnr, command)
   if entry_bufnr then
     edit_buffer(command, entry_bufnr)
   else
-
     -- check if we didn't pick a different buffer
     -- prevents restarting lsp server
     if vim.api.nvim_buf_get_name(0) ~= filename or command ~= "edit" then
       filename = path.normalize(vim.fn.fnameescape(filename), vim.loop.cwd())
       vim.cmd(string.format("%s %s", command, filename))
     end
+  end
 
-    if row and col then
-      local ok, err_msg = pcall(a.nvim_win_set_cursor, 0, {row, col})
-      if not ok then
-        log.debug("Failed to move to cursor:", err_msg, row, col)
-      end
+  if row and col then
+    local ok, err_msg = pcall(a.nvim_win_set_cursor, 0, {row, col})
+    if not ok then
+      log.debug("Failed to move to cursor:", err_msg, row, col)
     end
   end
 end
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -926,6 +926,30 @@ internal.tagstack = function(opts)
     }):find()
 end
 
+internal.jumplist = function(opts)
+  opts = opts or {}
+  local jumplist = vim.api.nvim_eval('getjumplist()')[1]
+
+  for _, value in ipairs(jumplist) do
+    value.text = ''
+  end
+
+  -- reverse the list
+  local sorted_jumplist = {}
+  for i = #jumplist, 1, -1 do
+    sorted_jumplist[#sorted_jumplist+1] = jumplist[i]
+  end
+
+  pickers.new(opts, {
+      prompt_title = 'Jumplist',
+      finder = finders.new_table {
+        results = sorted_jumplist,
+        entry_maker = make_entry.gen_from_jumplist(opts),
+      },
+      previewer = previewers.vim_buffer_qflist.new(opts),
+      sorter = conf.generic_sorter(opts),
+    }):find()
+end
 
 local function apply_checks(mod)
   for k, v in pairs(mod) do
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -928,7 +928,7 @@ end
 
 internal.jumplist = function(opts)
   opts = opts or {}
-  local jumplist = vim.api.nvim_eval('getjumplist()')[1]
+  local jumplist = vim.fn.getjumplist()[1]
 
   for _, value in ipairs(jumplist) do
     value.text = ''
@@ -937,7 +937,7 @@ internal.jumplist = function(opts)
   -- reverse the list
   local sorted_jumplist = {}
   for i = #jumplist, 1, -1 do
-    sorted_jumplist[#sorted_jumplist+1] = jumplist[i]
+    table.insert(sorted_jumplist, jumplist[i])
   end
 
   pickers.new(opts, {
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -916,39 +916,36 @@ internal.tagstack = function(opts)
   end
 
   pickers.new(opts, {
-      prompt_title = 'TagStack',
-      finder = finders.new_table {
-        results = tags,
-        entry_maker = make_entry.gen_from_quickfix(opts),
-      },
-      previewer = previewers.vim_buffer_qflist.new(opts),
-      sorter = conf.generic_sorter(opts),
-    }):find()
+    prompt_title = 'TagStack',
+    finder = finders.new_table {
+      results = tags,
+      entry_maker = make_entry.gen_from_quickfix(opts),
+    },
+    previewer = conf.qflist_previewer(opts),
+    sorter = conf.generic_sorter(opts),
+  }):find()
 end
 
 internal.jumplist = function(opts)
   opts = opts or {}
   local jumplist = vim.fn.getjumplist()[1]
 
-  for _, value in ipairs(jumplist) do
-    value.text = ''
-  end
-
   -- reverse the list
   local sorted_jumplist = {}
   for i = #jumplist, 1, -1 do
+    jumplist[i].text = ''
     table.insert(sorted_jumplist, jumplist[i])
   end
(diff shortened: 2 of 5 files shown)


lua/telescope/actions/set.lua
```diff
@@ -113,19 +113,18 @@ action_set.edit = function(prompt_bufnr, command)
...
 
-    if row and col then
-      local ok, err_msg = pcall(a.nvim_win_set_cursor, 0, {row, col})
-      if not ok then
-        log.debug("Failed to move to cursor:", err_msg, row, col)
-      end
+  if row and col then
+    local ok, err_msg = pcall(a.nvim_win_set_cursor, 0, {row, col})
+    if not ok then
+      log.debug("Failed to move to cursor:", err_msg, row, col)
```
smithbm2316, line 127: I'm not sure if this was an intentional change, or if you have an auto-formatter that maybe did this, but why did you move this `if row and col then` block outside of the `else` block above? Was this change necessary to make your new picker work?

lua/telescope/builtin/internal.lua
```diff
@@ -926,6 +926,30 @@ internal.tagstack = function(opts)
     }):find()
 end
 
+internal.jumplist = function(opts)
+  opts = opts or {}
+  local jumplist = vim.api.nvim_eval('getjumplist()')[1]
```
smithbm2316, line 931: Just as an fyi, you should be able to use `vim.fn` here to call the `getjumplist()` function. `vim.fn` is the Neovim Lua api for calling Vimscript functions in Lua, so you could change that line to:
```lua
local jumplist = vim.fn.getjumplist()[1]
```

lua/telescope/builtin/internal.lua
```diff
@@ -926,6 +926,30 @@ internal.tagstack = function(opts)
...
+
+  for _, value in ipairs(jumplist) do
+    value.text = ''
+  end
+
+  -- reverse the list
+  local sorted_jumplist = {}
+  for i = #jumplist, 1, -1 do
+    sorted_jumplist[#sorted_jumplist+1] = jumplist[i]
```
smithbm2316, line 940: I believe you should be able to replace the line in the for loop with `table.insert(sorted_jumplist, jumplist[i])` if you'd like. `table.insert` is a part of Lua's table library which inserts an item into a table. If you only give two arguments to it, it will accept the table to insert into and the item you want to insert, inserting that item at the end of the table (similar to the `append()` method in Python for example). If you give 3 args, it would take the table to insert into, the position in the table to insert the item, and then the item to insert. Just a little bit easier to read I think using `table.insert` instead :smiley:

smithbm2316 (commented):
I love this! Afaik you've done a great job for this being your first experience writing lua for a neovim plugin :smile:. Thanks for the contribution and for taking an interest! I've left a few comments of minor suggestions and clarifications, but this looks pretty nice to me. I'm a pretty new maintainer here and am still learning the codebase myself, so I'll wait for feedback from one of the other maintainers before considering merging this.

You mentioned that `getjumplist()` doesn't include the file text, can you describe why this is an issue for what you built here? I noticed you can see the location in the jumplist in the preview window, what did you have in mind to implement if you were able to get access to the file text?

lua/telescope/actions/set.lua
caojoshua: This is intentional. Its not necessary for the picker to work, but is necessary to set the cursor to `row, col`. Otherwise, it just jumps to where the buffer was previously left off, or line 0 of the buffer.

Regardless of whether we want this picker, this change should be merged. For example, lets say we are telescoping the quickfix list, and an open buffer has `N` entries. If we select any of those entries, the cursor wherever the buffer was left previously. With this change, the cursor would be set at the correct `row, col`.

caojoshua:
Thanks for the review!

Let me clarify. For example, quickfix list does:
```
    return displayer {
      line_info,
      entry.text:gsub(".* | ", ""),
      filename,
    }
```
Where `entry.text` holds the actual line of code/text in the file. This works because the items returned from `vim.fn.getqflist()` have an `item` field. `vim.fn.getjumplist()` does not do the same.

Instead, I've implemented
```
    return displayer {
      line_info,
      filename,
    }
```
I would prefer it to also include the text, but there does not seem to be an easy way to do it.

smithbm2316:
> Where `entry.text` holds the actual line of code/text in the file. This works because the items returned from `vim.fn.getqflist()` have an `item` field. `vim.fn.getjumplist()` does not do the same.
> 
> I would prefer it to also include the text, but there does not seem to be an easy way to do it.

Oh I see! Unfortunately I don't know of any other vimscript functions that might make that possible. I'm relatively new to Lua and Neovim plugin development myself, so I'm still learning the APIs. With the new changes this LGTM, as I mentioned before I'll wait for another maintainer to give this the green light before merging.

smithbm2316:
@tami5 @elianiva any chance one of you could take a peek at this PR? If you think it looks good I can merge, I'm still not too familiar with the `make_entry.lua` section of the codebase yet.

lua/telescope/builtin/internal.lua
```diff
@@ -926,6 +926,30 @@ internal.tagstack = function(opts)
...
+
+  for _, value in ipairs(jumplist) do
+    value.text = ''
+  end
+
+  -- reverse the list
+  local sorted_jumplist = {}
+  for i = #jumplist, 1, -1 do
+    table.insert(sorted_jumplist, jumplist[i])
+  end
```
Conni2461, line 941: Can't we merge these two iterations

lua/telescope/builtin/internal.lua
```diff
@@ -926,6 +926,30 @@ internal.tagstack = function(opts)
...
+
+  pickers.new(opts, {
+      prompt_title = 'Jumplist',
```
Conni2461, line 944: wrong indent

lua/telescope/builtin/internal.lua
```diff
@@ -926,6 +926,30 @@ internal.tagstack = function(opts)
...
+      finder = finders.new_table {
+        results = sorted_jumplist,
+        entry_maker = make_entry.gen_from_jumplist(opts),
+      },
+      previewer = previewers.vim_buffer_qflist.new(opts),
```
Conni2461, line 949: should be `previewer = conf.qflist_previewer(opts),`

lua/telescope/actions/set.lua
Conni2461: I think this fixes: https://github.com/nvim-telescope/telescope.nvim/issues/794

lua/telescope/builtin/internal.lua
Conni2461: Oh i think its also wrong for `tagstack`, can you fix that as well

lua/telescope/builtin/internal.lua
Conni2461: Indent in `tagstack` is also wrong.

lua/telescope/actions/set.lua
caojoshua: oh sweet didn't know there was already an issue for this. Yes I was able to reproduce the issue and fix it.

tami5 (approved):

smithbm2316 (approved):

smithbm2316:
@caojoshua thanks for all the hard work on this! I'll merge this once #821 is resolved (that is what is causing the two tests to fail atm)

Conni2461 (approved):
LGTM.

Can we get a mention of that picker in the README?

caojoshua:
Thanks everyone. Left another commit for adding things to README

lua/telescope/make_entry.lua
```diff
@@ -1130,5 +1130,57 @@ function make_entry.gen_from_git_status(opts)
...
+    local line_info = {table.concat({entry.lnum, entry.col}, ":"), "TelescopeResultsLineNr"}
+
+    return displayer {
+      line_info,
+      filename,
+    }
+  end
+
+  return function(entry)
+    local filename = entry.filename or vim.api.nvim_buf_get_name(entry.bufnr)
```
Conni2461, line 1165: Just tested it. I am getting: `Invalid buffer id: 1` here if i had startify open before.
So: `nvim` -> `:Telescope find_files` -> `open file` -> `:Telescope jumplist` (error)
If i do: `nvim file` -> `:Telescope jumplist` works great

We should do
```lua
if not vim.api.nvim_buf_is_valid(entry.bufnr) then
  return
end
```
before this line. Thoughts?

Conni2461: Otherwise works great :) thanks

caojoshua: Added this to the commit.

I personally don't use startify, and haven't been able to reproduce a scenario where this makes/breaks anything. I think its OK, unless there is some other sort of verification we can do.

Conni2461:
Thanks all the work. Let's get this merged :)

PR #817, opened 2021-05-05
Accept env vars in `get_os_command_output`

morgsmccauley:


Conni2461:
Is there a reason why we need this in core? This util was previously just a shorthand for:
```lua
local f = assert(io.popen(cmd, 'r'))
local s = assert(f:read('*a'))
f:close()
```
Then i moved it to `plenary.job`.
If you want more complicated job stuff, you should probably write your own job rather than using this shorthand. :)

morgsmccauley:
Sorry for the no context PR. Main reason is for creating a silent commit squashing action for the `git_commit` picker by setting `GIT_SEQUENCE_EDITOR` i.e.
```lua
        local _, ret, stderr = utils.get_os_command_output(
          {
            'git',
            'rebase',
            '-i',
            '--autosquash',
            '--autostash',
            commit..'~1'
          },
          cwd,
          { GIT_SEQUENCE_EDITOR = ':' }
        )
```

I'm happy to create my own function for handling this if you don't think this belongs in core. Feel free to close :)

tjdevries:
I think at that point, might as well just use `Job:new { ... }:sync()`. Thanks for the PR but I'll close for now.

PR #819, opened 2021-05-06
fix(lsp): handle errors from buf_request_sync calls

williamboman:
If `buf_request_sync` runs into an error, or timeout, it'll return `(nil, err)`.


tami5:
LGTM 

cc @elianiva

elianiva (approved):
LGTM

tami5:
Thanks @williamboman

PR #820, opened 2021-05-06
Fix oldfiles cwd_only that include backslashes

l-kershaw:
Fixes an issue that stops the `cwd_only` option working for `oldfiles` when the `cwd` includes backslashes.

---

The results table has filenames with double escaped backslashes, but the cwd has only single escaped backslashes.
This problem stopped this option from working at all on Windows.


Conni2461:
LGTM Thanks :)

We have currently CI problems because of #821. I will merge as soon as these problems are resolved

Issue #818, opened 2021-05-05
Display Configuration


l-kershaw:
Will talk about width, but same logic applies to height.

**Is your feature request related to a problem? Please describe.**

Two different variables are used to calculate the width of a picker
- `width`
  - a top level option/default
  - default value: `0.75`
- `width_padding`
  - an entry in the `layout_defaults` option/default
    - allows you to specify a default value directly within `layout_defaults`
    - also allows you to specify a custom value for each `layout_strategy`
  - default value: hard-coded in each strategy

Problems:
- These values can be calculated from each other by `width_padding = math.floor((max_columns - width)/2)` or `width = max_columns - 2 * width_padding`
- Therefore most of the layout strategies ignore the `width` option and just use the `width_padding`, then calculating the appropriate `width`.
- This can make it confusing to users, as you would expect changing the `width` option to affect the width of the picker.

**Describe the solution you'd like**

A hierarchy of options/defaults.
When calculating the width of a picker with strategy `strat`, use the first non-null value in the following list to calculate both the `width` and the `width_padding`.
- `opt.width`
- `opt.layout_config[strat].width`
- `opt.layout_config[strat].width_padding`
- `opt.layout_config.width`
- `opt.layout_config.width_padding`
- `default.width` : default value=`nil`
- `default.layout_config[strat].width` : default value=`nil`
- `default.layout_config[strat].width_padding` : default value=`nil`
- `default.layout_config.width` : default value=`nil`
- `default.layout_config.width_padding` : default value=`math.max(max_columns*0.1,5)`

Advantages:
- Since `width` is still a top-level option, you can specify a custom width of the picker easily in any custom mappings.
- Since there is now a pre-defined order, you know that if you specify a width for a picker, it will actually affect the width, regardless of the layout.

Disadvantages:
- This is a lot of options that only affect one thing.
  - Like, a lot, a lot.

**Describe alternatives you've considered**

- Remove the `width_padding` options/defaults entirely.
  - Since `width_padding` can be calculated easily from `width`, this vastly simplifies the code required.
  - This would be a breaking change for some people's configs, especially as it affects the more commonly used layout strategies.
  - This makes it slightly more complicated for a user to specify that they want an `n` column gap on each side of the picker.
- Remove the `width` options/defaults entirely.
  - Again, since `width` can be calculated easily from `width_padding`, this simplifies the code required.
[...]

Issue #821, opened 2021-05-07
Telescope broken on neovim HEAD


clason:
Neovim just merged https://github.com/neovim/neovim/pull/13664, which breaks Telescope: Opening any finder (e.g., just `:Telescope`) yields
```
E5108: Error executing lua .../packer/start/plenary.nvim/lua/plenary/window/border.lua:138: Vim(autocmd):E680: <buffer=6>: invalid buffer number
```

I suspect this is due to https://github.com/neovim/neovim/pull/13664/commits/97abf8ee63994c19c2119e34bc7a7aa66457a7f2, which changed how autocommands are triggered in popups.

clason:
That doesn't fix it, unfortunately; then I get
```
E5108: Error executing lua ...ck/packer/start/telescope.nvim/lua/telescope/pickers.lua:343: Invalid window id:
 1006
```

Conni2461:
Thanks for testing :) I am pretty sure that got us further. I will try to track it down tomorrow morning. I wanted to do some telescope work this weekend anyway, so its at least a little bit maintained while tj is on vacation.

clason:
In the meantime, for those coming here:
```
git checkout 4be0e92db01a502863ac4bb26dd0fee16d833145
make
make install
```
"unbreaks" telescope.

Conni2461:
https://github.com/nvim-lua/plenary.nvim/pull/155 and https://github.com/nvim-lua/popup.nvim/pull/13 to fix this issue.

@clason I can merge anything into plenary but i don't have write rights for popup. Is there any chance that you can merge things into popup? Otherwise i will write tj in private

tjdevries:
🎶 🎶 🎶 
Listen, baby, ain't no mountain high
Ain't no valley low, ain't no river wide enough, baby
If you need me, call me, no matter where you are
No matter how far, don't worry, baby
Just call my name, I'll be there in a hurry
You don't have to worry 
🎶 🎶 🎶

clason:
aaaand it's fixed :)

raka26:
Thanks for your hard work!

Issue #796, opened 2021-04-25
Preview window cannot be scrolled when layout_strategy = "flex"


yutkat:
### Description

When I try to scroll the preview window with layout_strategy = "flex", I get the following error.
`E5108: Error executing lua ...vim/plugged/telescope.nvim/lua/telescope/actions/set.lua:140: attempt to index field 'layout_config' (a nil value)`

`layout_strategy = "horizontal",` works as expected.

**Expected Behavior**

scroll up/down in the preview window

**Actual Behavior**

display this error message
`E5108: Error executing lua ...vim/plugged/telescope.nvim/lua/telescope/actions/set.lua:140: attempt to index field 'layout_config' (a nil value)`

### Details

<details><summary>Reproduce</summary>

1. nvim -nu test.vim
2. `:Telescope buffers`
3. display this error message `E5108: Error executing lua ...vim/plugged/telescope.nvim/lua/telescope/actions/set.lua:140: attempt to index field 'layout_config' (a nil value)`
</details>

<details><summary>Environment</summary>

- nvim --version output: v0.5.0-dev+1272-g4570d04b1
- Operating system: ArchLinux
- Telescope commit: 6fd1b3bd255a6ebc2e44cec367ff60ce8e6e6cab

</details>

<details><summary>Configuration</summary>
<p>

```viml
set nocompatible hidden laststatus=2

if !filereadable('/tmp/plug.vim')
  silent !curl --insecure -fLo /tmp/plug.vim
    \ https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
endif

source /tmp/plug.vim
call plug#begin('/tmp/plugged')
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-telescope/telescope.nvim'
call plug#end()

autocmd VimEnter * PlugClean! | PlugUpdate --sync | close
lua << EOF

require('telescope').setup{
  defaults = {
    vimgrep_arguments = {
      'rg',
      '--color=never',
      '--no-heading',
      '--with-filename',
      '--line-number',
      '--column',
      '--smart-case'
    },
    prompt_position = "bottom",
    prompt_prefix = ">",
    selection_strategy = "reset",
    sorting_strategy = "descending",
    layout_strategy = "flex",
    layout_defaults = {},
    file_ignore_patterns = {},
    shorten_path = true,
    winblend = 0,
    width = 0.75,
    preview_cutoff = 120,
[...]

Issue #765, opened 2021-04-16
Bottom horizontal split theme


AnatoleLucet:
**Is your feature request related to a problem? Please describe.**
This is just a feature I think some people might like (including myself).

**Describe the solution you'd like**
I'd be nice to have a theme where the results are shown in a horizontal split at the bottom of the screen.
For e.g. something like CoC's fuzzy finder:

![image](https://user-images.githubusercontent.com/35486736/115040105-5d8ff100-9ed1-11eb-8b45-1f45f129cf09.png)

I'd personally prefer this layout in a few cases like the `commands` built-in.

tjdevries:
How about something like this:

https://github.com/nvim-telescope/telescope.nvim/pull/771

It's just a rough draft

AnatoleLucet:
It's looking great, thanks!
If the input is on the top, maybe the results list could go from top to bottom / stick to the top of the "results buffer"?
I also think that an horizontal line on the top'd nice to mark the separation between telescope and the background (like the borders of the other themes).

elianiva:
> If the input is on the top, maybe the results list could go from top to bottom / stick to the top of the "results buffer"?

this part is configurable

tjdevries:
OK, you can try this now with:

`lua require('telescope.builtin').find_files(require('telescope.themes').get_ivy())`

Let me know if that is what you're looking for. It still allows for all the normal telescope configurations.

tjdevries:
![image](https://user-images.githubusercontent.com/4466899/118375264-c29d3c00-b58e-11eb-9cfd-0a587b970c92.png)

![image](https://user-images.githubusercontent.com/4466899/118375269-ca5ce080-b58e-11eb-9591-a64f446a16d1.png)

smithbm2316:
@tjdevries thanks for the hard work on this new layout, I'm really looking forward to playing around with it! Is there a particular difference between setting your default `layout_strategy` to `bottom_pane` versus using the `ivy` theme, other than all of your pickers would be using the `bottom_pane` strategy by default? Or is it otherwise the same pretty much?

Hope you had a wonderful vacation! I'm looking forward to popping into a Teej stream again soon 😃

tjdevries:
`layout_strategies` are _only_ for how the windows are laid out.

`themes` are for combining `layout_strategies`, `sorting_strategies`, `borders` and more to make a cohesive unit that is easy for end users to consu.e

[...]

PR #823, opened 2021-05-07
Update display config methods

l-kershaw:
Plan:

- [x] Implement hierarchy of options that affect window size. _(requires fuller testing)_
- [x] Remove `width_padding` and `height_padding` option from layout strategies.
- [x] Add new option for `width` and `height` allowing a table of the form {`padding`=foo} where foo is one of the other options, and then define width in terms of padding.
- [x] Add documentation for the new method of defining dimensions of the picker.

For more details on the general idea, see the discussion in [this issue](https://github.com/nvim-telescope/telescope.nvim/issues/818).

---

Feedback is very welcome 🙂


Conni2461:
Thanks for all the work here. I know that tj also had plans to change how layouts work. So its hard for me to comment on your proposal. I know there is this: https://github.com/nvim-telescope/telescope.nvim/pull/773 and https://github.com/nvim-telescope/telescope.nvim/pull/771

Its definitely one of the areas telescope needs improvement. So thanks again.

PS: Ignore tests, #821 is currently breaking everything. But make sure that your fork has github actions enabled (or permissions) so `docgen` runs. It updates the vim documentation based on function headers. :)

tjdevries:
This looks cool, but wait a week for me to take a look when I'm back before merging. It's OK if we make some breaking changes, but I'd prefer to only break it once for a bit 😂

Thanks for the PR, I like where this is going.

l-kershaw:
Couple of comments on defaults affecting layout:
- I think we should rename `layout_defaults` to `layout_config`
  - Since this option is already contained in the `default` block of options in `setup()`, I don't think repeating the word default is necessary
  - This means that the key used for these matches in setting defaults and in setting options for a particular command
- I think that the defaults that we have for layout are good in that it's clear how they work, but they probably aren't the best default options in terms of usability.
  - For example, for particularly thin windows, it would be better to define width in terms of padding, but for wider windows it's better to use a percentage.
  -  We could change the defaults to be functions that take into account the size of the Neovim window, which I think would make them better for users, but probably harder to understand and/or document.

What are people's thoughts on this?

smithbm2316:
> * I think we should rename `layout_defaults` to `layout_config`
>   * Since this option is already contained in the `default` block of options in `setup()`, I don't think repeating the word default is necessary
>   * This means that the key used for these matches in setting defaults and in setting options for a particular command

I think this is a good idea. I was trying to debug an issue recently with the `layout_config` causing an issue with the `flex` layout, and it was a bit confusing to me bouncing back and forth between places that said `layout_defaults` and `layout_config`. I think one name would make it much simpler.

> * I think that the defaults that we have for layout are good in that it's clear how they work, but they probably aren't the best default options in terms of usability.
>   * For example, for particularly thin windows, it would be better to define width in terms of padding, but for wider windows it's better to use a percentage.
>   * We could change the defaults to be functions that take into account the size of the Neovim window, which I think would make them better for users, but probably harder to understand and/or document.

I like the sound of this, but don't really know enough about how the size of Telescope is currently calculated to have a very meaningful opinion on this aspect. I'll let others hopefully give more constructive feedback on this point.

Thanks for all of your hard work lately @l-kershaw, really looking forward to seeing some of the code you've written getting merged!

l-kershaw:
Yeh, I was having the same issue. I have changed the name so it matches now

---

> I like the sound of this, but don't really know enough about how the size of Telescope is currently calculated to have a very meaningful opinion on this aspect. I'll let others hopefully give more constructive feedback on this point.

I think this is fairly common, particularly as some layout strategies ignore some options that it seems like they shouldn't.
E.g. on the current master branch, the horizontal layout ignores the `width` option, and only changes based on the `width_padding` option within `layout_config`. If this is not present, it has a hard-coded option within the function, rather than being set as a default with the rest of the defaults. This function does depend on the width of the window, but it is not clear how unless you dig into the code.

Hopefully making it clearer how this is calculated will make it easier for users to customise as they wish.
That's the aim of this PR 🙂

README.md
```diff
@@ -146,7 +146,7 @@ require('telescope').setup{
     selection_strategy = "reset",
     sorting_strategy = "descending",
     layout_strategy = "horizontal",
-    layout_defaults = {
+    layout_config = {
```
tjdevries, line 149: if we make this change, we should error in the setup if `layout_defaults` is passed. otherwise uses will be confused.

lua/telescope/config/resolve.lua
```diff
@@ -88,8 +93,7 @@ That's the next step to scrolling.
 local get_default = require('telescope.utils').get_default
 
 local resolver = {}
-
-local _resolve_map = {
+local _simple_resolve_map = {
```
tjdevries, line 96: this is probably easier to understand by just doing `_resolve_map = {}` and then each of the following with `_resolve_map[function() ...] = ...` etc.

I don't think we need to do the copy thing, right?

Or is this because only some keys are considered "simple"

lua/telescope/pickers.lua
```diff
@@ -95,19 +95,20 @@ function Picker:new(opts)
...
-      (config.values.layout_defaults or {})[layout_strategy]
+      config.values.layout_config
     ) or {},
 
     window = {
       -- TODO: This won't account for different layouts...
-      -- TODO: If it's between 0 and 1, it's a percetnage.
-      -- TODO: If its's a single number, it's always that many columsn
+      -- TODO: If it's between 0 and 1, it's a percentage.
+      -- TODO: If it's a single number, it's always that many columns
```
tjdevries, line 104: I think maybe we can remove these TODOs... this is implemented now, right?

lua/telescope/pickers.lua
```diff
@@ -95,19 +95,20 @@ function Picker:new(opts)
...
       -- TODO: If it's a list, of length 2, then it's a range of min to max?
       height = get_default(opts.height, 0.8),
       width = get_default(opts.width, config.values.width),
 
       get_preview_width = get_default(opts.preview_width, config.values.get_preview_width),
 
+      -- TODO: figure out the intention of this option. It doesn't currently do anything
```
tjdevries, line 111: might not be useful anymore :)

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -101,39 +104,89 @@ local is_borderless = function(opts)
   return opts.window.border == false
 end
 
+-- TODO: make an automatic way to create this list
+--      could just use layout_strategies???
+local layout_list = {horizontal=1,vertical=1,flex=1,center=1,current_buffer=1}
```
tjdevries, line 109: can check if the key is in `layout_strategies` maybe? (if that's what it does)

tjdevries:
This looks good. I have been meaning to do this for awhile, and I like where it's going. Cool idea with `{ padding = ... }`, which will allow us to extend other weird situations in the future if people would like configure it.

I think one thing is probably just writing a bit more docs on what you expect to have happen with `{ padding = ... }` so people can understand it. For example, `number of columns = 100` and then show how different values could play out.

Nice work & thanks!

tjdevries:
If you're interested, you can port the changes I had here as well : https://github.com/nvim-telescope/telescope.nvim/pull/773

Which will make the docs quite easy to add and added some other documentation as well.

tjdevries:
P.S. Sorry, just added a new layout strategy :laughing:

lua/telescope/config/resolve.lua
l-kershaw: Originally, I was getting some issues where the `padding` case in `_resolve_map` was referring back to `_resolve_map`.

I don't seem to be getting those issues now, so I've changed to just using `_resolve_map` instead of `_simple_resolve_map`.
I'll keep an eye out for any issues re-emerging.

l-kershaw:
@tjdevries I've made the changes you suggested, thanks for the feedback.

---

> If you're interested, you can port the changes I had here as well : #773
> 
> Which will make the docs quite easy to add and added some other documentation as well.

If I'm being honest, I don't 100% understand what is happening in your PR.
I'll have another look when I get a chance, but I think the docs we have in this PR are probably sufficient for the moment.

> P.S. Sorry, just added a new layout strategy 😆

No worries 😅 
I think I've got it working with the resolver and stuff now, but if people find any issues I can have another look at it

---

FYI: due to GitHub actions issues (see [here](https://www.githubstatus.com/incidents/zbpwygxwb3gw)) the docs file has not been updated yet

l-kershaw:
I think I have got the key features from #773 implemented now, but I may have misunderstood that PR, so would be good if someone can check 🙂 

---

On another note, does anyone know if there is a way to force a newline within an itemize/enumerate for the docgen?
This is a bit of an issue for the resolver documentation, as just having it without these line breaks is a bit difficult to read.

I could wrap the whole thing in `<pre> ... </pre>` but this seems like something that would make maintenance of the documentation more difficult in the long run.

Conni2461:
> On another note, does anyone know if there is a way to force a newline within an itemize/enumerate for the docgen?

Can you give me an example. Because i have written the whole markdown formatter/renderer maybe we can support it. I remember you can force a `newline` with `<br>` at the end of the line but i can't remember if i implemented that for `itemize`/`enumerate`.

Example would be good so i have a test case when implementing it :)

l-kershaw:
@Conni2461 There is an example [here](https://github.com/nvim-telescope/telescope.nvim/blob/97dd75e8c7b371485ff77bf8ae9e4ced4496b544/lua/telescope/config/resolve.lua#L153) within this PR.

---

For a more minimal example:
```
1.  key idea of point one <br>
    A very long and detailed explanation about point one that is longer than the width of the docs file.
2. key idea of point two <br>
    An even longer and more detailed explanation about point two that would be far too long for
    use in a docs file without a break somewhere.
```

Ideally the docgen would output something like:
```
1.  key idea of point one
    A very long and detailed explanation about point one that is longer than
    the width of the docs file.
2. key idea of point two
    An even longer and more detailed explanation about point two that would
    be far too long for use in a docs file without a break somewhere.
```

Conni2461:
Okay thanks. I will try to implement this in the next ~3days

Conni2461:
Lunch break implementation. https://github.com/tjdevries/tree-sitter-lua/pull/23 I couldn't stop thinking about it :rofl: 

Also thanks for pushing that formatter/renderer to its limits. You are the first one in these two months since i replaced tjs previous method. That really helps improving it :)

lua/telescope/config.lua
```diff
@@ -49,6 +49,10 @@ config.descriptions = {}
 function config.set_defaults(defaults)
   defaults = defaults or {}
 
+  if defaults.layout_default then
+    error("'layout_default' is no longer a valid key. See 'layout_config'")
```
tjdevries, line 53: `See 'layout_config'` is not a super helpful error message IMO.

I was thinking as well, perhaps we should just warn them for now and then do `defaults.layout_config = defaults.layout_defaults` or similar.

This way we don't crash everyone's telescope right away on download. (we can error in one week or something)

lua/telescope/config.lua
```diff
@@ -85,15 +89,58 @@ function config.set_defaults(defaults)
...
+    each strategy.
+
+    Allows setting defaults for all strategies as top level options and
+    for overriding for specific options.
+    For example, the default values below set the default width to 80% of
+    the screen width for all strategies except 'center', which has width
+    of 50% of the screen width.
+
+    Default:
+    {
```
tjdevries, line 124: we can we get this generated from the value above?

like `layou_config_default = { ... } `, and then use `vim.inspect` so that it doesn't get  out of date?

l-kershaw: I'm not sure what the best way to do this is.
Also, it _might_ make it easier for the explanation that goes along with it to get out of date.

tjdevries:
@l-kershaw do you mind if I push a few things up to this branch? I have some ideas after looking at the code that would take a lot of comments to explain :laughing:

l-kershaw:
@tjdevries Yeh, that's not a problem 👍

tjdevries:
hehe, OK, I'm getting a bit carried away, but hopefully this should make things much simpler. I'll send the commits up today or tomorrow (hopefully). If you can hold off on doing any commits for that long, that'd be great, cause otherwise I think we're going to generate some confusing conflicts.

tjdevries:
OK, some stuff is broken still -- so it's not ready to merge, but figured I should push.

I broke many things in terms of config :) I added changelog doc inside of telescope docs that we can link to, which I think will be helpful, cause this will probably end up being pretty different.

@l-kershaw interested in your thoughts.

tjdevries:
Related issues (I think I found all the open ones) 

https://github.com/nvim-telescope/telescope.nvim/issues/796

https://github.com/nvim-telescope/telescope.nvim/issues/765

https://github.com/nvim-telescope/telescope.nvim/issues/667

https://github.com/nvim-telescope/telescope.nvim/issues/195

l-kershaw:
@tjdevries I've managed to look through most of it and have a few comments/questions:
- What do you intend the `shared_options` table ([here](https://github.com/nvim-telescope/telescope.nvim/blob/5c45e797ad6e5efb20d2cdb5872ad9bfd26c8ebd/lua/telescope/pickers/layout_strategies.lua#L119)) to be? Is it meant to be a list of keys that are used by all layouts or keys that are used by more than one layout?
- I found the use of the `height` key in the center layout a bit confusing, because the height of the whole layout doesn't correspond to the `height` key. e.g. if you set `height=0.5` the layout actually takes up the top 75% of the screen (roughly). I'm not sure what the best way to sort this is if we want to get rid of the `results_height` key.
  - Personally, I think it's best to keep the `results_height` option for this layout.
- Using `vim.tbl_deep_extend` [here](https://github.com/nvim-telescope/telescope.nvim/blob/5c45e797ad6e5efb20d2cdb5872ad9bfd26c8ebd/lua/telescope/config.lua#L65), means that either we have to have all our defaults as "top level" (so all layouts must have the same width for example) or if we have different, for example `width` for different layouts, if a user sets width as a "top level" option within their setup, it won't do anything.
  - I thought about extending the defaults here and couldn't come up with a simple (and more importantly, user understandable) way of doing it where users can just set the things they want to change, and also where "top level" options are respected
- I think we should probably put explanations for what we intend all of our helper functions to do in comments above them (e.g. `get_valid_configuration_keys`, `validate_layout_config`). I'm starting to lose track of what they are all meant to be doing, and I wrote most of them 🤣

tjdevries:
My next plans are:

- refactor a bunch of the code that picks out the options so that we can test it outside of opening up actual pickers.
  - so this way we can test what happens with "top level", "nested", etc. and make sure that we're picking the values that we hope
- center is just weird right now. we should make it act more like the other ones if possible.

(we can chat about this maybe on Friday? I'll be on stream. if you want we can talk about it live via discord or just via chat or I can record something or whatever :laughing: )

l-kershaw:
I've also just noticed that you removed the stuff [here](https://github.com/nvim-telescope/telescope.nvim/blob/2c1d42fe62f95c8437ee6d747dbbbe71792fbdfb/lua/telescope/pickers/layout_strategies.lua#L91) in my most recent commit allowing you to have keys in your options that aren't valid for the current layout but are for another. It handles this by just ignoring these keys without error (unlike keys that are invalid for all layouts).

This seems like it would be helpful, but I suppose it could encourage us to be a bit sloppy with what we input into `validate_layout_config`.

Was this an intentional choice to remove this functionality or just a biproduct of the general refactoring?
Should we add it back or leave as is?

doc/telescope_changelog.txt
```diff
@@ -0,0 +1,37 @@
...
+        `layout_config`.
+
+    Previously, these keys were passed via `picker:find(opts)`, but should be
+    passed via `opts.layout_config` now.
+        - {height}
+        - {width}
+
+    These keys are removed:
+        - {results_height}: This key is no longer valid. Instead, use `height`
+          and the corresponding `preivew_*` options for the layout strategy to
```
smithbm2316, line 30: *preview

doc/telescope.txt
```diff
@@ -1168,99 +1246,164 @@ All layout strategies are functions with the following signature:
...
+        - Where to place prompt window.
+        - Available Values: 'bottom', 'top'
 
 
 layout_strategies.center()                        *layout_strategies.center()*
-    Centered layout wih smaller default sizes (I think)
+    Centered layout with a combined block of the prompt and results aligned to
+    the middle of the screen. The preview window is then placed in the
+    remaining space above. Particularly useful for creating dropdown menus (try
+    using `theme=get_dropdown`).
```
smithbm2316, line 1309: We may want to consider linking to a part of the documentation that talks about using a theme here, as I think there is generally some confusion on how to use a theme for new users.

doc/telescope.txt
```diff
@@ -1168,99 +1246,164 @@ All layout strategies are functions with the following signature:
...
-layout_strategies.flex()                            *layout_strategies.flex()*
-    Swap between `horizontal` and `vertical` strategies based on the window
-    width
-     - Supports `vertical` or `horizontal` features
 
-    Uses:
-     - flip_columns
-     - flip_lines
+layout_strategies.bottom_pane()              *layout_strategies.bottom_pane()*
+    Bottom pane can be used to create layouts similar to "ivy".
```
smithbm2316, line 1404: @tjdevries are you planning to add `bottom_pane` docs in this PR, or would you like help on that? I know that it's a new option so it might not be ready to be documented yet

smithbm2316 (commented):
Just noted some small questions/typos. Overall really think this is going to be a great change and is looking great to me!

l-kershaw:
I have changed the `config.set_defaults` function so that it allows you to pass in some defaults, instead of having to use the normal ones.
I think this will be helpful for setting up tests for how we set options.

doc/telescope.txt
l-kershaw: I have added links [here](https://github.com/nvim-telescope/telescope.nvim/blob/6ca31e2d0d804af82258451a288dfda99cb821bd/lua/telescope/pickers/layout_strategies.lua#L307) but docgen doesn't seem to be working for me right now.

smithbm2316: Oh I see, I missed that in my review. That's a bit odd why it's not picking them up

l-kershaw: You didn't miss them, I added them after your review

Yeh, not sure why it's not updating

l-kershaw:
One of the `layout_config` tests fails ([this](https://github.com/nvim-telescope/telescope.nvim/blob/904b8eb87e244e4bdf854c178eada70ac064c1ba/lua/tests/automated/layout_strategies_spec.lua#L75) one).

I think this is due to the use of `vim.tbl_deep_extend` [here](https://github.com/nvim-telescope/telescope.nvim/blob/904b8eb87e244e4bdf854c178eada70ac064c1ba/lua/telescope/config.lua#L77).

This test checks for what I would expect to happen if I set `height` in my setup() `layout_config`.

l-kershaw:
Along with the issue with `vim.tbl_deep_extend` [here](https://github.com/nvim-telescope/telescope.nvim/blob/904b8eb87e244e4bdf854c178eada70ac064c1ba/lua/telescope/config.lua#L77), I don't really understand why we have `config.values[name]` in this line. It seems to also be in the `first_non_null` case below (though in a different order of preference), and I'm not sure why it's there either.

@tjdevries, you seem to be the last person who edited these bits.
Do you know the reasoning for including this here?

l-kershaw:
The latest commit replaces the use of `vim.tbl_deep_extend` in the `get` function of `config.set_defaults` with a custom function that overwrites some of the values at "level 2" in the defaults with the corresponding values at "level 1" in the user options.

For example, if the defaults were:
```
{
  height=0.8,
  vertical = {
    height=0.9,
    width=0.8
  }
}
```
and the user options were:
```
{
  height={padding=2}
}
```
the output would be:
```
{
  height={padding=2},
  vertical = {
    height={padding=2},
    width=0.8
  }
}
```

This system makes more sense to me as it allows users to overwrite options for all layouts more easily, if for example they wanted to define sizes in terms of padding. This fixes the failed test from the previous commit.

Since this is a more _unusual_ way of defining this option, I would appreciate feedback from anyone with thoughts on the matter 🙂 

---

I also changed the order of preference for this part so that it matched the historical one used [here](https://github.com/nvim-telescope/telescope.nvim/blob/557485e793d93d595cd2966dc8cb3205a5633db3/lua/telescope/config.lua#L106) in `first_non_null`.
I think the use of `config.values` here is to allow this `set_defaults` function to be called multiple times to change just a few options without having to input all of them again, but it would be helpful if someone who was involved with Telescope at the time ([September 2020](https://github.com/nvim-telescope/telescope.nvim/commit/11a3c706093f6656ce33dabe0810f96490403818)) could explain it? (@tjdevries?, @Conni2461?)

tjdevries:
I can try and take a look this week. Thanks for keeping on working on it :) I'll try and answer the questions as well.

lua/telescope/config.lua
```diff
@@ -77,44 +89,49 @@ function config.set_defaults(defaults)
     end
   end
 
-  set("sorting_strategy", "descending", [[
+  for key, info in pairs(tele_defaults) do
+    set(key, info[1], info[2])
+  end
+end
+
+telescope_defaults["sorting_strategy"] = { "descending", [[
```
tjdevries, line 128: why did you make this change? I don't remember talking about it (just looking for context)

tjdevries:
I will try and take a loo at the logic for geting the configuration -- I will at least write out how the options are chosen and then we can make sure we agree and write some tests.

lua/telescope/config.lua
l-kershaw: You're right, we didn't discuss this bit.

This was part of the switch I made to allow `config.set_defaults` to be passed in the Telescope defaults as well as the users defaults. If nothing is passed in for Telescope defaults, then we use the table `telescope_defaults`, which contains all of the previous defaults.

This makes it easier to test the inheritance in `layout_strategies_spec.lua` [here](https://github.com/nvim-telescope/telescope.nvim/blob/9699e68379996e0167c9e97feaa3601463eb358a/lua/tests/automated/layout_strategies_spec.lua#L41).
With this set up we can use different defaults for "testing" compared to "production", which makes it a bit easier to write (and understand) the tests.

tjdevries:
I think the tests should not be necessariyl about the "does the config load correctly" but "do we merge the tables".

We can export a function `_merge_config_tables` that doesn't require any global state to check the results.

I will see if I can do anything related to this today.

doc/telescope.txt
tjdevries: Yup :) added. Probably could use some more updates once we're done by just checking to make sure we've included all the options.

tjdevries:
Some stuff I'd like to see done before this PR is merged:
- [ ] Update readme. I want readme to have less in it and point more to the appropriate help sections (e.g. `:help telescope.layouts` or whatever.) This is because we don't auto-generate the readme, but we can autogenerate the docs. And I only want one place of truth for the layout information as much as possible (some stuff is fine to stay cause it gives people an idea of what they _can_ do with telescope in the readme).
- [ ] I think I want to see the `telescope_defaults` look more like this, rather than the current way.
```lua
local telescope_defaults = {
  sorting_strategy = { ... }
}
```
- [ ] I think `preview_cutoff` doesn't make sense the way it is currently. So, `center` should not do cutoff based on `columns`, but rather on `rows`, right?


@l-kershaw what else do you think?

l-kershaw:
Re: your first comment
- I'm not sure that checking if we merged the config tables correctly is sufficient. We should check if we are correctly interpreting the user provided `override` (the config passed when the Telescope is called, after it has been `setup`).
We can add some tests for checking the merging explicitly as well if we want, although I'm not sure it is necessary as this is sort of inherently checked with the current tests.

Re: your second comment
- I agree with removing some stuff from the readme and pointing to the docs instead. This will definitely make it easier for us to update stuff with less confusion in future.
- I'm a bit confused by your second point. I _think_ you mean that we should have the default `layout_config` separated by `layout_strategy`, which is definitely something we can do in the scope of this PR.
If this isn't what you mean, could you please clarify?
- Regarding `preview_cutoff` we could change this to a resolvable setting where users could input things like `{columns=120}`, `{rows=50}` or just a function that takes as inputs the maximum number of columns and rows and returns a boolean. We could even allow just a number and keep the current functionality for columns if we wanted to implement this change with minimal breaking of configs (although this would add complexity that I'm not sure is necessary).
This would then allow us to set the `center` layout to have a cutoff in terms of rows, for example.

I can't think of any additional aims for this PR atm, but happy to discuss if anyone else has suggestions.

l-kershaw:
I have done a first pass on implementing the ability to resolve `preview_cutoff`, which allows us to have cutoffs in terms of the number of lines in the window and things like that.

@tjdevries could you have a look to see if this is the sort of thing that you wanted?

tjdevries:
So for preview cutoff, I was thinking we could just have different help text for vertical vs. horizontal vs. center (etc.). It would just say that it does rows in one and columns in the other. I don't think we need to make a completely new style of resolver or anythinga

tjdevries:
(I'll respond to other comments later)

Conni2461:
Default theme in setup was suggested yesterday after i merged picker configuration. See https://github.com/nvim-telescope/telescope.nvim/pull/883#issuecomment-857956727

Could be part of this PR as well. Thoughts?

l-kershaw:
Not sure why the linter is complaining as I can't find any lines that are too long in the file specified 🤷 

---

@tjdevries I've reverted the resolve setup and switched to comparing for lines with the `center` layout (and added a check for lines with the `vertical` layout too`). Is this more what you were thinking?

@Conni2461 I like the idea, but I'm not sure if it should be part of this PR. It doesn't rely on any of the new stuff here, or vice versa, so IMO it should be a separate PR. If people have strong preferences for it being in this one we can do, I'm just a bit concerned about scope creep with this PR, as it's already quite big 😅

Conni2461:
You can rebase in ~5 min again. Then linting should be fixed. I merged something with a lint error (i missed it because of docgen) but the same person had another PR open that was almost done and linting will be addressed there. I will merge it after i have tested it.

Yeah separate PR for that theme thing is probably better. We wanted smaller PRs now anyway. :rofl:

l-kershaw:
I can't figure out why, but the help tags between `action_state.get_current_picker()` and `resolver.win_option` don't seem to be working properly. Neovim doesn't seem to think they exist, despite items on either side of them in `doc/telescope.txt` working fine.

Possibly something to do with running out of space on the `action_state.get_current_picker()` line???
I'm not familiar with all the nitty gritty of how vim help files are processed, so not really sure.

Any ideas?

tjdevries:
Let's do theme in a different PR -- It shouldn't be a breaking change, right? So we can do that part later? I don't want to add anything else to this if we can help it.

tjdevries:
OK, I think this is very very close to merging -- but I don't want to merge before I'm gone for a week in case something goes weird. So maybe we wait until after that and then we can merge when I'll be around

Conni2461:
replaced by #922 

Thanks for your work @l-kershaw

Issue #672, opened 2021-03-19
Add method of removing the selected entry from the picker


itmecho:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Similar to the `get_sselected_entry()` action, it would be great to have a `remove_selected_entry()` action which removes the selected entry from the list. 

**Describe alternatives you've considered**
I thought about closing telescope and then rerunning the `buffers` command to get the new list but that seems more complex than it needs to be.

**Additional context**
I want to add a mapping to the builtin `buffers` command so that I can close buffers from telescope. I've got the following which works but the buffer stays in the telescope list after it's been deleted.

```lua
M.buffers = function()
    tb.buffers(
        {
            attach_mappings = function(_, map)
                map(
                    "i",
                    "<c-d>",
                    function(_)
                        local selection = require("telescope.actions").get_selected_entry()
                        if pcall(vim.api.nvim_buf_delete, selection.bufnr, {}) then
                            print(string.format("Deleted buffer %d: %s", selection.bufnr, selection.filename))
                            -- Delete buffer here
                        end
                    end
                )
                return true
            end
        }
    )
end
```

Issue #621, opened 2021-03-06
Question: closing buffers on the buffers picker


dimas-cyriaco:
Sorry if this is not the correct place for asking question, but, is there a mapping for closing the selected buffer on the builtin buffers picker? If not, how could I create one?

Thank you for the awesome plugin!

Conni2461:
No there is no such mapping yet but you can inject your own mapping. Something like this should be possible:

```lua
local action_state = require('telescope.actions.state')
local actions = require('telescope

local m = {}
m.my_buffer = function()
  require('telescope.builtin').buffers{
    attach_mappings = function(prompt_bufnr, map)
      local delete_buf = function()
        local selection = action_state.get_current_selection()
        -- depending if you want to close or not, include this or not
        actions.close(prompt_bufnr)
        -- print(vim.inspect(selection))
        -- better print selection before first running this. I am not sure if it have a bufnr or if this field is named differently
        vim.api.nvim_buf_delete(selection.bufnr, { force = true })
      end

      -- mode, key, func
      -- this is just an example
      map('n', 'd', delete_buf)
    end
  }
end
return m
```

If more people find this a useful action we can include in core, but i am not sure about default mapping for it. Because its so easy to just map something yourself

Note: No idea if this works or not. Its written inside github :rofl:

oberblastmeister:
I actually think that this is pretty useful

Conni2461:
We would need a refresh for picker that allows us to take out only one element. Then we can get rid of `close`

dimascyriaco:
Sorry, but it's not clear to me where do I put this code, and how do I tell Telescope to use this `my_buffer` function.

I tried to put this simplified version directly on `init.lua` but it does not look to be the correct way, since it behaves strangely:

```lua
  require'telescope.builtin'.buffers{ attach_mappings = function(_, map)
    map('i', '<c-u>', function()
      print('inside test function!')
    end)

    return true
  end}
```

It seams that this code just tells telescope to open this picker (as it automatically opens when I resource the file, and the mapping works) but if I open the picker with `:Telescope buffers`, the mapping does not work.

What am I missing?

Conni2461:
Its not a configuration its a wrapper around our function. So the way you do this, is you put this function into a lua module (own file) put this file into `lua/dimascyriaco/telescope.lua` and then you can map `require('dimascyriaco/telescope').my_buffers()`

dimascyriaco:
[...]

PR #828, opened 2021-05-09
feat: delete entry from the picker without closing telescope

dhruvmanila:
This will allow users to delete entries  from the picker without closing telescope.

Discussion ref: https://github.com/nvim-telescope/telescope.nvim/pull/803#pullrequestreview-654985978
fixes: #689 
fixes: #621 
fixes: #672  

This function in and of itself will just remove the entry, what I'm proposing next is:
Idea: Should we provide a callback parameter which will be called with the removed entry, so that the user or the specific picker can do exactly what needs to be done outside of telescope? Eg., delete the buffer, delete the file, etc.

The selection strategy is temporarily set to "row" as that will help keep the selection in the same position instead of resetting to the top. If there is any other way to achieve this, do suggest :)

Thanks @Conni2461 for helping!

https://user-images.githubusercontent.com/67177269/117562588-a9434e00-b0bd-11eb-9181-275e82cce317.mov

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -497,7 +497,7 @@ end
 --- selections made using multi-select. It can be used to define actions
 --- such as deleting buffers or files.
 ---
---- Example usage: >
+--- Example usage:
 --- <pre>
 --- actions.delete_something = function(prompt_bufnr)
 ---    local current_picker = action_state.get_current_picker(prompt_bufnr)
@@ -506,7 +506,9 @@ end
 ---    end)
 --- end
 --- </pre>
---- <
+---
+--- Example usage in telescope:
+---   - `actions.delete_buffer()`
 ---@param delete_cb function: called with each deleted selection
 function Picker:delete_selection(delete_cb)
   vim.validate { delete_cb = { delete_cb, "f" } }
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -491,17 +491,19 @@ function Picker:hide_preview()
   -- 2. Resize prompt & results windows accordingly
 end
 
--- A simple interface to remove an entry from the results window without closing
--- telescope.
--- @param delete_cb function
+---A simple interface to remove an entry from the results window without closing
+---telescope. The `delete_cb` function will be called with each deleted selection.
+---@param delete_cb function
 function Picker:delete_selection(delete_cb)
+  vim.validate { delete_cb = { delete_cb, "f" } }
   local original_selection_strategy = self.selection_strategy
   self.selection_strategy = "row"
-  delete_cb = (delete_cb and type(delete_cb) == "function") and delete_cb or function() end
 
   local delete_selections = self._multi:get()
+  local used_multi_select = true
   if vim.tbl_isempty(delete_selections) then
     table.insert(delete_selections, self:get_selection())
+    used_multi_select = false
   end
 
   local selection_index = {}
@@ -519,7 +521,7 @@ function Picker:delete_selection(delete_cb)
     delete_cb(selection)
   end
 
-  if #delete_selections > 1 then
+  if used_multi_select then
     self._multi = MultiSelect:new()
   end
 
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -498,9 +498,8 @@ function Picker:delete_selection()
   local original_selection_strategy = self.selection_strategy
   self.selection_strategy = "row"
 
-  local row = self:get_selection_row()
-  local index = self:get_index(row)
-  table.remove(self.finder.results, index)
+  local selection = self:get_selection()
+  table.remove(self.finder.results, selection.index)
   self:__on_lines(nil, nil, nil, 0, 1)
 
   vim.schedule(function()
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -491,9 +491,23 @@ function Picker:hide_preview()
   -- 2. Resize prompt & results windows accordingly
 end
 
----A simple interface to remove an entry from the results window without closing
----telescope. The `delete_cb` function will be called with each deleted selection.
----@param delete_cb function
+-- TODO: update multi-select with the correct tag name when available
+--- A simple interface to remove an entry from the results window without
+--- closing telescope. This either deletes the current selection or all the
+--- selections made using multi-select. It can be used to define actions
+--- such as deleting buffers or files.
+---
+--- Example usage: >
+--- <pre>
+--- actions.delete_something = function(prompt_bufnr)
+---    local current_picker = action_state.get_current_picker(prompt_bufnr)
+---    current_picker:delete_selection(function(selection)
+---      -- delete the selection outside of telescope
+---    end)
+--- end
+--- </pre>
+--- <
+---@param delete_cb function: called with each deleted selection
 function Picker:delete_selection(delete_cb)
   vim.validate { delete_cb = { delete_cb, "f" } }
   local original_selection_strategy = self.selection_strategy
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -491,24 +491,28 @@ function Picker:hide_preview()
   -- 2. Resize prompt & results windows accordingly
 end
 
--- WIP
--- TODO(dhruvmanila): update the selection when we're at the last row or the
--- first depending on the sorting_strategy
-function Picker:delete_selection()
+-- A simple interface to remove an entry from the results window without closing
+-- telescope.
+-- @param delete_cb function
+function Picker:delete_selection(delete_cb)
   local original_selection_strategy = self.selection_strategy
   self.selection_strategy = "row"
 
-  local actual_index
+  local selection_index
   local selection = self:get_selection()
   for index, entry in ipairs(self.finder.results) do
-    if entry.index == selection.index then
-      actual_index = index
+    if entry == selection then
+      selection_index = index
       break
     end
   end
 
-  table.remove(self.finder.results, actual_index)
+  table.remove(self.finder.results, selection_index)
   self:__on_lines(nil, nil, nil, 0, 1)
+
+  if delete_cb and type(delete_cb) == "function" then
+    delete_cb(selection)
+  end
 
   vim.schedule(function()
     self.selection_strategy = original_selection_strategy
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -497,23 +497,33 @@ end
 function Picker:delete_selection(delete_cb)
   local original_selection_strategy = self.selection_strategy
   self.selection_strategy = "row"
+  delete_cb = (delete_cb and type(delete_cb) == "function") and delete_cb or function() end
 
-  local selection_index
-  local selection = self:get_selection()
-  for index, entry in ipairs(self.finder.results) do
-    if entry == selection then
-      selection_index = index
-      break
+  local delete_selections = self._multi:get()
+  if vim.tbl_isempty(delete_selections) then
+    table.insert(delete_selections, self:get_selection())
+  end
+
+  local selection_index = {}
+  for result_index, result_entry in ipairs(self.finder.results) do
+    if vim.tbl_contains(delete_selections, result_entry) then
+      table.insert(selection_index, result_index)
     end
   end
 
-  table.remove(self.finder.results, selection_index)
-  self:refresh()
-
-  if delete_cb and type(delete_cb) == "function" then
+  -- Sort in reverse order as removing an entry from the table shifts down the
+  -- other elements to close the hole.
+  table.sort(selection_index, function(x, y) return x > y end)
+  for _, index in ipairs(selection_index) do
+    local selection = table.remove(self.finder.results, index)
     delete_cb(selection)
   end
 
+  if #delete_selections > 1 then
+    self._multi = MultiSelect:new()
+  end
+
+  self:refresh()
   vim.schedule(function()
     self.selection_strategy = original_selection_strategy
   end)
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -508,7 +508,7 @@ function Picker:delete_selection(delete_cb)
   end
 
   table.remove(self.finder.results, selection_index)
-  self:__on_lines(nil, nil, nil, 0, 1)
+  self:refresh()
 
   if delete_cb and type(delete_cb) == "function" then
     delete_cb(selection)
@@ -667,10 +667,10 @@ function Picker:refresh(finder, opts)
   end
   if opts.reset_prompt then self:reset_prompt() end
 
-  self.finder:close()
   if finder then
-      self.finder = finder
-      self._multi = MultiSelect:new()
+    self.finder:close()
+    self.finder = finder
+    self._multi = MultiSelect:new()
   end
 
   self.__on_lines(nil, nil, nil, 0, 1)
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -498,8 +498,16 @@ function Picker:delete_selection()
   local original_selection_strategy = self.selection_strategy
   self.selection_strategy = "row"
 
+  local actual_index
   local selection = self:get_selection()
-  table.remove(self.finder.results, selection.index)
+  for index, entry in ipairs(self.finder.results) do
+    if entry.index == selection.index then
+      actual_index = index
+      break
+    end
+  end
+
+  table.remove(self.finder.results, actual_index)
   self:__on_lines(nil, nil, nil, 0, 1)
 
   vim.schedule(function()
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -491,6 +491,23 @@ function Picker:hide_preview()
   -- 2. Resize prompt & results windows accordingly
 end
 
+-- WIP
+-- TODO(dhruvmanila): update the selection when we're at the last row or the
+-- first depending on the sorting_strategy
+function Picker:delete_selection()
+  local original_selection_strategy = self.selection_strategy
+  self.selection_strategy = "row"
+
+  local row = self:get_selection_row()
+  local index = self:get_index(row)
+  table.remove(self.finder.results, index)
+  self:__on_lines(nil, nil, nil, 0, 1)
+
+  vim.schedule(function()
+    self.selection_strategy = original_selection_strategy
+  end)
+end
+
 
 function Picker.close_windows(status)
   local prompt_win = status.prompt_win
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -369,6 +369,15 @@ actions.open_loclist()                                *actions.open_loclist()*
 
 
 
+actions.delete_buffer({prompt_bufnr})                *actions.delete_buffer()*
+    Delete the selected buffer or all the buffers selected using multi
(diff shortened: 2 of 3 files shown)


dhruvmanila:
There is this problem:
It seems to be breaking sometimes when deleting the entries in the middle of a search. I will take a look at it soon

dhruvmanila:
No problem, take your time and thanks for taking a look at it :)

I am using refs to find out the index of the selection in the results table and also provided an option `delete_cb` which will be called with the deleted selection. With this simple interface we can now define actions like:

```lua
    vim.api.nvim_buf_delete(selection.bufnr, { force = true })
  end)
end

-- could do the same for file_browser where we delete a file/directory
```

Also, we can think of doing multiple deletes using multi-select once this is well tested.

tom-anders:
This looks really cool! I recently implemented something similar for my [telescope-vim-bookmarks extension](https://github.com/tom-anders/telescope-vim-bookmarks.nvim) where I have an action that either delete all currently multi-selected bookmarks or the bookmark under the cursor (if there's no multi selection). There I used the `action.enhace( post = ... )` interface to just refresh the picker after deleting the bookmarks.

But this here seems much more convenient, looking forward to it being merged, especially with multi-selection support.

tjdevries:
can we use `self:refresh(...)` inside of this call to re-use some of the code between these two items?

dhruvmanila:
I believe it could be used but it closes the finder here: https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/pickers.lua#L642

On that note, why are we closing the finder if the new finder is optional?

tjdevries:
> Edit: closing the finder only if the new finder is given will give us the ability to use the function `self:refresh()` instead of `self:__on_lines(...)`

Yeah, we can probably make that change here if you want.

I'd prefer to re-use refresh (we can extend refresh in whatever ways we need to make that happen)

dhruvmanila:
I am also not liking the way of finding the selection entry in the results, as mentioned [here](https://github.com/nvim-telescope/telescope.nvim/pull/828#issuecomment-836547929) I tried a bunch of ways but comparing it directly in a for loop seems the only way to get the correct index. It's fine for single selection but in case of multi-selection, we would have to do compare each result entry with every selection from multi-select.

Is there any possibility where we add a special field for each entry in the result table identifying its position like `result_idx` or something like that? This would make it extremely easy to remove the entry.

Conni2461:
> I believe it could be used but it closes the finder here

Just bad programming by me :rofl: 

Sorry that i haven't looked at it earlier, it was a crazy week and then i wanted to look at it Saturday and the whole `gitsigns.nvim` + `fzf-native` ffi fiasco started. :)

Great work here. I just tested it. (Also i thought we could change the `self.selection_strategy` back with `completion_callback`, but i think do lines can run multiple times last time i check, so it might not work. Like being able to set a special `completion_callback` that gets poped after running it ones. Idk)

Multi selection seems like a good idea, (maybe as option that can be enabled/disabled or as section function). Maybe someone has a use case in the future where that person doesn't want it to be the whole multi selection.

dhruvmanila:
I tested with `completion_callback` and it works although it gets added every time I call `delete_selection` and thus the callbacks keep increasing, currently there's no way to remove the called function. 

We could reset the callback table, but then what about any existing implementation, extensions and user configs which are using this?

Something like this:
```lua
{ <function 1> }
{ <function 1>, <function 2> }
{ <function 1>, <function 2>, <function 3> }
{ <function 1>, <function 2>, <function 3>, <function 4> }
{ <function 1>, <function 2>, <function 3>, <function 4>, <function 5> }
```

Also, I am brainstorming a bit on how to do multi-selection and here's what I'm trying:
- Get the selections from `self._multi:get()`, if it is empty then make a table of the current entry
- Loop through the results and store the result index for the corresponding selection in a table
- Sort the result index table in reverse order and start removing them (reverse because removing an element shifts the other elements to close the hole)
- If selections were from multi select then reset `self._multi` it
- Refresh
- Reset the `selection_strategy`

lua/telescope/pickers.lua
```diff
@@ -491,6 +491,44 @@ function Picker:hide_preview()
   -- 2. Resize prompt & results windows accordingly
 end
 
+-- A simple interface to remove an entry from the results window without closing
```
Conni2461, line 494: Emmy documentation uses 3 slashes 

https://github.com/tjdevries/tree-sitter-lua/blob/master/HOWTO.md#parameter

Maybe at some point we can export the documentation for `pickers.lua`

lua/telescope/pickers.lua
```diff
@@ -491,6 +491,44 @@ function Picker:hide_preview()
...
+-- telescope.
+-- @param delete_cb function
+function Picker:delete_selection(delete_cb)
+  local original_selection_strategy = self.selection_strategy
+  self.selection_strategy = "row"
+  delete_cb = (delete_cb and type(delete_cb) == "function") and delete_cb or function() end
```
Conni2461, line 500: I think we can just make the cb mandatory and error if none is given

lua/telescope/pickers.lua
```diff
@@ -491,6 +491,44 @@ function Picker:hide_preview()
...
+
+  -- Sort in reverse order as removing an entry from the table shifts down the
+  -- other elements to close the hole.
+  table.sort(selection_index, function(x, y) return x > y end)
+  for _, index in ipairs(selection_index) do
+    local selection = table.remove(self.finder.results, index)
+    delete_cb(selection)
+  end
+
+  if #delete_selections > 1 then
```
Conni2461, line 522: What if the multi selection was only one? `self._multi:get()` actually returned something and we need to create a new one, right? Maybe just set a flag if you insert into `delete_selections`? Might be wrong tho

Conni2461:
> I tested with completion_callback and it works although it gets added every time I call delete_selection and thus the callbacks keep increasing, currently there's no way to remove the called function.

Yeah i meant adding something that would pop the function after executing but we can just keep `vim.schedule` for now. 

Do you want to integrate both your prs together? or whats your plan?

lua/telescope/pickers.lua
dhruvmanila: Ah yes, that is correct. Thanks for pointing it out. I will use a flag to see whether we used `multi_selection` or not and reset it accordingly.

dhruvmanila:
> Do you want to integrate both your prs together? or whats your plan?

Yeah, I will close my other PR and add the same here.

lua/telescope/pickers.lua
```diff
@@ -491,6 +491,60 @@ function Picker:hide_preview()
   -- 2. Resize prompt & results windows accordingly
 end
 
+-- TODO: update multi-select with the correct tag name when available
+--- A simple interface to remove an entry from the results window without
+--- closing telescope. This either deletes the current selection or all the
+--- selections made using multi-select. It can be used to define actions
+--- such as deleting buffers or files.
+---
+--- Example usage: >
```
dhruvmanila, line 500: Do code blocks work? In vim help it is denoted with enclosing angle brackets: `> ... <`

elianiva: You can use `<pre>` tags for code blocks

dhruvmanila: Ah, no what I mean is does it highlight in the `help` buffer. [This previewers code block](https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/previewers/init.lua#L75) does not highlight.

elianiva: yeah, apparently they don't..

Conni2461 (approved):
LGTM

lua/telescope/pickers.lua
Conni2461: `<pre>` only disables the markdown formatter.

:thinking: Maybe i should add that to `docgen` :thinking:

dhruvmanila:
Ugh, first rebased on the wrong master, updated to the latest one, should be good to go.

lua/telescope/pickers.lua
dhruvmanila: Yeah, that would be good. The highlights will help separate out the code, maybe `<code>`?

Conni2461:
Okay this is really good work. I just tested it and i like it. Thank you very much :)

One thing left to ask is: Is it the right decision to not map it with `attach_mappings` to `builtin.buffers`? In my opinion yes but we should ask other core member as well.

CC @clason @tami5 @elianiva

clason:
I think this is a sufficiently niche (but useful!) function that not mapping it by default is OK.

(On the other hand, since this doesn't actually _do_ anything to the buffers etc., it'd be safe to map it -- I'd be more careful with a mapping that could actually delete files ;))

But in general my opinion is that workflows and personal preferences vary so wildly that default mappings annoy at least as many people as they help; and if you set them deliberately, you are more aware of what is going on.

tami5:
> Okay this is really good work. I just tested it and i like it. Thank you very much :)
> 
> One thing left to ask is: Is it the right decision to not map it with `attach_mappings` to `builtin.buffers`? In my opinion yes but we should ask other core member as well.
> 
> CC @clason @tami5 @elianiva

+1 I agree it should be mapped by default as it sensible and basic action to have in buffer pickers.

dhruvmanila:
Appreciate everyone's thoughts on this!

Personally, I use `delete_buffers` extensively and so I remapped it to `<C-x>` and remapped `<C-s>` to horizontal split, but that's again my own preference.

At first I thought of not to remap it thinking in the same line as @clason:
> But in general my opinion is that workflows and personal preferences vary so wildly that default mappings annoy at least as many people as they help; and if you set them deliberately, you are more aware of what is going on.

But, I also have seen multiple instances where users have requested for this feature to be implemented:
- #621 
- #672
- #689
- #873 (already closed)

I have linked those issues to be closed with this PR but feel free to exclude them as necessary.

At the end, I do think we should implement the default keybinding for it and the default one should be seems plausible and intuitive. Some ideas:
- `<M-w>` / `<C-S-w>` (recommened [here](https://github.com/nvim-telescope/telescope.nvim/pull/803#issuecomment-828195798) by @tami5)
- `<C-r>` for remove?

clason:
Let me reiterate (one last time) my suggestion to _not_ add a default binding: Anything you pick has a good chance to conflict with a _different_ user binding (for example, what happens if you map it to `<c-r>`, but I have already mapped this to -- fictitious example -- "reload"?) This will trigger sudden errors and require me to add a non-default mapping for something I don't want to use at all.

(This is a generic comment, not related specifically to this action.)

dhruvmanila:
Sorry for the misunderstanding and thank you for the clarification.

> And are you sure that these comments specifically request a default mapping for that feature (and not just the feature itself)? ;)

Ah, right, no they only specify the feature :)

I understand the reason for not providing a default mapping and I'm fine with that. This PR should be good to go then?

smarquez1:
What about an example of how to add a mapping for this in the README? That'd be usefull for new users (and even experienced ones like me) to easily enable it

clason:
of course, the possible function to map (with an example mapping) should be documented

Conni2461:
What about realizing https://github.com/nvim-telescope/telescope.nvim/issues/582 now? To make everything simpler for end users and we can be confident in not mapping things. 

I just did a super simple prototype that works, missing theme and mappings. Can anyone think of a better interface than tamis proposal? (Also i am not sure about the `- new custom picker` thingy but should be doable as well)

Conni2461:
Re: I think this would make the experience of setting a new action for a specific picker pretty good: https://github.com/nvim-telescope/telescope.nvim/pull/883

dhruvmanila:
> of course, the possible function to map (with an example mapping) should be documented

Yeah that sounds good. Should I add the documentation to the `:help` docs as I don't see any reference to actions in the README? Also, with the changes from #883, should we wait until that merges?

> Re: I think this would make the experience of setting a new action for a specific picker pretty good: #883

This is really good work and would make configuring Telescope simple and intuitive!

Conni2461:
I added documentation for #883. I would ask tj what he thinks and then we can merge yours either before mine or after. Doesn't really matter but this way we only really need to document the new way :)

dhruvmanila:
Sounds good to me 👍

Conni2461:
I'll test it with my configuration PR that i just merged and then i'll merge this and update the comment on how to set it up.

dhruvmanila:
Awesome, thanks everyone for their guidance and feedback! :)

Now, onto formatting the config in the new way! :D

mroavi:
Hi @dhruvmanila. Thank you for this, it's awesome! I was really missing a nice way of deleting buffers.

There is only one minor thing that is a bit annoying: if you delete the last entry, then the selection moves back to the first entry, while it would be much better if it stayed in the last. Should I open an issue for this?

fdschmidt93:
Not needed, you can do something like this (just tested it, works)

```lua
require("telescope").setup {
  pickers = {
    buffers = {
      show_all_buffers = true,
      mappings = {
        i = {
          ["<c-d>"] = actions.delete_buffer + actions.move_to_top,
        }
      }
    }
  }
}
```

mroavi:
I think you misunderstood. The default behavior is great: when you delete an entry, the selection stays where it is. The problem is that this rule is broken whenever you delete the last entry in the menu. Currently, it jumps back to the first entry while (in my opinion) it should stay in the last.

l-kershaw:
@mroavi If I understand what you want correctly, you can achieve this by setting your `scroll_strategy` to `limit`, although this then affects your ability to cycle from one end of the list to another.

mroavi:
@l-kershaw you are completely right. Didn't know about the possibility to configure scroll behavior. Please ignore my suggestion.

l-kershaw:
No worries 🙂

fdschmidt93:
Yes, on second thought what I suggested is pretty off :laughing: cool you've got it solved!

ssh352:
> I'll test it with my configuration PR that i just merged and then i'll merge this and update the comment on how to set it up.
> 
> WORKS Example config:
> 
> ```lua
> require("telescope").setup {
>   pickers = {
>     buffers = {
>       show_all_buffers = true,
>       sort_lastused = true,
>       theme = "dropdown",
>       previewer = false,
>       mappings = {
>         i = {
>           ["<c-d>"] = "delete_buffer",
>         }
>       }
>     }
>   }
> }
> ```

Thanks, this however doesn't delete the current buffer even if you select to delete it.

Kadrian:
For info, I was happy to find that mapping this to `normal mode` + `d` works as well

```
mappings = {
        n = {
          ["d"] = "delete_buffer",
        }
      }
```

sleewoo:
this check added after feature landing seems breaking stuff, at least for `find_files` and `live_grep`

https://github.com/nvim-telescope/telescope.nvim/blob/2d92125620417fbea82ec30303823e3cd69e90e8/lua/telescope/actions/init.lua#L1103

Screenshot_20230823_201253

copied the function to my local setup, commented callback content and works well

Screenshot_20230823_202627

jamestrew:
@sleewoo I think you're probably trying to delete an entry that doesn't exist as a buffer. The `delete_buffer` action isn't really designed to work with pickers like `find_files` and `live_grep` since unless you've opened nearly every file in the cwd, most of the entry result from those pickers won't exist as buffers.

This action is probably best used with something like the `buffers` picker which lists all open buffers.  

Probably could do with better error handling but considering the purpose of this action and the way you're using it, I think this is low-priority. 

Also, by commenting out the contents of the callback function, you've broken the intended use case of the action -- deleting an open buffer.

sleewoo:
@jamestrew yes, reverted my dirty hack as realized i try to delete buffers that does not exists.
what i was trying to do is out of scope for telescope.
thanks.

yuhsienchiang:
Hello @dhruvmanila, 
I'm quite new to neovim plugin implementation and lua, and recently I am implementing my own telescope extension for toggleterm. I ran into an issue and couldn't solve it.  Could you give my some guides on how to keep the telescope open when deleting the terminal buffer?

My extension is to list out all the terminal buffers created using toggleterm.  By pressing <CR>, the selected terminal buffer(s) will be opened or focused.  This part works fine.

I also added a keymap <C-d> to remove(closed) the terminals.  The terminals can be removed(closed) successfully, however, the telescope window we be closed every time I press <C-d> to perform the action.  I would like the remove action performs just like the actions.delete_buffer, which will not close telescope.

The code below is my implementation.  I wrote my code based on your implementation of actions.delete_buffer, but still didn't get the desired result.  Did I miss anything or this is a limitation for handling terminal buffers?

Thank you so much!

```
local terms = require("toggleterm.terminal")
local pickers = require("telescope.pickers")
local finders = require("telescope.finders")
local conf = require("telescope.config").values

local M = {}

M.term_select = function(opts)
	opts = opts or {}
	local terminals = terms.get_all(true)
	pickers.new(opts, {
		prompt_title = "Toggleterm",
		finder = finders.new_table({
		    results = terminals,
		    entry_maker = function(entry)
			return {
			    value = entry,
			    text = tostring(entry.bufnr),
			    display = tostring(entry.id .. ": " .. entry:_display_name()),
			    ordinal = tostring(entry.id),
			}
		    end,
		}),

                    -- close the prompt window
		    actions.close(prompt_bufnr)

		    -- get the selected terminal, which is the table returned by entry_maker
		    if selection == nil then
			return
		    end

		    -- perform action
		    if selection.value:is_open() then
			selection.value:focus()
		    else
			selection.value:open()
		    end
		end)

		map({ "i", "n" }, "<C-d>", function()
			local ok = pcall(vim.api.nvim_buf_delete, selection.value.bufnr, { force = true })
			return ok
						
		    end)
		end)

		return true
	    end,
	}):find()
end

return M

```

ryanmsnyder:
@yuhsienchiang I'm having the same issue. Were you able to find a solution?

ryanmsnyder:
I've confirmed the issue of telescope closing after a terminal buffer is deleted (using `delete_selection` with a callback) is specific to terminal buffers.

I've eliminated some possibilities for this problem. The issue doesn't seem to be in the `delete_selection` method directly (i.e. it isn't related to telescope [removing the entry from the finder](https://github.com/nvim-telescope/telescope.nvim/blob/74ce793a60759e3db0d265174f137fb627430355/lua/telescope/pickers.lua#L815)). It definitely _is_ related to the actual deletion of the buffer that's provided in the callback to `delete_selection` - it's a result of calling `vim.api.nvim_buf_delete`. However, I haven't been able to figure out why deleting a terminal buffer causes telescope to close.

I also tried to determine if this is related to `toggleterm`. My thinking was that maybe there was some post-processing/cleanup from `toggleterm` after a terminal buffer was deleted that was causing the focus to be removed from telescope and, therefore, closing telescope. I was hopeful when I found that `toggleterm` [listens for the `TermClose` event](https://github.com/akinsho/toggleterm.nvim/blob/c80844fd52ba76f48fabf83e2b9f9b93273f418d/lua/toggleterm/terminal.lua#L154), which is fired by neovim when a terminal buffer is deleted. However, this didn't appear to be the culprit either.

I then assumed it had to do with neovim's handling of the deletion of terminal buffers. To test that assumption, I opened a terminal buffer by running the `:terminal` command (bypassing `toggleterm`). After retrieving the terminal buffer's `bufnr` (`:echo bufnr('')`), I opened toggleterm and switched to normal mode. With the telescope window still open, I deleted the terminal buffer (`:lua vim.api.nvim_buf_delete(term_bufnr, { force = true })`). I was surprised that the telescope window remained open.

It must have to do with `toggleterm` but I haven't been able to figure out what could be causing it. 

What makes it more difficult to diagnose is that telescope isn't _always_ closed after deleting a terminal buffer that was created with `toggleterm`. There are some scenarios where it doesn't close it but I haven't been able to identify a useful pattern.

PR #831, opened 2021-05-09
fix: Cleanup on picker close

### LINKED ISSUE #822: Telescope doesn't clean up properly on <c-c>
### Description

It seems that the Telescope buffer isn't cleaned up properly on `<c-c>` in a buffer with `nobuflisted` and `hidden` set. No preview window must be shown, since it isn't using `TelescopeNormal`.

**Expected Behaviour**

Telescope should clean up properly on `<c-c>` and bring you back to the same buffer where `:Telescope` was executed from.

**Actual Behavior**

You end up up in a different buffer (8 instead of 1 following the test.vim) which obviously belonged to Telescope, since the background is still green (because `'winhl'`). And the original buffer (1) was deleted.

### Details

0. I did not find a way to disable the preview in the docs, so make sure to use a small enough window...
1. nvim -nu test.vim
2. `:set nobuflisted`
3. `:Telescope`
4. `<c-c>`
</details>

- nvim --version output: `NVIM v0.5.0-dev+1307-g17434b88b`
- Operating system: `macOS 11.3`
- Telescope commit: `88f7b27`

</details>

<p>

```viml

endif

source /tmp/plug.vim
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
call plug#end()

highlight TelescopeNormal guibg=green

" using Telescope default settings otherwise
```

l-kershaw:
I think this fixes #822 and other related cleanup issues.
_(Note: I haven't tested all the setups in the related issues, but it seems to fix all the issues I have encountered.)_

---

**What I think the problem was:**
- On closing the picker, the autocommand defined [here](https://github.com/nvim-telescope/telescope.nvim/blob/e2907fc0f225a7bf33ace6915bc6b55ed1e10b31/lua/telescope/pickers.lua#L451) would be activated.
- This calls `pickers.on_close_prompt`, which calls `Picker.close_windows`
- In layouts without a preview, `status.preview_win` would be nil.
- `del_win` would be called for the preview window, using the nil `win_id`.
- The check for valid win_id [here](https://github.com/nvim-telescope/telescope.nvim/blob/e2907fc0f225a7bf33ace6915bc6b55ed1e10b31/lua/telescope/pickers.lua#L505) would still pass (not really sure why it's set up this way).
- `bufnr` would then be set to the current buffer number (the one that telescope was opened from), and this buffer would be deleted.

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -502,7 +502,7 @@ function Picker.close_windows(status)
   local preview_border_win = status.preview_border_win
 
   local function del_win(name, win_id, force, bdelete)
-    if not vim.api.nvim_win_is_valid(win_id) then
+    if win_id == nil or not vim.api.nvim_win_is_valid(win_id) then
       return
     end
 


Conni2461:
CC @mhinz

Conni2461:
Okay i tested it. It works for me so i am inclined to just merge right now. It potentially fixes a lot of things.
So good job. Thank you very much :)

mhinz:
Yup, this fixes it for me. Thank you!

PR #800, opened 2021-04-27
feat: add git_stash picker

amirrezaask:
Adds git_stash picker

Resolves https://github.com/nvim-telescope/telescope.nvim/issues/799

PR #836, opened 2021-05-11
add git_stash field to builtin

saadparwaiz1:
PR #800 add git\_stash picker. However, it's not added as a field in builtin.


Conni2461:
Oops missed that :rofl: 

Thanks for adding it :)

PR #839, opened 2021-05-13
Consistent filepath display and code cleanup.

caojoshua:
## Changes
1. Default overridable values for `hide_filename`, `shorten_path`, and `tail_path`
2. Consistent use of aforementioned values across entry makers
3. extract copy pasted path transformations into a common utility function

### Caveats
* This will probably change filepath display for some pickers for many users.
* I tried some basic manual testing for many of the affected pickers, but may have missed some

### Open Questions
* I am not sure how to deal with some of the existing pickers that have default values for `shorten_path` and `tail_path`
  * ie. lsp references sets `shorten_path` to true by default
  * IMO, these pickers should instead follow global values, and the user can just pass in options to override defaults as needed

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -114,13 +114,15 @@ telescope.setup({opts})                                    *telescope.setup()*
           Into your config.
 
 
-        otherwise, just set the mapping to the function that you want it to be.
+        otherwise, just set the mapping to the function that you want it to
+        be.
 
             ...,
             ["<C-i>"] = require('telescope.actions').select_default,
             ...,
 
-        If the function you want is part of `telescope.actions`, then you can simply give a string.
+        If the function you want is part of `telescope.actions`, then you can
+        simply give a string.
           For example, the previous option is equivalent to:
 
             ...,
@@ -135,6 +137,21 @@ telescope.setup({opts})                                    *telescope.setup()*
             ["kk"] = { "<cmd>echo \"Hello, World!\"<cr>", type = "command" },)
             ...,
 
+
+                                          *telescope.defaults.path_display*
+    path_display: ~
+        Determines how file paths are displayed
+
+        path_display can be set to an array with a combination of:
+        - "hidden"    hide file names
+        - "tail"      only display the file name, and not the path
+        - "absolute"  display absolute paths
+        - "shorten"   only display the first character of each directory in
+                      the path
+
+        path_display can also be set to 'hidden' string to hide file names
+
+        Default: {}
 
                                          *telescope.defaults.prompt_prefix*
     prompt_prefix: ~
@@ -728,14 +745,9 @@ builtin.tags({opts})                                          *builtin.tags()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {ctags_file}    (string)   specify a particular ctags file to use
-        {show_line}     (boolean)  if true, shows the content of the line the
-                                   tag is found on in the picker (default is
-                                   true)
-        {shorten_path}  (boolean)  if true, makes file paths shown in picker
-                                   use one letter for folders (default is true)
-        {hide_filename} (boolean)  if true, hides the name of the file in the
-                                   current picker (default is false)
+        {ctags_file} (string)   specify a particular ctags file to use
+        {show_line}  (boolean)  if true, shows the content of the line the tag
+                                is found on in the picker (default is true)
 
 
 builtin.current_buffer_tags({opts})            *builtin.current_buffer_tags()*
@@ -1045,12 +1057,6 @@ builtin.tagstack({opts})                                  *builtin.tagstack()*
     Parameters: ~
         {opts} (table)  options to pass to the picker
 
-    Fields: ~
-        {shorten_path}  (boolean)  if true, makes file paths shown in picker
-                                   use one letter for folders (default is true)
-        {hide_filename} (boolean)  if true, hides the name of the file in the
-                                   current picker (default is true)
-
 
 builtin.jumplist({opts})                                  *builtin.jumplist()*
     Lists items from Vim's jumplist, jumps to location on `<cr>`
@@ -1067,10 +1073,6 @@ builtin.lsp_references({opts})                      *builtin.lsp_references()*
 
     Parameters: ~
         {opts} (table)  options to pass to the picker
-
-    Fields: ~
-        {shorten_path} (boolean)  if true, makes file paths shown in picker use
-                                  one letter for folders (default is false)
 
 
 builtin.lsp_definitions({opts})                    *builtin.lsp_definitions()*
@@ -1134,13 +1136,7 @@ builtin.lsp_workspace_symbols({opts})        *builtin.lsp_workspace_symbols()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {shorten_path}    (boolean)       if true, makes file paths shown in
-                                          picker use one letter for folders
-                                          (default is false)
         {ignore_filename} (string)        file(s) to ignore
-        {hide_filename}   (boolean)       if true, hides the name of the file
-                                          in the current picker (default is
-                                          false)
         {symbols}         (string|table)  filter results by symbol kind(s)
 
 
@@ -1153,10 +1149,6 @@ builtin.lsp_dynamic_workspace_symbols({opts})*builtin.lsp_dynamic_workspace_symb
 
     Parameters: ~
         {opts} (table)  options to pass to the picker
-
-    Fields: ~
-        {hide_filename} (boolean)  if true, hides the name of the file in the
-                                   current picker (default is false)
 
 
 builtin.lsp_document_diagnostics({opts})  *builtin.lsp_document_diagnostics()*
@@ -1173,9 +1165,6 @@ builtin.lsp_document_diagnostics({opts})  *builtin.lsp_document_diagnostics()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {hide_filename}  (boolean)        if true, hides the name of the file
-                                          in the current picker (default is
-                                          false)
         {severity}       (string|number)  filter diagnostics by severity name
                                           (string) or id (number)
         {severity_limit} (string|number)  keep diagnostics equal or more severe
@@ -1205,9 +1194,6 @@ builtin.lsp_workspace_diagnostics({opts})*builtin.lsp_workspace_diagnostics()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {hide_filename}  (boolean)        if true, hides the name of the file
-                                          in the current picker (default is
-                                          false)
         {severity}       (string|number)  filter diagnostics by severity name
                                           (string) or id (number)
         {severity_limit} (string|number)  keep diagnostics equal or more severe
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -492,7 +492,7 @@ files.current_buffer_tags = function(opts)
   return files.tags(vim.tbl_extend("force", {
     prompt_title = 'Current Buffer Tags',
     only_current_file = true,
-    hide_filename = true,
+    path_display = 'hidden',
   }, opts))
 end
 
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -177,6 +177,20 @@ local telescope_defaults = {
     Default: true]]
   },
 
+  path_display = { {}, [[
+    Determines how file paths are displayed
+
+    path_display can be set to an array with a combination of:
+    - "hidden"    hide file names
+    - "tail"      only display the file name, and not the path
+    - "absolute"  display absolute paths
+    - "shorten"   only display the first character of each directory in the path
+
+    path_display can also be set to 'hidden' string to hide file names
+
+    Default: {}]]
+  },
+
   borderchars = { { "─", "│", "─", "│", "╭", "╮", "╯", "╰" } },
 
   get_status_text = {
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -184,7 +184,8 @@ local telescope_defaults = {
     - "hidden"    hide file names
     - "tail"      only display the file name, and not the path
     - "absolute"  display absolute paths
-    - "shorten"   only display the first character of each directory in the path
+    - "shorten"   only display the first character of each directory in
+                  the path
 
     path_display can also be set to 'hidden' string to hide file names
 
@@ -261,13 +262,15 @@ local telescope_defaults = {
       Into your config.
 
 
-    otherwise, just set the mapping to the function that you want it to be.
+    otherwise, just set the mapping to the function that you want it to
+    be.
 
         ...,
         ["<C-i>"] = require('telescope.actions').select_default,
         ...,
 
-    If the function you want is part of `telescope.actions`, then you can simply give a string.
+    If the function you want is part of `telescope.actions`, then you can
+    simply give a string.
       For example, the previous option is equivalent to:
 
         ...,
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -5,8 +5,6 @@ local strings = require('plenary.strings')
 
 local Path = require('plenary.path')
 
-local get_default = utils.get_default
-
 local treesitter_type_highlight = {
   ["associated"] = "TSConstant",
   ["constant"]   = "TSConstant",
@@ -68,17 +66,13 @@ do
     local cwd = vim.fn.expand(opts.cwd or vim.fn.getcwd())
 
     local disable_devicons = opts.disable_devicons
-    local shorten_path = opts.shorten_path
 
     local mt_file_entry = {}
 
     mt_file_entry.cwd = cwd
     mt_file_entry.display = function(entry)
       local hl_group
-      local display = path.make_relative(entry.value, cwd)
-      if shorten_path then
-        display = utils.path_shorten(display)
-      end
+      local display = utils.transform_path(opts, entry.value)
 
       display, hl_group = utils.transform_devicons(entry.value, display, disable_devicons)
 
@@ -136,7 +130,6 @@ do
   end
 
   --- Special options:
-  ---  - shorten_path: make the path appear short
   ---  - disable_coordinates: Don't show the line & row numbers
   ---  - only_sort_text: Only sort via the text. Ignore filename and other items
   function make_entry.gen_from_vimgrep(opts)
@@ -145,7 +138,6 @@ do
     opts = opts or {}
 
     local disable_devicons = opts.disable_devicons
-    local shorten_path = opts.shorten_path
     local disable_coordinates = opts.disable_coordinates
     local only_sort_text = opts.only_sort_text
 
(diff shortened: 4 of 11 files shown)


lua/telescope/config.lua
```diff
@@ -113,6 +113,10 @@ function config.set_defaults(defaults)
   set("border", {})
   set("borderchars", { '─', '│', '─', '│', '╭', '╮', '╯', '╰'})
 
+  set("hide_filename", false)
```
tjdevries, line 116: can we change this to `hide_path` if we have the other values?

my other concern is that we're going to explode into many different `X_path` options... I wonder if there is a way we can get ahead of this but still keep it simple for people to use. perhaps some ENUM style item for path displaying? I'm not sure.

caojoshua: I was going with `hide_filename` because some of the pickers already use it, but `hide_path` sounds better.

Enum in this case would work, since the options are mutually exclusive, but its possible that we include additional paths that may break that assumption. For example, I thought that we could maybe add a `absolute_path` option, in case user prefers absolute/relative paths. If this were the case, we would need separate enum values for `shorten_path=true`, `absolute_path=true`, and `shorten_path && absolute_path`. 

I ended up not including example option, since I'm not sure if people want that. But maybe there are other cases that breaks mutual exclusivity, I just can't think of them. If there's nothing else, I can go with the enum idea.

tjdevries: We could do "enum" style, that is more like "flags".

So we can have them be binary flags see if the flag is set. that would allow us the flexibility in the future and for the 99% use case, users just say "telescope.enum_things.path_options.HIDE_PATH" or something.

I can think about this more. gotta run right now

caojoshua: I created another commit that tries out adhoc enums. Configs can be set with something like

```
local defaults = {}
defaults[require('telescope.types').path_display_options.SHORTEN_PATH] = true
require('telescope').setup{
    defaults = defaults
}
```

I'm honestly not feeling like this makes the world better, but maybe it can inspire some ideas.

caojoshua: Another idea. Could have user specify something like:

```
local defaults = {}
defaults.path_display_options = {
  require(tele.configs).SHORTEN_PATH,
  require(tele.configs).SOME_OTHER_OPT,
}
require('telescope').setup{ defaults=defaults }
```
And the path transformation function can just look through the given options. Again, not sure how nice this is to use. Open to any opinions.

tami5:
😟 enum, seriously. I don't see any value of enums vs strings. It will just break setups.

Strings doesn't require importing stuff and support future abstraction of configuration setup

tjdevries:
I will send some comments maybe later today, but probably tomorrow. This isn't exactly what I meant. No worries though 😁

tjdevries:
```lua
local make_enum = function(t)
  -- todo: enum maker... (i have implementation elsewhere)
  return t
end

local types = {}

types.FileDisplay = make_enum {
  HIDE_PATH = 1,
  SHORTEN_PATH = 2,
  TAIL_PATH = 4,
  ABSOLUTE_PATH = 8,
  -- ... future optiosn possible
}

require('telescope.builtin').find_files {
  -- This would make paths short & absolute
  file_display = types.FileDisplay.SHORTEN_PATH + types.FileDisplay.ABSOLUTE_PATH,
}

require('telescope.builtin').find_files {
  -- This would hide the path.
  -- It shouldn't matter what else you add here, as it will just hide the path.
  file_display = types.FileDisplay.HIDE_PATH,
}

-- This method lets us continue to add other options to the display without having to create
-- an ever growing number of fields and have people be confused about which one each does.
```

This is more along the lines of what I was thinking.

We can use bit flags to check whether a value is present or not and do things with it depending on how it works.

@tami5 does this make more sense?

tami5:
> ```lua
> 
> local make_enum = function(t)
> 
>   -- todo: enum maker... (i have implementation elsewhere)
> 
>   return t
> 
> end
> 
> 
> 
> local types = {}
> 
> 
> 
> types.FileDisplay = make_enum {
> 
>   HIDE_PATH = 1,
> 
>   SHORTEN_PATH = 2,
> 
>   TAIL_PATH = 4,
> 
>   ABSOLUTE_PATH = 8,
> 
>   -- ... future optiosn possible
> 
> }
> 
> 
> 
> require('telescope.builtin').find_files {
> 
>   -- This would make paths short & absolute
> 
>   file_display = types.FileDisplay.SHORTEN_PATH + types.FileDisplay.ABSOLUTE_PATH,
> 
> }
> 
> 
> 
> require('telescope.builtin').find_files {
> 
>   -- This would hide the path.
> 
>   -- It shouldn't matter what else you add here, as it will just hide the path.
> 
>   file_display = types.FileDisplay.HIDE_PATH,
> 
> }
> 
> 
> 
> -- This method lets us continue to add other options to the display without having to create
> 
> -- an ever growing number of fields and have people be confused about which one each does.
> 
> ```
> 
> 
> 
> This is more along the lines of what I was thinking.
> 
> 
> 
> We can use bit flags to check whether a value is present or not and do things with it depending on how it works.
> 
> 
> 
> @tami5 does this make more sense?

😭 idk @tjdevries, I definitely prefer if `file_display` simply receive a list of strings, with strict checking of string keys, so that spelling mistakes would error out.

Don't get me wrong I absolutely like like enums, but for lua and configs; 1. I don't see value in it, 2. it feels more robust and requires extra code.

Your call.

caojoshua:
@tjdevries I personally would slightly lean towards @tami5 's opinion. As a telescope user, configs will be a little easier to have raw strings then getting the enum type. I can see both sides though, since enums are more "proper" and can help catch configuration errors. Either works for me, I can go with what the maintainers prefer.

tjdevries:
How do you do this with strings: `  file_display = types.FileDisplay.SHORTEN_PATH + types.FileDisplay.ABSOLUTE_PATH,`

Perhaps we just enumerate all the string options "shortened_absolute", "shortened_relative", etc.? It seems a bit difficult to check them all (and prone to error as time goes on).

tjdevries:
Ah, so you're saying `file_display = { "short", "absolute" }` like this? I suppose I'm fine with that as long as we validate all the strings at start time. I don't mind it this way. (should also allow just passing one string: `file_display = "hidden"` as well, for ease of use and less confusion)

caojoshua:
seems like @tjdevries, @Conni2461, and myself are in alignment to have the `file_display` be a list of strings(will probably actually rename to `path_display`). Should be a simple change, but won't get to it until later this week.

caojoshua:
OK I actually just pushed a commit for using raw strings instead of enums. Apparently I don't know how to work with git well enough. I rebased against master and got a bunch of stuff not related to this PR. I'll try to figure it out tomorrow or the day after.

caojoshua:
Cleaned things up and should be reviewable now.

lua/telescope/config.lua
```diff
@@ -113,6 +113,8 @@ function config.set_defaults(defaults)
   set("border", {})
   set("borderchars", { '─', '│', '─', '│', '╭', '╮', '╯', '╰'})
 
+  set("path_display", {})
```
tami5, line 116: 🤔 How can we validate this? @tjdevries. I failed to figure out a method of validation under the current structure. Any ideas.

I think validation should also be out of the scope of this PR

tami5:
LGTM

lua/telescope/utils.lua
```diff
@@ -40,6 +40,15 @@ utils.reversed_ipairs = function(t)
   return reversedipairsiter, t, #t + 1
 end
 
+utils.contains = function(t, val)
```
tjdevries, line 43: we already have `vim.tbl_contains`

lua/telescope/make_entry.lua
```diff
@@ -935,7 +894,7 @@ end
 
 function make_entry.gen_from_lsp_diagnostics(opts)
   opts = opts or {}
-  opts.tail_path = utils.get_default(opts.tail_path, true)
+  opts.tail_path = get_default(opts.tail_path, true)
```
tjdevries, line 897: why do we still have `tail_path`?

tjdevries:
Only other thing that I notice right away is that we should warn people if they pass something like `shorten_path` as a key (so check that it's not nil). If it's nil, you should advise them how to fix (by passing "shorten" or whatever).

lua/telescope/config.lua
tjdevries, line 116: This needs a description here explaining how to use it -- otherwise people won't know what to do.

Also, we should make sure that every reference to `shorten_path`, including in the README is removed.

caojoshua: I added [this wiki page](https://github.com/nvim-telescope/telescope.nvim/wiki/Path-Display-Configuration). Feel free to edit anyone.

I'm still working on the PR, but plan on linking this from the README, since this setting is too much to be explained in one line.

tami5: > I added [this wiki page](https://github.com/nvim-telescope/telescope.nvim/wiki/Path-Display-Configuration). Feel free to edit anyone.
> 
> I'm still working on the PR, but plan on linking this from the README, since this setting is too much to be explained in one line.

😍 😍

caojoshua:
Added more changes
* cleaned up a bunch of old path options. Tried to maintain current functionality. Seems fine with some testing on my end.
* I've noticed that for making paths relative, we sometimes get error `viml functions must not be called in a lua loop callback`. I've so far only noticed this for live_grep. I've commented out that section for now, and want to see if the reviewers better understand this problem.
* editing docs, including a link to [this wiki](https://github.com/nvim-telescope/telescope.nvim/wiki/Path-Display-Configuration)

lua/telescope/utils.lua
```diff
@@ -147,6 +147,38 @@ utils.path_tail = (function()
...
+  local transformed_path = path
+
+  if vim.tbl_contains(path_display, "tail") then
+    transformed_path = utils.path_tail(transformed_path)
+  else
+    -- TODO: absolute/relative paths. we can't call vimL functions in lua loop callback.
+    -- if not vim.tbl_contains(path_display, "absolute") then
+    --   local cwd = vim.fn.expand(opts.cwd or vim.fn.getcwd())
+    --   transformed_path = pathlib.make_relative(transformed_path, cwd)
+    -- end
```
l-kershaw, line 172: I think we might be able to create a workaround here using `vim.loop.cwd` and `vim.loop.chdir` to store the original working directory (in an absolute form), switch to the new one (relative to the original working directory), get the new working directory (in an absolute form), and then switch back to the original working directory.

I'm not sure how computationally efficient this is, but it avoids the use of viml in this function.

---

Since I couldn't find the place where this is actually used in a lua loop callback, this is the best idea I can come up with to solve this issue.

l-kershaw (commented):
Looks very useful 🙂

---

Only additional suggestion would be that I think we can remove `get_default` on line 7 of `make_entry.lua` as it is no longer used in that file. (Not sure why GitHub isn't letting me review that line but 🤷)

lua/telescope/utils.lua
Conni2461: Its not save to use `vim.loop.chdir` because it doesn't trigger autocmds. So if you have fugitive installed and you do things with `vim.loop.chdir` you will get a bunch of errors. (last time i checked)

But `vim.loop.cwd` is fine.

doc/telescope.txt
```diff
@@ -502,11 +502,6 @@ builtin.tags({opts})                                          *builtin.tags()*
         {show_line}     (boolean)  if true, shows the content of the line the
                                    tag is found on in the picker (default is
                                    true)
-        {shorten_path}  (boolean)  if true, makes file paths shown in picker
```
Conni2461, line 505: please don't modify `doc/telescope.txt` by hand. Its autogenerated. You need to adjust `lua/telescope/builtin/init.lua`

And than you need to make sure that " Allow all actions " and " Read and write permissions " are set in your fork settings, actions.

lua/telescope/utils.lua
Conni2461: You can also check if you can call viml with

```lua
if not vim.in_fast_event() then
  -- do viml stuff here
else 
  -- graceful alternative
end
```

l-kershaw: I didn't get any errors from fugitive when using a little test script I wrote for myself as proof of concept (I didn't do extensive tests though).
I wouldn't expect there to be any errors from using `vim.loop.chdir` to switch working directory, as long as we ensure we switch back to the original working directory afterwards (although I'm not very experienced using `vim.loop`, so may be wrong).

---

I agree that using `vim.in_fast_event()` to check if we can just use the viml would be good too.

doc/telescope.txt
caojoshua: Not sure what the purpose of those repo settings, but they are set already. I'll fix the docs though.

Conni2461: I am just mentioning it because the CI should run docgen and revert the changes. But we have a lot of problems with that lately. Here it worked https://github.com/nvim-telescope/telescope.nvim/pull/888 and here it didn't https://github.com/nvim-telescope/telescope.nvim/pull/887

:|

caojoshua:
Any of the maintainers want to take a look at this?

lua/telescope/builtin/files.lua
```diff
@@ -496,7 +496,6 @@ files.current_buffer_tags = function(opts)
   return files.tags(vim.tbl_extend("force", {
     prompt_title = 'Current Buffer Tags',
     only_current_file = true,
-    hide_filename = true,
```
Conni2461, line 499: I think it still makes sense to hide the filename for `current_buffer_tags`. I don't know the new interface yet but you update that part?

lua/telescope/utils.lua
```diff
@@ -147,6 +147,37 @@ utils.path_tail = (function()
...
+  return path_display == nil or path_display == "hidden" or
+    type(path_display) ~= "table" or vim.tbl_contains(path_display, "hidden")
+end
+
+utils.transform_path = function(opts, path)
+  if utils.is_path_hidden(opts) then
+    return ''
+  end
+
+  local path_display = utils.get_default(opts.path_display, require('telescope.config').values.path_display)
```
Conni2461, line 162: It feels weird that we are doing `utils.get_default` twice. Once in is_path_hidden and again immediately again

lua/telescope/utils.lua
```diff
@@ -147,6 +147,37 @@ utils.path_tail = (function()
...
+  local transformed_path = path
+
+  if vim.tbl_contains(path_display, "tail") then
+    transformed_path = utils.path_tail(transformed_path)
+  else
+    if not vim.in_fast_event() and not vim.tbl_contains(path_display, "absolute") then
+      local cwd = vim.fn.expand(opts.cwd or vim.fn.getcwd())
+      transformed_path = pathlib.make_relative(transformed_path, cwd)
+    end
```
Conni2461, line 171: I am pretty sure this is not optimal. First of all `vim.loop.cwd()` is always better than `vim.fn.getcwd()` because it can run in an fast event. Second you don't need to expand `cwd` so you could do this:
- `if not opts.cwd` then we can always do `local cwd = vim.loop.cwd()`
- `if opts.cwd` then we need to make sure that we are not `in_fast_event`. if not then `vim.fn.expand`

lua/telescope/utils.lua
caojoshua: Yes, feels a little awkward, but I think it makes sense given that `is_path_hidden` should be separated into a general util function. 

I could do 
```
  return (path_display == nil or path_display == "hidden" or
    type(path_display) ~= "table" or vim.tbl_contains(path_display, "hidden")), path_display
```

and transform_path could do 
```
local is_path_hidden, path_display = utils.is_path_hidden(opts)
```

and other functions could just use the first value.

I'm personally iffy on this approach, since returning the extra value will not align with the name of the function, but open to other opinions/suggestions.

lua/telescope/utils.lua
caojoshua: * +1 on `vim.loop.cwd()`
* I think I get what you mean for the other points. Pushing the change

lua/telescope/utils.lua
l-kershaw: Maybe you could make `path_display` an optional parameter for `is_path_hidden`?
Then if it is passed to `is_path_hidden`, we use it, if not, we calculate it in the way you have done.

This means we can calculate it once in `transform_path`, before using `is_path_hidden`, then pass it to `is_path_hidden` where it is not calculated for a second time.

This allows the current functionality of `is_path_hidden` to be retained when no `path_display` parameter is passed, but avoids having to calculate the same thing multiple times in quick succession.

_(Hope this comment makes sense, it feels a little rambly, but not sure how else to phrase it 😅)_

caojoshua: Thanks @l-kershaw , I've added this to the most recent commit. I'm not 100% sold on it, since it feels a little awkward that after passing `path_display`, `opt` argument is no longer used. But feels slightly better than the current version.

l-kershaw: @caojoshua I agree, it does feel a little awkward.
Maybe we could make it clearer for future development by adding documentation to `is_path_hidden` to explain the intention.

(and/or)

Perhaps we could add a check that errors if both `opt` and `path_display` are non-`nil`, as it is ambiguous which parameter should take precedence in that situation?
Although this is making what should be quite a simple function even more complicated, so I don't really know what's best here.

lua/telescope/config.lua
```diff
@@ -101,6 +101,8 @@ function config.set_defaults(defaults)
...
+
```
Conni2461, line 105: Can we get some documentation for `path_display` similar to `entry_prefix`? (Only for `path_display`, we have another pr that adds documentation for some of the other config options)

caojoshua: Not sure I understand your question, but we do have https://github.com/nvim-telescope/telescope.nvim/wiki/Path-Display-Configuration. This PR updates the main README to the wiki.

Conni2461: Okay first of all i wanna get this merged today because neovim 0.5 comes out tomorrow and this is a breaking backwards compatibility pr.

Yes we have documentation in the wiki and in the readme but some people dont always have access to that and vim has a great builtin tool for documentation. We are generating vim documentation from code. This is a occasion something like that:

```lua
set("path_display", {}, [[
  This documentation is autogenerated and will end up in the vim documentation!!!

  So we should do at least say here what this config field does and what the
  available options are. Similar to the other fields like sorting_strategy, 
  selection_strategy, ....
]])
```

If you don't have the time to finish it today, also because there will be merge conflicts after https://github.com/nvim-telescope/telescope.nvim/pull/922 is merged, I can finish some of this stuff i this merge request, so i am not stealing you contribution anyway thanks :)

caojoshua: I think its best if the maintainers finish this part. I can't guarantee when I can get to this. And this will avoid more back-and-forth in the PR.

caojoshua: I went ahead and added this kind of documentation, and fixed a bunch of merge conflicts.

Conni2461:
I generated docs, written changelog and added the old options to deprecated file (because its a breaking change).
Also fixed an issue with hidden + tags.

I'll merge tomorrow, its too late for me now. :)

Thanks for adding the docs and all your work here @caojoshua :smiley:

PR #842, opened 2021-05-13
Change default behavior of git_files to be the same as git ls-files' default and fzf.vim's GFiles

### LINKED ISSUE #826: Change default of git_files to show_untracked=false
Problem: `git_files` has an unexpected default of listing untracked files.
Suggestion: make the default of `git_files` *not* list untracked files.

`:Telescope git_files` currently lists all files in a git repo folder; the default is to also list untracked files. This is not the same default as git and other tools have. For example

 - `git ls-files` lists all tracked files in the repo and a flag is needed to list untracked files as well.
 - `fzf.vim`'s `:GFiles` lists tracked files (has `:GFiles?` to list untracked files as well)

It is possible to list tracked files with `lua require"telescope.builtin".git_files({ show_untracked = false }) ` but I think that the default behavior is unexpected. The first time I tried telescope I thought that `git_files` was broken because it acted just like `find_files`.

Related: There is already `git_status` which can be used to find new untracked files in the repo.

smithbm2316:
This closes #826, just wanted to put up this PR as a way for us to discuss this. While I don't personally ever have a use case that I _don't_ want to see untracked files by git when fuzzy searching in my projects, I do think that this default does make more sense, considering how fzf.vim's `:GFiles` command and `git ls-files` work by default. We do have a lot of people transitioning from fzf who may find this default confusing. Perhaps it could be documented in more detail as well if we feel that's necessary, I'd be happy to make that happen. 

cc @mhinz @kaddkaka @clason

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -977,7 +977,7 @@ builtin.git_files({opts})                      *telescope.builtin.git_files()*
                                         (default: true)
         {show_untracked}     (boolean)  if true, adds `--others` flag to
                                         command and shows untracked files
-                                        (default: true)
+                                        (default: false)
         {recurse_submodules} (boolean)  if true, adds the
                                         `--recurse-submodules` flag to command
                                         (default: false)
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -22,7 +22,7 @@ git.files = function(opts)
     return
   end
 
-  local show_untracked = utils.get_default(opts.show_untracked, true)
+  local show_untracked = utils.get_default(opts.show_untracked, false)
   local recurse_submodules = utils.get_default(opts.recurse_submodules, false)
   if show_untracked and recurse_submodules then
     utils.notify("builtin.git_files", {
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -132,7 +132,7 @@ builtin.current_buffer_tags = require_on_exported_call("telescope.builtin.files"
 ---@param opts table: options to pass to the picker
 ---@field cwd string: specify the path of the repo
 ---@field use_git_root boolean: if we should use git root as cwd or the cwd (important for submodule) (default: true)
----@field show_untracked boolean: if true, adds `--others` flag to command and shows untracked files (default: true)
+---@field show_untracked boolean: if true, adds `--others` flag to command and shows untracked files (default: false)
 ---@field recurse_submodules boolean: if true, adds the `--recurse-submodules` flag to command (default: false)
 ---@field git_command table: command that will be exectued. {"git","ls-files","--exclude-standard","--cached"}
 builtin.git_files = require_on_exported_call("telescope.builtin.git").files


clason:
I don't have any skin in this game, but (as usual) I'd counsel towards consistency.

Specifically (as I commented on Gitter), I believe there is a choice here:
1. `find_files` and `git_files` (and others) are just wrappers for external commands, in which case they should behave by default exactly like the external commands without additional arguments;
2. these implement their own, differing, functionality, in which case this should be documented as clearly as possible.

There are good arguments for either choice, but it'd be good to pick one _globally_ and stick with it across commands. And in the second case, it would be good to reduce overlap as much as possible -- if `find_files` and `git_files` behave the same in many contexts, there's not really any reason (and, indeed, danger of confusion) to have both.

On the other hand, I don't think we need to feel pressured to follow `fzf.vim`'s behavior -- telescope is by now very much its own thing. That being said, a "transitioning from FZF" wiki page that lists telescope equivalents (if necessary with appropriate options) for common FZF commands would be quite useful. (I'm thinking of something like <https://numpy.org/doc/stable/user/numpy-for-matlab-users.html>.)

smithbm2316:
> I don't have any skin in this game, but (as usual) I'd counsel towards consistency.

Likewise for me, I just know from the discussion on Gitter you had some good thoughts that resonated with me.

> Specifically (as I commented on Gitter), I believe there is a choice here:
>  1. `find_files` and `git_files` (and others) are just wrappers for external commands, in which case they should behave by default exactly like the external commands without additional arguments;
>  2. these implement their own, differing, functionality, in which case this should be documented as clearly as possible.
>
> There are good arguments for either choice, but it'd be good to pick one _globally_ and stick with it across commands. And in the second case, it would be good to reduce overlap as much as possible -- if `find_files` and `git_files` behave the same in many contexts, there's not really any reason (and, indeed, danger of confusion) to have both.

I personally would lean towards the first option in general. I think it certainly would reduce confusion moving forward if we were able to make that a consistent tenant of how we approach managing Telescope's builtin features. Especially now as we've really reached a point where we have the majority of pickers that I think anyone would use on a regular basis now integrated into core.

> On the other hand, I don't think we need to feel pressured to follow `fzf.vim`'s behavior -- telescope is by now very much its own thing. That being said, a "transitioning from FZF" wiki page that lists telescope equivalents (if necessary with appropriate options) for common FZF commands would be quite useful. (I'm thinking of something like https://numpy.org/doc/stable/user/numpy-for-matlab-users.html.)

I really like this idea a lot. Regardless of what we choose to do moving forward I think this is a great shout, I'll add it to the Documentation projects kanban board as a To Do, I'd be happy to work on something like this once I finish documenting some of the more pressing functionalities. My point of discussion here was not so much "let's follow what fzf does", rather that AFAIK it tends to abide by the first option you mentioned, which adds some extra validity to consider moving forward with the first option as our typical pattern.

kaddkaka:
Telescope just being a wrapper around other tools is a good way to keep telescope from surprising users. Here I guess the underlying tool is `git ls-files`, Perhapa the commit message should reflect that, mention git instead of fzf that is.

I like this change.

clason:
> I personally would lean towards the first option in general. I think it certainly would reduce confusion moving forward if we were able to make that a consistent tenant of how we approach managing Telescope's builtin features. Especially now as we've really reached a point where we have the majority of pickers that I think anyone would use on a regular basis now integrated into core.

In this case, I would suggest revisiting the fallbacks in `find_files` and `grep_string` -- currently, e.g., the first uses `fd`, `fdfind`, `rg`, or `find` (in this order) depending on what's available, but these have more or less subtly different behaviors which I think confuses people. Better to be explicit here as well -- either by making `fd` a hard requirement or adding a separate fallback function that people should call instead if it's not available?

smithbm2316:
> In this case, I would suggest revisiting the fallbacks in `find_files` and `grep_string` -- currently, e.g., the first uses `fd`, `fdfind`, `rg`, or `find` (in this order) depending on what's available, but these have more or less subtly different behaviors which I think confuses people. Better to be explicit here as well -- either by making `fd` a hard requirement or adding a separate fallback function that people should call instead if it's not available?

Definitely something that might be worth revisiting. However, this may be something to put on the backburner until TJ gets back from vacation and we can loop him in here. Changing these kinds of defaults would be a significant change that I'd rather make sure we get his feedback on before proceeding. 

I don't personally use `git_files` and am at the point where I know enough about the basics of the Telescope codebase that these kinds of confusions don't affect me as much anymore, so I can't say I have a really _strong_ opinion on what choice we end up making. Abiding by the defaults of the external program we use for different pickers makes more sense to my head, but I am not married to the idea. The key is that I make sure we document these aspects in more detail to try and reduce confusion for future users.

tjdevries:
I'm happy to have us just pick one and document it well in the docs somehow. I don't really care either way as long as we have it documented (since I think people will expect it both ways basically haha)

tjdevries:
Wiki page for transition from fzf is a great idea though

Conni2461:
I am going to merge this into dev branch. We then have a dev MR that will document all breaking changes prior to the merge. Will hit master in about 4 months.

Thanks for the PR :)

PR #849, opened 2021-05-16
Fix string.find() matching for only_cwd option

thornomad:
This fixes an issue with the `builtin.buffers` `only_cwd` option that occurs when the `cwd` has an `-` in the name (or any other character that triggers pattern matching).

`string.find()` defaults to _pattern_ matching (rather than string literal matching).

This fix asks `string.find()` to interpret the arguments as literal strings rather than patterns.

Reference: https://stackoverflow.com/a/15258515/181902


Conni2461:
Good catch.

Thanks :)

smithbm2316:
Thanks @thornomad for taking the time to debug and track this down! Was gonna merge this but it looks like Conni beat me to it :rofl:

PR #853, opened 2021-05-21
fix: attempt to concatenate when entry.kind == nil or symbol_type == nil

### LINKED ISSUE #763: Treesitter `make_entry.gen_from_treesitter` does not account for possible value `nil` of `entry.kind`
### Description
Using `<cmd>Telescope treesitter<cr>` in some typescript files I get the following error:

```
E5108: Error executing lua ...k/packer/opt/telescope.nvim/lua/telescope/make_entry.lua:544: attempt to concatenate field 'kind' (a nil value)
```

Code in question:

```lua
return {
    -- ...
    ordinal = node_text .. " " .. entry.kind,
    -- ...
}      
```

It seems that some Treesitter parsers do not necessarily attach a kind to entries.
Seems to happen especially to auto-typed parameters and non-trivial type defs, at least using Typescript.
Using a default value if `entry.kind == nil` (for example "unknown") fixes the error.

opsound:
Fix #763 #874. I'm also hitting the same error when I attempt to `:Telescope treesitter` in .c files.

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -424,7 +424,7 @@ function make_entry.gen_from_lsp_symbols(opts)
     if not opts.ignore_filename and filename then
       ordinal = filename .. " "
     end
-    ordinal = ordinal ..  symbol_name .. " " .. symbol_type
+    ordinal = ordinal ..  symbol_name .. " " .. (symbol_type or "unknown")
     return {
       valid = true,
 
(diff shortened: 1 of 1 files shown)


Conni2461:
We have a similar issue at line 427. Mind fixing it there as well? Would also fix https://github.com/nvim-telescope/telescope.nvim/issues/874

opsound:
SGTM -- updated L427

PR #865, opened 2021-05-26
feat: Added "Silent" theme

torbratsberg:
Simple theme for quick operations where you barely need to look at the Telescope window.

### DIFF
--- a/README.md
+++ b/README.md
@@ -523,6 +523,7 @@ nnoremap <Leader>f :lua require'telescope.builtin'.find_files(require('telescope
 or use with command:
 ```vim
 Telescope find_files theme=get_dropdown
+Telescope colorscheme theme=get_silent
 ```
 
 Themes should work with every `telescope.builtin` function.  If you wish to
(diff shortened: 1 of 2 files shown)


disrupted:
I like it, can imagine this really well for LSP code actions. Bonus points if we figure out how to place the popup relative to the current cursor location.

torbratsberg:
@disrupted I agree! I couldn't find a way to set it relative to the cursor with the current configuration options, but that was the initial plan.

Conni2461:
Isn't this just `require('telescope.themes').get_dropdown{ previewer = false }` with a different result_height? Also might be interesting for you but there is a huge refactor of how layouts work here: https://github.com/nvim-telescope/telescope.nvim/pull/823

tjdevries:
Yeah, this seems _very_ close to the dropdown theme?

torbratsberg:
It is very similar to the dropdown theme, but intended to be used for smaller and quicker actions. I just saw this PR #878 which has almost the same theme as Silent, but relative to the cursor which is probably a better layout for this kind of theme. So this PR can be closed.

gokulsoumya:
> I like it, can imagine this really well for LSP code actions. Bonus points if we figure out how to place the popup relative to the current cursor location.

There's an open PR for cursor relative positioning - #878

Issue #617, opened 2021-03-06
Feature: System of default and common keybinds


timsofteng:
Hello. I was really impressed by telescope.

 I have several ideas how to improve it and pickers which I will try to describe in this and another issues.

#KEYBINDING

 In my opinion keybinding is very important part in software development. It's not only the thing which everyone can customize. It's part of architecture and philosophy of soft. Especially for telescope which architecture (ecosystem of pickers) is complex and modular.
 My idea is that the **main keyboard shortcuts should be consistent from picker to picker**. 
 I guess it's necessary to design picker's feature with a look at the common shortcuts

 That's why I want to introduce you my ideas of common shortcuts and hope we all can discuss about it. Some of them I have seen in another applications and another I just suggest by myself.

 *(note I will mention keys in normal mode which means for insert mode we just should add ctrl key to shortcuts)*

## LIST OF BINDS

###movements
 1. Down (next) - `j`.
 2. Up (previous) - `k`.
 3. Back - `h`.
 4. Forward - `l`.

 It's very traditional vim movement. Back and forward actions could be used in file browser to change directory or go to higher level. Or from git branch to the list of commits and back (it's something like pickers chain which I'll try to describe in the next issue). Also `back` key can clean up all from search field at first if it's not the mpty.

 Additionally we will have empty `n` and `p` keys which we can use in another useful actions. E.g press `n` in file browser to create file or directory which you can choose in addition dialog in the vim' bottom *Press f to create file, press d to create directory, press b to create bookmark*. It's kind of nnn's FM working.
 Another example is press n to create new git branch etc.
 `P` key we can you to paste something (commits cherry picking, files in file browser).

###go up one level in the filesystem
 Press `backspace` to go up one level from current directory. The difference between `backspace` and `h` is that with `h` you can't go up if this level is out of scope of your working directory. With backspace you will can. **BUT** without `cwd`. *(Check how fern.vim work, it's same).*

###change working directory ("cwd")
 *`* or `~`. Just press it to change working directory *(e.g in file browser)*

###mark all searched instances
 I'm not sure is it bind exists or not but something like `a` key to mark all will be fine.

###disable ESC
 `Escape` key can be used as closed of some deep sub-actions of pickers. I guess it's more useful.
 Close telescope by `ctrl-c`

###deletion
 Simple `d` key. When you need to remove something just press `d`.

###hidden files
 Just press `.` (dot) to enable dot files in pickers *(file browser, file picker, etc)*

[...]

PR #868, opened 2021-05-27
Send to loclist

runiq:
Pretty simple all in all. Tested and appears to work.

I'm not sure about the mappings, though; I overwrote the `complete_tag` insert mode mapping. Please advise :)

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -313,6 +313,28 @@ actions.add_to_qflist()                              *actions.add_to_qflist()*
 
 
 
+actions.send_selected_to_loclist()        *actions.send_selected_to_loclist()*
+    Sends the selected entries to the location list, replacing the previous
+    entries.
+
+
+
+actions.add_selected_to_loclist()          *actions.add_selected_to_loclist()*
+    Adds the selected entries to the location list, keeping the previous
+    entries.
+
+
+
+actions.send_to_loclist()                          *actions.send_to_loclist()*
+    Sends all entries to the location list, replacing the previous entries.
+
+
+
+actions.add_to_loclist()                            *actions.add_to_loclist()*
+    Adds all entries to the location list, keeping the previous entries.
+
+
+
 actions.smart_send_to_qflist()                *actions.smart_send_to_qflist()*
     Sends the selected entries to the quickfix list, replacing the previous
     entries. If no entry was selected, sends all entries.
@@ -321,6 +343,18 @@ actions.smart_send_to_qflist()                *actions.smart_send_to_qflist()*
 
 actions.smart_add_to_qflist()                  *actions.smart_add_to_qflist()*
     Adds the selected entries to the quickfix list, keeping the previous
+    entries. If no entry was selected, adds all entries.
+
+
+
+actions.smart_send_to_loclist()              *actions.smart_send_to_loclist()*
+    Sends the selected entries to the location list, replacing the previous
+    entries. If no entry was selected, sends all entries.
+
+
+
+actions.smart_add_to_loclist()                *actions.smart_add_to_loclist()*
+    Adds the selected entries to the location list, keeping the previous
     entries. If no entry was selected, adds all entries.
 
 
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -330,6 +330,11 @@ actions.open_qflist()                                  *actions.open_qflist()*
 
 
 
+actions.open_loclist()                                *actions.open_loclist()*
+    Open the location list
+
+
+
 
 ================================================================================
                                                              *telescope.builtin*
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -544,18 +544,22 @@ actions.add_to_qflist = function(prompt_bufnr)
   send_all_to_qf(prompt_bufnr, 'a')
 end
 
+ --- Sends the selected entries to the location list, replacing the previous entries.
 actions.send_selected_to_loclist = function(prompt_bufnr)
   send_selected_to_qf(prompt_bufnr, 'r', 'loclist')
 end
 
+ --- Adds the selected entries to the location list, keeping the previous entries.
 actions.add_selected_to_loclist = function(prompt_bufnr)
   send_selected_to_qf(prompt_bufnr, 'a', 'loclist')
 end
 
+ --- Sends all entries to the location list, replacing the previous entries.
 actions.send_to_loclist = function(prompt_bufnr)
   send_all_to_qf(prompt_bufnr, 'r', 'loclist')
 end
 
+ --- Adds all entries to the location list, keeping the previous entries.
 actions.add_to_loclist = function(prompt_bufnr)
   send_all_to_qf(prompt_bufnr, 'a', 'loclist')
 end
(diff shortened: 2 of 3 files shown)


tjdevries:
Hmmm, I'd probably say for now we can just merge without any default mapping changes -- I don't like changing default mappings that have already been taken, so I don't think I want to do change `<c-l>` at this point.

runiq:
Fair, I'll remove the mappings then

fdschmidt93:
When I introduced `<C-l>`, the motivation was to have a binding (on my home row :) that plays well with `<C-n>`, `<C-p>` for speedy filtering of diagnostics and symbols. I personally like the mnemonic in vim's spirit with  `<C-l>`/`<A-l>` parallel to `<C-q>`/`<A-l>` to send to loclist. Sending to qflist appears to be a widely used feature (frequently mentioned on reddit, Primeagen, etc.) and similarly sending to loclist might be as well.

What about remapping `complete_tag` to `<C-i>`? I guess (a) adding tabs insert mode with `<C-i>` is expendable and (b) anyone who currently uses `<C-l>` (my unfortunate hunch is not too may people are yet aware of it :x) I suppose will figure out that the mapping was rebound to `<C-i>` quickly after the fact.

clason:
I strongly feel that a plugin should not register specific keymappings -- your motivation and mnemonics are likely very different from mine (and we even speak the same mother tongue). 

Rather, telescope has a nice capability of registering keymappings in its setup function, and I see no reason why that can't be used for "send to loclist", just like it is for other actions?

fdschmidt93:
> I strongly feel that a plugin should not register specific keymappings -- your motivation and mnemonics are likely very different from mine (and we even speak the same mother tongue).

Absolutely fair!

> Rather, telescope has a nice capability of registering keymappings in its setup function, and I see no reason why that can't be used for "send to loclist", just like it is for other actions?

Mhm, I entirely see your points. However, I guess the same thing can be said about `<C-l>` for `complete_tag` as is (or another global mapping like `<C-q>`). A global mapping for `complete_tag` was more a practical decision (@Conni2461 agreed on gitter :sweat_smile:) such that users would not have to attach mappings for _every_ function in case they'd want it changed since it's universally applicable to `lsp_{document,workspace}_{diagnostics,symbols}` (and if I wasn't preoccupied, to maybe `tags` and `treesitter` pickers too :). Following that line of thinking, sending results to loclist is even more universally useful than `complete_tag`; but again, your view is at least as fair as mine.

clason:
Had I noticed, I would have spoken against the `complete_tag` mapping as well :)

I think the way to go here is to have a single `register_default_mappings` function or key (possibly dict) that can be suggested in the README and registers a set of such mappings. If they are completely opt-in, people who feel strongly are much less likely to complain.

The usefulness of  `global` actions is a valid point; but I would rather realize them through a corresponding global setting (maybe taking `actions` as keys?)

fdschmidt93:
> Had I noticed, I would have spoken against the complete_tag mapping as well :)

Here I was, thinking you'd be one of the primary users of that functionality as you initially requested diagnostics and filtering ;) :smile: 

> I think the way to go here is to have a single register_default_mappings function or key (possibly dict) that can be suggested in the README and registers a set of such mappings. If they are completely opt-in, people who feel strongly are much less likely to complain.

> The usefulness of global actions is a valid point; but I would rather realize them through a corresponding global setting (maybe taking actions as keys?)

Something along these lines sounds good.

https://github.com/nvim-telescope/telescope.nvim/blob/904f84942384f988271cd40cc4b195263fe0763a/lua/telescope/mappings.lua#L9 It looks like default mappings as a whole can already be overriden, though single keymaps cannot (as also suggested by see the [top-line comment](https://github.com/nvim-telescope/telescope.nvim/blob/904f84942384f988271cd40cc4b195263fe0763a/lua/telescope/mappings.lua#L1)

So, (a) we'd have some barebones default mappings (e.g. select vertically with `<C-v>`), (b) expose suggested default mappings with `register_default_mappings`, and (c) enable registering global mappings via a function akin to `attach_mappings` (to be added, I'll see what I can do). That begs the question whether `register_default_mappings` should be a separate config entry or just a proposal for the registering function laid out in the documentation for the `attach_mappings` global-equivalent.

> (And this is a good discussion to have now as telescope is moving towards 1.0 and full documentation.)

Absolutely!

Conni2461:
FYI i did not really agree with it. originally i said that i didn't like it and we should just map it for the specific functions: https://github.com/nvim-telescope/telescope.nvim/pull/626#discussion_r590649792 I just realized after that on gitter that its painful to remap `attach_mappings` (this still applies to that day. You would need to wrap the function in your own function with opts). It was more like a fine, whatever, doesn't really matter. (i can't really stuff that is so opinionated, you always find someone who has a problem with it. Like this guy who wanted to change "everything" https://github.com/nvim-telescope/telescope.nvim/issues/617 :rofl: ).

I also don't like mapping specific actions. I had a discussion with prime in a pr, where i introduced a action in a previous pr and didn't map (for `git_branches`) and he opened a pr a day later or so mapping that action. In my opinion, specific actions shouldn't be mapped and everyone should be able to easily remap them. For example similar to what tami suggested here: https://github.com/nvim-telescope/telescope.nvim/issues/582#issue-819110334

Also would we have a different conversation if `<c-l>` mapping was not made global? We would be at that point where this pr would be merged and someone would open a issue in 5 days saying that `<c-l>` doesn't work for `lsp_*_symbols` because the local mapping overrides the global one.

To come back to this PR. If you ask me we can make `send_to_loclist` global for `<c-l>`. But how we should handle completion, idk. That depends on how many people actually know about that feature. Otherwise code looks good :) Thanks @runiq

clason:
> Here I was, thinking you'd be one of the primary users of that functionality as you initially requested diagnostics and filtering ;) 😄

Really? I can't remember 😬  (Yesterday I was a different person -- Heraclitus and all that... ;))

> It looks like default mappings as a whole can already be overriden, though single keymaps cannot (as also suggested by see the [top-line comment](https://github.com/nvim-telescope/telescope.nvim/blob/904f84942384f988271cd40cc4b195263fe0763a/lua/telescope/mappings.lua#L1)
> So, (a) we'd have some barebones default mappings (e.g. select vertically with `<C-v>`), (b) expose suggested default mappings with `register_default_mappings`, and (c) enable registering global mappings via a function akin to `attach_mappings` (to be added, I'll see what I can do). That begs the question whether `register_default_mappings` should be a separate config entry or just a proposal for the registering function laid out in the documentation for the `attach_mappings` global-equivalent.

I think this would be something to ask @tjdevries input on!

runiq:
Oh wow what happened here? I'm sorry everyone, didn't mean to step on anyone's toes, sepecially not yours, @fdschmidt93 :( It was a bad call to just override an existing binding, I'm terribly sorry for that.

FWIW I really really really do *not* mind *not* mapping this action.

clason:
I thought you _did_ remove the mapping (so no fault of yours)?

runiq:
Yeah I did remove it now, but I feel like I should have asked in the first place, is what I mean

fdschmidt93:
@clason [here you go](https://github.com/nvim-telescope/telescope.nvim/issues/313#issuecomment-779468961) :)

@runiq Absolutely no worries! I guess I got the ball rolling advocating in favor of mapping `<C-l>` to `send_to_loclist`

clason:
Ah, OK, it was a _suggestion_ -- that sounds more like something I'd do ;)

tjdevries:
After you resolve conflicts this is fine to merge.

runiq:
@tjdevries Done.

Conni2461 (approved):
LGTM Thanks :)

PR #878, opened 2021-05-31
feat: add cursor layout

### LINKED ISSUE #496: feat: layout relative to current cursor position
**Why?**
When using things like LSP related pickers, I don't usually want them to take all of the space on my screen, especially things like the `lsp_code_actions` picker would be a lot better to use if it was smaller and relative to my cursor's position.

**How would I see it implemented?**
I actually already started trying to implement it. Here's the layout I had in mind:
```
+--------------+---------------------+
|    Prompt    |      Previewer      |
+--------------+      Previewer      |
|    Result    |      Previewer      |
|    Result    |      Previewer      |
+--------------+---------------------+
```
With the top left being the cursor's position. The previewer would be optional, but I put it here because that's how the Visual Studio IDE displays it on code actions.

**What knowledge I'm missing to make it work**
As said above, I already have a branch that has the basic idea, my issue is with getting the current cursor position.
I've gone through the help and found a bunch of things like `nvim_win_get_position`, `nvim_win_get_cursor`, `getline`, `wincol`, `screencol`, etc.
But nothing actually allows me to get the position of the cursor on screen without counting the folds, the fact that the buffer can be scrolled, splits, tabline, sign column, line number column, etc.
If someone knows what functions to get the position or knows what functions to use and calculate the cursor position, then I'd be able to do it myself.
Another thing that I would need to know is the size of the current neovim instance, so that if the coordinates are out of bound, I can position everything at the right place. I think these are given by the `lines` and `columns` options from what I understand.

Luxed:
Closes #496.

Here's how it looks like when space is not an issue, my cursor is on the first "t" of the "test" variable name:
image

Here's how it looks like when space is restricted in the bottom right (cursor at the same place):
image

What happens when there is not enough room is that the floating window gets put above the _cursor line_ and gets fixed to the right (with 1 extra space to give it some room).


dhruvmanila:
Nice job! Are you calculating the corner the window should be anchored to? Like in floating windows either one of the corners (`NW`, `NE`, `SW`, `SE`) is made an anchor point to the cursor. This is calculated using the space available for the telescope window w.r.t. the main Neovim window. You might want to take a look at `vim.lsp.util.make_floating_popup_options()`

Also, I think the layout updates should be put on hold for now until #823 is finalized and merged.

Luxed:
> Nice job! Are you calculating the corner the window should be anchored to? Like in floating windows either one of the corners (NW, NE, SW, SE) is made an anchor point to the cursor.

I am calculating the position so that the cursor is above the NW corner.
If the size of the floating window would be bigger than the available width, I then anchor it to the right of the screen (leaving 1 character of space).
If the size of the floating window would be bigger than the available height, I then anchor the bottom of the floating window above the cursor line.

Maybe it would be better for the window to always anchor to the NW, NE and SE points?

> You might want to take a look at `vim.lsp.util.make_floating_popup_options()`

I remember looking into it, but it was so long ago that I'll look into it again.

> Also, I think the layout updates should be put on hold for now until #823 is finalized and merged.

Of course, I didn't see this was ongoing.

l-kershaw:
@Luxed just a heads up that #823 has now been merged, so this PR is no longer blocked.

If you need any help with the new `layout_config` stuff, feel free to ping me 🙂

Luxed:
> @Luxed just a heads up that #823 has now been merged (or at least its successor #922 has), so this PR is no longer blocked.
> 
> If you need any help with the new `layout_config` stuff, feel free to ping me slightly_smiling_face

It ended up being a lot easier than I expected. Everything is working as it did before now.

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -391,6 +391,91 @@ layout_strategies.center = make_documented_layout("center", vim.tbl_extend("erro
...
+
+  prompt.height = 1
+  results.height = height
+  preview.height = results.height + prompt.height + bs
+
+  -- If previewer is activated, it should take 2/3 of the space (width)
+  local width_left = self.previewer and math.floor(max_width / 3) or max_width
+  prompt.width = width_left
+  results.width = prompt.width
+  preview.width = self.previewer and (2*width_left) or 0
```
l-kershaw, line 440: Is there a reason we are hard coding the width of the previewer?
We could include a `preview_width` key in a similar way to `layout_strategies.horizontal`.

l-kershaw (commented):
One small nitpick/question regarding `preview_width`, but overall looks very good 🙂

lua/telescope/pickers/layout_strategies.lua
Luxed: Do you mean if `self.previewer` is false?
I'm guessing this is just a leftover from testing. I can remove it since it shouldn't affect anything I think?

l-kershaw: Sorry for the confusion, I mean if `self.previewer` is `true`.

If `self.previewer` is `false`, then yes, `width_left` should be `max_width`.
But if the user wants a previewer and `self.previewer` is `true`, then the width of the previewer should be configurable.
I would suggest looking at the implementation of `preview_width` in `layout_strategies.horizontal` as a guide for how it could be implemented in `layout_strategies.cursor`.

Luxed: Ah yes, I think I understand now, I'm going to look into that!

Luxed: I think I've implemented it correctly now.

l-kershaw (approved):
I think I'm happy with this now 👍 

I'll run this PR for a little bit and if I don't find any problems I'll merge this later today/tomorrow.

Melkster:
This looks awesome. How do I tell `vim.lsp.buf.code_action()` to use this layout?

gegoune:
@Melkster https://github.com/nvim-telescope/telescope-ui-select.nvim

Melkster:
@gegoune Thanks!

PR #883, opened 2021-06-01
feat: set defaults for each picker in telescope setup

Conni2461:
Close https://github.com/nvim-telescope/telescope.nvim/issues/582

Its so long that i wrote telescope code. LUL 

I don't like the idea of custom pickers, the one that @tami5 proposed. I think for that people should just continue to go with own wrapper functions. But this helps clean up other things. Like this e.g.

```lua

require("telescope").setup {
  pickers = {
    buffers = {
      show_all_buffers = true,
      sort_lastused = true,
      theme = "dropdown",
      previewer = false,
      mappings = {
        i = {
          ["<c-d>"] = actions.delete_buffer,
        },
        n = {
          ["<c-d>"] = actions.delete_buffer,
        }
      }
    }
  }
}
```

I am not sure if this is the best way to implement this feature

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -37,6 +37,8 @@
 ---       mappings = {
 ---         i = {
 ---           ["<c-d>"] = require("telescope.actions").delete_buffer,
+---           -- or right hand side can also be a the name of the action as string
+---           ["<c-d>"] = "delete_buffer",
 ---         },
 ---         n = {
 ---           ["<c-d>"] = require("telescope.actions").delete_buffer,
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -14,9 +14,37 @@
 ---
 --- <pre>
 --- :lua require('telescope.builtin').live_grep({
----    prompt_title = 'find string in open buffers...',
----    grep_open_files = true
----  })
+---   prompt_title = 'find string in open buffers...',
+---   grep_open_files = true
+--- })
+--- -- or with dropdown theme
+--- :lua require('telescope.builtin').find_files(require('telescope.themes').get_dropdown{
+---   previewer = false
+--- })
+--- </pre>
+---
+--- You can also pass default configurations to builtin pickers. These options will also be added if
+--- the picker is executed with `Telescope find_files`.
+---
+--- <pre>
+--- require("telescope").setup {
+---   pickers = {
+---     buffers = {
+---       show_all_buffers = true,
+---       sort_lastused = true,
+---       theme = "dropdown",
+---       previewer = false,
+---       mappings = {
+---         i = {
+---           ["<c-d>"] = require("telescope.actions").delete_buffer,
+---         },
+---         n = {
+---           ["<c-d>"] = require("telescope.actions").delete_buffer,
+---         }
+---       }
+---     }
+---   }
+--- }
 --- </pre>
 ---
 --- This will use the default configuration options. Other configuration options are still in flux at the moment
@@ -325,4 +353,46 @@ builtin.lsp_document_diagnostics = require('telescope.builtin.lsp').diagnostics
 ---@field hide_filename boolean: if true, hides the name of the file in the current picker (default is false)
 builtin.lsp_workspace_diagnostics = require('telescope.builtin.lsp').workspace_diagnostics
 
+local apply_config = function(mod)
+  local pickers_conf = require('telescope.config').pickers
+  for k, v in pairs(mod) do
+    local pconf = vim.deepcopy(pickers_conf[k] or {})
+    if pconf.theme then
+      local theme = pconf.theme
+      pconf.theme = nil
+      pconf = require("telescope.themes")["get_" .. theme](pconf)
+    end
+    if pconf.mappings then
+      local mappings = pconf.mappings
+      pconf.mappings = nil
+      pconf.attach_mappings = function(_, map)
+        for mode, tbl in pairs(mappings) do
+          for key, action in pairs(tbl) do
+            map(mode, key, action)
+          end
+        end
+        return true
+      end
+    end
+    mod[k] = function(opts)
+      opts = opts or {}
+      if pconf.attach_mappings and opts.attach_mappings then
+        local attach_mappings = pconf.attach_mappings
+        pconf.attach_mappings = nil
+        local opts_attach = opts.attach_mappings
+        opts.attach_mappings = function(prompt_bufnr, map)
+          attach_mappings(prompt_bufnr, map)
+          return opts_attach(prompt_bufnr, map)
+        end
+      end
+
+      v(vim.tbl_extend("force", pconf, opts))
+    end
+  end
+
+  return mod
+end
+
+-- We can't do this in one statement because tree-sitter-lua docgen gets confused if we do
+builtin = apply_config(builtin)
 return builtin
--- a/lua/telescope/mappings.lua
+++ b/lua/telescope/mappings.lua
@@ -126,6 +126,8 @@ local telescope_map = function(prompt_bufnr, mode, key_bind, key_func, opts)
       return
     elseif key_func.type == "action_key" then
       key_func = actions[key_func[1]]
+    elseif key_func.type == "action" then
+      key_func = key_func[1]
     end
   end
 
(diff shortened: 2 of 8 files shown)


tami5:
THANKS SO MUCH CONNI. I'm really disappointed 😢 that I haven't gotten around and implemented this. but awesome work, I wouldn't have done it better

for mapping I suggest we simplify it a bit

```
mappings = {
  ['n|<C-d>'] = 
}
```

I already implemented this in some other projects, so I can share a code snippet here

This better because it save a lot of space. what do u think

Conni2461:
> for mapping I suggest we simplify it a bit

I would not do that for a number of reasons
1. Simpler for one mapping, but for that you can also just do this ` i = { ["<c-d>"] = actions.delete_buffer, }` but for multiple mappings it introduces additional ways on how the configuration could break, that would not happen with a table. So for each mapping you have a chance to introduce a typo or forget the mode etc.
2. This would be a now mapping interface which would differ from the current one used in the `defaults` tbl
3. So if we would do this style in both we would break backwards compatibility 
4. If we add it additionally we introduce possible ways how the configuration code could break and just add confusion to why we support multiple ways
5. I think we gain nothing from changing it. So nay from my side

dhruvmanila:
Wow, looking forward to configuring in this way!

Maybe we could provide a field to define the keymap for opening the picker? The user can also provide the default keymap options to pass. This would simplify defining the dozen key bindings for all the pickers ;)

```lua
require("telescope").setup {
  pickers = {
    buffers = {
      open_with = "<C-b>",
      ...
      }
    }
  }
}
```

Or, instead of mapping each picker individually, a table specifically for mapping all the pickers, similar to that of [gitsigns](https://github.com/lewis6991/gitsigns.nvim#usage):

```lua
require("telescope").setup {
  picker_keymaps = {
    noremap = true,
    silent = true,
    buffers = "<C-b>", 
    -- or ["<C-b>"] = "buffers",
    -- or ["<C-b>"] = require('telescope.builtin').buffers,
    -- or ["<C-b>"] = "<cmd>lua require('telescope.builtin').buffers()<cr>",
    }
  }
}
```

tjdevries:
Oh, this gave me an idea.

What if we also added the ability for mappings right hand side to just be the key from actions as a shorthand, for very simple setup for people not used to Lua.

So instead of `actions.delete_buffer`, it would instead be `"delete_buffer" `.

Thoughts?

lua/telescope/builtin/init.lua
```diff
@@ -322,4 +352,50 @@ builtin.lsp_document_diagnostics = require('telescope.builtin.lsp').diagnostics
...
+      pconf = require("telescope.themes")["get_" .. theme](pconf)
+    end
+    if pconf.mappings then
+      local mappings = pconf.mappings
+      pconf.mappings = nil
+      pconf.attach_mappings = function(_, map)
+        for mode, tbl in pairs(mappings) do
+          for key, action in pairs(tbl) do
+            if type(action) == "string" then
+              map(mode, key, require("telescope.actions")[action])
```
tjdevries, line 371: we should assert that this exists: `map(mode, key, assert("telescope.actions")[action], string.format("'%s' is not a valid action", action))`

lua/telescope/mappings.lua
```diff
@@ -191,7 +191,11 @@ mappings.apply_keymap = function(prompt_bufnr, attach_mappings, buffer_keymap)
       local key_bind_internal = a.nvim_replace_termcodes(key_bind, true, true, true)
       if not applied_mappings[mode][key_bind_internal] then
         applied_mappings[mode][key_bind_internal] = true
-        telescope_map(prompt_bufnr, mode, key_bind, key_func)
+        if type(key_func) == "string" then
```
tjdevries, line 194: maybe we should just actually do this check above at  the top of the function, inside of the map function we create on line 104 -- that way any time someone uses a string, it defaults to searching `actions`.

lua/telescope/builtin/init.lua
Conni2461: We kinda do in the `telescope.actions` table. You will get `error("Key does not exist for 'telescope.actions': " .. tostring(k))`.

If you want something additional here we can also do that

tjdevries: Ah, nice, I forgot that we already did that. Then nevermind :)

lua/telescope/mappings.lua
```diff
@@ -111,51 +111,58 @@ local telescope_map = function(prompt_bufnr, mode, key_bind, key_func, opts)
...
+        opts or {
+          silent = true
+        }
+      )
+      return
+    elseif key_func.type == "action_key" then
+      key_func = actions[key_func[1]]
+    end
+  end
+
```
Conni2461, line 131: Post here:

I only changed the indentation. From if-else to, if end and the `keymap` branch returns

dhruvmanila:
I have a question: I know that concatenating actions works because of `__add` but will it be the same for concatenating action strings?
```lua
mappings = {
  i = {
    -- This works
    ["<c-d>"] = actions.send_to_qflist + actions.open_qflist,
    -- But does won't work
    -- E5108: Error executing lua [string ":lua"]:1: attempt to perform arithmetic on a string value
    ["<c-d>"] = "send_to_qflist" + "open_qflist",
  }
}
```
Maybe for multiple actions, we can ask for a table of strings like so: `{ "send_to_qflist", "open_qflist" }` and the actions will be executed in that order.

Conni2461:
Okay the current interface that is implemented is:

```lua
mappings = {
  i = {
    -- current
    ["<c-d>"] = actions.smart_send_to_qflist + ....,
    -- same as
    ["<c-d>"] = { actions.smart_send_to_qflist + ...., type = "action" },

    -- is basicall this:

    -- for viml
  }
}
```

Plus for actions should still work. But i am not sure if we should do `"send_to_qflist" + "open_qflist"`, i will talk with tj about it

tjdevries:
> Wow, looking forward to configuring in this way!
> 
> Maybe we could provide a field to define the keymap for opening the picker? The user can also provide the default keymap options to pass. This would simplify defining the dozen key bindings for all the pickers ;)
> 
> ```lua
> require("telescope").setup {
>   pickers = {
>     buffers = {
>       open_with = "<C-b>",
>       ...
>       }
>     }
>   }
> }
> ```
> 
> Or, instead of mapping each picker individually, a table specifically for mapping all the pickers, similar to that of [gitsigns](https://github.com/lewis6991/gitsigns.nvim#usage):
> 
> ```lua
> require("telescope").setup {
>   picker_keymaps = {
>     noremap = true,
>     silent = true,
>     buffers = "<C-b>", 
>     -- or ["<C-b>"] = "buffers",
>     -- or ["<C-b>"] = require('telescope.builtin').buffers,
>     -- or ["<C-b>"] = "<cmd>lua require('telescope.builtin').buffers()<cr>",
>     }
>   }
> }
> ```

I think we could do something like this, but it's out of scope for this PR. In fact, what I'd prefer to do is probably wait til I get a little farther on something like https://github.com/tjdevries/vimterface.nvim and consider hooking up to that plugin rather than re-creating it all here.

lua/telescope/builtin/init.lua
```diff
@@ -322,4 +352,46 @@ builtin.lsp_document_diagnostics = require('telescope.builtin.lsp').diagnostics
...
+        end
+        return true
+      end
+    end
+    mod[k] = function(opts)
+      opts = opts or {}
+      if pconf.attach_mappings and opts.attach_mappings then
+        local attach_mappings = pconf.attach_mappings
+        pconf.attach_mappings = nil
+        local opts_attach = opts.attach_mappings
```
tjdevries, line 381: is it possible for this to be `nil`?

tjdevries: Oh, I should just read the code above LUL

Conni2461: There is an if check above but i can make it more explicit with `~= nil` because true would also pass here

Shatur:
@Conni2461, thank you very much, I really missed this feature.

Can we add the ability to specify a default `theme`?

After this PR we can do the following:

```lua
require("telescope").setup {
  pickers = {
    buffers = {
      theme = "dropdown",
    }
  }
}
```

But it would be nice to be able to also specify a theme for each buffer:

```lua
require("telescope").setup {
  defaults = {
    buffers = {
      theme = "dropdown",
    }
  }
}
```

Conni2461:
@Shatur Currently Layout Configuration is completely being reworked here #823 
This could be part of that PR as well. I suggested it at least

Issue #880, opened 2021-06-01
shorten_path: Support length and option to exlcude positions from shortening


tom-anders:
**Is your feature request related to a problem? Please describe.**
opts.shorten_path currently always shorten the folders to a single character. This can sometimes be confusing when working in a large project

**Describe the solution you'd like**
Say I have this file `some/long/path/file.txt`.
Currently it would get shortened to `s/l/p/file,txt`. I'd like the option to specify the length to which the path gets shortened, e.g. with lenghth=2 it would look like this: `so/lo/pa/file.txt`. There already seems to be such a feature in plenary.vim (https://github.com/nvim-lua/plenary.nvim/pull/78), but it's not used in Telescope yet I think.

Also, it would be nice to have an option to exlcude e.g. the first directory from the shortening, which would also be useful in large projects that are split up into modules. In my example above this would yield `some/l/p/file.txt`.

PR #886, opened 2021-06-02
feat: Add length option for `shorten_path`

l-kershaw:
WIP

Adding option to provide a length for `shorten_path`, not just a boolean, as suggested in #880.
Implemented using `Path:shorten` from plenary as the function in telescope had gotten out of date in comparison.

Todo:
- [x]  Update documentation

---

Feedback welcome 🙂

### DIFF
--- a/lua/telescope/path.lua
+++ b/lua/telescope/path.lua
@@ -21,29 +21,12 @@ path.make_relative = function(filepath, cwd)
   return filepath
 end
 
-path.shorten = (function()
-  if jit then
-    local ffi = require('ffi')
-    ffi.cdef [[
-    typedef unsigned char char_u;
-    char_u *shorten_dir(char_u *str);
-    ]]
-
-    return function(filepath)
-      if not filepath then
-        return filepath
-      end
-
-      local c_str = ffi.new("char[?]", #filepath + 1)
-      ffi.copy(c_str, filepath)
-      return ffi.string(ffi.C.shorten_dir(c_str))
-    end
-  else
-    return function(filepath)
-      return filepath
-    end
-  end
-end)()
+-- In most cases it is better to use `utils.path_shorten`
+-- as it handles cases for `len` being things other than
+-- a positive integer.
+path.shorten = function(filepath,len)
+  return require'plenary.path'.new(filepath):shorten(len)
+end
 
 path.normalize = function(filepath, cwd)
   filepath = path.make_relative(filepath, cwd)
--- a/lua/telescope/path.lua
+++ b/lua/telescope/path.lua
@@ -21,12 +21,29 @@ path.make_relative = function(filepath, cwd)
   return filepath
 end
 
--- In most cases it is better to use `utils.path_shorten`
--- as it handles cases for `len` being things other than
--- a positive integer.
-path.shorten = function(filepath,len)
-  return require'plenary.path'.new(filepath):shorten(len)
-end
+path.shorten = (function()
+  if jit then
+    local ffi = require('ffi')
+    ffi.cdef [[
+    typedef unsigned char char_u;
+    char_u *shorten_dir(char_u *str);
+    ]]
+
+    return function(filepath)
+      if not filepath then
+        return filepath
+      end
+
+      local c_str = ffi.new("char[?]", #filepath + 1)
+      ffi.copy(c_str, filepath)
+      return ffi.string(ffi.C.shorten_dir(c_str))
+    end
+  else
+    return function(filepath)
+      return filepath
+    end
+  end
+end)()
 
 path.normalize = function(filepath, cwd)
   filepath = path.make_relative(filepath, cwd)
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -222,6 +222,21 @@ telescope.setup({opts})                                    *telescope.setup()*
         - "row"
         - "closest"
 
+                                           *telescope.defaults.shorten_len*
+    shorten_len: ~
+        Determines the length of the shortened portions of a filepath.
+
+        e.g. for a path like
+          `alpha/beta/gamma/delta.txt`
+        setting `shorten_len = 1` will give a path like:
+          `a/b/g/delta.txt`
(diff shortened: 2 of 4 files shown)


Conni2461:
2 Things:
- It was always the idea to move from telescope.path to plenary.path and @anott03 has done a lot of work in plenary.path and here https://github.com/nvim-telescope/telescope.nvim/pull/473 I do not know why this PR was never finished, it was around that time when i took a break. So you would need to speak with him how everything should continue
- There is also this PR https://github.com/nvim-telescope/telescope.nvim/pull/839 which simplifies the whole display path thing and would make this PR way simpler. So i wanna at least wait till this is done.

Otherwise yes cool idea thanks :)

anott03:
@Conni2461 I don't really know what happened with #473 either - I ended up taking a bit of a break as well. If I'm not mistaken, it was before TJ refactored actions so there's likely some work to be done before we can merge it.

l-kershaw:
Ah, okay. I'll have a look at #473 and #839 🙂

caojoshua:
fyi https://github.com/nvim-telescope/telescope.nvim/pull/839 is merged. This PR is unblocked. This would be a nice change :)

l-kershaw:
I think we're also still waiting on #473 before this PR can proceed.
Although that PR hasn't been active in nearly a month, so not sure what is best to do at this point.

Conni2461:
Yeah i haven't received an answer. Hmm i am thinking of just finishing it. I think anott03 is a little bit frustrated with me. I haven't done the best job of helping with that PR.

If i have a evening of free time next week, i will spend it on 473. I dont wanna block stuff that is actually super cool :)

anott03:
@l-kershaw apologies for blocking you so long with #473! It's merged now so this should be good to go (I think Conni had some minor fixes he wanted to make – just double check that those are done). Let me know if you ever want help or feedback with this!

l-kershaw:
@anott03 no worries at all! Developer time is a scarce resource in open source projects, and I appreciate all of the time you spent on #473 🙂 

---

I've reimplemented the `shorten_len` option, but I'm not sure I'm happy with the way the user configures it.
Since #839 has implemented `path_display` as a sort of "flag list", we can't pass a length in with the option.
So the `shorten_len` is separate from the rest of the `path_display` stuff, which feels weird.

Does anyone have any thoughts on better ways to implement this option?

anott03:
> Does anyone have any thoughts on better ways to implement this option?

I'll back at my desk later tonight and I'll see if I can come up with any ideas then :)

lua/telescope/path.lua
```diff
@@ -21,29 +21,12 @@ path.make_relative = function(filepath, cwd)
...
-      ffi.copy(c_str, filepath)
-      return ffi.string(ffi.C.shorten_dir(c_str))
-    end
-  else
-    return function(filepath)
-      return filepath
-    end
-  end
-end)()
+-- In most cases it is better to use `utils.path_shorten`
```
Conni2461, line 24: I wouldn't update `telescope.path` anymore. Its deprecated now and we will remove it next week or so, with a bunch of other deprecated code. See `:help telescope.changelog` we will keep the `util.path_shorten` shorthand tho :)

Conni2461:
For interface i have 2 thoughts. `path_display = { "shorten", 2 }` for example (so index after shorten is potential len or `path_display = { "shorten" }` (still valid) + support for `path_display = { "shorten" = { len = 2 } }` and i would prefer len to be always a number, so we dont need `tonumber`. The util shorthand can then be only `Path:new(filename):shorten(len)` because plenary accepts either nil or a number > 1 and already does the assert https://github.com/nvim-lua/plenary.nvim/blob/8bae2c1fadc9ed5bfcfb5ecbd0c0c4d7d40cb974/lua/plenary/path.lua#L346-L352

Conni2461:
Thoughts on both interface ideas? You can also always ask tj if you are stuck, he will then make the decision :laughing:

lua/telescope/path.lua
l-kershaw: That's fair, I think this was a holdover from the previous attempt when I rebased. Have reverted this file now.

l-kershaw:
@Conni2461 I don't really like either of your suggested configuration methods, they both feel weird to me (not that my method is any better, mind you). There seem to be two contradictory goals with configuring this:

1. keeping all the configuration for how the path is displayed in the `path_display` object.
2. keeping the `path_display` object as some sort of "flag list" of strings.

I'm not sure what to do, since these goals don't seem compatible and I don't really know which one to sacrifice in favour of the other. @tjdevries do you have any thoughts on this?

---

Regarding the change to `utils.path_shorten` I agree that we should remove all of this logic from telescope as it is handled on the plenary side. However, since this basically amounts to `utils.path_shorten` being a straight wrapper (with no additional stuff) for `Path:shorten`, is it actually worth keeping `utils.path_shorten` in telescope at all?

anott03:
A quick project wide search for `path_shorten` only yielded one result – it's definition – so I see no reason to keep the function there (and even if there were uses, I'd say we should switch those to plenary and remove it from utils).

Conni2461:
We will keep the shorthand for now because we dont wanna break extensions, we can deprecate it later :)

In hindsight this `path_display = { shorten = { len = 2 } }` is stupid. We can also do this `path_display = { shorten = 2 }`.
I know that `path_display` is a list of strings but i kinda like to keep all the display path related stuff under one key and i am not saying remove the other version (list of strings)

Also we had a discussion here https://github.com/nvim-telescope/telescope.nvim/issues/961 where i said `path_display` could either be a table or a `function(opts, path)` so more advanced people have 100% control of the displaying

l-kershaw:
@Conni2461 I think that your `path_display = { shorten = 2 }` style is the best option so far.
I've implemented it and updated the docs accordingly. I also added a deprecation warning for `utils.path_shorten`.

I don't think there's a better way to deal with this option, so I'm happy to just keep it like this.

---

@Conni2461 @anott03 is there anything else you think I should change?

PR #890, opened 2021-06-03
feat: Enable individual options for mappings

waldson:
Currently `telescope_map()` has an unused parameter for passing [mapping](https://neovim.io/doc/user/map.html#:map-arguments) options for `nvim_buf_set_keymap`.  This PR allows mappings to have custom options:

```lua
require('telescope').setup{
    defaults = {
        mappings = {
            i = {
                ["<C-j>"] = {
                    action = actions.move_selection_next,
                    options = {
                        nowait = true,
                    }
                },
                ["<C-l>"] = {
                    -- prefix the query with the current buffer's directory
                    action = "nvim_feedkeys(nvim_replace_termcodes('<C-u>' .. expand('#:h') .. '/', v:true, v:false, v:true), 'i', v:true)",
                    options = {
                        expr = true
                    }
                },
                -- nothing is changed in regular mappings
                ["<C-c>"] = actions.close,
            }
        }
    }
}
``` 

The `silent` and `noremap` are enabled by default (like it is now).


Conni2461:
LUL life is weird sometimes. Nobody touched that code in half a year and not even a day after i talked with tj how we could change that and make it more easy there is a proposal to make it more powerful. Don't get me wrong. I like more powerful :rofl: 

But we need to find a good API for it: See this PR https://github.com/nvim-telescope/telescope.nvim/pull/883 and mostly this commit https://github.com/nvim-telescope/telescope.nvim/pull/883/commits/169342cf7c8c3a07a4dafb9905cce0faa4c7a951 from this morning (life is weird).

Our interface proposal was:
```lua
mappings = {
  i = {
    -- current
    ["<c-d>"] = actions.smart_send_to_qflist,

    -- is basicall this:

    -- for viml
  }
}
```

waldson:
Hey @Conni2461, thanks for the fast reply. I didn't saw that proposal and it seems way more robust than mine. Have you decided how the options will be handled there? Do you think a `options` key is viable? E.g:

```lua
mappings = {
  i = {
    ["<c-d>"] = { "smart_send_to_qflist", type = "action_key", options = { nowait = true }},
  }
}
```

Conni2461:
We have not talked about options yet. I think we should do it. It looks good to do it in the table (but personally i prefer opts > options, but thats just something super small). 
I would suggest that you can finish this PR after my initial new API is in master. But i also already wrote TJ just to check if he is fine with the API.

waldson:
I've just subscribed to your PR and I'll continue after it's merged. Thanks!

> opts > options

No problem with that. Actually, I think `opts` is better because it will have the same signature as `nvim_buf_set_keymap`'s.

rafi:
#883 has been merged.

tjdevries:
Do you need anything from me on this PR? I think if we do:

`    ["<c-d>"] = { "smart_send_to_qflist", type = "action_key", opts = { nowait = true }},`

That would be really cool. The only thing might be maybe instead of using `opts`, we should say `map_opts` or something to differentiate this from potential other things that might happen later. I don't want to limit ourselves to only configuring the options that are sent to `nvim_set_keymap` and it's not super obvious which `opts` this should be. So maybe `keymap_opts` or `map_opts` or something like that.

If we implement that, that would be really cool!

waldson:
Thanks @rafi and @tjdevries. Working on it now. I'm gonna use `keymap_opts` to be more clear.

waldson:
@tjdevries I've updated the code. Is there anything else that I should do?
 I can't test it because `telescope_map` is local. I was thinking about exposing an option just like `default_mappings` to be able to test it, but I'm not sure it will be that useful. 

Thanks.

Conni2461:
Do you need anything from me?

Conni2461:
finished it

PR #888, opened 2021-06-03
docs: Add documentation for function `buffers`

Gwarnifarn:
Two options from within the `buffers` function were not
documented. Added documentation for these two options.

PR #887, opened 2021-06-03
docs: update builtin.file_browser options

davidmh:
Removes the unsued `search_dirs` option, and adds `cwd` and `depth`.

PR #902, opened 2021-06-09
feat: add utility for fetching buffer directory

logan-connolly:
# Motivation

The [tpope/vim-vinegar plugin](https://github.com/tpope/vim-vinegar) offers great file/directory navigation and creation features. This is particularly helpful when you want to make edits to files in close proximity to your open buffer. 

It accomplishes this using the builtin `netrw` vim script. Telescope's `builtin.file_browser` does basically all I want in regards to file manipulation and navigation. The only thing that is missing is that it does not have the option of opening up in a *relative* location to your open buffer.

# Initial Solution

Add a `relative (boolean)` option to `builtin.file_browser` function. So when `true`, the file browser will open at current buffer's directory (overriding `cwd` option):

```lua
require('telescope.builtin').file_browser({ relative = true })
```

# Updated Solution

Add `buffer_dir()` utility function that fetches the directory of your current buffer, which can then be passed  to the `cwd` field in a picker. That way you can get this functionality anywhere `cwd` is supported.

```lua
require('telescope.builtin').file_browser({ cwd = require'telescope.utils'.buffer_dir() })
require('telescope.builtin').find_files({ cwd = require'telescope.utils'.buffer_dir() })
require('telescope.builtin').grep_string({ cwd = require'telescope.utils'.buffer_dir() }
-- etc
```

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -231,13 +231,7 @@ files.file_browser = function(opts)
   opts = opts or {}
 
   opts.depth = opts.depth or 1
-  opts.relative = opts.relative or false
   opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or vim.loop.cwd()
-
-  if opts.relative then
-    opts.cwd = utils.get_current_buffer_dir()
-  end
-
   opts.new_finder = opts.new_finder or function(path)
     opts.cwd = path
     local data = {}
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -231,7 +231,10 @@ files.file_browser = function(opts)
   opts = opts or {}
 
   opts.depth = opts.depth or 1
-  opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or vim.loop.cwd()
+	opts.relative = opts.relative or false
+	opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or vim.loop.cwd()
+	opts.cwd = utils.get_current_buffer_dir() and opts.relative or opts.cwd
+
   opts.new_finder = opts.new_finder or function(path)
     opts.cwd = path
     local data = {}
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -233,7 +233,10 @@ files.file_browser = function(opts)
   opts.depth = opts.depth or 1
 	opts.relative = opts.relative or false
 	opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or vim.loop.cwd()
-	opts.cwd = utils.get_current_buffer_dir() and opts.relative or opts.cwd
+
+	if opts.relative then
+		opts.cwd = utils.get_current_buffer_dir()
+	end
 
   opts.new_finder = opts.new_finder or function(path)
     opts.cwd = path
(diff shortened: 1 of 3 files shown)


logan-connolly:
The only thing I could not get working is the documentation generation. Did I understand correctly that it would be generated in a workflow job automatically?

tjdevries:
(I have to click magic button to prevent you from mining crypto on github's dime haha)

Conni2461:
> (I have to click magic button to prevent you from mining crypto on github's dime haha)

I don't think this is it. The docgen action should run in your fork, because otherwise we don't have permissions to write to your repo/branch. He figured it out https://github.com/nvim-telescope/telescope.nvim/pull/888#issuecomment-854135107 and he didn't https://github.com/nvim-telescope/telescope.nvim/pull/887 so i am not sure why its currently randomly broken for some people.

Also about the PR, i am not sure if we should do this. There is very little benefit over doing this `require('telescope.builtin').file_browser({ cwd = vim.fn.expand('%:p:h') })` but if he do it for `file_browser` we are indirectly forced to do it for all file related pickers. Thoughts @tjdevries ?

logan-connolly:
@Conni2461 thanks for the tip. I made sure read/write was allowed and forced pushed an amended commit. Perhaps that solves the docgen issue.

I agree it is a minor change and I considered not opening the pull request. However, then I considered how long it took me to figure out that `vim.fn.expand('%:p:h')` was the command I was looking for, and decided a layer of abstraction could be good. 

Especially for people who want to do telescope(y) things in the directory of their current buffer, which is why I added the `utils.get_current_buffer_dir()`. I am not necessarily sold on `relative` as the param name, but it is the best I could think of :smile:

Conni2461:
I see what you are saying. Its probably more user friendly. So i think we should add it to every picker that makes sense. Currently thinking about:
- find_files
- grep_string
- live_grep

logan-connolly:
Yeah, those pickers make sense to me as well, but I guess it should be decided if we _really_ need an additional option? Because `cwd` really is the only underlying option that is necessary, making `relative` seem redundant. And by selecting certain pickers now, users may ask for `relative` options in different pickers, which would be a hassle.

Maybe the solution is to just add the utility function for fetching current buffer directory and documenting how one can configure `cwd` to use this functionality:

```lua
local utils = require('telescope.utils')
require('telescope.builtin').file_browser({ cwd = utils.get_current_buffer_directory() })
```

I think this is a much more general solution to the actual problem of not knowing the vim command for fetching current buffer directory. What do you think?

logan-connolly:
@Conni2461 I went with the more general approach, updating   the PR  accordingly. I also shortened the name of the function to just `utilsl.buffer_dir()` and updated the documentation for those four pickers that you mentioned.

l-kershaw (commented):
Looks very helpful, thanks 🙂 

Only improvement I can think of would be adding docs for the `cwd` option in the other `builtin` pickers that accept it (I think it's only the "git_..." ones, but would be good to double check the rest).

After that, I'm happy to merge this (unless @Conni2461 had anything else they wanted doing here).

logan-connolly:
> Only improvement I can think of would be adding docs for the `cwd` option in the other `builtin` pickers that accept it (I think it's only the "git_..." ones, but would be good to double check the rest).

@l-kershaw thanks for the review! I double-checked to make sure that I documented the builtins where it made sense, and it looks like it really is just the git_ ones remaining.

Conni2461 (approved):
Thanks :)

PR #903, opened 2021-06-09
feat: add filtering by symbol(s) to lsp_{document,workspace}_symbols

### LINKED ISSUE #901: Add options to pre-filter LSP Document Symbols by type
I would like to be able to quickly jump to a method/function within the file I am currently in.

The builtin lsp_document_symbol viewer is a good start, but I would like to filter out all properties and constants as they drown out the functions I am actually searching for. Vista has the option to filter out specific types in it's finder and I would like to see the same option here: 
`let g:vista_ignore_kinds=["Variable", "Constant", "Property"] `

I've tried Vista with the finder option which is OK, but I would prefer to have the preview functionality of Telescope. Also Vista is often out of sync with the current buffer.

**Additional context**
Alternative solutions could be to just add the ability to provide the filters in the options when calling the function. Also, I would be happy to impliment this myself if I can get an example of how to filter out the results by symbol type.

Thanks!

fdschmidt93:
Closes #901 

If we have something like this for diagnostics eventually, might as well have it for lsp symbols I suppose.

e: thanks @Conni2461  -- docgen for multiple types seems to work nicely :)

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -82,7 +82,7 @@ utils.quickfix_items_to_entries = function(locations)
   return results
 end
 
-utils.filter_symbols = function(opts, results)
+utils.filter_symbols = function(results, opts)
   if opts.symbols == nil then
     return results
   end
@@ -92,34 +92,46 @@ utils.filter_symbols = function(opts, results)
   if type(opts.symbols) == "string" then
     opts.symbols = string.lower(opts.symbols)
     if vim.tbl_contains(valid_symbols, opts.symbols) then
-      for _, result in ipairs(results) do
+      for _, result in pairs(results) do
         if string.lower(result.kind) == opts.symbols then
           table.insert(filtered_symbols, result)
         end
       end
+    else
+      print(string.format("%s is not a valid symbol per `vim.lsp.protocol.SymbolKind`", opts.symbols))
     end
   elseif type(opts.symbols) == "table" then
     opts.symbols = vim.tbl_map(string.lower, opts.symbols)
-    for _, symbol in ipairs(opts.symbols) do
+    local mismatched_symbols = {}
+    for _, symbol in pairs(opts.symbols) do
       if vim.tbl_contains(valid_symbols, symbol) then
         for _, result in pairs(results) do
           if string.lower(result.kind) == symbol then
             table.insert(filtered_symbols, result)
           end
         end
+      else
+        table.insert(mismatched_symbols, symbol)
+        mismatched_symbols = table.concat(mismatched_symbols, ", ")
+        print(string.format("%s are not valid symbols per `vim.lsp.protocol.SymbolKind`", mismatched_symbols))
       end
     end
   else
-    return filtered_symbols
+    print("Please pass filtering symbols as either a string or a list of strings")
+    return
   end
 
+  local current_buf = vim.api.nvim_get_current_buf()
   if not vim.tbl_isempty(filtered_symbols) then
     -- filter adequately for workspace symbols
-    for _, symbol in ipairs(filtered_symbols) do
-      symbol['bufnr'] = vim.uri_to_bufnr(vim.uri_from_fname(symbol.filename))
+    local filename_to_bufnr = {}
+    for _, symbol in pairs(filtered_symbols) do
+      if filename_to_bufnr[symbol.filename] == nil then
+        filename_to_bufnr[symbol.filename] = vim.uri_to_bufnr(vim.uri_from_fname(symbol.filename))
+      end
+      symbol['bufnr'] = filename_to_bufnr[symbol.filename]
     end
     table.sort(filtered_symbols, function(a, b)
-      local current_buf = vim.api.nvim_get_current_buf()
       if a.bufnr == b.bufnr then
         return a.lnum < b.lnum
       end
@@ -131,8 +143,12 @@ utils.filter_symbols = function(opts, results)
       end
       return a.bufnr < b.bufnr
     end)
-  end
   return filtered_symbols
+  end
+  -- only account for string|table as function otherwise already printed message and returned nil
+  local symbols = type(opts.symbols) == 'string' and opts.symbols or table.concat(opts.symbols, ', ')
+  print(string.format("%s symbol(s) were not part of the query results", symbols))
+  return
 end
 
 local convert_diagnostic_type = function(severity)
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -82,6 +82,59 @@ utils.quickfix_items_to_entries = function(locations)
   return results
 end
 
+utils.filter_symbols = function(opts, results)
+  if opts.symbols == nil then
+    return results
+  end
+  local valid_symbols = vim.tbl_map(string.lower, vim.lsp.protocol.SymbolKind)
+
+  local filtered_symbols = {}
+  if type(opts.symbols) == "string" then
+    opts.symbols = string.lower(opts.symbols)
+    if vim.tbl_contains(valid_symbols, opts.symbols) then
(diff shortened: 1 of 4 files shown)


lua/telescope/utils.lua
```diff
@@ -82,6 +82,59 @@ utils.quickfix_items_to_entries = function(locations)
   return results
 end
 
+utils.filter_symbols = function(opts, results)
+  if opts.symbols == nil then
+      return results
```
Conni2461, line 87: Wrong formatting

Conni2461 (approved):
LGTM besides the two nits. Thanks :)

You have to rebase and can you also fix the linting error that was introduced in the previous PR i missed it. https://github.com/nvim-telescope/telescope.nvim/runs/2794891282?check_suite_focus=true Maybe we can trim the docs down by 5 chars. Sadly multiline annotations are not yet possible. (They kinda suck, either tj or i have on this sooner rather than later. Probably me :rofl:)

fdschmidt93:
Huh, is something run with CI? Linting didn't seem to be running on the other PR nor is any job being run here. Otherwise linter should also have complained about wrongful use of global I presume :thinking: 

Anyways, issues should now be fixed :)

lua/telescope/utils.lua
```diff
@@ -82,6 +82,59 @@ utils.quickfix_items_to_entries = function(locations)
   return results
 end
 
+utils.filter_symbols = function(opts, results)
```
Conni2461, line 85: i actually think `results, opts` makes more sense.

lua/telescope/utils.lua
```diff
@@ -82,6 +82,59 @@ utils.quickfix_items_to_entries = function(locations)
...
+  if opts.symbols == nil then
+    return results
+  end
+  local valid_symbols = vim.tbl_map(string.lower, vim.lsp.protocol.SymbolKind)
+
+  local filtered_symbols = {}
+  if type(opts.symbols) == "string" then
+    opts.symbols = string.lower(opts.symbols)
+    if vim.tbl_contains(valid_symbols, opts.symbols) then
```
Conni2461, line 94: Just rambling here: Does it make sense to give a small error message if you miss typed a symbol? For both string and table. I just had a typo and that resulted into no results -> no picker (which is weird) does this make sense?

lua/telescope/utils.lua
```diff
@@ -82,6 +82,59 @@ utils.quickfix_items_to_entries = function(locations)
...
+      if vim.tbl_contains(valid_symbols, symbol) then
+        for _, result in pairs(results) do
+          if string.lower(result.kind) == symbol then
+            table.insert(filtered_symbols, result)
+          end
+        end
+      end
+    end
+  else
+    return filtered_symbols
```
Conni2461, line 113: Okay if symbols is not a string or a table we shouldn't return the filtered_symbols in my opinion. Also not silently

lua/telescope/utils.lua
```diff
@@ -82,6 +82,59 @@ utils.quickfix_items_to_entries = function(locations)
...
+        if string.lower(result.kind) == opts.symbols then
+          table.insert(filtered_symbols, result)
+        end
+      end
+    end
+  elseif type(opts.symbols) == "table" then
+    opts.symbols = vim.tbl_map(string.lower, opts.symbols)
+    for _, symbol in ipairs(opts.symbols) do
+      if vim.tbl_contains(valid_symbols, symbol) then
+        for _, result in pairs(results) do
```
Conni2461, line 105: `ipairs`

lua/telescope/utils.lua
```diff
@@ -82,6 +82,59 @@ utils.quickfix_items_to_entries = function(locations)
...
+      end
+    end
+  else
+    return filtered_symbols
+  end
+
+  if not vim.tbl_isempty(filtered_symbols) then
+    -- filter adequately for workspace symbols
+    for _, symbol in ipairs(filtered_symbols) do
+      symbol['bufnr'] = vim.uri_to_bufnr(vim.uri_from_fname(symbol.filename))
```
Conni2461, line 119: Hmm do we want to cache `vim.uri_to_bufnr(vim.uri_from_fname(symbol.filename))` (filename -> bufnr). This might happen a lot

lua/telescope/utils.lua
```diff
@@ -82,6 +82,59 @@ utils.quickfix_items_to_entries = function(locations)
...
+    end
+    table.sort(filtered_symbols, function(a, b)
+      local current_buf = vim.api.nvim_get_current_buf()
```
Conni2461, line 122: Its weird to get the `current_buf` for each comparison. Can we do it outside of the function

Conni2461:
Sorry i have more nits. Should have seen that earlier

fdschmidt93:
No worries, I was a bit tipsy yesterday (like I'm now) :laughing: as I have lots of work and just wanted to distract myself a bit with open source :sweat_smile: 

I hopefully now adequately account for errors, but one "open" scenario remains --

Consider a scenario for which we
* We have matched valid symbols -> results shown
* We have invalid symbols -> message printed
* We have unmatched valid symbols -> omitted silently for the time being

On the other hand, consider
* We have invalid symbols -> message printed
* No matched valid symbols -> message printed

I think this is the right level of verbosity (only communicating valid symbols if no valid symbols has been matched), as in that's the case the user expects.

lua/telescope/utils.lua
```diff
@@ -82,6 +82,75 @@ utils.quickfix_items_to_entries = function(locations)
...
+  if opts.symbols == nil then
+    return results
+  end
+  local valid_symbols = vim.tbl_map(string.lower, vim.lsp.protocol.SymbolKind)
+
+  local filtered_symbols = {}
+  if type(opts.symbols) == "string" then
+    opts.symbols = string.lower(opts.symbols)
+    if vim.tbl_contains(valid_symbols, opts.symbols) then
+      for _, result in pairs(results) do
```
Conni2461, line 95: Okay again why is this a `pairs` over a `ipairs`? Isn't `results` a list? `1 ... n` so `ipairs` should be better

lua/telescope/utils.lua
```diff
@@ -149,7 +218,7 @@ utils.diagnostics_to_tbl = function(opts)
   local buffer_diags = opts.get_all and vim.lsp.diagnostic.get_all() or
     {[current_buf] = vim.lsp.diagnostic.get(current_buf, opts.client_id)}
   for bufnr, diags in pairs(buffer_diags) do
-    for _, diag in pairs(diags) do
+    for _, diag in ipairs(diags) do
```
Conni2461, line 221: I actually think this occasion should be `pairs`.

Short explanation:
`ipairs` does this:
- start from idx 1 til tbl[idx] == nil
- if `buffer_diags now: `{[current_buf] = vim.lsp.diagnostic.get(current_buf, opts.client_id)}` and `current_buf ~= 1` `ipairs` does nothing.
- `pairs` will iterate over all elements in the table, just not sorted.
- So `ipairs` if it is a list starting at 1 ... n else `pairs`

Conni2461:
LGTM thanks :)

thraizz:
How much work would it be to have this lovely feature for dynamic workspace symbols? I might try to make this my first PR that is not just docs :joy_cat:

fdschmidt93:
I think the below snippet
https://github.com/nvim-telescope/telescope.nvim/blob/162cf7b5343eaf97c32c97af6a37452f9a3a9ca1/lua/telescope/builtin/lsp.lua#L280-L292

needs to be applied here right thereafter before returning `locations`:
https://github.com/nvim-telescope/telescope.nvim/blob/162cf7b5343eaf97c32c97af6a37452f9a3a9ca1/lua/telescope/builtin/lsp.lua#L328

PR #904, opened 2021-06-10
Fix navigation link for LSP pickers

jpe90:
Fixes the broken "LSP" hyperlink in the "built in support" section of the README


Conni2461:
Thanks :)

Issue #907, opened 2021-06-12
Ripgrep 13 breaks telescope


clason:
This is just a short note in case other people hit this: Ripgrep has made some changes in version 13 (which has just hit homebrew); one of them seems to have broken `:Telescope grep_string` and friends (no matches show up). Not sure yet what is to be done.

It's not the `--vimgrep` change, likely as this looked at first glance. Best guess right now is

> PERF https://github.com/BurntSushi/ripgrep/discussions/1657: Check if a file should be ignored first before issuing stat calls.

clason:
@Conni2461's fix indeed fixes the main issue. There's another regression in ripgrep that was responsible for my breakage: I had a grep that had telescope do _ALL_ the filtering via
```
builtin.grep_string { shorten_path = true, word_match = "-w", only_sort_text = true, search = '' }
```
and the `-w` in combination with `''` now produces a segfault:
```
thread '<unnamed>' panicked at 'assertion failed: self.start <= end', /private/tmp/ripgrep-20210612-52810-yuhb53/ripgrep-13.0.0/crates/matcher/src/lib.rs:131:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at 'assertion failed: self.start <= end', /private/tmp/ripgrep-20210612-52810-yuhb53/ripgrep-13.0.0/crates/matcher/src/lib.rs:131:9
thread '<unnamed>' panicked at 'assertion failed: self.start <= end', /private/tmp/ripgrep-20210612-52810-yuhb53/ripgrep-13.0.0/crates/matcher/src/lib.rs:131:9
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Any', crates/ignore/src/walk.rs:1302:31
thread '<unnamed>' panicked at 'assertion failed: self.start <= end', /private/tmp/ripgrep-20210612-52810-yuhb53/ripgrep-13.0.0/crates/matcher/src/lib.rs:131:9
thread '<unnamed>' panicked at 'assertion failed: self.start <= end', /private/tmp/ripgrep-20210612-52810-yuhb53/ripgrep-13.0.0/crates/matcher/src/lib.rs:131:9
thread '<unnamed>' panicked at 'assertion failed: self.start <= end', /private/tmp/ripgrep-20210612-52810-yuhb53/ripgrep-13.0.0/crates/matcher/src/lib.rs:131:9
[...]

PR #908, opened 2021-06-12
fix: ripgrep 13

Conni2461:
This at least fixes things for linux.

I don't know whats wrong with mac.

See https://github.com/nvim-telescope/telescope.nvim/issues/907


clason:
Confirmed on macOS as well.

PR #913, opened 2021-06-14
fix: make_relative doesn't check path sep

bi0ha2ard:
A while ago I noticed a bug with find_file with a cwd argument set:

If you're editing `/tmp/foo/a`
and call `:Telescope find_files cwd /tmp/foo_bar` and select a file from that folder, it will edit that file as `bar/file`.

I've tracked it down to make_relative not checking for the path separator after the cwd properly:

```
:lua print(require'telescope.path'.make_relative("/tmp/foo_bar/b", "/tmp/foo"))
```
prints `bar/b`, while it should not find a relative path. The issue is that a path can only be made relative if the character at the `_` is a path separator.

Out of curiosity, i checked `plenary.path`, and it seems to have the same bug:

```
:lua print(require'plenary.path':new("/tmp/foo_bar/b"):make_relative("/tmp/foo/"))
```

prints `bar/b`

### DIFF
--- a/lua/telescope/path.lua
+++ b/lua/telescope/path.lua
@@ -7,17 +7,11 @@ path.home = vim.fn.expand("~")
 path.make_relative = function(filepath, cwd)
   if not cwd or not filepath then return filepath end
 
+  if cwd:sub(#cwd, #cwd) ~= path.separator then
+    cwd = cwd .. path.separator
+  end
   if filepath:sub(1, #cwd) == cwd  then
-    local offset =  0
-    -- if  cwd does ends in the os separator, we need to take it off
-    if cwd:sub(#cwd, #cwd) ~= path.separator then
-      offset = 1
-    end
-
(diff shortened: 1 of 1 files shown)


Conni2461:
If `plenary.path` has the same bug i personally think this should be fixed there because we are planing and working on replacing `telescope.path` with `plenary.path` here https://github.com/nvim-telescope/telescope.nvim/pull/473 I should help finish it asap.

I can merge stuff in plenary as well, and i think that fix looks good. Make sure you add a test case, when fixing it in plenary ;)

Good catch btw

bi0ha2ard:
Yep, I opened https://github.com/nvim-lua/plenary.nvim/pull/168 which should fix it in plenary.

Conni2461:
Thanks for fixing this in plenary. I am closing this here and try to get #473 done 

Thanks again :)

PR #922, opened 2021-06-17
feat: Consistent and sensible layout_config

tjdevries:
Moving to this PR from https://github.com/nvim-telescope/telescope.nvim/pull/823 because of some problems with being able to merge back, but I didn't want to lose the history from that PR.

This should also hopefully make it easier for people to comment on the PR and give us some feedback before merging.

cc @l-kershaw @Conni2461

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -23,7 +23,8 @@ telescope.setup({opts})                                    *telescope.setup()*
     default_mappings: ~
         Not recommended to use except for advanced users.
 
-        Will allow you to completely remove all of telescope's default maps and use your own.
+        Will allow you to completely remove all of telescope's default maps
+        and use your own.
 
 
                                  *telescope.defaults.dynamic_preview_title*
@@ -85,7 +86,8 @@ telescope.setup({opts})                                    *telescope.setup()*
           mode = { ..keys }
         }
 
-        where {mode} is the one character letter for a mode ('i' for insert, 'n' for normal).
+        where {mode} is the one character letter for a mode
+        ('i' for insert, 'n' for normal).
 
         For example:
 
@@ -97,20 +99,20 @@ telescope.setup({opts})                                    *telescope.setup()*
 
 
         To disable a keymap, put [map] = false
-              So, to not map "<C-n>", just put
+          So, to not map "<C-n>", just put
 
-                  ...,
-                  ["<C-n>"] = false,
-                  ...,
+              ...,
+              ["<C-n>"] = false,
+              ...,
 
-              Into your config.
+          Into your config.
 
 
         otherwise, just set the mapping to the function that you want it to be.
 
-                  ...,
-                  ["<C-i>"] = actions.select_default
-                  ...,
+            ...,
+            ["<C-i>"] = actions.select_default
+            ...,
 
 
                                          *telescope.defaults.prompt_prefix*
@@ -1211,10 +1213,11 @@ resolver.resolve_height()                          *resolver.resolve_height()*
        This means total height as a fixed number.
     3. function
        Must have signature: function(self, max_columns, max_lines): number
-    4. table of the form: {padding = `foo`} <br> where `foo` has one of the
-       previous three forms. <br> The height is then set to be the remaining
-       space after padding. For example, if the window has height 50, and the
-       input is {padding = 5}, the height returned will be `40 = 50 - 2*5`
+    4. table of the form: {padding = `foo`}
+       where `foo` has one of the previous three forms.
+       The height is then set to be the remaining space after padding. For
+       example, if the window has height 50, and the input is {padding = 5}, 
+       the height returned will be `40 = 50 - 2*5`
 
     The returned function will have signature: function(self, max_columns,
     max_lines): number
@@ -1230,10 +1233,11 @@ resolver.resolve_width()                            *resolver.resolve_width()*
        This means total width as a fixed number.
     3. function
        Must have signature: function(self, max_columns, max_lines): number
-    4. table of the form: {padding = `foo`} <br> where `foo` has one of the
-       previous three forms. <br> The width is then set to be the remaining
-       space after padding. For example, if the window has width 100, and the
-       input is {padding = 5}, the width returned will be `90 = 100 - 2*5`
+    4. table of the form: {padding = `foo`}
+       where `foo` has one of the previous three forms.
+       The width is then set to be the remaining space after padding. For
+       example, if the window has width 100, and the input is {padding = 5}, 
+       the width returned will be `90 = 100 - 2*5`
 
     The returned function will have signature: function(self, max_columns,
     max_lines): number
@@ -1630,7 +1634,7 @@ layout_strategies.horizontal()                *layout_strategies.horizontal()*
         - How tall to make Telescope's entire layout
         - See |resolver.resolve_height()|
       - mirror: Flip the location of the results/prompt and preview windows
-      - scroll_speed: The speed when scrolling through the previewer
+      - scroll_speed: The number of lines to scroll through the previewer
       - width:
         - How wide to make Telescope's entire layout
         - See |resolver.resolve_width()|
@@ -1673,7 +1677,7 @@ layout_strategies.center()                        *layout_strategies.center()*
         - How tall to make Telescope's entire layout
         - See |resolver.resolve_height()|
       - mirror: Flip the location of the results/prompt and preview windows
-      - scroll_speed: The speed when scrolling through the previewer
+      - scroll_speed: The number of lines to scroll through the previewer
       - width:
         - How wide to make Telescope's entire layout
         - See |resolver.resolve_width()|
@@ -1708,7 +1712,7 @@ layout_strategies.vertical()                    *layout_strategies.vertical()*
         - How tall to make Telescope's entire layout
         - See |resolver.resolve_height()|
       - mirror: Flip the location of the results/prompt and preview windows
-      - scroll_speed: The speed when scrolling through the previewer
+      - scroll_speed: The number of lines to scroll through the previewer
       - width:
         - How wide to make Telescope's entire layout
         - See |resolver.resolve_width()|
@@ -1734,7 +1738,7 @@ layout_strategies.flex()                            *layout_strategies.flex()*
         - How tall to make Telescope's entire layout
         - See |resolver.resolve_height()|
       - mirror: Flip the location of the results/prompt and preview windows
-      - scroll_speed: The speed when scrolling through the previewer
+      - scroll_speed: The number of lines to scroll through the previewer
       - width:
         - How wide to make Telescope's entire layout
         - See |resolver.resolve_width()|
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1067,8 +1067,8 @@ builtin.lsp_document_symbols({opts})          *builtin.lsp_document_symbols()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {ignore_filename} (type)    string with file to ignore
-        {symbols}         (string)  table
+        {ignore_filename} (type)          string with file to ignore
+        {symbols}         (string|table)  filter results by symbol kind(s)
 
 
 builtin.lsp_workspace_symbols({opts})        *builtin.lsp_workspace_symbols()*
@@ -1082,13 +1082,14 @@ builtin.lsp_workspace_symbols({opts})        *builtin.lsp_workspace_symbols()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {shorten_path}    (boolean)  if true, makes file paths shown in picker
-                                     use one letter for folders (default is
-                                     false)
-        {ignore_filename} (string)   file(s) to ignore
-        {hide_filename}   (boolean)  if true, hides the name of the file in the
-                                     current picker (default is false)
-        {symbols}         (string)   table
+        {shorten_path}    (boolean)       if true, makes file paths shown in
+                                          picker use one letter for folders
+                                          (default is false)
+        {ignore_filename} (string)        file(s) to ignore
+        {hide_filename}   (boolean)       if true, hides the name of the file
+                                          in the current picker (default is
+                                          false)
+        {symbols}         (string|table)  filter results by symbol kind(s)
 
 
 builtin.lsp_dynamic_workspace_symbols({opts})*builtin.lsp_dynamic_workspace_symbols()*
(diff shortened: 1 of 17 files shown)


lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -529,6 +529,23 @@ previewers.git_branch_log = defaulter(function(opts)
...
+        value = entry.value,
+        bufname = self.state.bufname,
+        cwd = opts.cwd
+      })
+      putils.regex_highlighter(self.state.bufnr, 'diff')
+    end
+  }
+end, {})
+
+previewers.git_commit_diff = defaulter(function(opts)
```
Conni2461, line 548: This whole block needs to go. Probably a automerge issue. It reimplements `git_commit_diff` which is no longer a thing. (And it actually uses a stash call :rofl:)

lua/telescope/deprecated.lua
```diff
@@ -0,0 +1,46 @@
+
+local deprecated = {}
```
Conni2461, line 2: deprecated is currently only checked for options directly passed to a builtin (and then picker) but we could also check if the setup table has one of the deprecated keys. Help people clean that up ;)

lua/telescope/config/resolve.lua
```diff
@@ -88,46 +93,81 @@ That's the next step to scrolling.
...
+--- The input must take one of four forms:
+--- 1. 0 <= number < 1 <br>
+---     This means total height as a percentage.
+--- 2. 1 <= number <br>
+---     This means total height as a fixed number.
+--- 3. function <br>
+---     Must have signature:
+---       function(self, max_columns, max_lines): number
+--- 4. table of the form:
+---       {padding = `foo`} <br>
```
Conni2461, line 163: This resolves in "stupid" `<br>` in the docgen output. I never said my formatter is prefect :rofl: but these two lines will wrap anyways and the issue is fixed if you put the `{padding = `foo`} <br>` immediatly after the `:`

lua/telescope/config/resolve.lua
```diff
@@ -138,6 +178,24 @@ resolver.resolve_height = function(val)
...
+--- The input must take one of four forms:
+--- 1. 0 <= number < 1 <br>
+---     This means total width as a percentage.
+--- 2. 1 <= number <br>
+---     This means total width as a fixed number.
+--- 3. function <br>
+---     Must have signature:
+---       function(self, max_columns, max_lines): number
+--- 4. table of the form:
+---       {padding = `foo`} <br>
```
Conni2461, line 191: Same as above :)

doc/telescope.txt
```diff
@@ -1118,6 +1197,56 @@ action_state.get_current_picker({prompt_bufnr})*action_state.get_current_picker(
...
+       input is {padding = 5}, the width returned will be `90 = 100 - 2*5`
+
+    The returned function will have signature: function(self, max_columns,
+    max_lines): number
+
+
+
+resolver.win_option()                                  *resolver.win_option()*
+    Win option always returns a table with preview, results, and prompt. It
+    handles many different ways. Some examples are as follows:
```
Conni2461, line 1245: Here is a lot more documentation in the source file. Not really important for the enduser so we could either make the two lines into only 2 `-` or we make the other 19 lines into 3 `-` (with some `<br>`s) up to you

lua/telescope/config.lua
```diff
@@ -15,13 +16,36 @@ local function first_non_null(...)
...
+end
+
+local sorters = require "telescope.sorters"
 
 -- TODO: Add other major configuration points here.
 -- selection_strategy
 
 local config = {}
 
+config.smarter_depth_2_extend = smarter_depth_2_extend
```
Conni2461, line 47: do we need this? I don't see the function used outside of `config` but if we do at some point we should put it in utils instead

lua/telescope/config.lua
```diff
@@ -15,13 +16,36 @@ local function first_non_null(...)
...
+      for k, v in pairs(val) do
+        result[key][k] = first_non_null(priority[k], v)
+      end
+    end
+  end
+  result = vim.tbl_deep_extend("keep", priority, result)
+  return result
+end
+
+local sorters = require "telescope.sorters"
```
Conni2461, line 40: Maybe we should just have all requires at the top

lua/telescope/config.lua
```diff
@@ -34,131 +58,273 @@ function config.set_pickers(pickers)
...
+      end
+
+      return string.format("%s / %s", xx, yy)
+    end,
+  },
+
+  dynamic_preview_title = {
+    false,
+    [[
+Will change the title of the preview window dynamically, where it
```
Conni2461, line 187: Some of the descriptions are not indented (I am not the biggest fan of starting at col 1 here) :smile:

lua/telescope/config.lua
```diff
@@ -34,131 +58,273 @@ function config.set_pickers(pickers)
...
+  mappings = {
+    {}, [[
+Your mappings to override telescope's default mappings.
+
+Format is:
+{
+  mode = { ..keys }
+}
+
+where {mode} is the one character letter for a mode ('i' for insert, 'n' for normal).
```
Conni2461, line 218: Longer than 80

lua/telescope/config.lua
```diff
@@ -34,131 +58,273 @@ function config.set_pickers(pickers)
...
+          ...,
+
+      Into your config.
+
+
+otherwise, just set the mapping to the function that you want it to be.
+
+          ...,
+          ["<C-i>"] = actions.select_default
+          ...,
```
Conni2461, line 243: This whole part is weirdly indented. I was also thinking maybe we should make this more complete.
Like in the builtin.init i put in the pickers configuration where i showed off more mappings:
```lua
      mappings = {
        i = {
          ["<c-d>"] = require("telescope.actions").delete_buffer,
          -- or right hand side can also be a the name of the action as string
          ["<c-d>"] = "delete_buffer",
        },
      }
```

Maybe we should also show the string option here and e.g.
```lua
["jj"] = { "<esc>", type = "command" },
["kk"] = { "<cmd>echo \"Hello, World!\"<cr>", type = "command" },
```

lua/telescope/config.lua
```diff
@@ -34,131 +58,273 @@ function config.set_pickers(pickers)
...
+]],
+  },
+
+  default_mappings = {
+    nil,
+    [[
+Not recommended to use except for advanced users.
+
+Will allow you to completely remove all of telescope's default maps and use your own.
```
Conni2461, line 252: To long

lua/telescope/config.lua
```diff
@@ -34,131 +58,273 @@ function config.set_pickers(pickers)
...
+]],
+  },
+
+  generic_sorter = { sorters.get_generic_fuzzy_sorter },
+  prefilter_sorter = { sorters.prefilter },
+  file_sorter = { sorters.get_fuzzy_file },
```
Conni2461, line 258: Stupid question. Do we want to switch to lua fzy as default sorter? Its faster, matches better? I know this is unrelated, currently just thinking about it :rofl:

lua/telescope/config.lua
```diff
@@ -34,131 +58,273 @@ function config.set_pickers(pickers)
...
-    Default: '  ']])
-  set("initial_mode", "insert")
+    Default: '  ']]
+  },
+
+  initial_mode = { "insert" },
 
-  set("border", {})
-  set("borderchars", { '─', '│', '─', '│', '╭', '╮', '╯', '╰'})
+  border = { {} },
```
Conni2461, line 168: Are we doing anything with that. I am currently confused by that.

If this thing is nil we are not calling the local `get_border_size` in layout_strategies.lua function and if i am using a theme (dropdown) its true, so `get_border_size` will be called. I can set it to false, obv and that works as expected but maybe we should set it to true on default. And give it a description. Thoughts?!

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -17,129 +17,249 @@
...
--- there are a few things that contribute to the border
-local is_borderless = function(opts)
-  return opts.window.border == false
+  return valid_configuration_keys
 end
 
+--@param strategy_name string: the name of the layout_strategy we are validating for
+--@param configuration table: table with keys for each option available
+--@param values table: table containing all of the non-default options we want to set
+--@param default_layout_config: table with the default values to configure layouts
```
Conni2461, line 74: Missing type.

Also we can have real emmy documentation for local functions

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -17,129 +17,249 @@
...
+--@return table: table containing the combined options (defaults and non-defaults)
+local function validate_layout_config(strategy_name, configuration, values, default_layout_config)
+  assert(strategy_name, "It is required to have a strategy name for validation.")
+  local valid_configuration_keys = get_valid_configuration_keys(configuration)
+
+  -- If no default_layout_config provided, check Telescope's config values
+  default_layout_config = default_layout_config or require('telescope.config').values.layout_config
```
Conni2461, line 81: For stuff like that we should probably slowly move to `vim.F.if_nil` Just to make it even more robust

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -17,129 +17,249 @@
...
+      end
+    end
+
+    return val
+  end
+
+  -- Always set the values passed first.
+  for k in pairs(values) do
+    if not valid_configuration_keys[k] then
+      -- TODO: At some popint we'll move to error here,
```
Conni2461, line 120: typo

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -17,129 +17,249 @@
...
+  return result
 end
 
-local layout_strategies = {}
+-- List of options that are shared by more than one layout.
+local shared_options = {
+  width = { "How wide to make Telescope's entire layout", "See |resolver.resolve_width()|" },
+  height = { "How tall to make Telescope's entire layout", "See |resolver.resolve_height()|" },
+  mirror = "Flip the location of the results/prompt and preview windows",
+  scroll_speed = "The speed when scrolling through the previewer",
```
Conni2461, line 146: I don't think this is good explained. Maybe "amount of lines to scroll"

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -414,7 +554,73 @@ layout_strategies.current_buffer = function(self, _, _)
     results = results,
     prompt = prompt,
   }
-end
+end)
+
+--- Bottom pane can be used to create layouts similar to "ivy".
+---
+--- For an easy ivy configuration, see |themes.get_ivy()|
+layout_strategies.bottom_pane = make_documented_layout('bottom_pane', vim.tbl_extend("error", shared_options, {
```
Conni2461, line 562: `bottom_pane` exists twice. They look identical on first glance. But one should still be removed :)

Conni2461 (changes requested):
I am not really proud of that review. I nitpicked way to much about comments and documentation. And had not really much to say about code, because its already really really good. Also i need way more time to get into how the whole math works to make any valid comments about that.

The two main things that needs to be fixed is remove duplicated code:
- remove a mysterious new duplicated previewer with new name :rofl: 
- remove one of the duplicated `bottom_pane` in layout_strategies

That are the two things that need to be fixed before merge. All other things can be address at a later time. :)

I'll run the PR for the next week and will look for issues

lua/telescope/config.lua
l-kershaw: I think this was a leftover from my initial attempts at setting up tests in `layout_strategies_spec.lua`.
It can be safely removed.

lua/telescope/pickers/layout_strategies.lua
l-kershaw: I think "The number of lines to scroll through the previewer" is the clearest way to phrase it

doc/telescope.txt
l-kershaw: I agree that this is not as important for the end user, so would be fine to get rid if we want.

l-kershaw:
@Conni2461
I agree with most of your suggestions, even if they are "nitpicky".
It's better to nitpick before we implement breaking changes to minimise confusion when it goes live.

I've 👍 for stuff that I simply agree with and commented where I have more to say.

---

Since we've moved to a new PR, I no longer have write access, so can't implement any of your suggested changes.
So the changes will either have to wait for @tjdevries to do them (not sure exactly when he's going away, but probably busy getting ready to go if not already away) or for someone else with write access to this branch.

tjdevries:
Oh dang, I forgot that you wouldn't be able to write to this. Sorry, if I get on an actual computer this week I'll see if I can figure out a way to do that 😂

Conni2461:
@tjdevries i don't see a way to only add him to one branch/PR, so we can either add him as collaborator to the repository or make him a member of the org. Its up to you :) I can add him, if you don't find a PC ;)

tjdevries:
Yeah, go ahead and add. 😁 Thanks.

l-kershaw:
Thanks for adding me to the org :)

---

I've sorted some of the simpler issues that @Conni2461 mentioned.
Some of the other ones I'm not sure the best way to deal with and/or need more discussion.

lua/telescope/config.lua
l-kershaw: I think that this should probably be just a boolean.

I've changed the default to `true` and added some documentation.
I'm happy with the situation now, but will wait for someone else to "Resolve" once they're happy with it 🙂

l-kershaw:
I've implemented some of the other changes requested now.

---

The only two remaining are:
- checking `deprecated` at setup
- switching to "lua fzy" as the default sorter

I will have a think about how best to check `deprecated`  in setup over the next couple of days.
I don't have any strong feelings on changing the default sorter.

lua/telescope/deprecated.lua
```diff
@@ -0,0 +1,59 @@
...
+  if #messages > 0 then
+    table.insert(messages, 1, "Deprecated window options. Please see ':help telescope.changelog'")
+    vim.api.nvim_err_write(table.concat(messages, "\n \n   ") .. "\n \nPress <Enter> to continue\n")
+  end
+end
+
+deprecated.layout_configuration = function(user_defaults)
+  if user_defaults.layout_defaults then
+    if user_defaults.layout_config == nil then
+      log.warn "Using 'layout_defaults' in setup() is deprecated. Use 'layout_config' instead."
```
l-kershaw, line 50: We currently use both `nvim_err_write` and `log.warn` for deprecation messages.
I think we might also use `log.info` elsewhere (despite our log level being at "warn") and also just straight `error`.

Do we want to standardise which method we are using?

l-kershaw:
@tjdevries @Conni2461
Do you guys have any thoughts on standardising our deprecation warnings (and if so, to what)?

Also, is there anything else that we want to change in this PR before it is ready to merge?

Conni2461:
Yeah we can talk about standardising for error messages and warning but this is out of scope for this PR. For the deprecated.lua file i prefer one way of doing it, i have no preferences in which way. For other deprecated messages we have, for example in `actions/init.lua` we currently use `error` but tj already removed them in one PR that is not finished (don't ask me for the number :rofl:)

I have not looked at it yet again but i didn't have big issues while using it. I 100% prefer to get this merged before neovim 0.5 gets released. It would be awful if new people check this out, have a config they feel comfortable with and we change everything within a couple of days.

I could take another look at the source code tomorrow morning :)

tjdevries:
I plan on merging this today or tomorrow. :)

Conni2461 (approved):

dhruvmanila:
So, I finally tried this and there seems to be one issue, I could be wrong.

<details>
<summary>Minimal config</summary>

```vim
map Q <nop>

" Download the plugin manager
  silent !curl -fsSLo /tmp/plug.vim
endif

source /tmp/plug.vim

" Plugins to test
Plug 'nvim-telescope/telescope.nvim', {'branch': 'layout_config'}

call plug#end()

" Load the plugins (packadd <plugin_name>)
packadd telescope.nvim

" For lua only plugins
lua << EOF

require("telescope").setup {
  defaults = {
    layout_config = {
      horizontal = {
        -- preview_width = 0.55,
      },
    },
  },
  pickers = {
    help_tags = {
      layout_config = {
        preview_width = 0.65,
      },
    },
  }
}

EOF

```

</details>

If I have the `defaults.layout_config.horizontal.preview_width` set then the `pickers.help_tags.layout_config.preview_width` is not being picked up. Settings for individual pickers should override the default one, right?

Using the above config and opening `Telescope help_tags` will give you the appropriate `preview_width` configured for that picker, but when you uncomment the line in the `defaults` table, the default preview_width will be taken into calculations.

| | |
| --- | ----------- |
| Screenshot 2021-07-01 at 14 54 18  |  Screenshot 2021-07-01 at 14 53 59 |

Conni2461:
@l-kershaw thanks for your work here :)

Issue #42, opened 2020-09-08
Allow "scrolling" through results in the list. 


rockerBOO:
Right now you are limited to the results in the list. 

If I am on vertical I am only showing 8 results by default. 

I sometimes use the fuzzy to give me a short list to browse through. 

![](https://i.imgur.com/4i8UZIy.png)

## Solutions

Allow us to go up on the list beyond the presented list.

I think limiting the the total results would be fine. 100 total results available for instance. Unless we want to encourage fuzzy browsing which then no limit (high limit) might be good.

clason:
You know, until now I didn't realize that you _couldn't_ scroll through the results list...

(I think I'd rather do lazy loading, i.e., if you scroll to the top-most result, retrigger the search with the current typed pattern and cut off at `2*results.height`.)

tjdevries:
Yes, it's a good idea. It might require a bit of time to think about. I have implemented `lines_manager` to (potentially) be able to handle this kind of thing at some point in the future.

I think maybe first step would be:
1. make it so that we can have top->down or down->top results sorting. This will require me to find all the places I have ssumptions about lines and sorting spots.
2. then, after doing that, we should be able to do some tricks to just move past the visible area.

tami5:
I'm starting to use loclist to get tags (from markdown into location) and use it as an outline. This feature is really useful

mizlan:
> Yes, it's a good idea. It might require a bit of time to think about. I have implemented `lines_manager` to (potentially) be able to handle this kind of thing at some point in the future.

@tjdevries Where can I find `lines_manager`? Is it on a particular branch or fork?

apazzolini:
Until this feature is built, are the full results accessible programmatically?

Specifically, I'd like to be able to trigger live_grep and send all results (including the ones outside the visible range) to the quickfix list.

VVKot:
@apazzolini you can do exactly that with `<C-q>` https://github.com/nvim-telescope/telescope.nvim/blob/f92b9b1fae70d5fac681a29f0df64549c399f18f/lua/telescope/mappings.lua#L29

apazzolini:
> actions.send_to_qflist + actions.open_qflist

Oh, I didn't realize that sent invisible results as well. Thanks!

VVKot:
You're welcome. I remembered that I actually use `smart_send_to_qflist` instead, which sends all results if there is no selection - but only selected rows if there is. Check it out!

[...]

PR #293, opened 2020-11-27
feat: quickfix

tjdevries:
I'm working on changing structure of how we store and manage entries.

Once I do that, we can store all the entries.

Once I do that, it should be relatively trivial to turn all the entries into quickfix items.

(still WIP and playing around with new ideas / simple data structures)

PR #927, opened 2021-06-20
docs: functions in `entry_manager.lua`, `pickers.lua` and `linked_list.lua`

l-kershaw:
Adds documentation for functions  in `entry_manager.lua` and in `pickers.lua`.

Was having a think about how to implement scrolling of results (as discussed in  #42) and realised I did not understand the purpose of most of the functions in these files. This PR is the result of me trying to figure this out.
(_It is also hopefully a way for me to be told where I have misunderstood things_ 🤣)

---

TODO:

- [x] finish off documenting `pickers.lua`
- [x] add documentation for `linked_list.lua`


lua/telescope/entry_manager.lua
```diff
@@ -26,6 +26,14 @@ if past loop of must have scores,
 local EntryManager = {}
 EntryManager.__index = EntryManager
 
+--- Create a new EntryManager object
+---@param max_results number: the number of results to keep sorted at the head of the linked list
+---@param set_entry function: handler function for object using the EntryManager
+---@param info table: (optional) table containing information to keep track of
+---   @key looped number: number of existing entries checked when adding new entries
+---   @key inserted number: TODO: figure out if this is used anywhere
```
Conni2461, line 34: I think is currently unused. Tj never did anything with it

lua/telescope/entry_manager.lua
```diff
@@ -65,18 +78,28 @@ function EntryManager:get_container(index)
...
 end
 
+--- Get the score in the EntryManager corresponding to `index`
+---@param index number
+---@return number
 function EntryManager:get_score(index)
   return self:get_container(index)[2]
 end
 
+--TODO: documentation
```
Conni2461, line 95: Ordinal is the text that is filtered on. So you can have a different display than what you filter on. Right now tj isn't doing anything with `get_ordinal` tho

lua/telescope/entry_manager.lua
```diff
@@ -65,18 +78,28 @@ function EntryManager:get_container(index)
...
 function EntryManager:get_ordinal(index)
   return self:get_entry(index).ordinal
 end
 
+--- Get the index of the given entry in the EntryManager
+---@param entry table: table with information about the given entry
+---@return integer/nil: the index of the entry if is present, nil otherwise
```
Conni2461, line 102: About emmy documentation. You can have multiple types for `@param` but not for `@return` seems like an oversight. Will open an issue, maybe i implement it but if i do we will end up with `---@return number|nil` (to keep consistency with `@param`) (there is no integer type in emmy doc its always number)

lua/telescope/entry_manager.lua
```diff
@@ -126,6 +169,17 @@ function EntryManager:_append_container(picker, new_container, should_update)
   end
 end
 
+-- Adds `new_container` to the associated linked list.
```
Conni2461, line 172: 3 `-`

lua/telescope/algos/linked_list.lua
```diff
@@ -2,6 +2,49 @@
...
+--- │ │item│ │───►   ───►│   │tracked│   │───►   ───►│ │item│ │
+--- │ └────┘ │           │   └───────┘   │           │ └────┘ │
+--- └────────┘           └───────────────┘           └────────┘
+--- </pre>
+---@brief ]]
+
+
+--- Create a new linked list
+---@param opts table
+---   @key track_at number: the index of the node to track
```
Conni2461, line 46: `@key` is currently not a thing, but that also doesn't matter because right now we are not exporting documentation for that file

lua/telescope/pickers.lua
```diff
@@ -561,7 +585,9 @@ function Picker:delete_selection(delete_cb)
   end)
 end
 
-
+--- Closes the windows for the prompt, results and preview
+---@param status table: table containing information on the picker
+--- and associated windows. Generally obtained from `state.get_status`
```
Conni2461, line 590: fyi there is currently no multiline annotations (doesnt matter as long as we do not generate docs from it)

lua/telescope/pickers.lua
```diff
@@ -975,6 +1049,7 @@ function Picker:close_existing_pickers()
   end
 end
 
+-- TODO: document what the purpose of this is
```
Conni2461, line 1052: returns a function that will set the virtual text `n / m` thing

lua/telescope/pickers.lua
```diff
@@ -1013,6 +1088,7 @@ function Picker:get_status_updater(prompt_win, prompt_bufnr)
 end
 
 
+-- TODO: document what the purpose of this is
```
Conni2461, line 1091: returns a function that will process a element. Increase number `m` and `n` if it isnt filter (virtual text one). and basically run the file ignore check (decrease `m` again because if true we ignore the entry) and run the sorters score function (which can be anything, substr, fzy, fzf, ...)

lua/telescope/pickers.lua
```diff
@@ -1054,6 +1130,7 @@ function Picker:get_result_processor(find_id, prompt, status_updater)
   end
 end
 
+-- TODO: document what the purpose of this is
```
Conni2461, line 1133: Runs after all elements are scored/processed. Handles things like, what should happen with selection index:
- `reset` will set the selection to row 1 (or `opts.default_selection_index`) back
- `closest` sets it to index 1 if prompt != "" else `opts.default_selection_index` ( `builtin.buffers` make use of `default_selection_index` with its `opts.sort_last_used` option)
- row tries to keep the row (might be broken, dont wanna blame tj but i think he broke it some time ago and i was too lazy to fix it) i reported it here: https://github.com/nvim-telescope/telescope.nvim/pull/293#issuecomment-757136468 he ignored me (as always, he is just above my pay grade :rofl:, if it isn't obvious i am memeing right now)
- follow tries to follow the currently selected entry

other things that happen in the returned method is:
- status_updater, update the n/m virtual text thingy
- set `current_line` so some actions can make use of it
- clear extra rows
- do highlighting
- call finish on sorter

lua/telescope/pickers.lua
```diff
@@ -963,6 +1035,8 @@ function Picker:register_completion_callback(cb)
   table.insert(self._completion_callbacks, cb)
 end
 
+-- TODO: document what the purpose of this is
+--  Currently only used in `Picker: get_result_completor`
```
Conni2461, line 1039: completion callbacks can be registered via `opts.on_complete` basically functions that run after all items are proceeded once. Currently used in testing, but users could do anything with it.
and `_on_complete` is basically only calling all these registered callbacks

lua/telescope/pickers.lua
```diff
@@ -975,6 +1049,7 @@ function Picker:close_existing_pickers()
   end
 end
```
Conni2461, line 1050: I cant comment for `close_existing_pickers` but it does what the function tells you. If someone what start telescope while telescope is still open it would close the already existing instance. Like if a user wants to go from one picker into another with an action.

(We should probably do `actions._close(prompt_bufnr, true)` tho, with pcall obv, because this function is currently used to fix an bug we have, also see `actions.run_builtin`

Conni2461:
This is mostly good but you always have to ask @tjdevries if you are working with systems that low level. He did wrote EVERYTHING here i also only read the code :laughing: 

This is probably another good point to mention. I still want to move linked list into plenary (as mentioned here as well https://github.com/nvim-telescope/telescope.nvim/pull/293#issuecomment-757136468) [plenary.collections](https://github.com/nvim-lua/plenary.nvim/tree/master/lua/plenary/collections) is currently a little "empty" :laughing:

lua/telescope/entry_manager.lua
tjdevries: I think I used it do some debugging for some testing, can probably be removed.

lua/telescope/algos/linked_list.lua
l-kershaw: That's fine, was just matching what had been done elsewhere

l-kershaw:
Closing in favour of #1554, as this PR is so out of date that the merge conflicts would be more trouble than its worth 😅

Issue #906, opened 2021-06-11
Buffer close and keep window open


dahmc:
**Is your feature request related to a problem? Please describe.**
I saw the commit https://github.com/nvim-telescope/telescope.nvim/commit/495f84fd35e7b5bb0bb670e232725924b3b85b88 which introduced the ability to delete buffers directly from the buffer list, which is awesome!

However, when I close it, the split gets closed as well. I usually have 3 splits next to each other, and I would love to keep them open.

**Describe the solution you'd like**
Currently I use https://github.com/romgrk/barbar.nvim for the bufferline/tabline, and it comes with a `:BufferClose` command that's basically what https://github.com/moll/vim-bbye does as well. It closes the buffer, but keeps the split open.

Is there a possibility to implement this in Telescope as well, maybe as either an option to the command or that we can run `:BufferClose` on the currently highlighted buffer in the list?

tami5:
> Not sure whether it should be part of telescope.nvim but here is a function that you can easily use to overwrite buffers to be deleted with empty buffers.

Oh that's very useful, thanks @fdschmidt93 . I believe it should. Although, commonly with plugins and hacks that closes a buffer without closing the window is reverting back to the previous buffer in the window list.

fdschmidt93:
> Although, commonly with plugins and hacks that closes a buffer without closing the window is reverting back to the previous buffer in the window list.

To clarify, what would the preferred logic be (doesn't have to mirror bbye exactly). A few variants come to mind -
* Just open the previous buffer (essentially `:bprevious` w/ buffer deletion included), in which case this `bprevious` buffer might already be opened somewhere
* Open the previous buffer that's not yet currently opened (at least in the current tab)
* Callback to pick a replacement buffer with telescope?

With an empty buffer being a fallback/selection option in all three cases I suppose.

In any case, I'm afraid I'm not versed in vimscript but from what I see vim-bbye (and hence barbar) does it a bit more complicated than we'd need to (and the logic isn't perfectly clear to me from reading vimscript I'm afraid :sweat_smile: ), though I might be mistaken. 

Happy to take a crack at it some time, once the desired feature is laid out a bit more clearly :)

dahmc:
Thanks @fdschmidt93 

I think just open the previous buffer, and fallback to empty if no more buffers sounds very reasonable. That would be awesome :)

fdschmidt93:
```lua
local action_state = require 'telescope.actions.state'
require'telescope'.setup {
[...]

Issue #793, opened 2021-04-22
Search refine like Ivy


baldore:
**Is your feature request related to a problem? Please describe.**
It's not, unless there's already a way to do this with Telescope.

**Describe the solution you'd like**
I was trying a little bit of Emacs and I found Ivy, which has pretty cool features. One I found really interesting is that you can refine the search multiple times by accepting the current results and filtering those results again and again. The way you do this is by pressing `shift + space`.

**Describe alternatives you've considered**
Tried with Fzf, which has a way to "ignore" results with `!`. This is not related to this feature request, but it should be great to know how to do this. Example: `!.test` will show all the results that DON'T HAVE `.test` extension.

**Additional context**
Example from Emacs:
![ivy](https://user-images.githubusercontent.com/3131037/115795891-a3433280-a385-11eb-977c-f66956e88281.gif)

As you can see, each time the search expression disappears, is because it starts from zero with the filtered results.

Let me know what do you think, also about the question of Fzf "ignore".

windwp:
it already have a PR #771

elianiva:
that PR only implements the layout, not the "refresh"

elianiva:
https://user-images.githubusercontent.com/51877647/115944245-cf25fb80-a4de-11eb-9d8b-0ff74f5af16f.mp4

I made a very rough proof of concept, someone needs to implement a better version

baldore:
@elianiva that looks pretty cool! Additionally to that, do you know how to do the ignore part? Regex maybe?

Conni2461:
I am not sure if this should be part of core. We can already do this with https://github.com/nvim-telescope/telescope-fzf-native.nvim just with space separated terms. Its also more powerful because it has suffix/prefix/exact/equal matching

tjdevries:
> I was trying a little bit of emacs... 

That's a ban from this repo! 😂

Yeah, the search refine would be cool. We can add this to 1.0 (might get pushed back depending on complexity)

VVKot:
As a temporary workaround you can send all results into a quickfix list and then call `builtin.quickfix` to get them back into Telescope.

thomkeh:
> As a temporary workaround you can send all results into a quickfix list and then call `builtin.quickfix` to get them back into Telescope.

[...]

PR #931, opened 2021-06-22
feat: syntactic sugar to map funcs over entries & selections

### LINKED ISSUE #900: Action: Toggle Select All results
I want to select all results with keybinding (e.g. git_status - to stage or unstage all results)

builtin action

write custom action (asking for advice):
So i need to get all results (from finder?) and cycle through and `toggle.selection` theme

smth like:
```
for i=1,number_of_results-1 do
  actions.toggle_selection
  actions.move_selection
end
actions.toggle_selection

```

fdschmidt93:
This PR closes #900 and implements
* toggling, selecting, and dropping all entries from multi selection
* carves out sugar to apply functions (and collect the result) on entries and multi-selections
* furthermore includes `vim-bbye` action from #906

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -144,45 +144,6 @@ function actions.toggle_selection(prompt_bufnr)
   current_picker:toggle_selection(current_picker:get_selection_row())
 end
 
---- Apply `f` to entries of current picker and returns list of mapped entries
---- `f` takes (entry, index, row) as viable arguments in order
----@param prompt_bufnr number: The prompt bufnr
----@param f function: function to apply on entries of picker
----@return table: result from mapped entries
-function actions.map_entries(prompt_bufnr, f)
-  local current_picker = action_state.get_current_picker(prompt_bufnr)
-  local results = {}
-  local index = 1
-  local row, result
-    -- indices are 1-indexed, rows are 0-indexed
-  for entry in current_picker.manager:iter() do
-    row = current_picker:get_row(index)
-    -- assign result to variable as function w/o return value
-    -- results in error upon table.insert as it's not technically nil but 'empty'
-    result = f(entry, index, row)
-    results[index] = result
-    index = index + 1
-  end
-  return results
-end
-
---- Apply `f` to multi selections of current picker and returns list of mapped selections.
----@param prompt_bufnr number: The prompt bufnr
----@param f function: function to apply on multi selection of picker
----@return table: result from `f` applied to multi selections
-function actions.map_selections(prompt_bufnr, f)
-  local current_picker = action_state.get_current_picker(prompt_bufnr)
-  local results = {}
-  local result
-  for _, selection in ipairs(current_picker:get_multi_selection()) do
-    result = f(selection)
-    -- assign result to variable as function w/o return value
-    -- results in error upon table.insert as it's not technically nil but 'empty'
-    table.insert(results, result)
-  end
-  return results
-end
-
 --- Multi select all entries, possibly including entries not visible in results popup.
 ---@param prompt_bufnr number: The prompt bufnr
 function actions.select_all(prompt_bufnr)
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -15,6 +15,7 @@ local utils = require('telescope.utils')
 local p_scroller = require('telescope.pickers.scroller')
 
 local action_state = require('telescope.actions.state')
+local action_utils = require('telescope.actions.utils')
 local action_set = require('telescope.actions.set')
 
 local transform_mod = require('telescope.actions.mt').transform_mod
@@ -159,7 +160,7 @@ function actions.map_entries(prompt_bufnr, f)
     -- assign result to variable as function w/o return value
     -- results in error upon table.insert as it's not technically nil but 'empty'
     result = f(entry, index, row)
-    table.insert(results, result)
+    results[index] = result
     index = index + 1
   end
   return results
@@ -195,33 +196,33 @@ function actions.select_all(prompt_bufnr)
       end
     end
   end
-  actions.map_entries(prompt_bufnr, select_all)
+  action_utils.map_entries(prompt_bufnr, select_all)
 end
 
 --- Drop all entries from multi selection.
 ---@param prompt_bufnr number: The prompt bufnr
 function actions.drop_all(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
-  local drop_all = function(entry, idx, row)
+  local drop_all = function(entry, _, row)
     current_picker._multi:drop(entry)
     if current_picker:can_select_row(row) then
       current_picker.highlighter:hi_multiselect(row, current_picker._multi:is_selected(entry))
     end
   end
-  actions.map_entries(prompt_bufnr, drop_all)
+  action_utils.map_entries(prompt_bufnr, drop_all)
 end
 
 --- Toggle multi selection for all entries.
 ---@param prompt_bufnr number: The prompt bufnr
 function actions.toggle_all(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
-  local toggle_all = function(entry, idx, row)
+  local toggle_all = function(entry, _, row)
     current_picker._multi:toggle(entry)
     if current_picker:can_select_row(row) then
       current_picker.highlighter:hi_multiselect(row, current_picker._multi:is_selected(entry))
     end
   end
-  actions.map_entries(prompt_bufnr, toggle_all)
+  action_utils.map_entries(prompt_bufnr, toggle_all)
 end
 
 function actions.preview_scrolling_up(prompt_bufnr)
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -143,6 +143,87 @@ function actions.toggle_selection(prompt_bufnr)
   current_picker:toggle_selection(current_picker:get_selection_row())
 end
 
+--- Apply `f` to entries of current picker and returns list of mapped entries
+--- `f` takes (entry, index, row) as viable arguments in order
+---@param prompt_bufnr number: The prompt bufnr
+---@param f function: function to apply on entries of picker
+---@return table: result from mapped entries
+function actions.map_entries(prompt_bufnr, f)
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local results = {}
+  local index = 1
+  local row, result
+    -- indices are 1-indexed, rows are 0-indexed
+  for entry in current_picker.manager:iter() do
+    row = current_picker:get_row(index)
+    -- assign result to variable as function w/o return value
+    -- results in error upon table.insert as it's not technically nil but 'empty'
+    result = f(entry, index, row)
+    table.insert(results, result)
+    index = index + 1
+  end
+  return results
+end
+
+--- Apply `f` to multi selections of current picker and returns list of mapped selections.
+---@param prompt_bufnr number: The prompt bufnr
+---@param f function: function to apply on multi selection of picker
+---@return table: result from `f` applied to multi selections
+function actions.map_selections(prompt_bufnr, f)
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local results = {}
+  local result
+  for _, selection in ipairs(current_picker:get_multi_selection()) do
+    result = f(selection)
+    -- assign result to variable as function w/o return value
+    -- results in error upon table.insert as it's not technically nil but 'empty'
+    table.insert(results, result)
+  end
+  return results
+end
+
+--- Multi select all entries, possibly including entries not visible in results popup.
+---@param prompt_bufnr number: The prompt bufnr
+function actions.select_all(prompt_bufnr)
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  -- indices are 1-indexed, rows are 0-indexed
+  local select_all = function(entry, _, row)
+    if not current_picker._multi:is_selected(entry) then
+      current_picker._multi:add(entry)
+      if current_picker:can_select_row(row) then
+        current_picker.highlighter:hi_multiselect(row, current_picker._multi:is_selected(entry))
+      end
+    end
+  end
+  actions.map_entries(prompt_bufnr, select_all)
+end
+
+--- Drop all entries from multi selection.
+---@param prompt_bufnr number: The prompt bufnr
+function actions.drop_all(prompt_bufnr)
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local drop_all = function(entry, idx, row)
+    current_picker._multi:drop(entry)
+    if current_picker:can_select_row(row) then
+      current_picker.highlighter:hi_multiselect(row, current_picker._multi:is_selected(entry))
+    end
+  end
+  actions.map_entries(prompt_bufnr, drop_all)
+end
+
+--- Toggle multi selection for all entries.
+---@param prompt_bufnr number: The prompt bufnr
+function actions.toggle_all(prompt_bufnr)
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local toggle_all = function(entry, idx, row)
+    current_picker._multi:toggle(entry)
+    if current_picker:can_select_row(row) then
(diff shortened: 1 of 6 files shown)


lua/telescope/actions/init.lua
```diff
@@ -143,6 +143,87 @@ function actions.toggle_selection(prompt_bufnr)
   current_picker:toggle_selection(current_picker:get_selection_row())
 end
 
+--- Apply `f` to entries of current picker and returns list of mapped entries
+--- `f` takes (entry, index, row) as viable arguments in order
+---@param prompt_bufnr number: The prompt bufnr
+---@param f function: function to apply on entries of picker
+---@return table: result from mapped entries
+function actions.map_entries(prompt_bufnr, f)
```
Conni2461, line 151: `actions` that have more than one parameter are usually prefixed with `_` because they are not "mapable". But this function makes even more sense in `actions.state` or a new `actions.util`

lua/telescope/actions/init.lua
```diff
@@ -143,6 +143,87 @@ function actions.toggle_selection(prompt_bufnr)
...
+  local index = 1
+  local row, result
+    -- indices are 1-indexed, rows are 0-indexed
+  for entry in current_picker.manager:iter() do
+    row = current_picker:get_row(index)
+    -- assign result to variable as function w/o return value
+    -- results in error upon table.insert as it's not technically nil but 'empty'
+    result = f(entry, index, row)
+    table.insert(results, result)
+    index = index + 1
```
Conni2461, line 163: We already have the index you can then also just do `results[index] = result` which is faster than `table.insert`

lua/telescope/actions/init.lua
```diff
@@ -664,6 +745,46 @@ actions.delete_buffer = function(prompt_bufnr)
   end)
 end
 
+--- Replace multi-selected buffers with previous buffers in the list, respectively.
+--- Falls back to empty unlisted scratch buffer as placeholder if no valid buffers found.
+--- See also: https://github.com/moll/vim-bbye
+---@param prompt_bufnr number: The prompt bufnr
+actions.bbye_buffer = function(prompt_bufnr)
```
Conni2461, line 752: I would prefer to not name that after another plugin

lua/telescope/mappings.lua
```diff
@@ -26,6 +26,7 @@ mappings.default_mappings = config.values.default_mappings or {
 
       ["<Tab>"] = actions.toggle_selection + actions.move_selection_worse,
       ["<S-Tab>"] = actions.toggle_selection + actions.move_selection_better,
+      ["<M-Tab>"] = actions.toggle_all,
```
Conni2461, line 29: I am currently at this point where i don't want to map new things on default. (Personally i don't have any vim mappings on MOD1)  and thats the problem with doing that. (I know tj mapped `<M-q>` but i also told him that this doesn't work for me and probably not for other people) yeah.

Leave it here for now. I'll think about it

Conni2461:
I am not sure if we need the bbye action or if we should work more with the wiki and have a page with action configurations similar to what fzf does. I don't think we can continue to add super "specific" actions for a "handful" people. We might just blow up the repository.

I'll need think about it and come back to you (I am super busy this week so it might take some time)

lua/telescope/actions/init.lua
fdschmidt93: Makes sense, any preference on `actions.state` or `actions.util`? Also makes prefixing with `_`  obsolete which I'd prefer.

fdschmidt93:
> I am not sure if we need the bbye action or if we should work more with the wiki and have a page with action configurations similar to what fzf does. 

I personally agree, though I'd be wary that lot of users wouldn't read a wiki (or something similar) and just raise another issue or reddit post (why is grepping slow on large repos? :laughing:) Maybe an extension that just loads all picker-specific actions appealing to the largest user base is most straightforward; users can then opt-out/remap if something doesn't suit them (and my guess is most actions would just be ignored since a user doesn't care for them).

> I'll need think about it and come back to you (I am super busy this week so it might take some time)

No worries, I'll pick this up again then.

lua/telescope/actions/init.lua
Conni2461: i prefer `actions/utils.lua`, make sure you add the file to `gendocs.lua` and write `@tag` and `@brief`

Conni2461:
I would prefer a good structured wiki. The users who, don't read manual will always exist, it doesn't matter if its in the docs as unmapped action, wiki as snippet or an extension as unmapped action. We just need to have a good wiki structure, with a table of content, etc and when the question/issues come we can just like into the wiki and close. 

Also, with a wiki, people can just setup things they actually want and maybe while copy and pasting the code, maybe they learn something :)

I try to set up the wiki that way on sunday. I haven't made any decisions on on `bbye` yet tho

fdschmidt93:
Any idea how to fix this "flickering" on `drop_all` that sometimes (not always) occurs? It's cleared once the cursor is moved. From the debugging I've tried it's related to clearing the extmarks [here](https://github.com/nvim-telescope/telescope.nvim/blob/e5bd4963da81b5d044749ee4507061801aeb0f78/lua/telescope/pickers/highlights.lua#L102), though I haven't yet had the time to think this through with a meaningful fix as I also haven't used that highlighting api yet. 


https://user-images.githubusercontent.com/39233597/123829515-d0bad800-d902-11eb-9473-1690b6999901.mp4

fdschmidt93:
Is #793 something we would want in telescope? With the `action_utils` we can easily refine in a simple manner.

```lua
function actions.refine(prompt_bufnr)
  -- select our filter entries
  actions.select_all(prompt_bufnr)
  current_picker:reset_prompt()
  -- invert selection
  actions.toggle_all(prompt_bufnr)
  -- remove inverted selection
  current_picker:delete_selection(function() end)
end
```
There is minimal flickering from intermediate highlighting (could always be made optional in `select_all` or `toggle_all`) and alphabetical (?) resorting of entries.

Happy to add it to the PR if desired.

Conni2461:
I still we shouldn't do #793 because its less performant than using and from fzf-native. But tj and i talked about being able to do `live_grep` or `lsp_dynamic_workspace_symbol` and then fuzzy find over the "live" results. So can you wait a little bit longer.

Also i think its out of scope for this PR, we should keep prs smaller. 

None the less you have the util functions twice now, once in the actions/init.lua file and once in actions/util.lua

fdschmidt93:
Ok, fixed. I didn't finally clean up yet as I was waiting on feedback for vim-bbye, whether to keep the mapping or not, and whether there's something to do about the mini-flickering as shown above or would be ok as is.

fdschmidt93:
To simplify things I've now removed the mapping and vim-bbye action. The mapping can of course be always set by the user and the action will make its way into the wiki eventually.

The highlighting issue referred to above is a separate issue, which stems from removal of highlighting extmarks, and any later fix of that should not effect this PR.

lua/telescope/actions/utils.lua
```diff
@@ -0,0 +1,58 @@
...
+    f = {f, "function"}
+  }
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local results = {}
+  local result
+  for _, selection in ipairs(current_picker:get_multi_selection()) do
+    result = f(selection)
+    -- assign result to variable as function w/o return value
+    -- results in error upon table.insert as it's not technically nil but 'empty'
+    table.insert(results, result)
```
Conni2461, line 53: Why not just `table.insert(results, f(selection))`

lua/telescope/actions/utils.lua
```diff
@@ -0,0 +1,58 @@
...
+---@param f function: Function to apply on entries of picker that takes (entry, index, row) as possible arguments
+---@return table: Result from `f` applied to entries
+function utils.map_entries(prompt_bufnr, f)
+  vim.validate{
+    f = {f, "function"}
+  }
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local results = {}
+  local index = 1
+  local row, result
```
Conni2461, line 25: Does this need to be outside of the loop?

lua/telescope/actions/utils.lua
```diff
@@ -0,0 +1,58 @@
...
+    -- indices are 1-indexed, rows are 0-indexed
+  for entry in current_picker.manager:iter() do
+    row = current_picker:get_row(index)
+    -- assign result to variable as function w/o return value
+    -- results in error upon table.insert as it's not technically nil but 'empty'
```
Conni2461, line 30: I dont get this comment

Conni2461: Like if `f` returns nil and you do `results[index] = nil` in this case. Its nil at this index. If f now only returns nil for some elements, then in the middle would be a nil -> only 1 .. middle will be iterated. Am i wrong?

Conni2461: If so and a function can return nil you should do a if not nil check and put line 32 33 in that check

lua/telescope/actions/init.lua
```diff
@@ -143,6 +144,50 @@ function actions.toggle_selection(prompt_bufnr)
...
+---@param prompt_bufnr number: The prompt bufnr
+function actions.toggle_all(prompt_bufnr)
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local toggle_all = function(entry, _, row)
+    current_picker._multi:toggle(entry)
+    if current_picker:can_select_row(row) then
+      current_picker.highlighter:hi_multiselect(row, current_picker._multi:is_selected(entry))
+    end
+  end
+  action_utils.map_entries(prompt_bufnr, toggle_all)
```
Conni2461, line 188: can you just do 
```lua
actions_utils.map_entries(prompt_bufnr, function(entry, _, row)

end)
```

applies to all your actions. no reason to have it as local function

lua/telescope/actions/init.lua
```diff
@@ -143,6 +144,50 @@ function actions.toggle_selection(prompt_bufnr)
   current_picker:toggle_selection(current_picker:get_selection_row())
 end
 
+--- Multi select all entries.
+--- - Note: selected entries may include results not visible in the results popup.
+---@param prompt_bufnr number: The prompt bufnr
+function actions.select_all(prompt_bufnr)
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  -- indices are 1-indexed, rows are 0-indexed
```
Conni2461, line 152: this comment is out of place here. `utils.map_entries` should mention it in its header.

lua/telescope/actions/utils.lua
Conni2461: Also the comment `assign result ...` does not apply here.

```lua
local results = {}

table.insert(results, nil)
table.insert(results, nil)
table.insert(results, 1)
table.insert(results, nil)
print(vim.inspect(results))
```

fdschmidt93: This is the point, also applies to the remark below. Users might write `f` that don't return any value and this was meant to account for that in a simple manner.

```lua

-- doesn't work and raises, nothing to insert into table
x = {}
y = function() end
table.insert(x, y())
print(vim.inspect(x))

-- works
x = {}
y = function() return nil end
table.insert(x, y())
print(vim.inspect(x))

-- result is now nil
x = {}
y = function() end
result = y()
table.insert(x, result)
print(vim.inspect(x))
```

lua/telescope/actions/utils.lua
fdschmidt93: Since we seem to be focused on maximum speed (`table.insert` vs `results[index] = result`) we can also avoid re-instantiating the variable. But happy to change it to e.g .`local row = current_picker:get_row(index)` then instead.

lua/telescope/actions/utils.lua
fdschmidt93: See the other comment -- it takes care of functions that do not have a return statement.

lua/telescope/actions/init.lua
fdschmidt93: Ok

lua/telescope/actions/init.lua
fdschmidt93: Genuinely curious, is there a difference apart from cosmetics and the function then not only being in scope of the `map_entries` call?

Conni2461: nope. in my opinion its just more readable. Its like a callback/lambda in any other language

lua/telescope/actions/utils.lua
Conni2461: Okay i see. Then just move the `local result` inside the for.

And maybe reframe the comment a little bit. Its was confusing for me :rofl:

lua/telescope/actions/utils.lua
Conni2461: Yeah dont try to cheat the jit compiler. We have one. it will take care of it. `table.insert` is a exception because it can't be jit compiled. In this case i prefer the more readable code

lua/telescope/actions/utils.lua
Conni2461: Yeah but this index will continue to insert things. Thats the point. You could get "holes" in your table, when doing it with `results[index] = result`

Conni2461:
```lua
local a = function()
  local results = {}

  local odd = function(i)
    if i % 2 == 1 then
      return i
    end
  end

  for i = 1, 10 do
    results[i] = odd(i)
  end

  P(results)
end

local b = function()
  local results = {}

  local odd = function(i)
    if i % 2 == 1 then
      return i
    end
  end

  for i = 1, 10 do
    local result = odd(i)
    table.insert(results, result)
  end

  P(results)
end

a()
b()
```

lua/telescope/actions/utils.lua
Conni2461: luajit wiki actually say dont do it. http://wiki.luajit.org/Numerical-Computing-Performance-Guide
this section. Close to the end of the page:
image

fdschmidt93: Ok will do, thanks for the reference!

lua/telescope/actions/utils.lua
fdschmidt93: > Yeah but this index will continue to insert things. Thats the point. You could get "holes" in your table, when doing it with results[index] = result

Fair enough, didn't consider it in that context, but that's also why I initially relied on table.insert ;)

lua/telescope/actions/utils.lua
fdschmidt93: See discussion below, we should probably also track the index here and only insert results for key-value table for which the function returned non-`nil` values.

lua/telescope/actions/utils.lua
Conni2461: tbh right now. idk what we want :rofl: 

You can still do it with an index but you need a if not nil check. I think about all that and maybe do a coin flip ;)

fdschmidt93: What about:
- default to just returning a table of values with `table.insert` as I had it -- that's most likely what's being used anyways
- do `index` and `result` key-value table when a tuple is returned

I don't think speed is a practical concern here, we then have a practical default, and if a user needs the index, he can have it.

fdschmidt93:
I circled a bit around our issue of how to collect the return values, only to realize the "burden" is probably best put to the user as it's too easy to just write what he/she want's to do with the output, as opposed to solving the problem for every case.

Stupidly funny side notes:
- Lua cannot store 1 as a key? `a()` from your above function returns something funny on careful consideration.
- If you ever cared on how to check how many values are returned, it's gets a bit [tedious](https://stackoverflow.com/questions/29769189/number-of-returned-values) in a practical use case

Conni2461:
> Lua cannot store 1 as a key? In practice it doesn't matter, but a() from your above function returns something funny on careful consideration.

Thats just false. Lua always stores things in keys. So a list starts has keys 1 ... n `inspect` just omits 1 .. first occurrence of nil because it can be omitted. A simple `for k,v pairs(tbl) do` should demonstrate that.

> If you ever cared on how to check how many values are returned, it's gets a bit tedious in a practical use case

for a list (a table starting with the key one) for `#` it returns how many values till first nil. That was whole point with the table with holes and why i said we need to change it but yeah

fdschmidt93:
> Thats just false. Lua always stores things in keys. So a list starts has keys 1 ... n inspect just omits 1 .. first occurrence of nil because it can be omitted. A simple for k,v pairs(tbl) do should demonstrate that.

Maybe a bit badly phrased, that's why I said it doesn't matter. But yeah, good to have it in full detail :laughing:

PR #940, opened 2021-06-26
fix: display issues in `:Telescope registers`

l-kershaw:
- Problems occured when a register contained newlines
- Caused issues in `Picker:set_selection`
- Couldn't move selection when a register with newlines was selected

I think that this problem wasn't noticed due to the clean-up for entries with newlines in `Picker:entry_adder`.


Conni2461:
Yeah there is something wackiness when you have a `\n` in your display. I also ran into this a while ago.
We could also do `content:gsub('\n','\\n')` might be a little bit clearer. Thoughts?

Otherwise this can be merged. Thanks :)

freezboltz (commented):
This makes sense 
Thanks @l-kershaw  :)

PR #950, opened 2021-07-03
New Example Config

max397574:
I got an error, when I used the configuration like in the example.
Here is the fix
Screen_Shot_2021-07-03_at_11 13 15
Screen_Shot_2021-07-03_at_11 13 29

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ As an example of using the `setup()` method, the following code configures
 
 ```lua
 require('telescope').setup{
-  defaults = {
+  config = {
     vimgrep_arguments = {
       'rg',
       '--color=never',


Conni2461:
The default config works as expected and your suggested change is not how you want to fix your issues. The key `defaults` didn't change its name to `config` but the key `layout_defaults` in the `defaults` table changed its name to `layout_config`. The default configuration already reflects this change. You want to read `:help telescope.changelog` to find out how to fix your issue.

TLDR:
- width is no longer a valid config key
- results_height is no longer a valid config key
- results_width is no longer a valid config key
- prompt_positions is no part of `layout_config`
- preview_cutoff is now part of `layout_config`

I am closing this PR because the suggested change is wrong. There is no `config` key in `telescope.setup` the name is still `defaults`

PR #953, opened 2021-07-03
feat: option to `include_extensions` in `builtin` picker

### LINKED ISSUE #925: Picker a la `builtin` but that also includes custom pickers
Telescope is great but at this point there are *lots* of pickers. This why `builtin` picker is amazing, I can bind one short cut and have easy access to all 49 (!) default pickers.

However, there is a problem. Custom pickers (see below) that are popping up are *not* included in this picker, which makes for a frustrating experience.

https://github.com/folke/todo-comments.nvim
https://github.com/folke/trouble.nvim
https://github.com/fhill2/telescope-ultisnips.nvim

A picker, either (renamed) builtin or a new one, that *does* include custom pickers. I would imagine all that needs to be done is exposing some plugin-like endpoint in telescope for pickers to "register" them - but I might be wrong.

I either do `:Telescope <Tab>` or bind a specific mapping for custom pickers, which is not ideal.

**Additional context**
I am not sure why but autocompletion in `:Telescope <Tab>` *does* recognize custom pickers.

l-kershaw:
Provide option to `include_extensions` in the `builtin.builtin` picker
- add option `include_extensions` which defaults to `false`
- if `include_extensions` is `true` then add functions from extensions to results
- update `actions.run_builtin` to check if extension function provided

---

Resolves #925

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -332,8 +332,17 @@ end
 actions.run_builtin = function(prompt_bufnr)
   local entry = action_state.get_selected_entry(prompt_bufnr)
 
-actions._close(prompt_bufnr, true)
-  require('telescope.builtin')[entry.text]()
+  actions._close(prompt_bufnr, true)
+  if string.match(entry.text," : ") then
+    -- Call appropriate function from extensions
+    local split_string = vim.split(entry.text," : ")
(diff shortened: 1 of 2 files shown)


Conni2461 (approved):
as always thanks :)

PR #954, opened 2021-07-03
Removed comment on nightly in `README`

paniash:
This PR highlights the latest stable release of neovim (v.0.5 is not nightly).

PR #956, opened 2021-07-04
Updated `README` with nightly as well

paniash:
I think it gives a false impression to users that the plugin only works with v0.5 and so to add to #954, this PR also adds nightly.


Conni2461:
I updated it by just appending or newer. That should be enough for now. Thanks anyways :)

PR #959, opened 2021-07-05
WIP: recalculate picker layout on `VimResized`

l-kershaw:
- implements `Picker:recalculate_layout` function
- adds an autocommand to call this function on the `VimResized` event
- relies on nvim-lua/plenary.nvim#180 and nvim-lua/popup.nvim#15

Currently has issues when the results window changes in height.

Demo:
telescope-resize

As shown in the demo, the picker still works fine when the results window is the same height, but when the height increases, you can no longer change the selection. In other tests, when the height decreases, you can change the selection but the displayed list is in the wrong position and updates as you select each one.

I will have a go at fixing the results window issues, but thought it would be good to get feedback on what is done so far 🙂

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -441,9 +441,13 @@ function Picker:find()
     [[  autocmd BufLeave <buffer> ++nested ++once :silent lua require('telescope.pickers').on_close_prompt(%s)]],
     prompt_bufnr)
 
+  local on_vim_resize = string.format(
+    [[  autocmd VimResized <buffer> ++nested :silent lua require('telescope.pickers').on_resize_window(%s)]],
+    prompt_bufnr)
   vim.cmd([[augroup PickerInsert]])
   vim.cmd([[  au!]])
   vim.cmd(    on_buf_leave)
+  vim.cmd(    on_vim_resize)
   vim.cmd([[augroup END]])
 
   self.prompt_bufnr = prompt_bufnr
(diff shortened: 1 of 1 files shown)


Conni2461:
My first thought that all we need to do is run `picker:__on_lines(nil, nil, nil, 0, 1)` after `picker:recalculate_layout()` but that did not work :| i will take a closer look at it later today. :) But looks already promising good job

l-kershaw:
I think this may require some fiddling with `EntryManager` and `LinkedList` as we would want to increase the number of entries that we have sorted when the height of the results window increases. It will probably require updating things like `entry_manager.max_results`, `entry_manager.worst_acceptable_score`, `linked_list.track_at`, `linked_list._tracked_node`, `linked_list.tracked`.

@Conni2461 could you have a look through #927 and make sure that I haven't misunderstood anything before I change stuff there? Don't want to start implementing stuff with errors in my understanding 😅

l-kershaw:
I'm happy to wait to finalise the implementation for this until after the async changes.

On a related note, could you point me in the direction of where I can find information about the new async stuff? I've seen a few people referring to "async2" but I can't figure out where to find out about it 😅

Conni2461:
This is async await 2: https://github.com/nvim-lua/plenary.nvim/pull/148

tj just suggested. Lets put the new async into `plenary.async` rather than `plenary.async_lib2` https://github.com/nvim-lua/plenary.nvim/tree/master/lua/plenary/async

and we keep the original async, till all plugins moved to the new variant. For example `gitsigns` already switched to the new version

KodyVB:
Sorry if this isn't the place to ask, but could someone give me a rough estimate of when this would get merged to the project? I've never gotten experience with big projects like this before, so I just don't know if I should be expecting it anywhere from "any day now" to "maybe in a few months". Anyways, keep up the good work, y'all.

l-kershaw:
I am currently waiting for the big async rewrite that TJ is working on [here](https://github.com/nvim-telescope/telescope.nvim/pull/987).
"Asyncifying" the handling of the results list will likely make it much easier to fix the issues when the results window changes height. Therefore its not worth writing a solution for these problems until after the rewrite is complete.

In terms of timeline, I'm not really sure sorry. I would expect this PR to take less than a week once the async rewrite is done, but I can't really tell how long that will take.

@KodyVB I hope this answers your question 🙂

KodyVB:
@l-kershaw I see, thanks for the reply!

l-kershaw:
Updated demo for this PR:
vimresize

I **_think_** I have now fixed all the issues I am aware of for this.
There are a few places where the implementation might not be 100% efficient, but there were lots of edge cases that came up that were hard to cover.

If people want to test this PR and give feedback that would be great 😊 
Make sure you update your `plenary` as well first, as you need the `popup.move` implementation that was merged earlier today.

---

@Conni2461 @tjdevries I would be particularly interested if either of you guys have any thoughts on this PR 🙂

l-kershaw:
I have refactored `find` and `recalculate_layout` slightly to use a helper function called `_create_window`.
This helper function is based on helper functions in #1305 by @johnybx (e.g. `_create_results_window`), but tweaked so that we only need one function instead of three (thanks for the inspiration johny 🙂).

This makes it easier to standardise the creation of picker windows in both `find` and `recalculate_layout`.
@tjdevries is this sufficient in terms of your request for shared code between the two?

---

I have also added a vague plan for `toggle_padding` to be used when you need to maximise the size of the picker.
This would probably be a good first PR for anyone who wants to try it; I'd be more than happy to mentor someone on this 🙂

fdschmidt93:
> I have also added a vague plan for toggle_padding to be used when you need to maximise the size of the picker.
This would probably be a good first PR for anyone who wants to try it; I'd be more than happy to mentor someone on this

You can add this to https://github.com/nvim-telescope/telescope.nvim/issues/1190 if you want :)

l-kershaw:
I have pushed a change to `plenary` that fixes that issue that @johnybx explained above.
This check is now handled on the `plenary` side, and behavior around this for popups matches that of vim now.
This means that we can keep the highlighting options in the simplified form we have here.

tjdevries:
Seems to work well for me -- code seems much less repeated now. We can always clean up a bit more later if we think of anything good.

LGTM (could get tested by maybe one more person and then  let's merge)

Conni2461:
Works great for me :+1: feel free to merge this as if you are done :) Thanks for the PR :)

PR #965, opened 2021-07-09
feat: hopping to entry in results

fdschmidt93:
Opening this for discussion as to whether we'd want something akin to hop.nvim in telescope. I understand it might be a bit more "controversial" and maybe not something for the plugin as a whole, but rather something for a telescope-actions wiki (or something similar).

(teles)hopping: when action is called list keys in order via extmarks to jump to entry corresponding to key of choice

https://user-images.githubusercontent.com/39233597/125062803-1455d480-e0af-11eb-8d16-b3c8f124b8ab.mp4

I wanted to learn about the extmark api and thought hopping to an entry might be nice use case, especially whenever you look for something that's already within your say top3-10 results. To that end, the video shows hopping to `lsp_document_symbols` filtered for functions and methods.

The implementation is not yet 100% polished, but happy to if it's something for telescope as a whole. Open TODOs/Questions:

- ~~Chaining `actions.hop` + `actions.select_default` doesn't work out of the box as the buffer previewer then complains that the buffer we hopped to is missing (latter action can be manually `vim.schedule`d to do everything in proper order I guess, but would be nice to not have to do that~~
- How to pass `opts` nicely to configure keys / highlight group / extmark position / ...

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -680,6 +680,55 @@ actions.smart_add_to_loclist = function(prompt_bufnr)
   smart_send(prompt_bufnr, 'a', 'loclist')
 end
 
+actions.hop = function(prompt_bufnr, opts)
+  opts = opts or {}
+  local escape = 27
+  local default_keys = {
+    "a", "s", "d", "f", "g", "h", "j", "k", "l", ";",
+    "q", "w", "e", "r", "t", "u", "v", "w",
+    "A", "S", "D", "F", "G", "H", "J", "K", "L", "W", "E", "R", "T", "U"}
+  opts.keys = vim.F.if_nil(opts.keys, default_keys)
+  opts.hl_group = vim.F.if_nil(opts.hl_group, "Search")
+
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local max_results = current_picker.max_results
+  local num_results = current_picker.manager:num_results()
+  local results_bufnr = current_picker.results_bufnr
+  local sorting_strategy = current_picker.sorting_strategy
+
+  local ns = vim.api.nvim_create_namespace "teleshopping"
+  local keyline = {}
+
+  for i = 1, math.min(num_results, max_results, #opts.keys) do
+    local key = opts.keys[i]
+    local linenr = sorting_strategy == "descending" and max_results - i or i
+    vim.api.nvim_buf_set_extmark(
+      results_bufnr,
+      ns,
+      linenr,
+      0,
+      { virt_text = { { key, opts.hl_group } }, virt_text_pos = "overlay" }
+    )
+    keyline[key] = linenr
+  end
+  -- ensure marks are drawn before getchar is executed
+  vim.cmd [[redraw]]
+
+  local char = vim.fn.getchar()
+  local key = vim.fn.nr2char(char)
+
+  -- TODO fix error of buffer previewer
+  if char ~= escape and keyline[key] ~= nil then
+    current_picker:set_selection(keyline[key])
+    if opts.callback then
+      opts.callback()
+    end
+  end
+
+  vim.api.nvim_buf_clear_namespace(results_bufnr, ns, 0, -1)
+end
+
+
 actions.complete_tag = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   local tags = current_picker.sorter.tags
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -704,6 +704,7 @@ end
 ---     },
 ---   }
 --- </pre>
+---@param prompt_bufnr number: The prompt bufnr
 ---@param opts table: options to pass to hop
 ---@field keys table: table of chars in order to hop to, roughly defaults to lower and upper-cased home row
 ---@field sign_hl string|table: hl group to link hop chars to; if table, must be two groups that are alternated between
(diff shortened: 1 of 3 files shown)


fdschmidt93:
@Conni2461 do you have any thoughts? many thanks in advance

lua/telescope/actions/init.lua
```diff
@@ -652,6 +652,55 @@ actions.smart_add_to_loclist = function(prompt_bufnr)
   smart_send(prompt_bufnr, 'a', 'loclist')
 end
 
+actions.hop = function(prompt_bufnr, opts)
+  opts = opts or {}
+  local escape = 27
+  local default_keys = {
+    "a", "s", "d", "f", "g", "h", "j", "k", "l", ";",
+    "q", "w", "e", "r", "t", "u", "v", "w",
```
Conni2461, line 660: w is mentioned twice.

Any reason to not include the whole alphabet?

```lua
for i = 97, 122 do -- a -> z
  print(i, vim.fn.nr2char(i))
end
```

Idk i am not in the hop.nvim world, idk what people do here :laughing:

lua/telescope/actions/init.lua
```diff
@@ -652,6 +652,55 @@ actions.smart_add_to_loclist = function(prompt_bufnr)
   smart_send(prompt_bufnr, 'a', 'loclist')
 end
 
+actions.hop = function(prompt_bufnr, opts)
```
Conni2461, line 655: Usually actions that accept a second argument are prefixed with `_` because they are not meant to be mapped.

I am thinking of having a `actions.hop(prompt_bufnr)` that calls `actions._hop` with the defaults you encoded below.
How do you feel about that?

Ref: see `actions._close`

lua/telescope/actions/init.lua
```diff
@@ -652,6 +652,55 @@ actions.smart_add_to_loclist = function(prompt_bufnr)
...
+    )
+    keyline[key] = linenr
+  end
+  -- ensure marks are drawn before getchar is executed
+  vim.cmd [[redraw]]
+
+  local char = vim.fn.getchar()
+  local key = vim.fn.nr2char(char)
+
+  -- TODO fix error of buffer previewer
```
Conni2461, line 692: I think you fixed that one :) good job.

Another option would be to teardown previewer before selection and set it to nil. Then it will not try to update the previewer. I think.

Conni2461:
It definitely looks cool, we could include that :) i am just not a hop guy :laughing: but with that we make moving more vim ish, at least for a part of the community (if that makes sense :speak_no_evil:)

clason:
Do https://github.com/ggandor/lightspeed.nvim next! :P

tami5:
WOW that seems a strong feature. Looking forward

lua/telescope/actions/init.lua
tami5: > ```lua
> for i = 97, 122 do -- a -> z
>   print(i, vim.fn.nr2char(i))
> end
> ```

This is black magic conni 😆 , did you sell your soul for it 🤣 🤣 🤣

fdschmidt93: I noticed I repeated some entries locally, but haven't converted the changes yet :sweat_smile: I think a manual configuration is preferred, since the table aligns with the home row. At the same time, I can imagine some users just like 1, 2, .., N better as that might more easily with the mental imagine of what constitutes an entry row.

In any case, I'll make it configurable :) and I think it might be cool to include an optional alternating highlighting of entry rows to even better visualize where the user would jump to

lua/telescope/actions/init.lua
fdschmidt93: Just to avoid ambiguity, without my fix the buffer previewer would have complained about something like this in default mappings 
```lua
i = {
  ['<C-space>'] = actions.hop + actions.select_default
}
```
Since -- my hunch -- the buffer was de facto teared down before the previewer executes the `vim.schedule`d function to show it (though as a non-CS person who's never really used async I might be wrong :).

Anyways, `actions.select_default` calls `actions.close` which includes a buffer preview teardown via ultimately calling `action_set.edit` which already and  probably just "too late" (due to scheduling) does what you're referring to? Consequently, tearing down the previewer in time I'm sure would avoid this but the validity check for the buffer previewer should not induce any adverse side effects and, if at all apart from this, silently avoid other similar hurdles in the future :sweat_smile: . So I'd prefer to leave it as with the check whether the buffer is valid in `buffer_previewer.lua`. 

(`<C-space>' works for me nicely since capslock is mapped to ctrl :laughing: )

lua/telescope/actions/init.lua
Conni2461: yep makes sense. I said i am not a hop person :laughing:

lua/telescope/actions/init.lua
Conni2461: Yeah i am fine with the current fix. Thanks :)

lua/telescope/actions/init.lua
fdschmidt93: I feel like I'm overthinking this. Is it straightforward for a user as you've proposed to pass his configuration for keys and hl group when setting mappings? 

I'm struggling to come up with the solution for what I'd love to do (inspired by `vim.lsp.handler` overrides) that would allow for constructing mappings of configurable actions akin to this

```
actions.hop(prompt_bufnr, opts)
  opts = opts or {}
  vim.validate{
    opts = {opts, "table"}
  }

  ...
end

i = {
  ['<C-space>'] = actions.with(actions.hop, {keys = {"1", "2", "3"}) + actions.select_default
}
```
That way we can introduce the convention that there are actions for which `opts` can be nicely overridden when mapped.
However, I'm mentally running against the wall right now on doing this real simple real clean real cool with the `action(prompt_bufnr)` pattern  :sweat_smile: 

I feel like having something like this would be really nice in the long run :) of course let's only go to that length if you agree that such a mapping pattern would be cool to have as there might be something easier to begin with.

Conni2461: We already have a "million" ways to setup actions, i would prefer not to add another one. Its already confusing for me, i do not know how a new user feels :laughing: 

This currently exist:
```lua
mappings = {
  i = {
    -- current
    ["<c-d>"] = actions.smart_send_to_qflist + ....,
    -- same as
    ["<c-d>"] = { actions.smart_send_to_qflist + ...., type = "action" }, 
    -- future opts = { nowait = true } in table for all tables. See #890 

    -- is basicall this:

    -- for viml
  }
}
```

fdschmidt93: I didn't see the forest for the trees last night :sweat_smile: 

I've changed it to what seemed most sensible in my self tests. I moved configuration and validation fully to `actions._hop` as otherwise partial overrides in configuration are a bit annoying.

fdschmidt93:
I've extended configuration such that (with the right highlighting groups) something along these lines is achievable for much better readability :)

hop_alternate

fdschmidt93:
During lunch break, I've actually thought of another cool extension to the idea of hopping to entries :sweat_smile: -- the hop loop: here's how you could now lever `actions.hop_toggle_select` to multi-select using hop.


https://user-images.githubusercontent.com/39233597/125447425-0e99bb6b-c0ec-46bb-8b92-f67b68cb3551.mp4

tjdevries:
So this is super, super, super cool.

I am watching the gifs and thinking "I definitely install this".

But, I think it will be better as an extension (you can add to wiki and I can create `nvim-telescope/telescope-hop.nvim` or similar. I think you're definitely missing out on a joke about zero-gravity or something though, since it's jumping & telescope together :laughing: ). I have a feeling that many users will want to tune this one way or another, add different default keys (different keyboards, languages, etc) and more, so it probably makes more sense to move that out and have a dedicated readme / help doc / etc. to read from there.

Does that sound fine to you?

clason:
"Moon hopping"?

fdschmidt93:
> But, I think it will be better as an extension [...] many users will want to tune this one way or another, add different default keys

Fully agree!
 
>  I can create nvim-telescope/telescope-hop.nvim

Sounds good, as per convention of other telescope extensions :)

I'll close the PR and will then continue over at the extension repo. I'm happy to then incorporate further ideas for `teles_hopping`. Let's see if we can come up with an even better pun by release of the extension :laughing:

tjdevries:
Awesome, feel free to ping me with any questions there. :) thanks for making it!

PR #972, opened 2021-07-10
Empty string text for closed buffer in tagstack picker.

caojoshua:
When there is an tagstack entry for a closed buffer, there is a `nil` error in [make_entry.lua#L312](https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/make_entry.lua#L312) because entry.text is nil. 

```
ordinal = (
        not opts.ignore_filename and filename
        or ''
        ) .. ' ' .. entry.text,
```

This fix enters empty text for entries with closed buffers. I think it makes sense, because most people are not regularly closing buffers, especially recently open ones, during their workflow. Therefore, people are unlikely to see entries for closed buffers high up on their Telescope list. So we use empty text as a way to avoid nil errors, but don't expect the empty text to be surfaced to the user often.

Steps to repro:
1. open buffer `A`
2. create a tagstack entry in `A` ie. builtin LSP jump to reference does so
3. close buffer `A`
4. `:Telescope tagstack`

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -924,7 +924,7 @@ internal.tagstack = function(opts)
       value.lnum - 1,
       value.lnum,
       false
-    )[1]
+    )[1] or ''
   end
 
   -- reverse the list
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -906,31 +906,30 @@ end
 
 internal.tagstack = function(opts)
   opts = opts or {}
-  local tagstack = vim.fn.gettagstack()
-  if vim.tbl_isempty(tagstack.items) then
+  local tagstack = vim.fn.gettagstack().items
+
+  local tags = {}
+  for i = #tagstack, 1, -1 do
+    local tag = tagstack[i]
+    tag.bufnr = tag.from[1]
+    if vim.api.nvim_buf_is_valid(tag.bufnr) then
+      tags[#tags + 1] = tag
+      tag.filename = vim.fn.bufname(tag.bufnr)
+      tag.lnum = tag.from[2]
+      tag.col = tag.from[3]
+
+      tag.text = vim.api.nvim_buf_get_lines(
+        tag.bufnr,
+        tag.lnum - 1,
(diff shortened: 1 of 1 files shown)


Conni2461:
If i close a, i actually get a `Invalid buffer id: 6` because you cant get the text of a closed/invalid buffer.

My proposed bugfix for this would be. If `nvim_buf_is_valid`, get text, else `""`. This does work for me but now i can't open it because open goes over the bufnr. Doing `value.bufnr = nil` if its not valid afterwards would make sure that the filepath is used again.

What do you think?

Conni2461:
So i just tested it and if the buffer is deleted we end up with nothing usable anymore (bufnr is wrong, ... take a look at `print(vim.inspect(value))` in the loop), so this tagstack item is useless for us. I would just ignore it `<c-t>` also says Buffer not found.

Still a good found. Thanks :)

caojoshua:
> So i just tested it and if the buffer is deleted we end up with nothing usable anymore (bufnr is wrong, ... take a look at `print(vim.inspect(value))` in the loop), so this tagstack item is useless for us. I would just ignore it `<c-t>` also says Buffer not found.
> 
> Still a good found. Thanks :)

TIL there is difference between `bdelete` and `bwipeout`.

Similar to our discussion on `:jumps` vs `vim.fn.jumplist()` in https://github.com/nvim-telescope/telescope.nvim/pull/973, `:tags` will not show entries for invalid buffers, while `vim.fn.gettagstack()` does return entries for invalid buffers. We should try to emulate `:tags` and remove invalid entries.

lua/telescope/builtin/internal.lua
```diff
@@ -906,31 +906,30 @@ end
...
-      value.lnum - 1,
-      value.lnum,
-      false
-    )[1] or ''
+      tag.text = vim.api.nvim_buf_get_lines(
+        tag.bufnr,
+        tag.lnum - 1,
+        tag.lnum,
+        false
+      )[1] or ''
```
Conni2461, line 926: same I think this or just doesnt do anything. doesnt matter tho :laughing:

Conni2461 (approved):
LGTM thanks :)

PR #973, opened 2021-07-10
Jumplist picker uses quickfix entry maker. Delete jumplist entry maker.

caojoshua:
When I authored #839, I didn't include the jumplist text because we wouldn't be able to get text for closed buffers. This is a non-issue because I've found in my workflow that that recent entries on the jumplist are always open in a buffer. I think that having the buffer text feels a little better when going through the jumplist picker. For entries that aren't open in a buffer, we include empty text, but we expect it to not be surfaced to the user very often.

Another benefit is that the jumplist picker now reuses the quickfix entry maker, as many other pickers already do. This means we can delete the jumplist entry maker.

This is all opinionated. Open to ideas.

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -951,9 +951,11 @@ internal.jumplist = function(opts)
   -- reverse the list
   local sorted_jumplist = {}
   for i = #jumplist, 1, -1 do
-    jumplist[i].text = vim.api.nvim_buf_get_lines(jumplist[i].bufnr, jumplist[i].lnum, jumplist[i].lnum+1,
-      false)[1] or ''
-    table.insert(sorted_jumplist, jumplist[i])
+    if vim.api.nvim_buf_is_valid(jumplist[i].bufnr) then
+      jumplist[i].text = vim.api.nvim_buf_get_lines(jumplist[i].bufnr, jumplist[i].lnum, jumplist[i].lnum+1,
+        false)[1] or ''
+      table.insert(sorted_jumplist, jumplist[i])
+    end
   end
 
   pickers.new(opts, {
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -951,7 +951,8 @@ internal.jumplist = function(opts)
   -- reverse the list
   local sorted_jumplist = {}
   for i = #jumplist, 1, -1 do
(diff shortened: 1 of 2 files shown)


lua/telescope/builtin/internal.lua
```diff
@@ -951,15 +951,16 @@ internal.jumplist = function(opts)
   -- reverse the list
   local sorted_jumplist = {}
   for i = #jumplist, 1, -1 do
-    jumplist[i].text = ''
+    jumplist[i].text = vim.api.nvim_buf_get_lines(jumplist[i].bufnr, jumplist[i].lnum, jumplist[i].lnum+1,
+      false)[1] or ''
     table.insert(sorted_jumplist, jumplist[i])
```
Conni2461, line 956: Hmm maybe we should only do it if the buffer is valid. before that we had the check in entry maker but now we should do it before we do `nvim_buf_get_lines` otherwise it throws an error
So lets only do it `vim.api.nvim_buf_is_valid(jumplist[i].bufnr` just in case

caojoshua: So apparently `:jumps` only outputs jumplist entries with valid buffers, while `vim.fn.getjumplist()` will output entries with invalid buffers.

Based on your suggestion, we should more closely emulate `:jumps`, which I agree with. Pushed the change.

lua/telescope/builtin/internal.lua
```diff
@@ -951,15 +951,18 @@ internal.jumplist = function(opts)
   -- reverse the list
   local sorted_jumplist = {}
   for i = #jumplist, 1, -1 do
-    jumplist[i].text = ''
-    table.insert(sorted_jumplist, jumplist[i])
+    if vim.api.nvim_buf_is_valid(jumplist[i].bufnr) then
+      jumplist[i].text = vim.api.nvim_buf_get_lines(jumplist[i].bufnr, jumplist[i].lnum, jumplist[i].lnum+1,
+        false)[1] or ''
```
Conni2461, line 956: I think this or just doesnt do anything. doesnt matter tho :laughing:

Conni2461 (approved):
LGTM thanks :)

lua/telescope/builtin/internal.lua
caojoshua: It does something in my testing. If I `bdelete` a buffer, than the buffer is still valid, but we can't get the lines for it. Only when I `bwipeout` a buffer does a buffer become invalid. I'm not 100% on the behavior of buffers in vim and neovim, but I think `bdelete` keeps a lot of the buffer data around, but unloads the file text from memory.

Issue #429, opened 2021-01-12
Pressing enter in an empty list and moving cursor left/right blocks backspace


OkanEsen:
### Description
As per: https://github.com/nvim-telescope/telescope.nvim/issues/231#issuecomment-758282094:

> [...] The second issue still persists though: just fill the input, until no entry is matched/visible and press enter. Cursors moves len(input) - 1 and moving the cursor to the end and pressing backspace is prevented.
> 
> Video: https://i.imgur.com/srnllRo.mp4

It actually doesn't matter, if the cursor is moved to the end or not, it happens always after the cursor move.

**Expected Behavior**
Backspace should not be prevented.

**Actual Behavior**
Backspace is being prevented.

### Details

See video or:

1. Open file finder
2. Type anything which results in no entries.
3. Press enter
4. Move the cursor left or right
5. Press backspace

Typing another char fixes the issue.

caojoshua:
https://github.com/nvim-telescope/telescope.nvim/pull/974 prevents the cursor from moving

> Backspace is being prevented.

see https://github.com/nvim-telescope/telescope.nvim/issues/852. upstream prompt buffer bug

Issue #852, opened 2021-05-19
Backspace doesn't work after normal mode


kuator:
Backspace doesn't work after normal mode
Reproduction steps:
1. `:Telescope find_files`
2. `search-pattern<esc>i`
3. try to delete inserted text with backspace, it doesn't work

LamprosPitsillos:
Having the same problem here...

elianiva:
should've been fixed by this PR https://github.com/neovim/neovim/pull/16421

kuator:
> should've been fixed by this PR [neovim/neovim#16421](https://github.com/neovim/neovim/pull/16421)

can confirm, it works

ryoppippi:
I have the same problem still rn

Conni2461:
Its a neovim bug which was fixed in the latest nightly. If you are still running 0.6.* of course you still experience this bug. This bug will be fixed with neovim 0.7 for you

ryoppippi:
Ok Thanks

PR #974, opened 2021-07-11
Remove center action after select_default action in default mappings.

caojoshua:
Partially fixes https://github.com/nvim-telescope/telescope.nvim/issues/429

default mapping change: `["<CR>"] = actions.select_default + actions.center` -> `["<CR>"] = actions.select_default`

When selecting to edit an empty entry, the telescope floating window does not close. Then, `actions.center` changes the prompt buffer to normal mode triggering prompt buffer bugs in https://github.com/nvim-telescope/telescope.nvim/issues/429 and https://github.com/nvim-telescope/telescope.nvim/issues/852.

I also think its awkward that by default, `select_default` mapping is followed by `actions.center`, while `select_(horizontal|vertical|tab)` default mappings are not. This change brings some consistency. The `actions.center` action is kept around in case users want to map to it.

If we want, we could add `vim.cmd(':normal! zz')` to the end of the edit function, but I feel its unnecessary.

### DIFF
--- a/lua/telescope/mappings.lua
+++ b/lua/telescope/mappings.lua
@@ -16,7 +16,7 @@ mappings.default_mappings = config.values.default_mappings or {
       ["<Down>"] = actions.move_selection_next,
       ["<Up>"] = actions.move_selection_previous,
 
-      ["<CR>"] = actions.select_default + actions.center,
+      ["<CR>"] = actions.select_default,
       ["<C-x>"] = actions.select_horizontal,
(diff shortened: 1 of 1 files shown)


elianiva (approved):
I agree with your reasoning. It's annoying when I accidentally pressed enter on an empty entry.

tjdevries:
Yeah, especially given the prompt buffer bugs, let's remove this and let people do it themselves if they want I think seems fine.

PR #976, opened 2021-07-11
Fix: Define undefined variable in config example

j0le:
Define the undefined variable “action_state”, which is used in a code
example for replacing picker mappings in the README.

The new definiton is added 7 lines above the usage of that variable:

``` lua
  local entry = action_state.get_selected_entry()
```

I haven't tested this yet. I just stumbled across this, while reading the documentation, and i wondered: “Where does this variable come from?”


elianiva:
LGTM, thanks.

PR #977, opened 2021-07-11
add use_regex option to live_grep

Kartiku:
Would be useful to have this option for live_grep as it is for grep_string


elianiva:
Hmm, I think the better solution is #670, enabling regex on live_grep was causing some issue with incomplete regex, that's why it got disabled.

Kartiku:
very nice.  Passing cmdline args will be useful to do things like filter grep by file extension.  Will close this

PR #978, opened 2021-07-12
Show results correctly when sorting_strategy=ascending is set

### LINKED ISSUE #840: Setting sorting to ascending with the previewer causes selection caret to disappear
### Description
Setting `sorting-strategy` to `ascending` with the previewer showing causes selection caret to disappear. If the `sorting-strategy` is set to `descending`, the selection caret does not disappear, and works as expected.

### To Reproduce

1. Set `sorting-strategy` to `ascending`

2. Execute `Telescope find_files` in a folder that has a lot of files and subfolders.

3. Make sure the the previewer is showing.

4. Search for something; the selection caret disappears.

### Video

In the first half of the video, I execute `Telescope find_files` in a folder that doesn't have subfolders. The selection caret does not disappear. The selection caret seems to disappear when there are a lot of files and subfolders and the previewer is showing.

ss

JRasmusBm:
**Why** is the change needed?

See #840

When setting sorting_strategy to ascending, the wrong line range is
shown in the result making the result impossible to follow.

**How** is the need addressed?

-   Basically adding the [fix](https://github.com/nvim-telescope/telescope.nvim/issues/840#issuecomment-867127533) that @jedrzejboczar suggested in the issue.

What **side-effects** could the change have?

- Can't think of any, are there better ways to do this?

Closes #840
Co-authored-by: yendreij <yendreij@gmail.com>

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -886,6 +886,8 @@ function Picker:entry_adder(index, entry, _, insert)
     if insert then
       if self.sorting_strategy == 'descending' then
         vim.api.nvim_buf_set_lines(self.results_bufnr, 0, 1, false, {})
+      else
+        vim.api.nvim_buf_set_lines(self.results_bufnr, self.max_results - 1, self.max_results, false, {})
       end
     end
 


rachit-ranjan16 (approved):

Conni2461:
I already said this in the issue, i can't reproduce. Also these changes so nothing for me, so technically i am fine with merging them but it seems that you have found the issue.

I am fine with setting a filetype for the results buffer, somewhere here https://github.com/nvim-telescope/telescope.nvim/blob/5692edd004fed1dbd55c90775c34a051298724e7/lua/telescope/pickers.lua#L473 `TelescopeResults` as name?

Would also be good if other people that experienced this issue could confirm that this issue is caused by fastfold

jedrzejboczar:
I can confirm that disabling FastFold seems to solve the issue, though I have no idea how it could cause it. I tested by disabling FastFold and reverting the changes included in this PR, so it looks like that's just FastFold.

As a side note abour reproducing the issue, for me this problem happens only on large code bases (e.g. with 2162 files found by `fd --type f | wc -l`) and only as long as the filtering query results in more results than there are lines in the results window.

JRasmusBm:
@Conni2461 Thank you for taking the time 😄

Here's how I figured it out: I have a very modular vim config so when I realized that it was not straightforward to reproduce I thought it might be something in my config. 

I then started with a minimal Telescope setup and that worked. Then I recursively deleted half of the files and checked if the problem persisted until I had narrowed it down to the one plugin. 😇

I agree that adding a file type and advising users to disable FastFold for the file type is probably the best solution, so I will change my PR to that. Should we also add it to the documentation somehow, or what is the best mechanism to help other people who stumble over the plugin config know what to do?

Conni2461:
I should have been more clear, i can not reproduce this without fastfold. Thats what i care about :laughing: but if you both have this problem because fastfold is loaded then filetype is the better solution.

You can add it here https://github.com/nvim-telescope/telescope.nvim/wiki/Known-Issues-and-Troubleshooting

JRasmusBm:
It's all good, that's how I understand you! 😄 I really like telescope and completely agree that we shouldn't add stuff to it that are unnecessary.

Like you said the best and smallest change is to add filetypes to the the result (and preview window for good measure). I'll make the change when I have time! Thanks for following up! 😇

JRasmusBm:
On second thought I won't change the preview filetype, it might be used to decide how to highlight that buffer :innocent:

JRasmusBm:
After some further reading I discovered that filetype skipping is broken in FastFold (https://github.com/Konfekt/FastFold/issues/40). 

I also read that the Neovim implementation of folds is more efficient anyways (https://www.reddit.com/r/neovim/comments/6fmddc/does_neovim_improve_on_vims_fold_implementation/) so I might actually just uninstall FastFold.

If @jedrzejboczar or someone else that has the issue wants to keep pushing for the change that's fine with me, but on consideration I'm happy to uninstall FastFold. :innocent:

Conni2461:
Filetype for Previewer is indeed more difficult, i would skip that. But can you please add the filetype for the Results buffer none the less? We have a similar issue with #991 and then we should probably disable folds for that filetype.

Would be cool if you could do that :) thanks

JRasmusBm:
@Conni2461 I'll definitely do that, but since that doesn't address #840 I will create a separate PR for that.

We should probably make a decision whether to merge this PR or to close #840 as a wontfix, though. 😇

JRasmusBm:
Force push: rebase to keep this branch up to date with master until a decision has been made.

Conni2461:
Yeah i dont wanna do this because i do not know if that has a negative effect down the line

shivamchandra75:
> I agree that adding a file type and advising users to disable FastFold for the file type is probably the best solution,

I don't understand what is fastFold and How to disable it.... I am on latest stable branch of telescope and this issue is from 2021 it should be solved by now.

jamestrew:
@shivamchandra75
Looks like FastFold is a vim plugin.

This supposedly can't be reproduced with telescope alone so you must have some configuration or plugin interfering with telescope behavior.

PR #979, opened 2021-07-12
feat: add Windows `where` file finder command

l-kershaw:
Adds the Windows `where` command as a default file finder.

Note, since `where` doesn't have many options:
- `hidden` and `follow` flags will not change anything
- `search_dirs` only supports a single search directory

I also added an extra check for the `find` command as it is different on Windows and Linux.

---

This allow Windows users to have a usable experience without installing external dependencies that are currently labelled as optional.

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -187,7 +187,6 @@ files.find_files = function(opts)
         end
       end
     elseif 1 == vim.fn.executable("where") then
-      print('using the `where` function')
       find_command = { 'where', '/r', '.', '*'}
       if hidden then
         print('The `hidden` key is not available for the Windows `where` command in `find_files`.')
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -7,6 +7,7 @@ local pickers = require('telescope.pickers')
 local previewers = require('telescope.previewers')
 local utils = require('telescope.utils')
 local conf = require('telescope.config').values
+local log = require('telescope.log')
 
 local scan = require('plenary.scandir')
 local Path = require('plenary.path')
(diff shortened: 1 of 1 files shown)


lua/telescope/builtin/files.lua
```diff
@@ -186,6 +186,26 @@ files.find_files = function(opts)
           table.insert(find_command, 2, v)
         end
       end
+    elseif 1 == vim.fn.executable("where") then
+      print('using the `where` function')
```
Conni2461, line 190: no print please for that :)

lua/telescope/builtin/files.lua
```diff
@@ -186,6 +186,26 @@ files.find_files = function(opts)
...
+        table.remove(find_command, 3)
+        local dir_num = 1
+        for _,v in pairs(search_dirs) do
+          if dir_num == 1 then
+            table.insert(find_command, 3, v)
+          else
+            print('Multiple `search_dirs` are not available for the Windows `where` command in `find_files`.')
+          end
+        end
+      end
```
Conni2461, line 208: hmm `search_dirs` are explicitly only meant for multiple directories. For single dir `cwd` should be preferred. Its set on `Job` level, so it should work without more work.

lua/telescope/builtin/files.lua
l-kershaw: Sorry, left it in from debugging 😅

lua/telescope/builtin/files.lua
l-kershaw: > hmm `search_dirs` are explicitly only meant for multiple directories. For single dir `cwd` should be preferred. 

Okay, so would be best to remove this `search_dirs` bit and replace with a similar message to the other parts, right?

> Its set on `Job` level, so it should work without additional work.

I'm not really sure what you mean by this, could you elaborate?

l-kershaw:
Also, just thought, would it be better to use the logging system with `log.warn` or something? Then the message to the user will give a bit more information?

lua/telescope/builtin/files.lua
Conni2461: sure.

we do `finder = finders.new_oneshot_job(find_command, opts),` which will start a job (uv.spawn + more) https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/finders/async_oneshot_finder.lua#L16

This is simplified. You end up with something like this:
```lua
local Job = require("plenary.job")

-- obv we dont have sync here
local result = Job:new({ 'where', '/r', '.', '*', cwd = cwd1}):sync()
```

But now you can change the `cwd1` var to change the directory and thats what happens with `Telescope find_files cwd=...` similar applies to other pickers. If you only wanna change one cwd, use cwd :) its just not documented, smith missed it when reading code, but thats fine

Feel free to ask more questions when you run into them :)

Ref job: https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/job.lua
Ref spec: https://github.com/nvim-lua/plenary.nvim/blob/master/tests/plenary/job_spec.lua

Conni2461:
Sure you can do `log.warn`. Works for me :)

Thanks for doing this btw. I dont know windows, so its good to have someone on the team that does :)

lua/telescope/builtin/files.lua
l-kershaw: Ah, I understand now, thanks 👍

l-kershaw:
I've swapped to using `log.warn` for the messages here.
I'll probably try to standardise our logging elsewhere when I get a chance.

---

One problem at the moment is that any messages appear then disappear immediately as we enter insert mode.
I can't think of a better solution at the moment, so this is probably good enough for now.

Conni2461 (approved):
LGTM. Thanks :)

Cant test it so if it works you can merge :)

PR #980, opened 2021-07-13
feat: enable to preview themes

xiyaowong:
Preview the selected theme, applies it on <cr> or restore original theme
automatically.

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -634,26 +634,72 @@ internal.buffers = function(opts)
 end
 
 internal.colorscheme = function(opts)
+  local before_color = vim.api.nvim_exec('colorscheme', true)
+  local need_restore = true
+
   local colors = vim.list_extend(opts.colors or {}, vim.fn.getcompletion('', 'color'))
 
-  pickers.new(opts,{
-    prompt = 'Change Colorscheme',
+  -- used for hiding the preview window
+  local deleted
+  local function del_win(win_id)
+    if win_id == nil or not vim.api.nvim_win_is_valid(win_id) then
+      return
+    end
+
+    local bufnr = vim.api.nvim_win_get_buf(win_id)
+    if vim.api.nvim_buf_is_valid(bufnr) and not vim.api.nvim_buf_get_option(bufnr, 'buflisted') then
+      vim.cmd(string.format('silent! bdelete! %s', bufnr))
+    end
+
+    if not vim.api.nvim_win_is_valid(win_id) then
+      return
+    end
+    pcall(vim.api.nvim_win_close, win_id, true)
+  end
+
+  local picker
+  picker = pickers.new(opts, {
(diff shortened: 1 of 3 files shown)


Conni2461:
Cool, works good so far, would also be cool to have the current file opened in the previewer buffer rather than closing the window. Will do code review tomorrow :)

CC @teto

xiyaowong:
yes, that' what I thought at first, but I found it's easier to close than to preview, so I just implement close. I will make some commits later:smile:

lua/telescope/builtin/internal.lua
```diff
@@ -634,26 +634,90 @@ internal.buffers = function(opts)
...
+  if vim.fn.buflisted(bufnr) ~= 1 then
+    local deleted = false
+    local function del_win(win_id)
+      if win_id == nil or not vim.api.nvim_win_is_valid(win_id) then
+        return
+      end
+
+      local bnr = vim.api.nvim_win_get_buf(win_id)
+      if vim.api.nvim_buf_is_valid(bnr) and not vim.api.nvim_buf_get_option(bnr, 'buflisted') then
+        vim.cmd(string.format('silent! bdelete! %s', bnr))
```
Conni2461, line 655: We have a util function for that `utils.buf_delete(bufnr)`

https://github.com/nvim-telescope/telescope.nvim/blob/37a3a68a78e15900254ae5d9d4f9f0b5100a1735/lua/telescope/utils.lua#L338-L350

tjdevries:
Only thing I'd add is to add an option to disable auto switching, because some colorschemes do not reload nicely, so it may be best to allow users to simply do the fuzzy finding over them.

Otherwise idea & impl seem pretty good to me

xiyaowong:
got it :)

Conni2461:
LGTM, can we get some docs here https://github.com/nvim-telescope/telescope.nvim/blob/138697980b68ce7d4f0661c752691ffe5dbb72b0/lua/telescope/builtin/init.lua#L255-L257 Like `---@field enable_preview boolean: ....`

I dont think `docgen` will run because its disabled for master (and you are working on master) but i will run it locally then :)

Thanks :)

xiyaowong:
Shall I edit something？Actually, I don't know about docgen 🤣 @Conni2461

Conni2461:
Yeah can you add the documentation there i will generate the vim docs from it ;) basically after 256 before 257 :laughing: If you are still not sure what i am talking about ill do it. Just tell me :laughing:

xiyaowong:
@Conni2461 There is another option called colors here, I think it's useless, and I keep it just for compatibility.

Conni2461:
I am actually not sure what it does :laughing: 

LGTM thanks for the pr :)

Conni2461:
To be more precise, i see what it does but i am not sure of the use case and why it was added. I cant remember

teto:
Just had the opportunity to test, that's dope ! Thanks a lot !

PR #986, opened 2021-07-14
add `search_dirs` to split_keywords table

### LINKED ISSUE #985: `search_dirs` option isn't parsed as a table
### Description
Running `:Telescope find_files search_dirs=~` throws an error because it fails to parse `search_dirs` as a table

**Expected Behavior**
`search_dirs` should be parsed correctly

**Actual Behavior**
The command above throws the following error:

```
E5108: Error executing lua ...ker/start/telescope.nvim/lua/telescope/builtin/files.lua:142: bad argument #1 to 'pairs' (table expected, got string)
```

jonathannerat:
Closes #985.

I might need some feedback on if this is the correctly way to deal with a new split keyword since we also have `command.register_keyword` but it isn't being used anywhere. Maybe it's for extensions only and since `search_dirs` is for builtins, it should be in the `split_keywords` table?

### DIFF
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -123,7 +123,8 @@ end
 local split_keywords = {
   ['find_command'] = true,
   ['vimgrep_arguments'] = true,
-  ['sections'] = true
+  ['sections'] = true,
+  ['search_dirs'] = true
 }
 
 function command.register_keyword(keyword)


Conni2461 (approved):
LGTM thanks :)

`register_keyword` was only meant for extensions but we never documented it :laughing:

PR #992, opened 2021-07-14
chore: changelog for 473

Conni2461:
CC @clason

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -1,7 +1,7 @@
 local has_devicons, devicons = pcall(require, 'nvim-web-devicons')
 
 local Path = require('plenary.path')
-local Job     = require('plenary.job')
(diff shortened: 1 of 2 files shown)


lua/telescope/utils.lua
```diff
@@ -251,6 +251,10 @@ utils.diagnostics_to_tbl = function(opts)
   return items
 end
 
+utils.path_shorten = function(file)
```
Conni2461, line 254: We should have kept this as shorthand, but make it work for the new interface

clason: That's probably the biggest one, yes.

clason:
Yes, that is very helpful, thank you!

clason:
What about `path_tail`?

Conni2461:
`path_tail` was always in utils.

PR #993, opened 2021-07-14
fix: set filetype of preview buffer

tdfirth:
# What does this PR address?

When opening telescope with a preview buffer, I get an `unknown filetype` error message. When opening telescope without the preview buffer showing (for example on a narrow window), I do not get such an error message.

# How does the PR fix this?

I have simply added one line in the buffer preview code to set the filetype of the preview buffer based on the filetype that was detected by plenary. This resolves the issue for me.

I'm not familiar with the code base (or writing vim plugins in general) so do let me know if there's a better way to accomplish this.

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -116,6 +116,8 @@ previewers.file_maker = function(filepath, bufnr, opts)
       end
     end
   end
+
+  pcall(vim.api.nvim_buf_set_option, bufnr, 'filetype', ft)
 end
 
 previewers.new_buffer_previewer = function(opts)


l-kershaw:
Hi! Thanks for writing a PR to fix your bug 🙂 

We normally recommend opening an issue for bugs first. This would then prompt you to give information about how to reproduce your error with a minimal vimrc and to provide more info about your configuration/general setup. This gives the maintainers more information about your bug, and makes it easier for us to figure out what is causing it.

Would you be able to open an issue for this and fill out the template? Then we can see if your changes fix the root cause of the issue or just prevent a symptom of a deeper problem.
Thanks 🙂

Conni2461:
This is not a bug. We are deliberately not doing this. Doing `setf` will run some autocmds which makes things slower. Sorry this is an issue with your setup. As @l-kershaw said please open an issue and we can help you figure out whats wrong :) I tried your vimrc (added telescope) when you submitted the PR yesterday and couldn't reproduce it. Now its lua based so its no longer that easy for me to try. So please try and issue :)

matu3ba:
@tdfirth This is a (to me at least) known issue: https://github.com/nanotee/nvim-lua-guide/issues/69.
And in neovim core: https://github.com/neovim/neovim/issues/14774#issuecomment-868461116

I suspect this requires figuring out the complete vim and lua settings load order of neovim.

PR #994, opened 2021-07-14
fix: prevent error if preview scroll is triggered but no preview is available

abecodes:
In some pickers there is no preview buffer available (e.g. symbols). Pressing <C-d> or <C-u> then results in an error.
Adding a check to make sure a previewer is available and it can scroll to prevent the throwing of an error.

### DIFF
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -155,7 +155,6 @@ action_set.scroll_previewer = function (prompt_bufnr, direction)
     return
   end
 
-
   local status = state.get_status(prompt_bufnr)
   local default_speed = vim.api.nvim_win_get_height(status.preview_win) / 2
   local speed = status.picker.layout_config.scroll_speed or default_speed
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -148,11 +148,19 @@ end
 ---@param direction number: The direction of the scrolling
 --      Valid directions include: "1", "-1"
 action_set.scroll_previewer = function (prompt_bufnr, direction)
+  local picker = action_state.get_current_picker(prompt_bufnr)
+
+  -- Check if we actually have a previewer
+  if (type(picker['previewer']) ~= "table" or picker.previewer['scroll_fn'] == nil) then
+    return
+  end
+
+
   local status = state.get_status(prompt_bufnr)
   local default_speed = vim.api.nvim_win_get_height(status.preview_win) / 2
   local speed = status.picker.layout_config.scroll_speed or default_speed
 
-  action_state.get_current_picker(prompt_bufnr).previewer:scroll_fn(math.floor(speed * direction))
+  picker.previewer:scroll_fn(math.floor(speed * direction))
 end
 
 -- ==================================================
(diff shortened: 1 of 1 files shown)


lua/telescope/actions/set.lua
```diff
@@ -148,11 +148,19 @@ end
 ---@param direction number: The direction of the scrolling
 --      Valid directions include: "1", "-1"
 action_set.scroll_previewer = function (prompt_bufnr, direction)
+  local picker = action_state.get_current_picker(prompt_bufnr)
```
Conni2461, line 151: Hmm i think i prefer `local previewer = action_state.get_current_picker(prompt_bufnr)` makes the things later a little bit more clear.

Otherwise LGTM thanks :)

lua/telescope/actions/set.lua
```diff
@@ -148,11 +148,19 @@ end
...
+
+  -- Check if we actually have a previewer
+  if (type(picker['previewer']) ~= "table" or picker.previewer['scroll_fn'] == nil) then
+    return
+  end
+
+
```
Conni2461, line 158: Not two empty lines here :)

lua/telescope/actions/set.lua
```diff
@@ -148,11 +148,19 @@ end
 ---@param direction number: The direction of the scrolling
 --      Valid directions include: "1", "-1"
 action_set.scroll_previewer = function (prompt_bufnr, direction)
+  local picker = action_state.get_current_picker(prompt_bufnr)
+
+  -- Check if we actually have a previewer
+  if (type(picker['previewer']) ~= "table" or picker.previewer['scroll_fn'] == nil) then
```
tjdevries, line 154: can you do `picker.previewer` and `picker.previewer.scroll_fn`

lua/telescope/actions/set.lua
abecodes: Good catch; )

lua/telescope/actions/set.lua
abecodes: Got your point, but previewer.previewer looks strange. I went with current_picker.

lua/telescope/actions/set.lua
abecodes: Done. Still new to lua ;)

lua/telescope/actions/set.lua
Conni2461: Oh sorry i meant to only get the previewer because we never use the current picker.
`local previewer = action_state.get_current_picker(prompt_bufnr).previewer`

Conni2461: But tbh its fine like this as well

Conni2461:
Thanks

lua/telescope/actions/set.lua
abecodes: That makes great sense, should have seen that when I was writing `previewer.previewer`.
Thanks a lot.

Conni2461: no worries, is was kinda my bad :laughing:

Conni2461:
LGTM Thank you for fixing this :)

PR #999, opened 2021-07-15
feat: add git reset action for git commits picker

joelpalmer:
**Adding a [git reset](https://git-scm.com/docs/git-reset) action and making it available in the `git_commits` picker.** 

This PR:
- Adds a `git_reset` action
- Adds the above action to the `git_commits` picker
- Maps `<c-r>` to call the action in either normal or insert mode

Like similar actions, the `git_reset` action has a confirmation along with handling of the error and success cases and their messages. The default for git reset is [--mixed](https://git-scm.com/docs/git-reset#Documentation/git-reset.txt---mixed). That behavior is continued here. If this PR is accepted, I'd like to follow-up with [--soft](https://git-scm.com/docs/git-reset#Documentation/git-reset.txt---soft) and [--hard](https://git-scm.com/docs/git-reset#Documentation/git-reset.txt---hard) options in a future PR. I personally have my local version hardcoded to `--hard`. But, that would probably surprise and upset many users if it were the default. Git's default is safe and friendly. 😄 

The reason for adding the new action to the `git_commits` picker is because it's Telescope's `git log` picker. It is convenient, intuitive and awesome to `reset` to the selected commit. I've been using it locally for a while and it saves me from having to copy/paste the SHA or count backwards from `HEAD`.  Who has time for all that? Telescope to the rescue as usual. 🔭

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,7 +468,7 @@ document symbols not recognized as methods by treesitter.
 
 | Functions                           | Description                                                                                                |
 |-------------------------------------|------------------------------------------------------------------------------------------------------------|
-| `builtin.git_commits`               | Lists git commits with diff preview and checks them out on `<cr>`                                          |
+| `builtin.git_commits`               | Lists git commits with diff preview, checkout action `<cr>`, reset mixed `<C-r>m`, reset soft `<C-r>s` and reset hard `<C-r>h` |
 | `builtin.git_bcommits`              | Lists buffer's git commits with diff preview and checks them out on `<cr>`                                 |
 | `builtin.git_branches`              | Lists all branches with log preview, checkout action `<cr>`, track action `<C-t>` and rebase action`<C-r>` |
 | `builtin.git_status`                | Lists current changes per file with diff preview and add action. (Multi-selection still WIP)               |
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -542,6 +542,28 @@ actions.git_rebase_branch = function(prompt_bufnr)
   end
 end
 
+--- Reset to selected git commit
+---@param prompt_bufnr number: The prompt bufnr
+actions.git_reset_branch = function(prompt_bufnr)
+  local cwd = action_state.get_current_picker(prompt_bufnr).cwd
+  local selection = action_state.get_selected_entry()
+
+  local confirmation = vim.fn.input('Do you really wanna reset to commit ' .. selection.value .. '? [Y/n] ')
+  if confirmation ~= '' and string.lower(confirmation) ~= 'y' then return end
+
+  actions.close(prompt_bufnr)
+  local _, ret, stderr = utils.get_os_command_output({ 'git', 'reset', selection.value }, cwd)
+  if ret == 0 then
+    print("Reset to: " .. selection.value)
+  else
+    print(string.format(
+      'Error when resetting to: %s. Git returned: "%s"',
+      selection.value,
+      table.concat(stderr, '  ')
+    ))
+  end
+end
+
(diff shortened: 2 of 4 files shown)


Conni2461 (approved):
LGTM can we get some docs for the key mappings here https://github.com/nvim-telescope/telescope.nvim/blob/138697980b68ce7d4f0661c752691ffe5dbb72b0/lua/telescope/builtin/init.lua#L143-L148

Similar to `git_branches`

Conni2461:
Also i am interested, how do you wanna do configuring of `--soft` / `--hard` with opts? So `require('telescope.builtin').git_commits({ reset = "hard" })`? Just curious right now because i cant think of a better idea

joelpalmer:
> LGTM can we get some docs for the key mappings here

Absolutely! Pushed. 

> Also i am interested, how do you wanna do configuring of `--soft` / `--hard` with opts? So `require('telescope.builtin').git_commits({ reset = "hard" })`? Just curious right now because i cant think of a better idea

That's a good idea for it. I was still thinking about it. 😄  

Thank you @Conni2461!

tjdevries:
I don't see any reason not to allow for not allowing to do `actions.git_reset_soft`, `actions.git_reset_mixed`, `actions.git_reset_hard` and just having them all share the same implementation except for the argument sent to git.

joelpalmer:
Sweet. I like it. I'll update this accordingly.

joelpalmer:
Updated per [the above comment](https://github.com/nvim-telescope/telescope.nvim/pull/999#issuecomment-881802351) from @tjdevries. 

This PR now has these actions (_which share implementation_) available in the `git_commits` picker: 
- `actions.git_reset_soft `
- `actions.git_reset_mixed` 
- `actions.git_reset_hard` 

Keymaps:
- `<C-r>m`: resets current branch to selected commit using _mixed_ mode
- `<C-r>s`: resets current branch to selected commit using _soft_ mode
- `<C-r>h`: resets current branch to selected commit using _hard_ mode

Docs have been updated to reflect these updates. 😃

l-kershaw:
I'm not sure what @tjdevries and @Conni2461 would prefer, but to me it seems like a better user experience to just have `<C-r>` mapped to mixed reset by default and no mappings for the other reset modes. We would then mention in the docs (though I'm not sure where is the best place for this) that this can be remapped to another mode, or removed to implement multiple keymappings in the users preferred way.

This then doesn't overcomplicate things for new users, but allows "power users" to easily configure things how they want. Although it is then very important to have it be easily discoverable in the documentation that this is something that they can configure.

joelpalmer:
>it seems like a better user experience to just have <C-r> mapped to mixed reset by default and no mappings for the other reset modes

Thanks for the feedback @l-kershaw! Either way works for me. As you mentioned or having all modes mapped.

> We would then mention in the docs (though I'm not sure where is the best place for this) that this can be remapped to another mode, or removed to implement multiple keymappings in the users preferred way.
> 
> This then doesn't overcomplicate things for new users, but allows "power users" to easily configure things how they want. Although it is then very important to have it be easily discoverable in the documentation that this is something that they can configure.

For now, to be "easily discoverable", I've added documentation for the mappings to the **Git Pickers** → `git_commits` [section of the README](https://github.com/joelpalmer/telescope.nvim/tree/add-reset-to-commits#git-pickers). The actions have documentation and also being mentioned in the README will aid in discoverability. This way, if we go with just  defaulting `<C-r>` to _mixed_, then we can at least leave a mention of the other available actions in that **Description** section of the README.

tami5:
LGTM, Thanks @joelpalmer.

joelpalmer:
Thank you @tami5!

PR #1000, opened 2021-07-15
Custom Path Display: Add ability to have a user defined function to format the path display

Nazeehe:
This PR adds the ability to have the user provide a function to format the Path Display. 

"path_display" can now be assigned a function for custom path display formatting. Example:

path_display = function(opts, filepath) 
    -- custom formatting logic
    return formattedPath
end

### DIFF
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -187,11 +187,13 @@ local telescope_defaults = {
     - "absolute"  display absolute paths
     - "shorten"   only display the first character of each directory in
                   the path
+    - "function"  use a custom function defined in path_display_custom
 
     path_display can also be set to 'hidden' string to hide file names
 
     Default: {}]]
   },
+  path_display_custom = { nil },
 
   borderchars = { { "─", "│", "─", "│", "╭", "╮", "╯", "╰" } },
 
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -187,19 +187,14 @@ local telescope_defaults = {
     - "absolute"  display absolute paths
     - "shorten"   only display the first character of each directory in
                   the path
-    - "function"  use a custom function defined in path_display_custom
 
     path_display can also be set to 'hidden' string to hide file names
 
-    Default: {}]]
-  },
+    path_display can also be set to a function for custom formatting of the
+    path display. Example:
 
-  path_display_custom = { nil, [[
-    Custom function for formatting the Path Display. Function called when path_display is set to 'function'
-
-    Example function that formats the path to "file.ext (full path)":
-
-        path_display_custom = function(opts, filepath)
+        -- Format path as "file.txt (path\to\file\)"
+        path_display = function(opts, filepath)
             iterator = string.gmatch(filepath, "(.-)([^\\/]-%.?([^%.\\/]*))$")
             result = filepath
             for key, value in iterator do
@@ -208,9 +203,9 @@ local telescope_defaults = {
                 end
             end
             return result
-        end
+        end,
 
-    Default: nil ]]
+    Default: {}]]
   },
 
   borderchars = { { "─", "│", "─", "│", "╭", "╮", "╯", "╰" } },
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -194,15 +194,9 @@ local telescope_defaults = {
     path display. Example:
 
         -- Format path as "file.txt (path\to\file\)"
-        path_display = function(opts, filepath)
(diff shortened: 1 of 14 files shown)


lua/telescope/config.lua
```diff
@@ -187,12 +187,32 @@ local telescope_defaults = {
     - "absolute"  display absolute paths
     - "shorten"   only display the first character of each directory in
                   the path
+    - "function"  use a custom function defined in path_display_custom
```
Nazeehe, line 190: Not sure about the naming here. Function? callback?

lua/telescope/config.lua
```diff
@@ -187,12 +187,32 @@ local telescope_defaults = {
...
 
     path_display can also be set to 'hidden' string to hide file names
 
     Default: {}]]
   },
 
+  path_display_custom = { nil, [[
```
Nazeehe, line 197: Naming convention here as well. Open to a better name.

lua/telescope/config.lua
Conni2461: My thought was not to introduce a new config field but make `path_display` either a table or a function. 
So in `utils.transform_path` right at the beginning, we can do 
```lua
if type(path_display) == "function" then
  return path_display(opts, path) 
end
```

@l-kershaw what do you think

Nazeehe: Interesting... from a "make code simple", I like it. Is it confusing for users though? that one field can be either a table or function?

l-kershaw: I agree with @Conni2461. I think it is better to allow `path_display` to be either a table or a function. This means that we keep all of the configuration for displaying paths in one place.

I think the main thing is to ensure that we document it clearly. I think the general layout of the docs for `resolver.resolve_width` would be a good starting point, and then we can adjust from there.

lua/telescope/config.lua
```diff
@@ -187,12 +187,32 @@ local telescope_defaults = {
...
+        path_display_custom = function(opts, filepath)
+            iterator = string.gmatch(filepath, "(.-)([^\\/]-%.?([^%.\\/]*))$")
+            result = filepath
+            for key, value in iterator do
+                if key ~= '' then
+                    result = value .. " (" .. key .. ")"
+                end
+            end
+            return result
+        end
```
l-kershaw, line 211: I think that this function is probably a bit too complicated to serve as an example in the docs.
Is there another function that you could can think of that is simpler but still useful in this context?

lua/telescope/config.lua
Nazeehe: Sounds good. I'll make the change and we can fine tune it. Thanks.

lua/telescope/config.lua
```diff
@@ -190,6 +190,21 @@ local telescope_defaults = {
...
+    path display. Example:
+
+        -- Format path as "file.txt (path\to\file\)"
+        path_display = function(opts, filepath)
+            iterator = string.gmatch(filepath, "(.-)([^\\/]-%.?([^%.\\/]*))$")
+            result = filepath
+            for key, value in iterator do
+                if key ~= '' then
+                    result = value .. " (" .. key .. ")"
+                end
```
Nazeehe, line 203: Let's make sure to review this example code as well. My Lua-fu is not strong yet. Seems that I need to make that check for '' since the match returns an extra empty value for some reason.

lua/telescope/config.lua
Nazeehe: Yeah... I see what you're saying. It can be as simple as:

path_display = function(opts, filepath)
-- custom formatting logic
return formattedPath
end

I'll see if I can think of something a bit more useful but not overly complex. Suggestions always welcome :)

Conni2461: Just saying but we have a `path_tail` util function so we could use that
```lua
  local tail = require("telescope.utils").path_tail(path)
  return string.format("%s (%s)", tail, path)
end
```

Just tried it works like a charm. So we are just working on finalizing the documentation. We have docgen in ci, if ci doesn't wanna work i run it locally and push the docgen changes to this branch when its done :)

Nazeehe: Nice! Much better and works fine on Windows as well. This would be a good example for the docs I would think. Simple enough.

Conni2461 (approved):
LGTM thanks :)

I only changed the formatting (we use 2 spaces) and generated docs

Nazeehe:
Thanks folks! Had fun doing this :)

DanielPower:
This works exactly as I had hoped it would. Thanks so much!

PR #1004, opened 2021-07-16
feat: hidden implies ignoring ignore files

lmarlow:
Include files hidden by `.gitignore` and other ignore files honored by `fd` and `rg`.


tjdevries:
Hmm... perhaps it would be better to add an `ignore` option or something? I don't know if hidden and ignored should be the same

lmarlow:
Closing in favor of #1010

PR #1005, opened 2021-07-16
Fix some typos in readme

Yagua:
fixed some typing errors in the repository readme.


tjdevries:
In general, we use `builtin` because that's how we import and use the files -- I don't thin we'll change it to the other way.

Also, `filesystem` is an acceptable spelling I think, so we'll probably leave that. If you remove those changes, I think I can merge. Thanks!

PR #1006, opened 2021-07-16
fix: error from hopping faster than buffer preview

fdschmidt93:
This got swept under the rug from closing #965 and turning it into a telescope extension :sweat_smile:


tjdevries:
LGTM! Thanks :)

PR #1007, opened 2021-07-17
Add ^ and $ to characters handled by regex escape

dimfeld:
This lets you use live grep to search for strings containing ^ or $. Thanks!


elianiva:
LGTM. Thanks! 

FWIW the `^` doesn't seem to get highlighted while the `$` does, but that has nothing to do with this PR, just letting people know :)

PR #1008, opened 2021-07-17
fix(themes): width and height does not exceed max

dhruvmanila:
For default values, the width and height of the telescope window should not exceed the max available columns (editor width) and lines (editor lines). This is similar to how it is in `get_dropdown`.

I have used `2` to subtract from the max available width/height to give a padding of 1 on each side.

### DIFF
--- a/lua/telescope/themes.lua
+++ b/lua/telescope/themes.lua
@@ -77,12 +77,12 @@ function themes.get_cursor(opts)
     results_title = false,
     layout_strategy = 'cursor',
     layout_config = {
-      width = function(_, _, _)
-        return 80
+      width = function(_, max_columns, _)
+        return math.min(max_columns - 2, 80)
       end,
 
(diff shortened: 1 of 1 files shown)


l-kershaw:
This is currently half handled by `resolve_height` and `resolve_width` within the layout strategies. The problem is that they do not take into account the border size, so it still looks weird.
You can see that this is part handled by checking that putting `height=2000` looks the same as `height=50000`, although it is wrong in both cases 😅 

I'm in the process of writing a PR that will fix this at the moment, and will link this PR when I submit it.

tjdevries:
@l-kershaw will your PR solve this problem then? If so, we should close this. Thanks @dhruvmanila either way :)

dhruvmanila:
No problem! If the linked PR fixes this, then feel free to close this one :)

Issue #751, opened 2021-04-11
resume previous list


sodiumjoe:
**Is your feature request related to a problem? Please describe.**
I'm switching from [denite](https://github.com/Shougo/denite.nvim), and the one thing I'm missing is its `-resume` flag, which simply restores the previous list, in the same filtered state/order. This allows me to, for example, search my project for a token, then go through the list and make modifications. The quickfix workflow from https://github.com/nvim-telescope/telescope.nvim/issues/52 is close, but it doesn't give me the ability to keep track of where in the list my current progress is.

**Describe the solution you'd like**
`:Telescope resume` to simply re-open my last Telescope search, in the same state.
`:Telescope resume cursor_pos=+1` to re-open my last Telescope search with the cursor one line above the previous state.
`:Telescope resume cursor_pos=-1` to re-open my last Telescope search with the cursor one line below the previous state.

**Describe alternatives you've considered**
`:Telescope next/previous` to take the default action on the next/previous item in the previous search, e.g. open the next buffer after `:Telescope buffers`.

**Additional context**
https://github.com/Shougo/denite.nvim/blob/master/doc/denite.txt#L905
https://github.com/Shougo/denite.nvim/blob/master/doc/denite.txt#L684

egidijusz:
This is also a thing I miss coming from doom-emacs. I'm so used to it that it's kind of a blocker to fully embrace Telescope.

[LeaderF](https://github.com/Yggdroot/LeaderF) has `--recall` option. So I'm currently using a mix of both Telescope and LeaderF. :sweat_smile:  Would love to fully switch to Telescope though!

YodaEmbedding:
In a similar vein, `:Telescope builtin mru=true` or `:Telescope builtin open_latest=true` would be nice too.

9ary:
This would also make telescope a solid replacement for [ack.vim](https://github.com/mileszs/ack.vim), if not even more ergonomic. The quickfix window doesn't interact very nicely with splits.

RRethy:
fzf has `<c-p>` and `<c-n>` to cycle through previous searches that would be nice in telescope and would sort of allow resuming a previous list.

Conni2461:
> fzf has `<c-p>` and `<c-n>` to cycle through previous searches that would be nice in telescope and would sort of allow resuming a previous list.

See https://github.com/nvim-telescope/telescope.nvim/pull/521 (the day will come when it will be merged)

[...]

PR #1014, opened 2021-07-17
feat: multi-select style actions

fdschmidt93:
This PR is a (slightly opinionated) attempt to pick up #807 with the design as laid out below. It should now be ready for review :)

The key question is whether we want `action_state.get_selected_entry` (ie what is considered the selected entry) to be overridden if the global `set_entry` is set (see `telescope.state` and `telescope.actions.state`). The idea is that most `telescope.actions` are performed on what is considered "selected". Thus, multi actions should as such not function any differently (to avoid implicit code duplication). 

I'll be on the road week of 19 July and probably have less time to look/work with comments up until after. but I'm very happy to get some thoughts in the mean time.

To illustrate, the PR would allow for the following:

```lua
local actions = require 'telescope.actions'
local action_utils = require 'telescope.actions.utils'
telescope.setup {
  defaults = {
    mappings = {
      i = {
        ["<C-v>"] = action_utils.with_selections(actions.select_vertical),
        -- obviously only for builtin.git_status :)
        ["<C-s>"] = action_utils.with_entries(actions.git_staging_toggle)
      },
    },
  },
```

In particular, this WIP:
* Cleans up `action_utils`; sorry @Conni2461 I didn't grok the `_` prefix in the original PR :sweat_smile: 
* Introduces the global `set_entry` that allows to temporarily override what `actions_state.get_selected_entry` returns
    * The idea is to keep `actions` on `selected_entry` as atomic as possible (that is, as `actions` on `selected_entry` are currently designed, they are meant to be run on a single entry)
* Introduces new `action_utils` for multi-select style actions
    * `action_utils.with_entries(action)`
    * `action_utils.with_selection(action, selection_index): to mimic `context` as I understood it from #807, but might be off
    * `action_utils.with_selections(action)`
    * `action_utils.cycle`: run `action` on an entry-by-entry basis, e.g. `actions.with_selections(actions.cycle({actions.select_vertical, actions.select_horizontal}))` with 2 selections would first select vertically and then horizontally
 
Some things I think we should discuss:  
- Should we change the signature of atomic actions that exclusively rely on `selected_entry` to `()` (to match `action_state.get_selected_entry`) for consistency (and to separate actions that actually require `prompt_bufnr`)
- We should maybe expand `telescope.actions` in more submodules in mid- to long-run, e.g. `telescope.actions.git`?
- TODO? Should we maybe add a `callback` that takes the list of selections/entries to allow for more complex logic? Eg, open 4 files in a vertical split and that reshape them to 4 squares layout

As a non-sensical showcase, here's all my files in `~/.config/nvim` opened with `actions.select_vertical` :laughing: 

https://user-images.githubusercontent.com/39233597/126034846-3f792d59-478f-42a0-97be-7ad4030a4e9c.mp4

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1482,7 +1482,7 @@ Functions to be used to determine the current state of telescope.
 Generally used from within other |telescope.actions|
 
 action_state.get_selected_entry()          *action_state.get_selected_entry()*
-    Get the current entry
+    Get the user-overriden or current entry
 
 
 
@@ -1555,29 +1555,56 @@ action_set.scroll_previewer({prompt_bufnr}, {direction})*action_set.scroll_previ
 ================================================================================
                                                        *telescope.actions.utils*
 
-Utilities to wrap functions around picker selections and entries.
+Utilities to wrap actions and functions around picker selections and entries.
+Generally used with other |telescope.actions| or custom functions that lever
+entries or selections.
+The functions without the `_` prefix are intended to be mapped at setup,
+whereas `functions` with the `_` prefix are intended to be used in custom
+actions. Refer to the respective function documentation for examples.
+The functions that take `action(s)` as input typically are |telescope.actions|
+that lever the `action_state.selected_entry`. In particular,
+`action_utils.with_{entries, selection, selections}` intermittently override
+the `selected_entry` to perform the |telescope.actions| accordingly. More
+generally, `actions` are functions are akin to the below example
+  local action_state = require "telescope.actions.state"
+  function(prompt_bufnr)
+    local entry = action_state.get_selected_entry()
+    -- lever entry for function
+    ...
+  end
 
-Generally used from within other |telescope.actions|
-
-utils.map_entries({prompt_bufnr}, {f})                   *utils.map_entries()*
-    Apply `f` to the entries of the current picker.
+action_utils._map_entries({prompt_bufnr}, {f})   *action_utils._map_entries()*
+    Apply `f` to the entries of the current picker and prompt.
+    - `f` takes (entry, index, row) as arguments.
     - Notes:
       - Mapped entries may include results not visible in the results popup.
       - Indices are 1-indexed, whereas rows are 0-indexed.
     - Warning: `map_entries` has no return value.
-      - The below example showcases how to collect results
-    Usage:
-        local action_state = require "telescope.actions.state"
-        local action_utils = require "telescope.actions.utils"
-        function entry_value_by_row()
-          local prompt_bufnr = vim.api.nvim_get_current_buf()
-          local current_picker = action_state.get_current_picker(prompt_bufnr)
-          local results = {}
-            action_utils.map_entries(prompt_bufnr, function(entry, index, row)
-            results[row] = entry.value
-          end)
-          return results
-        end
+      - The below example showcase how to collect results.
+    Example Usage: collect entries in key-value table
+      local action_state = require "telescope.actions.state"
+      local action_utils = require "telescope.actions.utils"
+      require("telescope").setup {
+        defaults = {
+          mappings = {
+            i = {
+              -- with `action_utils._map_entries`
+              ['<C-e>'] = function(prompt_bufnr)
+                  local results = {}
+                  action_utils._with_entries(prompt_bufnr, function(entry, entry_index)
+                  results[entry_index] = entry
+                  return results
+                end)
+              end,
+              -- with `action_utils.map_entries`
+              ['<C-e>'] = action_utils.with_entries(function(entry, entry_index)
+                  results[entry_index] = entry
+                  return results
+                end)
+            },
+          },
+        },
+      }
 
 
     Parameters: ~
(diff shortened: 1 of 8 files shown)


Conni2461:
First of all thanks for looking into this :) But I have a couple of issue with this approach. I try to make it as clear as possible:
- I don't like the fact that a user has to wrap the action in a `action_utils.with_selections` call. I think it should just work, for multiselection.
  - So assuming we just wrap our `select_*` mappings in that utils call, we no longer can use them without multiselection
  - Also what about bob? Who remapped `select_default` to `<c-space>` he now has to do manual work to get multiselection working.
  - talking about utils they should mainly be for developers and for endusers
  - For novice users its also another hurdle to have that include thats why we also introduced that you can just say `["<C-v>"] = "select_vertical" because we did got a lot of issues that actions is nil.
- Another issue is the global state. We will run into issue if the action needs more than just the entries.
  - For example the git onces access `picker.cwd`. 
  - So if someone writes a own staging but it closes the picker automatically in the end and now this is used in multiselection we no longer have access to `cwd` and it will error.
  - Tj tried to solve this issue in his approach
- Another question i am asking myself is how will this work with the other mapping approaches. Like `string` or table with vim command. Right now you accept a callable (function or metatable which implements `__call`)

Just to be clear. Right now i am only focusing on `multi selection` not on with all entries. Tbh i dont actually know how we can make this work.

To sum it up. I imagined something more "easy" (should just work) for more novice users. They see the commit multiselect actions, pull, and it should just work.

Disclaimer: Thats 100% my personal opinion, i will talk with tj about PR sometime soonish. :)

fdschmidt93:
Thanks for having a look :) I totally understand the comments around "UX" of this PR :sweat_smile: but I think this would be something we could address here with this approach :) Replying to the most important remarks also to add context to your discussions with TJ.

> talking about utils they should mainly be for developers and for endusers

Sorry, recency bias and wasn't sure where to put this :sweat_smile: I'd be ok with moving it elsewhere :)

> I don't like the fact that a user has to wrap the action in a action_utils.with_selections call. I think it should just work, for multiselection [...] Another question i am asking myself is how will this work with the other mapping approaches. Like string or table with vim command. Right now you accept a callable (function or metatable which implements __call)

Fully agree, in the easy cases it should be quite straightforward to implicitly switch to `action_utils.with_selections` if there a multi-selections. I'd have to look at the code (this is a part I'm not too familiar with), but I think this is something we can account for in `telescope.actions.mt` or in `telescope.mappings`? I'd be happy to add to this PR.

At the same time, I'm wondering whether there are cases in which users would want more explicit control (selected entry vs selections vs all entries), for instance for actions they'd write themselves where the above interface would be appreciated.

> Another issue is the global state. We will run into issue if the action needs more than just the entries.

Yeah, that's the truly tricky part I suppose :sweat_smile: though the only built-in `actions` that close the picker (ie de facto kill the global state accessible via `prompt_bufnr` and `picker`) are `select_*` I suppose? We could also cache the picker for the `with_{selections, entries}` loop and similarly override that `action_state.get_current_picker` favors a cached one for the passed `prompt_bufnr`, provided a cached one for that `prompt_bufnr` exists.

> So if someone writes a own staging but it closes the picker automatically in the end and now this is used in multiselection we no longer have access to cwd and it will error.

I'm cautiously optimistic caching the picker should account for such a scenario. :sweat_smile: 

> Disclaimer: Thats 100% my personal opinion, i will talk with tj about PR sometime soonish. :)

Yeah no worries; again, these comments are really fair and it's an important discussion to have :)

tjdevries:
OK, so conni and I talked for awhile to try and figure out our thoughts about this.

A lot of the things that I am thinking are missing from this design that are present in my original design sort of boil down to this:

In my design, we extend the current signature of actions from `function(prompt_bufnr)` to `function(prompt_bufnr, context)`, which allows all actions to be called from other lua code with bonus information. For example, one of the keys in `context` will be `entry`. When `entry` is passed, that instructs the `action` to NOT ask for the currently selected entry, but instead to just operate on whatever was selected. This makes it much simpler to do something like:

1. Gather the results you'd like to act on
2. Store the results in a list
3. Iterate over the list and perform action on them
4. Release list contents / memory

This means we could make something like `actions.multi_select_default` and `actions.multi_select_split`, etc. that will wrap the original values. We can also make, as we've done before, `actions.multi_select_smart_default` or whatever the names are which will choose between performing on multi-selection entries or single entry depending on if the user has selected anything.

This makes it so that users can still do something like `{ ["<c-v>"] = "multi_select_smart_default" }` and have it work. There are no new globals or special state and no hacks we have to add around keeping that state hidden from end users.

We can also add `"select_all_default"` or similar to add some of the concepts you are playing around here as well.

Does that make sense? (sorry, I wrote this quickly)

fdschmidt93:
Thanks for the guidance, @tjdevries and @Conni2461 :) happy have at this in a revised PR over the next two to three weeks (as I'll continue to travel). I think I better understand the requirements and know how to avoid some of the pitfalls now.

Closing the PR as it will be superseded.

Conni2461:
I am thinking a lot about this today and i have a couple more pointers.
- Keeping the last picker around in the global state might be super useful in the future. For example for resume #751. Tj did this in this attempt, you should give it a read.
  - It might not be that easy thought for pickers with more entries > 1M. Maybe we should then introduce a upper or so idk
  - For `new_job` or `one_shot_job` we could probably clean up all entries because we can easily reproduce them. (just some thoughts not sure tho). 
- About the utils you introduced. Maybe it would be better to make use of [plenary.iterators](https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/iterators.lua) ([spec](https://github.com/nvim-lua/plenary.nvim/blob/master/tests/plenary/iterators_spec.lua)). Its based on [luafun](https://github.com/luafun/luafun)
  - it should work out of the box with `current_picker:get_multi_selection()`
  - but I dont think it works with `manager.iter()` because its a custom linked list with a iter function but maybe we should make that work.
  - It would give use a bunch of cool things. So for example if `manager` would return a plenary iterator we could end up with something like this: `manager:piter():map(function(e) return e.ordinal end):tolist()`.

fdschmidt93:
> In my design, we extend the current signature of actions from function(prompt_bufnr) to function(prompt_bufnr, context), which allows all actions to be called from other lua code with bonus information. For example, one of the keys in context will be entry. When entry is passed, that instructs the action to NOT ask for the currently selected entry, but instead to just operate on whatever was selected. This makes it much simpler to do something like:

> 1. Gather the results you'd like to act on
> 2. Store the results in a list
> 3. Iterate over the list and perform action on them
> 4. Release list contents / memory

I've thought about this a bit more and I think the best thing is to merge our two approaches. What I did is very close to this but not as extensible; so I'll set up an `actions.iterators` to set up `multi_select_default` etc with `with_selections` etc. and leave what I've provided as a lower level api for anyone who needs it.

In addition, I'll try to address your concerns @Conni2461 about handing some of the more complicated custom actions and try to incorporate your ideas for much better UX :)

Issue #997, opened 2021-07-15
File mask in live_grep


igor-semenov:
**Is your feature request related to a problem? Please describe.**
I'd like to use live_grep to search through specific file types in project, for example, C++ source files. But usual live_grep searches in all of files.

**Describe the solution you'd like**
Maybe a parameter in live_grep, specifying file masks, for example: *.h,*.cpp

**Describe alternatives you've considered**
I'm using ag for now.

Nazeehe:
I like this idea. I messed around the code a bit and have this working:

Telescope live_grep grep_args=-tcs

this passes -tcs to Rg and scopes the search down to C# files only. For Ag, that would be:

Telescope live_grep grep_args=--csharp

Thoughts on this pattern?

igor-semenov:
I tried your solution:
`:Telescope live_grep grep_args=-tcpp`
but unfortunately it doesn't work for me, live_grep still searches in all of files, not in C++ sources.

Nazeehe:
Hey @igor-semenov, I have changed the PR (which you would need to apply to your branch as it's not checked in yet). When it's checked in (waiting on approval), you would add something like this to your telescope config:

``` lua
require("telescope").setup {
  defaults = { 
    -- your defaults for telescope
  },
  pickers = {
    live_grep = {
      -- your defaults for live_grep
        additional_args = function(opts)
            if opts.search_all == true then
                return {}
            end

            local args_for_ext = {
                ["cs"]  = "-tcs",
                ["cpp"] = "-tcpp",
                ["c"]   = "-tcpp",
                ["h"]   = "-tcpp",
                ["rs"]  = "-trust",
                ["lua"] = "-tlua"
            }
            return { args_for_ext[vim.bo.filetype] }
        end
    }
  }
}
```
This would make it so that when you run ":Telescope live_grep", live_grep code will call this function. This function is meant to add extra args to be passed to Rg. This particular function here will read the filetype of the buffer you're on now and will use the lookup table to match the args to the type. So in your case, if the buffer you're on is .cpp, it will add -tcpp to Rg. If you switch to a buffer with .rs, it will automatically add -trust. That's just my implementation of the feature using this function, you can do whatever you want of course.

igor-semenov:
@Nazeehe thank you for the explanation!

Nazeehe:
> @Nazeehe thank you for the explanation!

[...]

PR #1017, opened 2021-07-17
Auto file mask in live grep via additional args function in live_grep

Nazeehe:
Add the ability to have runtime additional args to live_grep via a user configurable function. This allows the user to add additional arguments to the grepper program based on custom logic. Example logic that filters search results automatically based on buffer filetype:

```lua
require("telescope").setup {
  defaults = { 
    -- your defaults for telescope
  },
  pickers = {
    live_grep = {
      -- your defaults for live_grep
        additional_args = function(opts)
            if opts.search_all == true then
                return {}
            end

            local args_for_ext = {
                ["cs"]  = "-tcs",
                ["cpp"] = "-tcpp",
                ["c"]   = "-tcpp",
                ["h"]   = "-tcpp",
                ["rs"]  = "-trust",
                ["lua"] = "-tlua"
            }
            return { args_for_ext[vim.bo.filetype] }
        end
    }
  }
}
```

The custom logic allows for reverting back to searching all files by passing search_all = true:
:lua require('telescope.builtin').live_grep({ search_all = true })

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -70,7 +70,7 @@ files.live_grep = function(opts)
   end
 
   local additional_args = {}
-  if opts.additional_args ~= nil then
+  if opts.additional_args ~= nil and type(opts.additional_args) == "function" then
       local context = {
           filetype = vim.bo.filetype,
           filename = vim.fn.expand(vim.api.nvim_buf_get_name(vim.api.nvim_get_current_buf()))
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -79,11 +79,11 @@ files.live_grep = function(opts)
       end
 
       local extraArgs = {}
-      local smart_mask = utils.get_default(opts.smart_mask, require('telescope.config').values.smart_mask)
+      local smart_mask = utils.get_default(opts.smart_mask, conf.smart_mask)
       local filetype = opts.current_filetype or ''
 
       if (smart_mask == true and filetype ~= '') then
-          local smart_mask_args = utils.get_default(opts.smart_mask_args, require('telescope.config').values.smart_mask_args)
+          local smart_mask_args = utils.get_default(opts.smart_mask_args, conf.smart_mask_args)
           table.insert(extraArgs, smart_mask_args[filetype])
       end
 
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -43,7 +43,6 @@ files.live_grep = function(opts)
   local vimgrep_arguments = opts.vimgrep_arguments or conf.vimgrep_arguments
   local search_dirs = opts.search_dirs
   local grep_open_files = opts.grep_open_files
-  local filetype = vim.bo.filetype
   opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or vim.loop.cwd()
 
   local filelist = {}
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -36,6 +36,7 @@ files.live_grep = function(opts)
   local vimgrep_arguments = opts.vimgrep_arguments or conf.vimgrep_arguments
   local search_dirs = opts.search_dirs
   local grep_open_files = opts.grep_open_files
+  local filetype = vim.bo.filetype
   opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or vim.loop.cwd()
 
   local filelist = {}
@@ -80,7 +81,6 @@ files.live_grep = function(opts)
 
       local extraArgs = {}
       local smart_mask = utils.get_default(opts.smart_mask, conf.smart_mask)
-      local filetype = opts.current_filetype or ''
 
       if (smart_mask == true and filetype ~= '') then
           local smart_mask_args = utils.get_default(opts.smart_mask_args, conf.smart_mask_args)
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -81,8 +81,9 @@ files.live_grep = function(opts)
       end
 
       local extraArgs = {}
-      local smart_mask = utils.get_default(opts.smart_mask, conf.smart_mask)
-      table.insert(extraArgs, smart_mask[filetype])
+      if (opts.smart_mask ~= nil) then
+          table.insert(extraArgs, opts.smart_mask[filetype])
+      end
 
       return flatten { vimgrep_arguments, extraArgs, prompt, search_list }
     end,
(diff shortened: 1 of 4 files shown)


Nazeehe:
My take on addressing feature request: https://github.com/nvim-telescope/telescope.nvim/issues/997

lua/telescope/builtin/init.lua
```diff
@@ -402,6 +402,8 @@ local apply_config = function(mod)
         end
       end
 
+      opts.current_filetype = vim.bo.filetype
```
Nazeehe, line 405: is there a better way to do this?

Conni2461: No its the correct way to do this, but why here? why at this place and not the beginning of `live_grep` this will now apply to all pickers and only `live_grep` needs it

lua/telescope/config.lua
```diff
@@ -282,6 +282,30 @@ local telescope_defaults = {
     { "rg", "--color=never", "--no-heading", "--with-filename", "--line-number", "--column", "--smart-case" },
   },
 
+  smart_mask = { false },
+
+  smart_mask_args = {
```
Conni2461, line 287: Both should not be config values. config is meant for options across more than one picker.

lua/telescope/builtin/files.lua
```diff
@@ -78,7 +78,16 @@ files.live_grep = function(opts)
         search_list = filelist
       end
 
-      return flatten { vimgrep_arguments, prompt, search_list }
+      local extraArgs = {}
+      local smart_mask = utils.get_default(opts.smart_mask, conf.smart_mask)
+      local filetype = opts.current_filetype or ''
```
Conni2461, line 83: yep just move local filetype = vim.bo.filetype outside of inner function `live_grepper`

lua/telescope/builtin/files.lua
```diff
@@ -78,7 +78,16 @@ files.live_grep = function(opts)
...
+
+      if (smart_mask == true and filetype ~= '') then
```
Conni2461, line 85: Only one option, so if we have a smart_mask table then its enabled, no table or table empty then disabled.

lua/telescope/builtin/init.lua
Nazeehe: I'll try again. I *think* the first time I used it, I did have it in there and it came out blank. I guessed it must be b/c the floating window is up and the buffer is now telescope. I was using vm.fn.expand("%:e") to get it though, maybe that was not the right way? I'll try and get back to you.

lua/telescope/builtin/files.lua
Nazeehe: Ah, but what happens when you _don't_ want smart mask to be applied? Example: In general, I want it to be on most of the time, but I can see some instances where I actually want it off right now since I do want to search all files. My thought process was passing the smart_mask as a param would allow me to map two bindings, one with mask on, one with off. 

Thoughts?

lua/telescope/builtin/files.lua
Nazeehe: Ah, that was it. _Outside_ of inner function did the trick. Thanks!

lua/telescope/builtin/init.lua
Nazeehe: Nevermind. Got it now thanks to your other comment.

lua/telescope/config.lua
Nazeehe: Mind elaborating on this? This is excellent for me to understand. If it's not a config value, how does the user pass the table to the code?

Conni2461: So we have this system.
```lua
require("telescope").setup {
  defaults = { 
    -- your defaults for telescope
  },
  pickers = {
    live_grep = {
      -- your defaults for live_grep
      smart_mask = {
        ["lua"] = "-tlua",
        ["cs"]  = "-tcs",
        ["cpp"] = "-tcpp",
        ["h"]   = "-tcpp"
      }
    }
  }
}
```
So if you now call `require('telescope.builtin').livegrep()` it will applies these `live_grep` defaults. If you then do `require('telescope.builtin').livegrep({ smart_mask = {} })` it would prefer the opts again.

So for your code. You only need to check for `opts.smart_mask`

Conni2461: The one thing i dont know how to handle yet would be the documentation. The documentation needs to happen in `builtin/init.lua` again but our docgen system can only understand single line annotations.

Nazeehe: Awesome, this was the missing piece for me. Much better now. Thanks!

Nazeehe: PR updated. Anything I can do to help with docs?

Nazeehe:
@Conni2461 thanks for all the feedback on this PR. I just updated it to latest master (there was a merge conflict). I am getting a stylua error at the moment that is unclear to me. Other than that, anything else we are waiting on for this? -Thanks!

tjdevries:
Just to note what we talked about on stream:

I think it might be interesting to instead (or maybe in addition) add a function here, instead of a table, that looks like:

```
-- context has keys like, "filetype", "bufnr", "bufname" ? idk exactly
function(context)
  return { "additional", "args", "here" }
end
```

And the option can be called "file_args"? Idk, that's not  the right name.

Nazeehe:
> Just to note what we talked about on stream:
> 
> I think it might be interesting to instead (or maybe in addition) add a function here, instead of a table, that looks like:
> 
> ```
> -- context has keys like, "filetype", "bufnr", "bufname" ? idk exactly
> function(context)
>   return { "additional", "args", "here" }
> end
> ```
> 
> And the option can be called "file_args"? Idk, that's not the right name.

Yeah, I like that idea. We would need clear docs around the usage pattern. This PR started in response to a telescope user asking for a way to grep on particular file extensions. Very valid and useful scenario. The change as is now is pretty clear in that respect. Changing it to a function would be great from a flexibility point of view where their request is possible to implement but at the cost of discoverability.

Nazeehe:
Ok, implemented feedback. Using a custom function now which is more flexible.

Conni2461:
Sorry for the long wait, tj granted me one week off from telescope and that was last week :laughing: (one week per year. not a good employer i know) :laughing: 

PR looks good to me. Does it make sense to also do the same for `grep_string`? If not i'll generate docs and merge :)

Nazeehe:
> Sorry for the long wait, tj granted me one week off from telescope and that was last week 😆 (one week per year. not a good employer i know) 😆
> 
> PR looks good to me. Does it make sense to also do the same for `grep_string`? If not i'll generate docs and merge :)

Welcome back! Take what you can get! :)

Good idea adding it to grep_string as well. Added. Things are good from my side now. Feel free to gen docs/etc when you have a chance. 

Thanks @Conni2461!

Conni2461:
LGTM Thanks :)

PR #1019, opened 2021-07-17
Set a filetype for the results buffer

### LINKED ISSUE #991: How do you disable folds inside pickers?
I have the following fold settings.

```
set foldclose=all
set foldopen=all
```

When I open pickers, navigating the result list in normal mode with j/k results in folds. Is this expected behavior, and is there a way to turn it off?

I've attached a video showing the issue.

https://user-images.githubusercontent.com/273399/125663225-fdbd8f06-9aef-40eb-9451-3938aaae03ff.mp4

JRasmusBm:
**Why** is the change needed?

So that it can be targeted with specific settings. These could include disabling
plugins for the buffer (as in #840) or disabling folding (as per #991).

**How** is the need addressed?

- Add a filetype to the buffer after it has been created.

Closes #991

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -471,6 +471,7 @@ function Picker:find()
 
   -- Do filetype last, so that users can register at the last second.
   pcall(a.nvim_buf_set_option, prompt_bufnr, 'filetype', 'TelescopePrompt')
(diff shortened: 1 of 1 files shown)


JRasmusBm:
@Conni2461 Is this sort of what you had in mind? Thank you for being so inviting to contribute! :smile:

Conni2461:
Thanks :) and thanks for the kind words

PR #1023, opened 2021-07-18
feat: check type of `path_display` and warn user if wrong

l-kershaw:
Saw a reddit post where people got confused when they gave a string for `path_display`.
Thought it would be a good idea to check for people.

---

Had to have `hidden` as a separate case, as it allows `path_display` to be the string "hidden".


Conni2461 (approved):
LGTM

PR #1026, opened 2021-07-18
fix: set a filename for entries in `file_browser`

l-kershaw:
This prevents an issue caused by [this](https://github.com/nvim-telescope/telescope.nvim/blob/8c3f2b630be0241fe10709e61ee9dab473518f32/lua/telescope/actions/set.lua#L116) on Windows, as filepaths will often contain ":", e.g. "C:\foo\bar". The result was that in the `file_browser` any time you open a file on the C drive, it will instead open a file called "C" 🤣.

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -394,9 +394,11 @@ builtin.file_browser({opts})                          *builtin.file_browser()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {cwd}      (string)  directory path to browse (default is cwd)
-        {depth}    (number)  file tree depth to display (default is 1)
-        {dir_icon} (string)  change the icon for a directory. default: 
+        {cwd}      (string)   directory path to browse (default is cwd)
+        {depth}    (number)   file tree depth to display (default is 1)
+        {dir_icon} (string)   change the icon for a directory. default: 
+        {hidden}   (boolean)  determines whether to show hidden files or not
+                              (default is false)
 
 
 builtin.treesitter()                                    *builtin.treesitter()*
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -94,7 +94,7 @@ action_set.edit = function(prompt_bufnr, command)
 
   local filename, row, col
 
-  if entry.filename then
+  if entry.path or entry.filename then
     filename = entry.path or entry.filename
 
     -- TODO: Check for off-by-one
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -297,9 +297,6 @@ files.file_browser = function(opts)
           if is_dir(t.value) then retpath = retpath .. os_sep end
           return retpath
         end
-        if k == "filename" then
-          return Path:new(t[1]):absolute()
-        end
 
         return rawget(t, rawget({ value = 1 }, k))
       end
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -304,6 +304,7 @@ files.file_browser = function(opts)
       return function(line)
         local tbl = {line}
         tbl.ordinal = Path:new(line):make_relative(opts.cwd)
+        tbl.filename = Path:new(line):absolute()
         return setmetatable(tbl, mt)
       end
     end
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -297,6 +297,9 @@ files.file_browser = function(opts)
           if is_dir(t.value) then retpath = retpath .. os_sep end
           return retpath
         end
+        if k == "filename" then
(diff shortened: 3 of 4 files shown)


Conni2461:
Does this work instead? Because i kinda wanna use metatable `__index` for this. There is no reason to run this for every entry. We should only do it if we need it.
 
```diff
diff --git a/lua/telescope/builtin/files.lua b/lua/telescope/builtin/files.lua
index 2156223..fe5e4a3 100644
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -290,7 +290,7 @@ files.file_browser = function(opts)
         if raw then return raw end
 
         if k == "path" then
-          local retpath = Path:new({t.cwd, t.value}):absolute()
+          local retpath = Path:new(t.value):absolute()
           if not vim.loop.fs_access(retpath, "R", nil) then
             retpath = t.value
           end
```

l-kershaw:
That's a fair point about not having to run `absolute` every time. I'll have a look at the metatable magic tomorrow.

l-kershaw:
@Conni2461 I've implemented some metatable magic so we only calculate `filename` when needed. I didn't really understand the `k == "path"` case, so just implemented the filename separately. I also think we need to set `filename` specifically because of [this](https://github.com/nvim-telescope/telescope.nvim/blob/8c3f2b630be0241fe10709e61ee9dab473518f32/lua/telescope/actions/set.lua#L97) check.

It seems to work in my tests, but I'm not really familiar with metatable magic, so would be good if you could make sure I haven't done anything stupid 😅

Conni2461:
I don't think we need to set the `filename` like that because the path part does exactly the same, its only that weird check in `action_set`. Does this patch work?
```diff
diff --git a/lua/telescope/builtin/files.lua b/lua/telescope/builtin/files.lua
index 2156223..f23e713 100644
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -290,7 +290,7 @@ files.file_browser = function(opts)
         if raw then return raw end
 
         if k == "path" then
-          local retpath = Path:new({t.cwd, t.value}):absolute()
+          local retpath = Path:new({t.value}):absolute()
           if not vim.loop.fs_access(retpath, "R", nil) then
             retpath = t.value
           end
@@ -298,7 +298,7 @@ files.file_browser = function(opts)
           return retpath
         end
 
-        return rawget(t, rawget({ value = 1 }, k))
+        return rawget(t, rawget({ value = 1, filename = 1 }, k))
       end
 
       return function(line)
```

alma3lol (approved):

alma3lol (commented):
Thanks. It works. TJ is in a vacation (status says), so I'll just keep using your fixed version until it's merged. Much appreciated.

l-kershaw:
@Conni2461 I decided it was easier to avoid any metatable magic, and simply check if either `path` or `filename` is provided, and prioritise `path` if both are present.

I also added some extra docs, as mentioned [here](https://github.com/nvim-telescope/telescope.nvim/issues/1050#issuecomment-887006766).

---

@mroavi could you check if this fixes your issue in #1053? I managed to find a way to reproduce on my machine, and this fixes it. If this does fix it for you, then I will link your issue so it is closed when this is merged.

mroavi:
It's fixed! (as far as I can tell) However, I am still having issues with `git_commits()` and `git_bcommits`. I have the feeling that the issue is related since I'm configuring these in a very similar way (see below). The issue is that occasionally the picker list appears empty when there should definitely be commit entries in it.

```lua
function M.git_commits()
	require("telescope.builtin").git_commits{
		cwd = vim.fn.expand("%:p:h"),
		layout_config = {
			prompt_position = "top",
		},
		sorting_strategy = "ascending",
	}
end

function M.git_bcommits()
	require("telescope.builtin").git_bcommits{
		cwd = vim.fn.expand("%:p:h"),
		layout_config = {
			prompt_position = "top",
		},
		sorting_strategy = "ascending",
	}
end


utils.remap("n", "<Leader>co", "<Cmd>lua require('mrv.plugins.telescope').git_commits()<CR>")
utils.remap("n", "<Leader>bc", "<Cmd>lua require('mrv.plugins.telescope').git_bcommits()<CR>")
```

l-kershaw:
@mroavi I don't think that this is a related issue, as it does not relate to the opening of the file on selection.

I can't see any other similar issues about `bcommits`, but would be good if you could double check.
If you can't find a similar issue, then please open another issue, and give a process to reproduce if you can come up with one.

mroavi:
If it's not a related issue then I would say that the original problem is fixed. I tried `git_status()` in several of my repos with modified files, and it worked fine in all of them. I will check for issues regarding the `git_bcommits()` problem like you suggest. Thanks a lot!

Conni2461 (approved):
LGTM i like that solution more :)

PR #1027, opened 2021-07-18
fix: check if tabline is present when calculating certain layouts

### LINKED ISSUE #995: Vertical padding does not take tabline into account
Hi there! Is there any setting to have both ( top and bottom space around telescope ) have equal padding? the bottom padding looks a bit more than the top one.

image

l-kershaw:
- Don't include tabline in the height of the window when present.
  - Only applies to `horizontal`, `vertical` and `center` strategies (also inherited by `flex`)
- ~~Tweaks height calculation for `center` strategy~~

Closes #995

### DIFF
--- a/lua/telescope/pickers/layout_strategies.lua
+++ b/lua/telescope/pickers/layout_strategies.lua
@@ -323,7 +323,7 @@ layout_strategies.horizontal = make_documented_layout('horizontal', vim.tbl_exte
     preview = self.previewer and preview.width > 0 and preview,
     results = results,
     prompt = prompt
-}
+  }
 end)
 
 --- Centered layout with a combined block of the prompt
@@ -600,7 +600,7 @@ layout_strategies.vertical = make_documented_layout("vertical", vim.tbl_extend("
   if tbln then
     prompt.line = prompt.line + 1
     results.line = results.line + 1
-    preview.line = preview.line + 1
+    preview.line = preview.line ~= nil and preview.line + 1
   end
 
   return {
--- a/lua/telescope/pickers/layout_strategies.lua
+++ b/lua/telescope/pickers/layout_strategies.lua
@@ -46,6 +46,15 @@ local get_border_size = function(opts)
   return 1
 end
 
+local calc_tabline = function(max_lines)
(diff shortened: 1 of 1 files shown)


l-kershaw:
I've removed the tweak for the `center` strategy.
I think I want the tweak to be more precise, and also apply it to some other layouts, so will put it in a separate PR when I get a chance.

l-kershaw:
I've found a problem with the vertical strategy, and some other refactoring to do.
Will hopefully fix today

Conni2461:
Are you still working on this?
My comment would be. "I think i see a reoccurring pattern, maybe a short local `function(max_lines) -> max_lines, tbln`"

Otherwise LGTM (you dont need to do it. I shouldn't be that nitpicky)

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -563,6 +597,12 @@ layout_strategies.vertical = make_documented_layout("vertical", vim.tbl_extend("
     prompt.line = results.line + results.height + 2
   end
 
+  if tbln then
+    prompt.line = prompt.line + 1
+    results.line = results.line + 1
+    preview.line = preview.line ~= nil and preview.line + 1
```
Conni2461, line 603: Hmm why is preview.line only not a thing for vertical. It seems like all other configs set it even if previewer is disabled. (Just asking stupid questions, you know way more about layout)

l-kershaw: Not a stupid question, it was bugging me too. Just couldn't figure out a way of refactoring it that didn't seem arbitrary.

Latest commit fixes this (and also allows the `mirror` option to work without a previewer, which will be very useful when we manage to get previewer toggle working).

l-kershaw:
@Conni2461 I have factored out the function you suggested (good spot!), also considered factoring out the "shift down a line if have a tabline", but thought it would be more trouble than it was worth.

I'm happy with this now, but if you could have a last look over before merging, that would be great 😊

Conni2461 (approved):
LGTM thanks :)

you can ship it if its done. I felt the same about the second refactoring. I also thought it would complicate things.

PR #1028, opened 2021-07-18
feat: buffers opts.sort_mru

### LINKED ISSUE #791: Telescope buffers should allow ordering by most recent buffer
Coming from Unite.vim, buffers should allow sorting by most recent instead of buffer number. 

Either a configuration or a option like `require('telescope.builtin').buffers(most_recent_order=1)`

Conni2461:
Close #791

Finish it tomorrow. Its sunday and i wanna eat my well deserved black pepper ramen.

I am missing docs

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -659,9 +659,12 @@ builtin.buffers({opts})                                    *builtin.buffers()*
         {only_cwd}              (boolean)  if true, only show buffers in the
                                            current working directory (default
                                            false)
-        {sort_lastused}         (boolean)  if true, sort the shown buffers so
-                                           that the last used one is selected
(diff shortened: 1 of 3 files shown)


caojoshua:
@Conni2461 Just seeing this merged, and I'm enjoying the change. Do you think it makes sense to remove `sort_lastused` buffer picker option? `sort_mru` just seems to be a better version.

Conni2461:
@caojoshua Why would you want to remove it? It still puts your cursor at previous used buffer and not at current buffer. I dont really mind it and actually use both. Also maybe that option is enough for some people

caojoshua:
OK, its not a big deal. I thought that `mru` was a superset of `lastused`, but did not realize the latter was responsible for cursor position.

PR #1034, opened 2021-07-19
fix: unknown filetype bug

### LINKED ISSUE #1033: bug: treesitter additional_vim_syntax
If `additional_vim_regex_highlighting` is enabled in [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter#modules) we are getting the error described here: https://github.com/nvim-telescope/telescope.nvim/pull/993 (unknown filetype). 

We do not set filetype to improve performance of buffer previewer (we dont want autocmds for buffer previewers) and i would like to keep this behavior but we need to find a way to deal with this. For now i suggest disabling this option but I need to look at `nvim-treesitter` source code first.

CC @tdfirth

Conni2461:
I dont like this fix that much tbh

Close #1033

### DIFF
--- a/lua/telescope/previewers/utils.lua
+++ b/lua/telescope/previewers/utils.lua
@@ -1,7 +1,7 @@
 local context_manager = require('plenary.context_manager')
 
 local has_ts, _ = pcall(require, 'nvim-treesitter')
-local _, ts_highlight = pcall(require, 'nvim-treesitter.highlight')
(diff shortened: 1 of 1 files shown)


Conni2461:
Not the best fix but that should give us more consistency (a little bit longer).
Also if treesitter is now disabled for a language (in `treesitter.setup`) its also disabled here.

PR #1037, opened 2021-07-21
fix: make path for send to qflist correct

seblj:
When using git_files and sending the to the qflist, it's not possible to cycle through them using `:cnext`. This PR should fix the path issues by making the path absolute with plenary. It looks like the quickfix-list handles relative path so it doesn't show the entire path if it doesn't need to


seblj:
Here is a video of how qflist handles the path

https://user-images.githubusercontent.com/5160701/126824026-738847a0-9574-4eff-bb60-62eadb046757.mov

tjdevries:
Thanks! :)

Issue #549, opened 2021-02-24
Error executing luv callback: attempt to index field '_tracked_node' (a nil va lue)                                                                                           


kevinpita:
### Description

I can't use telescope, when opening find files (just an example) the window appears but it doesn't work and I get tons of errors

**Expected Behavior**

Telescope to work

**Actual Behavior**

```
Error executing luv callback:
...ugged/telescope.nvim/lua/telescope/algos/linked_list.lua:77: attempt to index field '_tracked_node' (a nil va
lue)                                                                                                            
Press ENTER or type command to continue                                                                         
Error executing luv callback:
cannot resume dead coroutine
Press ENTER or type command to continue
Error executing vim.schedule lua callback: ...lugged/telescope.nvim/lua/telescope/pickers/scroller.lua:59: Must 
never return a negative row: { result = -2, args = { -1 1 -2 } }
```

after that an infinite loop of
```
Press ENTER or type command to continue
Error executing luv callback:
cannot resume dead coroutine
```

### Details

<details><summary>Reproduce</summary>

1. nvim -nu test.vim
2. :Telescope find_files
</details>

<details><summary>Environment</summary>

- nvim --version output:  NVIM v0.5.0-dev+1108-g0450e155d
- Operating system:  Linux ubuntudesktop 5.8.0-43-generic - Ubuntu 20.04.1
- Telescope commit: 1c5e42a

</details>

<details><summary>Configuration</summary>
<p>

```viml



call plug#begin('~/.vim/plugged')
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-telescope/telescope.nvim'
call plug#end()

nnoremap <leader>ff <cmd>Telescope find_files<cr>
nnoremap <leader>fg <cmd>Telescope live_grep<cr>
nnoremap <leader>fb <cmd>Telescope buffers<cr>
nnoremap <leader>fh <cmd>Telescope help_tags<cr>


lua << EOF
require('telescope').setup{
  defaults = {
    vimgrep_arguments = {
      'rg',
      '--color=never',
      '--no-heading',
      '--with-filename',
      '--line-number',
      '--column',
      '--smart-case'
    },
    prompt_position = "bottom",
    prompt_prefix = ">",
    initial_mode = "insert",
    selection_strategy = "reset",
    sorting_strategy = "descending",
    layout_strategy = "horizontal",
    layout_defaults = {
      -- TODO add builtin options.
[...]

PR #1039, opened 2021-07-22
feat: `layout_strategies` refactor

l-kershaw:
This PR has a few related aims:

- [x] handle layout configuration that is too big/small more gracefully, taking into account the presence of borders
- [x] standardise the way in which the `layout_strategies` are written (e.g. variable names, code structure)
- [x] (where possible) refactor so that `height` corresponds to the total height of the picker, similarly for `width`
  - I think this will probably warrant some discussion as in a few cases (e.g. `center` and `bottom_pane`) the `height` currently only corresponds to the number of results displayed.
  - My reasoning is that consistency is better, particularly that we now allow things like `{padding=...}`, which doesn't make sense when the height gives the number of results.
  - @tjdevries I _think_ this was something that we didn't 100% align on when discussed previously, but can't remember/find what you actually said

---

For the third item, I think all of the pickers (except `center`) should have the height correspond to the total height of the picker (including borders).
For `center`, since the preview just takes up the remaining space above the central block, I think height should correspond to the total height of the central block (including borders). This has the benefit of actually matching up with the height when the previewer is disabled.

---

Related issues/comments/PRs:

Resolves #1008
Reduces problems such as discussed [here](https://github.com/nvim-telescope/telescope.nvim/issues/549#issuecomment-885059513)


l-kershaw:
I think I've implemented everything I wanted to now.

Any feedback would be much appreciated 🙂

tjdevries:
Awesome, I will try and look this week 😁 thanks for your great work!

l-kershaw:
@tjdevries have you had a chance to look at this?

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -468,46 +488,54 @@ layout_strategies.cursor = make_documented_layout(
     local width_opt = layout_config.width
     local width = resolve.resolve_width(width_opt)(self, max_columns, max_lines)
 
-    local max_width = (width > max_columns and max_columns or width)
-
     local bs = get_border_size(self)
 
+    -- Cap over/undersized height
+    height = math.min(height, max_lines)
+    height = math.max(height, 2 + (3 * bs))
```
tjdevries, line 495:
```
height = bounded(height, max_lines, 2 + (3 *bs))

height = border_bounded(height, max_lines, 2)
-- height = math.min(height, max_lines)
--   height = math.max(height, offset + (3 * bs))
```

l-kershaw: Problem is the multiplier is different in different layouts

l-kershaw: But could factor into a function to make clearer what's going on

tjdevries:
LGTM after rebasing and making sure things still work (and doing that little local function we talked about)

l-kershaw:
A few things I still need to fix atm, will have another look tomorrow.

l-kershaw:
I can't figure out how I have broken the tests in `find_files_spec.lua`.
But being honest, I don't understand how those tests work anyway. I have tried a few things to adjust for the changes in height calculation, but none of them seem to work.

@tjdevries, I think you wrote those tests, can you figure out what is going wrong here?

l-kershaw:
At @Conni2461's suggestion I have tried switching to a different plenary branch in the hopes that this fixes the ci
It doesn't seem to, but I will leave it as is for the moment to see if we can diagnose the problem easier.

We just need to make sure that this branch is **NOT MERGED** until the ci has been switched back.

tjdevries:
Can you switch everything back to how it should be and then I'll see if I can get tests to pass again? There is kind of a lot happening to make the CI run at all :laughing: If I don't respond in here, ping me on discord or something so that I take a look. I want to merge this cause it's really good.

l-kershaw:
@tjdevries I have switched it back to the main plenary branch

l-kershaw:
It looks like the 0.5.0 tests now work, but the nightly release is not up, so can't be used in those ci tests.
I'll re-run those tests when the nightly release is back available (hopefully tomorrow), if it all works I'll merge then.

Thanks @tjdevries for spotting the mistake 🙂

Conni2461:
:tada: :tada: :tada: 

Good Job :)

clason:
Quick question: after this PR, the `cursor` theme went from 5 lines of results (which is OK) to 2 lines (which is a bit meager...)

I don't quite see why, nor how to change this back?

Conni2461:
We need to change it here: https://github.com/nvim-telescope/telescope.nvim/blob/37c9ab6688eb2cf46a22c70c8d46ca9d4e147e85/lua/telescope/themes.lua#L82-L84

Changing it to 9 will show 5 items again.

clason:
Yep -- but shouldn't I be able to override it in my setup?

l-kershaw:
You should still be able to override it, for example:
`:Telescope find_files theme=get_cursor layout_config={height=9}`
works as expected for me.

clason:
Yes, that works -- I was trying
```
defaults = {
    layout_config = {
      cursor = {
        height =10,
      },
    },
```

l-kershaw:
Ah, I think that is because themes override your defaults, but not the options you pass in.

Conni2461:
I think we should still go back to showing 5 elements on default.

PR #1044, opened 2021-07-24
refactor: remove unused parameter

elianiva:


l-kershaw:
LGTM, thanks!
Only thing to add is that there is another instance of using `prompt_bufnr` in the `WIP` file for the `reloader`.
I don't know if anyone is using/working on it, but would be good to sort everything at once 🙂

PR #1045, opened 2021-07-24
refactor: make luacheckrc stricter

MunifTanjim:
Instead of completely ignoring `212`, make it so that only arguments starting with `_` is ignored. Also do the same for `211` and `213`.

Code | Description
-- | --
211 | Unused local variable.
212 | Unused argument.
213 | Unused loop variable.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -8,7 +8,10 @@ self = false
 
 -- Glorious list of warnings: https://luacheck.readthedocs.io/en/stable/warnings.html
 ignore = {
-  "212", -- Unused argument, In the case of callback function, _arg_name is easier to understand than _, so this option is set to off.
+  -- Unused argument/variable starting with _
+  "211/_.*",
+  "212/_.*",
+  "213/_.*",
   "122", -- Indirectly setting a readonly global
 }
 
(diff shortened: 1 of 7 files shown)


Conni2461:
I am not sure i like prefixing unused arguments with `_`

MunifTanjim:
> I am not sure i like prefixing unused arguments with `_`

Ah, okay! I created this PR reading this comment: `-- Unused argument, In the case of callback function, _arg_name is easier to understand than _, so this option is set to off.`.

But if it's not really wanted, we can close the PR.

Conni2461:
Are you talking about a sumneko error? I do not have this issue

MunifTanjim:
> Are you talking about a sumneko error? I do not have this issue

So, these were my assumptions about the linting preference for this repo:

For a callback function that takes the argument `arg_one` and `arg_two`, if we only need to use `arg_two` - we don't like to do:

```lua
callback = function(_, arg_two) print(arg_two) end
```

instead we want to write:

```lua
callback = function(_arg_one, arg_two) print(arg_two) end
```

But then `luacheck` would consider this a lint error, because `_arg_one` is unused.

To tackle this `212` was completely disabled. So `luacheck` didn't check for unused arguments at all.

So, the change in this PR was to make `luacheck` check for unused arguments, but only if it doesn't have `_` prefix.

Anyway, seems like my assumptions are wrong. 😃  Closing the PR.

Issue #1018, opened 2021-07-17
using initial_mode = "normal" search results are not displaying property (small bug)


Adomas-beep-boop-maggots:
### Description

Using "initial_mode" option with builtin.buffers function causes telescope to
not refresh and therefore leaving empty "results" display

**Expected Behavior**

Expected "results" display when using "initial_mode" option in builtin.buffers
function exactly when new telescope search buffer lists window spawns, my
reason is because my buffer list is usually rather short, so id rather use hjkl
keys to navigate without pressing escape

**Actual Behavior**

*you're lol.
When telescope search buffer list spawns the results are empty, but when i go
to insert mode (press i) results of buffer list appear

### Details

<details><summary>Reproduce</summary>
i use "<leader>fb" keybind to call ":Telescope buffers" 
and have a initial_mode = "normal", option in buffers section 
(I'm not sure how to describe it programmatically sorry)
(don't worry I'll link my short telescope.rc.vim below)

</details>

<details><summary>Environment</summary>
I'm note sure what to write there...

- nvim --version output: 
- NVIM v0.5.0

- Operating system: 
- Arch linux
no. Arch GNU/linux lol

- Telescope commit: 
1. https://pastebin.com/eUrDF6GR ~/.config/nvim/after/plugin/telescope.rc.vim
2. :Telescope live_grep
3. https://i.imgur.com/tcaKrQD.png 
(it just hangs like that, but refreshes when in insert mode)
4.  https://i.imgur.com/lQbfgot.png

</details>

<details><summary>https://pastebin.com/eUrDF6GR</summary>
<p>

```viml
https://pastebin.com/eUrDF6GR
```

Adomas-beep-boop-maggots:
really sorry for formatting here this my first issue post.
placing things in brackets is hard

Adomas-beep-boop-maggots:
updated the imigur picture links. i copied the wrong ones at first...

JoseConseco:
I can confirm. Using  config option   :  initial_mode='normal'  -  wont initialize results list.   Only going into insert mode will populate the result list.   This issue is visible in eg: grep_string

Conni2461:
This bug will be fixed in #987 You can try if that works for you

Issue #777, opened 2021-04-19
regression when using no `prompt_prefix`


Conni2461:
Probably related #709. But have not testing it

`:Telescope find_files prompt_prefix=` doesn't show any files until you start typing.

Annoyingly pinging my old boss @tjdevries

Issue #750, opened 2021-04-11
initial_mode=normal doesn't populate selections


sodiumjoe:
### Description

`initial_mode=normal` doesn't populate the selections until you go into insert mode

**Expected Behavior**

`initial_mode=normal` should populate buffers exactly like `initial_mode=insert`, but in normal mode.

**Actual Behavior**

`initial_mode=normal` doesn't populate the selections until you go into insert mode

### Details

<details><summary>Reproduce</summary>

1. nvim -nu test.vim test.vim
2. :Telescope buffers initial_mode=normal
3. selections are empty
4. `i` to change to insert mode
5. selections are populated

<details><summary>Environment</summary>

- nvim --version output: 
```
NVIM v0.5.0-dev+370469b
Build type: Release
LuaJIT 2.1.0-beta3
Compilation: clang -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNDEBUG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/tmp/neovim-20210405-60261-zzx58z/build/config -I/tmp/neovim-20210405-60261-zzx58z/src -I/opt/homebrew/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -I/opt/homebrew/opt/gettext/include -I/tmp/neovim-20210405-60261-zzx58z/build/src/nvim/auto -I/tmp/neovim-20210405-60261-zzx58z/build/include
Compiled by joe@Josephs-MacBook-Air.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/HEAD-370469b_2/share/nvim"

Run :checkhealth or more info
```
- Operating system: macos big sur 11.2.2 (20D80) (M1 macbook air)
- Telescope commit: `5bd6f5c`

</details>

<details><summary>Configuration</summary>
<p>

```viml
[...]

PR #987, opened 2021-07-14
start: Working w/ async jobs

tjdevries:
This should just work by checking this only. No backwards incompat changes (hopefully)

----

old text

Fixes #750
Fixes #777 
Fixes #1018 

Latest Clip: https://clips.twitch.tv/SincereQuaintNostrilWholeWheat-9VR2k_is7r1F01VN <-- LiveGrep from home directory

Example of some cool improvements :) https://clips.twitch.tv/GrossRenownedOryxKappaClaus-mTa6exkBdFuym-F5

Will keep on working on this cause I'm (obviously) quite excited :laughing:

PR #1046, opened 2021-07-24
Small pickers refactor

vlmonk:
Fix these issues:

- https://github.com/nvim-telescope/telescope.nvim/issues/750
- https://github.com/nvim-telescope/telescope.nvim/issues/777
- https://github.com/nvim-telescope/telescope.nvim/issues/1018

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -719,7 +719,7 @@ function Picker:refresh(finder, opts)
     self._multi = MultiSelect:new()
   end
 
-  self.__on_lines(nil, nil, nil, 0, 1)
+  self.updater()
 end
 
 function Picker:set_selection(row)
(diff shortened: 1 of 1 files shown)


lua/telescope/pickers.lua
```diff
@@ -339,10 +339,14 @@ function Picker:find()
 
   -- Prompt prefix
   local prompt_prefix = self.prompt_prefix
+  a.nvim_buf_set_option(prompt_bufnr, "buftype", "prompt")
```
Conni2461, line 342: I would prefer only to set prompt as buftype if we actually have a prefix because prompt buffers are still broken in neovim, because of missing vim patches and you can workaround these issues if you are fine without a prompt prefix.

See https://github.com/nvim-telescope/telescope.nvim/labels/upstream

Conni2461:
Techincally this would work for me but tj is currently rewriting all of that here #987 so he needs to approve. Also its a good chance that everything might change after #987 is merged. All of that bugs are regressions introduced in his previous async attempt :laughing:

Conni2461:
I just tested all issue in https://github.com/nvim-telescope/telescope.nvim/pull/987 and they are fixed there so this PR is no longer needed.

Still thanks for looking into it :)

PR #1051, opened 2021-07-26
feat: allow caching and resuming picker

### LINKED ISSUE #751: resume previous list

fdschmidt93:
Will close #751

What?
- Introduce `resume` (resume last picker) and `pickers` (picker over cached pickers) `builtin` pickers

How?
- Modify `pickers.lua`  so picker attributes are not `nil`-ed on detach but stored in global `cached_pickers` table-stack instead
- `cache_limit` > 0 reduces # of cached entries via `LinkedList:truncate` (primarily to reduce size)
- New buffer previewer that writes original results in preview

Showcase

https://user-images.githubusercontent.com/39233597/127784598-e3d32413-2957-4fc5-8a1e-8ea6598b42b2.mp4

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -25,6 +25,28 @@ telescope.setup({opts})                                    *telescope.setup()*
 
         Default: true
 
+                                          *telescope.defaults.cache_picker*
+    cache_picker: ~
+        This field handles the configuration for picker caching.
+        By default it is a table, with default values (more below).
+        To disable caching, set it to false.
+
+        Caching preserves all previous multi selections and results and
+        therefore may result in slowdown or increased RAM occupation
+        if too many pickers (`cache_picker.num_pickers`) or entries
+        ('cache_picker.limit_entries`) are cached.
+
+        Fields:
+          - num_pickers:      The number of pickers to be cached.
+                              Set to -1 to preserve all pickers of your session.
+                              If passed to a picker, the cached pickers with
+                              indices larger than `cache_picker.num_pickers` will
+                              be cleared.
+                              Default: 1
+          - limit_entries:    The amount of entries that will be written in the
+                              Default: 1000
+
+
                                       *telescope.defaults.default_mappings*
     default_mappings: ~
         Not recommended to use except for advanced users.
@@ -367,8 +389,8 @@ builtin.live_grep({opts})                                *builtin.live_grep()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {cwd}             (string)    directory path to search from (default
-                                      is cwd, use utils.buffer_dir() to search
+        {cwd}             (string)    root dir to search from (default is cwd,
+                                      use utils.buffer_dir() to search
                                       relative to open buffer)
         {grep_open_files} (boolean)   if true, restrict search to open files
                                       only, mutually exclusive with
@@ -388,8 +410,8 @@ builtin.grep_string({opts})                            *builtin.grep_string()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {cwd}             (string)    directory path to search from (default
-                                      is cwd, use utils.buffer_dir() to search
+        {cwd}             (string)    root dir to search from (default is cwd,
+                                      use utils.buffer_dir() to search
                                       relative to open buffer)
         {search}          (string)    the query to search
         {search_dirs}     (table)     directory/directories to search in
@@ -408,9 +430,9 @@ builtin.find_files({opts})                              *builtin.find_files()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {cwd}          (string)   directory path to search from (default is
-                                  cwd, use utils.buffer_dir() to search
-                                  relative to open buffer)
+        {cwd}          (string)   root dir to search from (default is cwd, use
+                                  utils.buffer_dir() to search relative to
+                                  open buffer)
         {find_command} (table)    command line arguments for `find_files` to
                                   use for the search, overrides default config
         {follow}       (boolean)  if true, follows symlinks (i.e. uses `-L`
@@ -446,8 +468,8 @@ builtin.file_browser({opts})                          *builtin.file_browser()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {cwd}      (string)   directory path to browse (default is cwd, use
-                              utils.buffer_dir() to browse relative to open
+        {cwd}      (string)   root dir to browse from (default is cwd, use
+                              utils.buffer_dir() to search relative to open
                               buffer)
         {depth}    (number)   file tree depth to display (default is 1)
         {dir_icon} (string)   change the icon for a directory. default: 
@@ -673,6 +695,35 @@ builtin.search_history({opts})                      *builtin.search_history()*
     - Default keymaps:
       - `<C-e>`: open a search window with the text of the currently selected
         search result populated in it
+
+
+    Parameters: ~
+        {opts} (table)  options to pass to the picker
+
+
+builtin.resume({opts})                                      *builtin.resume()*
+    Opens the previous picker in the identical state (incl. multi selections)
+    - Notes:
+      - Requires `cache_picker` in setup or when having invoked pickers, see
+        |telescope.defaults.cache_picker|
+
+
+    Parameters: ~
+        {opts} (table)  options to pass to the picker
+
+    Fields: ~
+        {cache_index} (number)  what picker to resume, where 1 denotes most
+                                recent (default 1)
+
+
+builtin.pickers({opts})                                    *builtin.pickers()*
+    Opens a picker over previously cached pickers in there preserved states
+    (incl. multi selections)
+    - Default keymaps:
+      - `<C-x>`: delete the selected cached picker
+    - Notes:
+      - Requires `cache_picker` in setup or when having invoked pickers, see
+        |telescope.defaults.cache_picker|
 
 
     Parameters: ~
@@ -1576,6 +1627,16 @@ actions.cycle_previewers_next({prompt_bufnr})*actions.cycle_previewers_next()*
 actions.cycle_previewers_prev({prompt_bufnr})*actions.cycle_previewers_prev()*
     Cycle to the previous previewer if there is one available.
     This action is not mapped on default.
+
+
+    Parameters: ~
+        {prompt_bufnr} (number)  The prompt bufnr
+
+
+actions.remove_selected_picker({prompt_bufnr})*actions.remove_selected_picker()*
+    Removes the selected picker in |builtin.pickers|.
+    This action is not mapped by default and only intended for
+    |builtin.pickers|.
 
 
     Parameters: ~
--- a/lua/telescope/algos/linked_list.lua
+++ b/lua/telescope/algos/linked_list.lua
@@ -215,4 +215,41 @@ function LinkedList:ipairs()
   end
 end
 
+function LinkedList:truncate(max_results)
+  if max_results >= self.size then
+    return
+  end
+
+  local current_node
+  if max_results < self.size - max_results then
+    local index = 1
+    current_node = self.head
+    while index < max_results do
+      local node = current_node
+      if not node.next then
+        break
+      end
+      current_node = current_node.next
+      index = index + 1
+    end
+    self.size = max_results
+  else
+    current_node = self.tail
+    while self.size > max_results do
+      if current_node.prev == nil then
+        break
+      end
+      current_node = current_node.prev
+      self.size = self.size - 1
+    end
+  end
+  self.tail = current_node
+  self.tail.next = nil
+  if max_results < self.track_at then
+    self.track_at = max_results
+    self.tracked = current_node.item
+    self._tracked_node = current_node
+  end
+end
+
 return LinkedList
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -6,7 +6,9 @@ local make_entry = require "telescope.make_entry"
 local Path = require "plenary.path"
 local pickers = require "telescope.pickers"
 local previewers = require "telescope.previewers"
+local p_window = require "telescope.pickers.window"
 local sorters = require "telescope.sorters"
+local state = require "telescope.state"
 local utils = require "telescope.utils"
 
 local conf = require("telescope.config").values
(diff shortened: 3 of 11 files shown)


lua/telescope/pickers.lua
```diff
@@ -401,18 +405,31 @@ function Picker:find()
...
+        -- resume previous picker
+        local index = 1
+        for entry in self.manager:iter() do
+          self:entry_adder(index, entry, _, true)
+          index = index + 1
+        end
+        self.cached_picker = false
+        -- TODO(fdschmidt93) this does not work for live_grep
+        -- redraw highlighting for multi-selections
+        self.refresh()
```
fdschmidt93, line 432: I've been dreading already too long over this.

The straightforward approach of 

```lua
local index = 1
for entry in self.manager:iter() do
  self:entry_adder(index, entry, _, true)
  index = index + 1
  if self:can_select_row(row) and self._multi:is_selected(entry) then
    self.highlighter:hi_multiselect(row, true)
  end
end
```

doesn't work. I've tried a lot of other hacks I could think of but to no avail. In any case, selected entries in any case would be highlighted once you scroll over them.

lua/telescope/pickers.lua
```diff
@@ -401,18 +405,31 @@ function Picker:find()
...
+        -- resume previous picker
+        local index = 1
+        for entry in self.manager:iter() do
+          self:entry_adder(index, entry, _, true)
+          index = index + 1
+        end
+        self.cached_picker = false
+        -- TODO(fdschmidt93) this does not work for live_grep
+        -- redraw highlighting for multi-selections
+        self:refresh()
```
fdschmidt93, line 432: I've been dreading already too long over this.

The straightforward approach of 

```lua
local index = 1
for entry in self.manager:iter() do
  self:entry_adder(index, entry, _, true)
  index = index + 1
  if self:can_select_row(row) and self._multi:is_selected(entry) then
    self.highlighter:hi_multiselect(row, true)
  end
end
```

doesn't work. I've tried a lot of other hacks I could think of but to no avail. In any case, selected entries in any case would be highlighted once you scroll over them.

lua/telescope/algos/linked_list.lua
```diff
@@ -215,4 +215,25 @@ function LinkedList:ipairs()
   end
 end
 
+function LinkedList:truncate(max_results)
```
tjdevries, line 218: Hmm, we could decide whether to go from the front or the end based on the length and the max results.

For example, if we have 500,000 results and we limit 700,000 then there's actually no reason to go at all.

If we have 1,000,001 results and limit is 1,000,000 better to just truncate the last entry.

doc/telescope.txt
```diff
@@ -23,6 +23,25 @@ telescope.setup({opts})                                    *telescope.setup()*
...
+        Default: true
+
+                                           *telescope.defaults.cache_limit*
+    cache_limit: ~
+        Will restrict the number of cached entries.
+
+        Only relevant if `cache_picker = true`. Set to -1 to avoid
+        caching limit.
+
+        Default: 10000
```
tjdevries, line 35: I think our default could be even smaller if it affects performance at all.

fdschmidt93:
I'd be to get advice on whether the buffer previewer can be done more cleverly to universally account for (filtered) ascending or descending results. Lines apparently cannot be placed at an arbitrary row for empty buffers.

doc/telescope.txt
fdschmidt93: Agreed

tami5:
> It would be great to persist pickers across sessions by writing them to disk as we then can preserve multi selections for cool stuff.
> 
> 
> 
> However, it quite challenging:
> 
> - Pickers across projects are mostly irrelevant to one another -- requires likely something like git root resolved workspace storage
> 
> - Serializing lua tables appears tedious for that use case ... really happy to hear ideas on this as I think it's super cool :)

I'd say use sql.nvim, it is fully equipped with this kind of stuff 🤣. But @tjdevries hate being tide to yet another dependency. Alternatively, look at how I serialized and deserialized lua tables, it's simply. then maybe save and load then from a file.

btw great job @fdschmidt93

tjdevries:
@fdschmidt93 I think saving to disk would make a good extension :) (as @tami5 would predict that I would say). If we end up finding simple, OS agnostic, performant solutions over time there, then we can talk about moving it back into core, but I do not want to hold up a PR and/or create undue complexity for that feature at this time.

Additionally, if you make it an extension, you can require sql.nvim, or other luarocks packages if desired.

tjdevries:
I do not like that the pickers are cached forever. We should make a maximum size so that we release the pickers at some point in the future (for example, another config option is OK to do perhaps). We should set this to something small by default, perhaps only 1.

This will prevent keeping around 100,000 entries for the rest of your nvim editing session, which would be very confusing for people who just boot up telescope for the first time and live grep somewhere a few times.

tjdevries:
Also, if we add more entries, then I would move us towards doing something like having a `resume` option that takes several configuration options for pickers. Idk, I'd have to think a bit more about that generally then.

fdschmidt93:
Thanks @tami5 :)

I've briefly looked at `sql.nvim/json.lua` and if I understand all correctly it's not look I cannot easily dump a complex table like a picker and its finders, entries, and selections in there without significant wrappers or telescope.nvim changes to make my life easier. As in, the way to probably best use `sql.nvim` would be to parse entries, selections, and whatever required to reinstantiate the picker and then run `find`. Coming from Python, I was looking for something like `pickle` :sweat_smile: 

@tjdevries 

Agreed on the extension :)

Makes sense, I'll add a limit of # cached pickers and expose it as a config. For context, as is, it's not really an issue since the entries are truncated to the first 1K by default.

tjdevries:
I can't really imagine a usecase for looking at the 327th picker ago :laughing:

fdschmidt93:
We now
- Cache only `cache_n_pickers` (defaullt: 10, @tjdevries if you feel strongly about only the last, happy to change though from all my tests it should be fine)
- Can override `opts` of `builtin.resume` and `builtin.pickers` (latter passed onto `resume` ofc) consistent with other pickers

:)

lua/telescope/builtin/internal.lua
```diff
@@ -1,12 +1,15 @@
 local actions = require "telescope.actions"
 local action_set = require "telescope.actions.set"
 local action_state = require "telescope.actions.state"
+local config = require "telescope.config"
```
Conni2461, line 4: Not needed we already have `conf` imported (unless you need something else from config, doesn't look like it rn)

lua/telescope/config.lua
```diff
@@ -279,6 +279,42 @@ local telescope_defaults = {
   ]],
   },
 
+  cache_picker = {
```
Conni2461, line 282: Rather than having 3 (or more) values could we do something similar to `layout_config` and `history` where we have one table setting with multiple keys?

lua/telescope/builtin/internal.lua
fdschmidt93: Yes, slight oversight sorry :) fixed locally and to be updated once more comes in

lua/telescope/pickers.lua
```diff
@@ -490,6 +529,8 @@ function Picker:find()
 
   if self.initial_mode == "insert" then
     vim.cmd [[startinsert!]]
+    -- make starting in insert mode more robust
+    vim.cmd [[normal! A]]
```
Conni2461, line 533: Is this a workaround for opening a picker in an action if telescope wasn't closed with `actions._close(prompt_bufnr, true)`

lua/telescope/builtin/internal.lua
```diff
@@ -71,6 +74,56 @@ internal.builtin = function(opts)
...
+      entry_maker = make_entry.gen_from_picker(opts),
+    },
+    previewer = previewers.pickers.new(opts),
+    sorter = conf.generic_sorter(opts),
+    cache_picker = false,
+    attach_mappings = function(_)
+      actions.select_default:replace(function(prompt_bufnr)
+        local current_picker = action_state.get_current_picker(prompt_bufnr)
+        local selection_index = current_picker:get_index(current_picker:get_selection_row())
+        actions.close(prompt_bufnr)
```
Conni2461, line 118: You should close with `actions._close(prompt_bufnr, true)`

lua/telescope/pickers.lua
fdschmidt93: Sort of, yes 😅 as I didn't recall `_close` vs `close` in `builtin.internal.pickers`.

Not sure whether to keep this (since it shouldn't hurt?) or only change to `_close` in pickers action.

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -858,6 +860,77 @@ previewers.highlights = defaulter(function(_)
...
+      if self.state and self.state.last_set_bufnr and vim.api.nvim_buf_is_valid(self.state.last_set_bufnr) then
+        vim.api.nvim_buf_clear_namespace(self.state.last_set_bufnr, ns_telescope_multiselection, 0, -1)
+      end
+    end,
+
+    define_preview = function(self, entry, status)
+      putils.with_preview_window(status, nil, function()
+        local ns_telescope_entry = vim.api.nvim_create_namespace "telescope_entry"
+        local preview_height = vim.api.nvim_win_get_height(status.preview_win)
+        if not self.state.bufname then
```
Conni2461, line 892: Can we flip this to make it a little bit more readable. `if self.state.bufname then return end`. I currently having a hard time with the indents and all the ends :laughing:

lua/telescope/config.lua
fdschmidt93: I hope this is fine now; I'm afraid this increased control flow complexity quite a bit as I'm writing to the `pickers.cache_picker` during caching and `opts.cache_picker = false` can be validly passed to `builtin.{resume, pickers}`.

sodiumjoe:
Trying out this branch, a couple of notes few notes:
* it would be great if it would also resume the selection state
* it would be great if it would also resume the input buffer mode

The workflow i'm thinking of is quickfix-ish, i.e. I grep for something, then go through the list of results making edits. So when I resume, I want to be in normal mode, with the same selection highlighted.

fdschmidt93:
Many thanks for trying out the branch and the valuable feedback :) I've incorporated your suggestions as far as possible right now.

There still are a couple of issues around initial normal mode, most prominently wrt caching `default_text` does not play nicely and results are only shown once insert mode was entered. I guess most, if not all, issues around this will be fixed in #987 

I'll most likely have to rebase this branch then to #987 anyways and will validate then that these cases around caching work as expected.

Thanks again!

sodiumjoe:
Thanks, this is great! Now that I think about it, I reported https://github.com/nvim-telescope/telescope.nvim/issues/750 which might have something to do with the issues starting in normal mode. I'll continue testing this branch, but so far so good!

lua/telescope/config.lua
```diff
@@ -279,6 +290,33 @@ local telescope_defaults = {
...
+    {
+      num_pickers = 20,
```
tjdevries, line 295: I really think this should be 1 by default. If users are interested in storing lots of em, they can do that.

Otherwise, this lets them do the most common use case, which is resuming the last thing they did. I don't think this will happen often to store a bunch of them.

lua/telescope/pickers.lua
```diff
@@ -401,18 +408,53 @@ function Picker:find()
...
+        local ok, msg = pcall(function()
+          self.finder(prompt, process_result, vim.schedule_wrap(process_complete))
+        end)
 
-      if not ok then
-        log.warn("Failed with msg: ", msg)
+        if not ok then
+          log.warn("Failed with msg: ", msg)
+        end
+      else
```
tjdevries, line 425: Hmmm, this section makes me a bit nervous.

We have a few vim.schedules, which I'm hoping we should not have to do necessarily.

Also, with newest update (you'll have to pull from master again) you should be able to do something like await_scheduler() instead, if you really need to do scheduling to make this work out OK.

Also, I'd see if we can re-use more code here or abstract it into some helper function within picker so that we don't share too much state here accidentally.

Does that make sense?

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -234,6 +234,8 @@ previewers.new_buffer_previewer = function(opts)
         buf_delete(bufnr)
       end
     end
+    -- enable resuming picker with existing previewer to avoid lookup of deleted bufs
+    bufname_table = {}
```
tjdevries, line 238: I can't find anywhere that this is being used (on first search, just in github browser)

tjdevries:
Other than a few nitpicks, I just want to make sure that this still works with the newest async updates and maybe it will allow writing some of the code a bit easier and more straightforward in the pickers.lua section.

Once we do that, I think we can probably merge!

lua/telescope/previewers/buffer_previewer.lua
Conni2461: getting a new table ref here should be good. I think we can ignore that comment :laughing:

lua/telescope/pickers.lua
fdschmidt93: > re-use more code here

The function for highlighting was already there :laughing: I just hadn't seen it.

Yes, `await_scheduler` is required to make this work.

fdschmidt93:
I've rebased to latest master and addressed the comments. One thing that doesn't work (should it? also seems to affect master), is setting `default_text` for a picker if `initial_mode="normal"`, such that the `default_text` is shown in the prompt.

@sodiumjoe given your experience with this branch, could you also please retest it and report whether everything fine is on your end? If you are using the `pickers` picker then please set caching in your set up adequately, like so:

```lua
telescope.setup {
  defaults = {
    cache_picker = {
      num_pickers = 20 -- your preferred number here, values up to 100 should be perfectly fine; likely even much higher
    }
  }
}
```

doc/telescope.txt
```diff
@@ -25,6 +25,28 @@ telescope.setup({opts})                                    *telescope.setup()*
...
+        if too many pickers (`cache_picker.num_pickers`) or entries
+        ('cache_picker.limit_entries`) are cached.
+
+        Fields:
+          - num_pickers:      The number of pickers to be cached.
+                              Set to -1 to preserve all pickers of your session.
+                              If passed to a picker, the cached pickers with
+                              indices larger than `cache_picker.num_pickers` will
+                              be cleared.
+                              Default: 20
```
Conni2461, line 45: docs no longer align.

fdschmidt93: Sorry & thanks, fixed.

lua/telescope/config.lua
Conni2461: I like the config now. I am more concerned about a better user interface than a couple more ifs in code :laughing: 
Also i love this table config to group stuff since clason proposed it for history. way better than having multiple string values that are prefixed with something

Conni2461 (approved):
LGTM

Thanks :) and good job

fdschmidt93:
Thanks :)

max397574:
afaict it isn't possible but ig you could opena feature request issue for this
it sounds like something really useful @yyy33

yyy33:
> afaict it isn't possible but ig you could opena feature request issue for this it sounds like something really useful @yyy33

thank's , i opened it

PR #1058, opened 2021-07-28
fix(entry_to_qf): handle nil entry.cwd and absolute filenames

bi0ha2ard:
When entry.cwd is nil, filename was set to vim's cwd for the quickfix entries

This happens for example with `lsp_references` which returned entries like
```
filename = "/some/absolute/path.hpp",
cwd = nil
```
`Path:new(nil, "/some/absolute/path.hpp")` would turn into `Path:new(nil)` which turns into vim's current working directory with `:absolute()`, leading to the quickfix list items not jumping to the right place.

This should handle the edge-cases correctly, though I didn't find anything that produced a relative `entry.filename` and a `nil` `entry.cwd` at the same time, so I didn't test that explicitly.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -17,10 +17,9 @@ local p_scroller = require "telescope.pickers.scroller"
 local action_state = require "telescope.actions.state"
 local action_utils = require "telescope.actions.utils"
 local action_set = require "telescope.actions.set"
+local from_entry = require "telescope.from_entry"
 
 local transform_mod = require("telescope.actions.mt").transform_mod
-
-local Path = require "plenary.path"
 
 local actions = setmetatable({}, {
   __index = function(_, k)
@@ -585,7 +584,7 @@ end
 local entry_to_qf = function(entry)
   return {
     bufnr = entry.bufnr,
-    filename = Path:new(entry.cwd, entry.filename):absolute(),
+    filename = from_entry.path(entry, false),
     lnum = entry.lnum,
     col = entry.col,
     text = entry.text or entry.value.text or entry.value,
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -367,6 +367,8 @@ builtin.find_files({opts})                              *builtin.find_files()*
                                   flag for the `find` command)
         {hidden}       (boolean)  determines whether to show hidden files or
                                   not (default is false)
+        {no_ignore}    (boolean)  show files ignored by .gitignore, .ignore,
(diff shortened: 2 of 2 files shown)


levouh:
@Conni2461 or someone with the rights, you got a minute to review this one? I have a similar issue where the actual usage from something like `live_grep` works when going to a file directly (default `<CR>` action or similar), but if you pass `cwd` and `search_dirs` to `live_grep` and the `entry.filename` is already absolute, we end up going `cwd` + `absolute_path` when forming paths for the `quickfix` list which falls under this same kind of error.

lua/telescope/actions/init.lua
```diff
@@ -546,10 +546,19 @@ actions.git_staging_toggle = function(prompt_bufnr)
   end
 end
 
+local make_entry_fname_abs = function(entry)
```
levouh, line 549: Only comment here is that usually the prefixes for `make_entry_*` functions are the actual `entry_maker` functions themselves, so it might be confusing as this function isn't actually an `entry_maker` per se, but rather altering a attribute of an entry.

Also this might go better in `util.lua` or similar rather than here where we are declaring baseline actions?

levouh: Actually that second comment about putting this in `util.lua` might not make sense, as the `entry` argument is contextual in that it must be a table with `cwd` and `filename` keys, etc.

Conni2461:
I do not know who merged the previous thing (`Path:new(entry.cwd, entry.filename):absolute()`) in, but we have a util to get the right path, dont we?!
We use it everywhere in previewer and they work. https://github.com/nvim-telescope/telescope.nvim/blob/b742c50bf119d7c442a553cc52e8afde0086ae02/lua/telescope/previewers/buffer_previewer.lua#L316 and https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/from_entry.lua

Most `entry_maker` already have a `.path` which does the `cwd` concat when its needed, e.g. https://github.com/nvim-telescope/telescope.nvim/blob/b742c50bf119d7c442a553cc52e8afde0086ae02/lua/telescope/make_entry.lua#L150-L154

No harm against you, i think the previous pr didn't address the issue correctly. https://github.com/nvim-telescope/telescope.nvim/pull/1037

I think we should just do `from_entry.path(entry, false)`

bi0ha2ard:
> I think we should just do `from_entry.path(entry, false)`

That seems to work too. I've updated the PR to do that now.

Conni2461 (approved):
Thanks LGTM :)

PR #1060, opened 2021-07-29
Feat: multi-select action take 2

fdschmidt93:
Raw WIP, I'll update progress and TODOs (as I'm on and of travelling)

- [x] Change function signature of actions and iterators (to be further reviewed) to take `context` that can effecticely override `entry` and `picker` (think actions on multiple pickers incl. their selections & entries, will depend on #1051 )
- [x] Autosupport for builtin actions via `smart_`, `multi_`, and `entries_` through `__index` method
- [ ] Review and possibly move to `plenary.iterators` where applicable
- [ ] Refactor `actions.mt` to work with custom actions
- [ ] If autosupport is fine: refactor `actions` to only comprise atomic actions
- [ ] Maybe, we should include a general `actions` override in `telescope.setup` like mappings to better facilitate combining variants of actions

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1558,29 +1558,61 @@ action_set.scroll_previewer({prompt_bufnr}, {direction})*action_set.scroll_previ
 ================================================================================
                                                        *telescope.actions.utils*
 
-Utilities to wrap functions around picker selections and entries.
+Utilities to wrap actions and functions around picker selections and entries.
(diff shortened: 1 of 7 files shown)


fdschmidt93:
@tjdevries I'd be very happy to continue on this once #1093, #1143, or something in between gets merged as I've put in some time in multi actions and be optimistic to have this in good shape then quickly. For the time being, I've held back as iirc you've mentioned on stream you want to work on actions. Any preference?

PR #1066, opened 2021-07-31
feat: checkhealth

sunjon:
PR to add some health checks to sanity check common user issues.

### DIFF
--- a/autoload/health/telescope.vim
+++ b/autoload/health/telescope.vim
@@ -1,0 +1,3 @@
+function! health#telescope#check()
+  lua require 'telescope/health'.check_health()
+endfunction
--- a/autoload/health/telescope.vim
+++ b/autoload/health/telescope.vim
@@ -1,3 +1,3 @@
 function! health#telescope#check()
-  lua require 'telescope/health'.check_health()
+  lua require 'telescope.health'.check_health()
 endfunction
--- a/lua/telescope/health.lua
+++ b/lua/telescope/health.lua
@@ -1,4 +1,5 @@
 local fn = vim.fn
+local extension_module = require "telescope._extensions"
 local extension_info = require("telescope").extensions
 local is_win = vim.api.nvim_call_function("has", { "win32" }) == 1
 
@@ -58,12 +59,6 @@ local function lualib_installed(lib_name)
   return res
 end
 
--- local function get_healthcheck(ext_name)
---   local has_check, check = pcall(require, "telescope.extensions"[ext_name]['health'])
---     if(has_check) then return check end
---   return
--- end
-
 local M = {}
 
 M.check_health = function()
@@ -109,7 +104,8 @@ M.check_health = function()
   end
 
   -- Extensions
-  health_start "Installed extensions"
+  health_start "===== Installed extensions ====="
+
   local installed = {}
   for extension_name, _ in pairs(extension_info) do
     installed[#installed + 1] = extension_name
@@ -117,14 +113,13 @@ M.check_health = function()
   table.sort(installed)
 
   for _, installed_ext in ipairs(installed) do
-    local extension_healthcheck = extension_info[installed_ext].health
+    local extension_healthcheck = extension_module._health[installed_ext]
 
-    health_info(("`%s` installed"):format(installed_ext))
+    health_start(string.format("Telescope Extension: `%s`", installed_ext))
     if extension_healthcheck then
-      local announce_check = ("echo '  - extension healthcheck:'"):format(installed_ext)
-      vim.cmd(announce_check)
       extension_healthcheck()
-      vim.cmd "echo ''"
+    else
+      health_info "No healthcheck provided"
     end
   end
 end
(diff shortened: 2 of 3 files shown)


sunjon:
Not sure which of the other dependencies need a check. Suggestions welcome

elianiva:
maybe add `fd` and `find` check since they're the alternative for `find_files`

lua/telescope/health.lua
```diff
@@ -0,0 +1,115 @@
...
+  { lib = "plenary", optional = false },
+  {
+    lib = "nvim-treesitter",
+    optional = true,
+    info = "Install [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter) to enable preview highlight capabilities.",
+  },
+}
+
+local check_binary_installed = function(package)
+  local file_extension = jit.os == "Windows" and "exe" or ""
```
Conni2461, line 43: `jit.os` doesn't return Windows. something like `mingw.....` idk can we not rely on jit. just use `has`

lua/telescope/health.lua
```diff
@@ -0,0 +1,115 @@
...
+  local filename = package.name .. file_extension
```
Conni2461, line 44: Will result in `fdexe`. Missing dot

lua/telescope/health.lua
```diff
@@ -0,0 +1,115 @@
...
+      end
+    end
+  end
+
+  -- Extensions
+  health_start "Installed extensions"
+  local installed = {}
+  for extension_name, _ in pairs(extension_info) do
+    installed[#installed + 1] = extension_name
+  end
```
Conni2461, line 109: hmm what if extensions could hook into checkhealth? by optionally implementing a function? For example we then could check if everything is working (fzf calc something)

sunjon: I've implmented that. If an extension has exposed a `health` function, the parent Telescope healthcheck will call it.
All responsiblity of calling health_ok/health_error, etc is down to the extension, ie. no sanitization of what goes on in that function occurs.

Conni2461: Okay cool thanks :) i'll try it tomorrow with fzf-native and give a overall better feedback but right now code looks good :)

lua/telescope/health.lua
```diff
@@ -0,0 +1,133 @@
...
+  table.sort(installed)
+
+  for _, installed_ext in ipairs(installed) do
+    local extension_healthcheck = extension_info[installed_ext].health
```
Conni2461, line 121: sorry i am still kinda busy. The "issue" that i have is that health is now exported and will appear when doing. `Telescope frecency health`.

On the one hand its not the worst but on the other i am not sure i really like it. I had more something like this in mind (from extension side).

```lua
return require("telescope").register_extension {
  setup = function(ext_config, config)
    -- config
  end,
  exports = {
    -- the important public thins
  },
  health = function()
   local health_ok = vim.fn["health#report_ok"]
   health_ok "okay"
  end
}
```

for implementation we currently have in `_extensions` a `manager` table which holds all extensions. For health, we could just write that in a separate local table under extensions. Similar to `_loaded` or `_config`. Like `extension._health[name] = fn` and then you could all that here.

Does this make sense? Or am i overthinking this :rofl:

clason:
I think you need to rebase on latest master and then remove the check for popup :)

For `fd` et al., it might be helpful to state which executable is used (basically, replicating the logic): `fd` > `rg` >  `find` I believe? With some special case for Windows?

fdschmidt93:
With the recent issues of plenary (being updated and multiple runtimepaths, etc., e.g. see https://github.com/nvim-telescope/telescope.nvim/issues/1105), we could _with_ plenary maybe add checks that the plugins are found uniquely in the remote paths and up to the latest commits, with something like this:

```lua
local ts_utils = require "telescope.utils"

-- check unique path
local plugin_path = vim.tbl_filter(function(path, plugin_name)
  return path:find(plugin_name)
end, vim.api.nvim_list_runtime_paths())
assert(#plugin_path == 1, "Check uniqueness")

local sha = ts_utils.get_os_command_output({ "git", "rev-parse", "HEAD" }, plugin_path[1])[1]
-- below results in { "0b78fe699b9049b8f46942664027b32102979832\tHEAD" }, still needs parsing :)
local plenary_sha = ts_utils.get_os_command_output(
  { "git", "ls-remote", "https://github.com/nvim-lua/plenary.nvim.git", "HEAD" },
  vim.loop.cwd()
)
-- validate that they are the same
```
This of course depends on plenary, but if there is _some_ version of it in the runtimepath, this should work :laughing: which is required for telescope to start anyways. Thoughts?

I'll clean up the code example later, on the jump right now.

autoload/health/telescope.vim
```diff
@@ -0,0 +1,3 @@
+function! health#telescope#check()
+  lua require 'telescope/health'.check_health()
```
tjdevries, line 2: telescope.health instead of telescope/health please

lua/telescope/health.lua
tjdevries: Yes, adding a `health` argument for `register_extension` is very nice. I prefer this.

PR #1024, opened 2021-07-18
Rework documentation

Conni2461:
Ready for Review :)

Rework of readme and wiki (live) + more stuff for vimdoc.

Now that we have vimdocs it makes sense to move some of the more specific documentation in the readme to the vimdocs. That affects mostly setup defaults and other stuff. `vimdoc` > `README`

Basically we need a better introduction for new users.
Todo:
- [x] getting started in vimdoc !!!
- [x] I'll move the api to developers.md for now and write a gettings started guide for people who want to write their own pickers
- [x] Wiki rework (wiki is meant for community and all documentation for telescope should be part of readme.md, developers.md or
vimdoc. But we need to make that people can put configurations in the wiki, so i need to clean it up)
- [x] sorting for setup defaults currently isn't ideal
- [x] add missing opts for builtin

Close #506 (its clearer now that it is required to have ripgrep installed)
Close #806 (same)
Close #780 (document that it is a lua regex with lua documentation link)
Close #277 (this is mostly already done, after the documentation changes we can close it)
Close #596
Close #1197
Close #1204

CC @tami5 @smithbm2316 @rockerBOO 

Shorthand to new rendered documents:
- [README.md](https://github.com/Conni2461/telescope.nvim/blob/docs_rework/README.md)
- [developers.md](https://github.com/Conni2461/telescope.nvim/blob/docs_rework/developers.md)

PR #1073, opened 2021-08-03
docs: Comment out the custom action in mappings

forest0:
Just comment out the custom action in mappings to make it a working config without any change.

Last night, I install this wonderful plugin and give it a try. Several minutes later I want to change the default mapping,
so I open the doc and just copy the configs **without any change**, then I find that config does not work as I expect.
After about ten minutes struggle, I finally realize that config failed to work because `my_cool_custom_action` is not defined.

Maybe we can comment out these two lines, if someone needs to customize action, he will get a clue from the comment.
What's more, it will potentially reduce the unexpected surprise  for impatient people like me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,11 +327,11 @@ require('telescope').setup{
         ["<cr>"] = actions.select_default + actions.center,
 
         -- You can perform as many actions in a row as you like
-        ["<cr>"] = actions.select_default + actions.center + my_cool_custom_action,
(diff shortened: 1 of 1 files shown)


Conni2461:
Sorry i dont think this is useful. I am currently rewriting the whole documentation here #1024 because i see people just copy and pasting stuff in their config that isn't necessary to copy and paste, like the whole default configuration.
My take is, its called README not COPYME so closing

Issue #1036, opened 2021-07-21
Prompt on top on horizontal


ChrisTsou:
add option in the horizontal view for the positions of prompt and results to be exchanged, something like:
```
horizontal = {
    mirror_horizontal = true/false,  -- change Preview position left/right
    mirror_vertical = true/false        -- change prompt/Results top/down
}
```

l-kershaw:
If I understand what you are asking for correctly, this is already possible!
You can use the `prompt_position` key of `layout_config` and set it to "top" (see `:help layout_strategies.horizontal()` for more info).

Depending on your personal preferences, you may also want to change the `sorting_strategy` to "ascending" (note that this is not a part of `layout_config`).

Changing both of these settings will move the prompt to the top of the layout, and have the "best" results nearest the top.

philip1987:
I used to be able to change the prompt position through an ex command such as :Telescope find_files prompt_position=top.
Is there a new way to invoke the same behaviour with an ex command? This would be helpful when creating and testing custom Telescope pickers.

l-kershaw:
Unfortunately, it is not currently possible to configure layouts using the vimscript commands, as our command parser is not set up to handle vimscript dictionaries (although this is something I'm planning on looking at).

If you want to configure layouts for one call of a picker, then you need to use the lua api, for example:
```
:lua require('telescope.builtin').find_files({layout_config={prompt_position='top'}})
```

ChrisTsou:
That is exactly what i wanted! Sorry missed the option. Maybe added to the readme?

PR #1075, opened 2021-08-03
feat: allow tables in vimscript command parser

l-kershaw:
This PR allows users to pass vimscripts lists and dictionaries (which are converted to lua tables) when using the vimscript commands for telescope.

For example:
- `:Telescope find_files layout_config={'prompt_position':'top'}` will now move the prompt to the top
- `:Telescope find_files search_dirs=['~/foo','~/bar']` will set the `search_dirs` to `~/foo` and `~/bar`
- `:Telescope find_files search_dirs=~/foo,~bar` will still work too

Implementation:
- The implementation relies on `vim.fn.eval`, as it seemed better to me to rely on a robust existing method for converting between vimscript objects and lua objects, rather than try to make my own and miss some edge cases.
- This choice of implementation can cause some unusual methods of configuration for some things, e.g. `:Telescope find_files layout_config={'border':v:false}`, but these are idiosyncrasies of vimscript that I feel like you sign up for when using the vimscript api for telescope.

TODO:
- [x] docs

Related discussion:
- https://github.com/nvim-telescope/telescope.nvim/issues/1036#issuecomment-885491418
- https://github.com/nvim-telescope/telescope.nvim/issues/901#issuecomment-889802749

@Conni2461 could you have a look at this when you get a chance?

---

Also, I think docgen was not run on a previous PR before merge, as I definitely haven't changed anything with Telescopes git stuff 🤣

### DIFF
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -8,17 +8,25 @@
 --- The lua api is the more direct way to interact with Telescope, as you directly call the
 --- lua functions that Telescope defines.
 --- It can be called in a lua file using commands like:
+--- <pre>
 --- `require("telescope.builtin").find_files({hidden=true, layout_config={prompt_position="top"}})`
+--- </pre>
 --- If you want to use this api from a vim file you should prepend `lua` to the command, as below:
+--- <pre>
 --- `lua require("telescope.builtin").find_files({hidden=true, layout_config={prompt_position="top"}})`
+--- </pre>
 --- If you want to use this api from a neovim command line you should prepend `:lua` to
 --- the command, as below:
+--- <pre>
 --- `:lua require("telescope.builtin").find_files({hidden=true, layout_config={prompt_position="top"}})`
+--- </pre>
 ---
 --- The viml api is more indirect, as first the command must be parsed to the relevant lua
 --- equivalent, which brings some limitations.
 --- The viml api can be called using commands like:
+--- <pre>
 --- `:Telescope find_files hidden=true layout_config={"prompt_position":"top"}`
+--- </pre>
 --- This involves setting options using an `=` and using viml syntax for lists and
 --- dictionaries when the corresponding lua function requires a table.
 ---
@@ -27,7 +35,9 @@
 --- only want to search within the folder `/foo bar/subfolder/` you could not do that using the
 --- viml api, as the path name contains a space.
 --- Similarly, you could NOT set the `prompt_position` to `"top"` using the following command:
+--- <pre>
 --- `:Telescope find_files layout_config={ "prompt_position" : "top" }`
+--- </pre>
 --- as there are spaces in the option.
 ---
 ---@brief ]]
@@ -53,6 +63,7 @@ local split_keywords = {
   ["vimgrep_arguments"] = true,
   ["sections"] = true,
   ["search_dirs"] = true,
+  ["symbols"] = true,
 }
 
 -- convert command line string arguments to
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -53,7 +53,12 @@ local function convert_user_opts(user_opts)
       if ok then
         user_opts[key] = eval
       else
-        user_opts[key] = nil
+        eval = assert(loadstring("return " .. val))()
+        if type(eval) == "table" then
+          user_opts[key] = eval
+        else
+          user_opts[key] = nil
+        end
       end
     end,
   }
@@ -73,13 +78,8 @@ local function convert_user_opts(user_opts)
         user_opts[key] = vim.split(val, ",")
       end
     elseif default_opts[key] ~= nil then
-      -- vim.fn.input('1')
-      -- dump(key,val)
       _switch[type(default_opts[key])](key, val)
     else
-      -- vim.fn.input('2')
-      -- dump(key,val)
-      -- print('type',type(default_opts[key]))
       _switch["string"](key, val)
     end
   end
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -15,6 +15,13 @@ local bool_type = {
   ["true"] = true,
 }
 
+local split_keywords = {
+  ["find_command"] = true,
+  ["vimgrep_arguments"] = true,
+  ["sections"] = true,
+  ["search_dirs"] = true,
+}
+
 -- convert command line string arguments to
 -- lua number boolean type and nil value
 local function convert_user_opts(user_opts)
@@ -42,7 +49,12 @@ local function convert_user_opts(user_opts)
       end
     end,
     ["table"] = function(key, val)
-      user_opts[key] = vim.fn.eval(val)
+      local ok, eval = pcall(vim.fn.eval, val)
+      if ok then
+        user_opts[key] = eval
+      else
+        user_opts[key] = nil
+      end
     end,
   }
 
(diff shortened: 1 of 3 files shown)


tami5:
☺️ it doesn't feel right. idk I doubt if this going to be relied on. 

I vote for forgetting viml exists and just have simpler syntax

`key=\~/,1,false`
`key={key=~/path, key=true }`

Conni2461:
dont listen to tami, its his fault that docgen wasn't generated :laughing:
I think this looks great. We can't just ignore viml if we could we would have never written `:Telescope` commands

tami5:
> dont listen to tami, its his fault that docgen wasn't generated 😆
> I think this looks great. We can't just ignore viml if we could we would have never written `:Telescope` commands

😆 yah blame me for everything.

oh no viml is still valid, but not for user facing stuff. it easier to write `:Telescope find_files layout_config={border = false}` in the command line than `:Telescope find_files layout_config={'border':v:false}`.

I don't understand why I should as a user conform to viml syntax of defining dicts and lists while writing a simple one liner.

l-kershaw:
I was thinking about whether `lua` or `viml` would be better here, and had a bit of a "why not both?" moment.
So I added a fallback that evaluated it as lua. I've now realised that this doesn't play nicely with the rest of the command parser, as we are splitting stuff at `=`, which obviously messes stuff up when you try to do `layout_config={prompt_position='top'}`.
I'll have more of a think about it, and see if I can come up with a way to get both lua and viml syntax to work.

If it doesn't seem feasible, I think we should just keep it as a viml syntax for the viml api.

l-kershaw:
Okay, I think I've managed to get something that seems to work for all the cases I can think to test, excluding ones that I don't think are possible to fix in a robust way (mostly from the way that `<f-args>` works).

The following all work as I would expect:
- `:Telescope find_files search_dirs=~/foo,~/bar`
- `:Telescope find_files search_dirs=['~/foo','~/bar']`
- `:Telescope find_files search_dirs={'~/foo','~/bar'}`
- `:Telescope find_files layout_config={'prompt_position':'top'}`
- `:Telescope find_files layout_config=#{prompt_position:'top'}`
- `:Telescope find_files layout_config={prompt_position='top'}`
- `:Telescope find_files layout_config={width=50,height=10}`
- `:Telescope find_files layout_config={'width':50,'height':10}`
- `:Telescope find_files layout_config=#{width:50,height:10}`

Things that don't work as I would expect are ones that have spaces:
- `:Telescope find_files search_dirs=~/foo, ~/bar`
- `:Telescope find_files search_dirs=['~/foo', '~/bar']`
- `:Telescope find_files search_dirs={'~/foo', '~/bar'}`
- `:Telescope find_files layout_config={'prompt_position' : 'top'}`
- `:Telescope find_files layout_config=#{prompt_position : 'top'}`
- `:Telescope find_files layout_config={prompt_position = 'top'}`
- `:Telescope find_files layout_config={width=50, height=10}`
- `:Telescope find_files layout_config={'width':50, 'height':10}`
- `:Telescope find_files layout_config=#{width:50, height:10}`

I don't think that there is a good way of dealing with the spaces issue since `<f-args>` automatically splits at spaces and tabs, and trying to "sew things back together" in the right places would be a real pain to do for not much gain.

---

I think that this is now a reasonable way of parsing things that allows users to configure using whichever syntax they like (as long  as they don't use spaces 🤣).

@Conni2461 @tami5 when you get a chance, could you test this and let me know your thoughts?

l-kershaw:
@Conni2461 @tami5 I will assume that you haven't found anything egregiously wrong with this, but obviously let me know if you have any suggested changes.

---

I've added docs to give an overview of the differences between the lua and viml apis.
I'm now happy with this, so if people have any suggestions before this PR is merged, please let me know 🙂

tami5:
😍 😍 amazing work @l-kershaw . Give me one sec I'll test it out. sorry I almost forgot about the pr

tami5:
Everything seems to work pretty well for me. Although while testing and I think you've mentioned it `:Telescope find_files search_dirs=~/foo,~/bar` doesn't work.



> The following all work as I would expect:
> 
>     * `:Telescope find_files search_dirs=~/foo,~/bar`
> 
>     * `:Telescope find_files search_dirs=['~/foo','~/bar']`
> 
>     * `:Telescope find_files search_dirs={'~/foo','~/bar'}`
> 
>     * `:Telescope find_files layout_config={'prompt_position':'top'}`
> 
>     * `:Telescope find_files layout_config=#{prompt_position:'top'}`
> 
>     * `:Telescope find_files layout_config={prompt_position='top'}`
> 
>     * `:Telescope find_files layout_config={width=50,height=10}`
> 
>     * `:Telescope find_files layout_config={'width':50,'height':10}`
> 
>     * `:Telescope find_files layout_config=#{width:50,height:10}`
> 
> 
> Things that don't work as I would expect are ones that have spaces:
> 
>     * `:Telescope find_files search_dirs=~/foo, ~/bar`
> 
>     * `:Telescope find_files search_dirs=['~/foo', '~/bar']`
> 
>     * `:Telescope find_files search_dirs={'~/foo', '~/bar'}`
> 
>     * `:Telescope find_files layout_config={'prompt_position' : 'top'}`
> 
>     * `:Telescope find_files layout_config=#{prompt_position : 'top'}`
> 
>     * `:Telescope find_files layout_config={prompt_position = 'top'}`
> 
>     * `:Telescope find_files layout_config={width=50, height=10}`
> 
>     * `:Telescope find_files layout_config={'width':50, 'height':10}`
> 
>     * `:Telescope find_files layout_config=#{width:50, height:10}`

isn't the two lists identical?

tami5:
trying to run `Telescope find_files search_dirs={'~/.config', '~/.local'}`

results in `E5108: Error executing lua ...ck/packer/start/telescope.nvim/lua/telescope/command.lua:89: [string "return {'~/.config',"]:1: unexpected symbol near '<eof>'`

lua/telescope/command.lua
```diff
@@ -15,6 +48,13 @@ local bool_type = {
   ["true"] = true,
 }
 
+local split_keywords = {
+  ["find_command"] = true,
+  ["vimgrep_arguments"] = true,
+  ["sections"] = true,
+  ["search_dirs"] = true,
+}
```
Conni2461, line 56: I think you should add `symbols` to this map. https://github.com/nvim-telescope/telescope.nvim/issues/901#issuecomment-891670369

lua/telescope/command.lua
```diff
@@ -1,3 +1,36 @@
...
+
+---@brief [[
+---
+--- Telescope commands can be called through two apis,
+--- the lua api and the viml api.
+---
+--- The lua api is the more direct way to interact with Telescope, as you directly call the
+--- lua functions that Telescope defines.
+--- It can be called in a lua file using commands like:
+--- `require("telescope.builtin").find_files({hidden=true, layout_config={prompt_position="top"}})`
```
Conni2461, line 11: should we add `<pre>` to these long lines, so they are not broken up?
```lua
--- <pre>
--- ...
--- </pre>
```

Conni2461:
tami they are not the same. There is no space allowed as stated in the documentation. try without a space after the `,`.

PR looks good to thanks :)

tami5:
> tami they are not the same. There is no space allowed as stated in the documentation. First snippet has no spaces second one has. Try without a space after the `,`.
> 
> PR looks good to thanks :)

oh yahhh, now it works, I think we should marge after the requested edits

l-kershaw:
I have implemented your suggestions, thanks for the feedback guys 👍

Unless there are any other suggestions, I'm happy for this to be merged 🙂

tami5:
@l-kershaw GO AHEAD, well done

gizmo385:
Does the README need to be updated after this PR? It looks like `find_files` commands similar to the ones in the README are broken after this change:

```
E5108: Error executing lua ~/.vim/plugged/telescope.nvim/lua/telescope/command.lua:100: [string "return rg,--ignore--hidden,--files"]:1: unexpected symbol near '<eof>'
```

l-kershaw:
@gizmo385 this is a bug, sorry about that 😅
It is caused by the fact that having `-` makes the string in the error message an invalid lua expression.

I have an idea for fixing this, and should hopefully have a PR for this up soon (tomorrow at the latest)

PR #1076, opened 2021-08-04
fix: preview the correct man page section

zeertzjq:
With `sections=ALL`, telescope uses the same cache name for man pages with the same name in different sections, so the wrong man page is previewed if multiple man pages have the same name. This PR appends the section name to the cache name, separated by a `"/"` which doesn't appear in man page names.


doc/telescope.txt
```diff
@@ -636,6 +636,11 @@ builtin.man_pages({opts})                                *builtin.man_pages()*
     Parameters: ~
         {opts} (table)  options to pass to the picker
 
+    Fields: ~
+        {sections} (table)  a list of sections to search, use `{ "ALL" }` to
+                            search in all sections
+
+
```
Conni2461, line 643: Please dont manually edit `doc/telescope.txt` its autogenerated. You need to change it here https://github.com/nvim-telescope/telescope.nvim/blob/b7cd8c7699623c4b351009192dc9a03a54195e6b/lua/telescope/builtin/init.lua#L243-L245
Note: currently our parser can only handle single line annotations. Limitation i am already working on.
Also you should read CONTRIBUTING.md :)

Other stuff LGTM  thanks :)

Conni2461:
i do not know why docgen ci doesn't want to run for some people so i executed it locally. Thanks :)

PR #1084, opened 2021-08-05
feat: show keymaps for builtin actions

### LINKED ISSUE #1062: Show keymaps for current picker
I need to read manual or code directly to know the keymaps for a specify picker. It will be very helpful if there is a common way (like C-?) to show keymaps in the Previewer for current pick.

After `Telescope find_files`, press `C-?` could display keymaps of builtin.find_files picker in Previewer.

fdschmidt93:
Will close #1062

E: I've hid a lot of discussions to improve readability.

Previews v5 w/ popups and right titles (mode hint with `opts.only_show_current_mode`)

Prompt at bottom

keymaps_v5

Prompt at top

km_v5_b

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -990,14 +990,14 @@ actions.which_key = function(prompt_bufnr, opts)
   a.nvim_win_set_option(km_opts.border.win_id, "winhl", "Normal:" .. opts.border_hl)
   a.nvim_win_set_option(km_win_id, "winblend", opts.winblend)
 
-  vim.cmd(
-    string.format(
-      "autocmd BufLeave <buffer> ++once lua pcall(vim.api.nvim_win_close, %s, true); pcall(vim.api.nvim_win_close, %s, true); require 'telescope.utils'.buf_delete(%s)",
-      km_win_id,
-      km_opts.border.win_id,
-      km_buf
-    )
-  )
+  vim.cmd(string.format(
+    "autocmd BufLeave <buffer> ++once lua %s",
+    table.concat({
+      string.format("pcall(vim.api.nvim_win_close, %s, true)", km_win_id),
+      string.format("pcall(vim.api.nvim_win_close, %s, true)", km_opts.border.win_id),
+      string.format("require 'telescope.utils'.buf_delete(%s)", km_buf),
+    }, ";")
+  ))
 
   a.nvim_buf_set_lines(
     km_buf,
@@ -1036,14 +1036,14 @@ actions.which_key = function(prompt_bufnr, opts)
   -- only set up autocommand after showing preview completed
   if opts.close_with_action then
     vim.schedule(function()
-      vim.cmd(
-        string.format(
-          "autocmd User TelescopeKeymap ++once lua pcall(vim.api.nvim_win_close, %s, true); pcall(vim.api.nvim_win_close, %s, true); require 'telescope.utils'.buf_delete(%s)",
-          km_win_id,
-          km_opts.border.win_id,
-          km_buf
-        )
-      )
+      vim.cmd(string.format(
+        "autocmd User TelescopeKeymap ++once lua %s",
+        table.concat({
+          string.format("pcall(vim.api.nvim_win_close, %s, true)", km_win_id),
+          string.format("pcall(vim.api.nvim_win_close, %s, true)", km_opts.border.win_id),
+          string.format("require 'telescope.utils'.buf_delete(%s)", km_buf),
+        }, ";")
+      ))
     end)
   end
 end
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -10,16 +10,20 @@
 local a = vim.api
 
 local log = require "telescope.log"
+local config = require "telescope.config"
 local state = require "telescope.state"
 local utils = require "telescope.utils"
+local popup = require "plenary.popup"
 local p_scroller = require "telescope.pickers.scroller"
 
 local action_state = require "telescope.actions.state"
 local action_utils = require "telescope.actions.utils"
 local action_set = require "telescope.actions.set"
+local entry_display = require "telescope.pickers.entry_display"
 local from_entry = require "telescope.from_entry"
 
 local transform_mod = require("telescope.actions.mt").transform_mod
+local resolver = require "telescope.config.resolve"
 
 local actions = setmetatable({}, {
   __index = function(_, k)
@@ -847,8 +851,205 @@ actions.remove_selected_picker = function(prompt_bufnr)
   end
 end
 
+--- Display the keymaps of registered actions similar to which-key.nvim.<br>
+--- - Notes:
+---   - The defaults can be overridden via |action_generate.toggle_registered_actions|.
+---@param prompt_bufnr number: The prompt bufnr
+actions.which_key = function(prompt_bufnr, opts)
+  opts = opts or {}
+  opts.max_height = utils.get_default(opts.max_height, 0.4)
+  opts.only_show_current_mode = utils.get_default(opts.only_show_current_mode, true)
+  opts.mode_width = utils.get_default(opts.mode_width, 1)
+  opts.keybind_width = utils.get_default(opts.keybind_width, 7)
+  opts.name_width = utils.get_default(opts.name_width, 30)
+  opts.column_padding = utils.get_default(opts.column_padding, "  ")
+  opts.column_indent = table.concat(utils.repeated_table(utils.get_default(opts.column_indent, 4), " "))
+  opts.line_padding = utils.get_default(opts.line_padding, 1)
+  opts.separator = utils.get_default(opts.separator, " -> ")
+  opts.close_with_action = utils.get_default(opts.close_with_action, true)
+  opts.normal_hl = utils.get_default(opts.normal_hl, "TelescopePrompt")
+  opts.border_hl = utils.get_default(opts.border_hl, "TelescopePromptBorder")
+  opts.winblend = utils.get_default(opts.winblend, config.values.winblend)
+
+  -- close on repeated keypress
+  local km_bufs = (function()
+    local ret = {}
+    local bufs = a.nvim_list_bufs()
+    for _, buf in ipairs(bufs) do
+      for _, bufname in ipairs { "_TelescopeWhichKey", "_TelescopeWhichKeyBorder" } do
+        if string.find(a.nvim_buf_get_name(buf), bufname) then
+          table.insert(ret, buf)
+        end
+      end
+    end
+    return ret
+  end)()
+  if not vim.tbl_isempty(km_bufs) then
+    for _, buf in ipairs(km_bufs) do
+      utils.buf_delete(buf)
+      local win_ids = vim.fn.win_findbuf(buf)
+      for _, win_id in ipairs(win_ids) do
+        pcall(a.nvim_win_close, win_id, true)
+      end
+    end
+    return
+  end
+
+  local displayer = entry_display.create {
+    separator = opts.separator,
+    items = {
+      { width = opts.mode_with },
+      { width = opts.keybind_width },
+      { width = opts.name_width },
+    },
+  }
+
+  local make_display = function(mapping)
+    return displayer {
+      { mapping.mode, utils.get_default(opts.mode_hl, "TelescopeResultsConstant") },
+      { mapping.keybind, utils.get_default(opts.keybind_hl, "TelescopeResultsVariable") },
+      { mapping.name, utils.get_default(opts.name_hl, "TelescopeResultsFunction") },
+    }
+  end
+
+  local mappings = {}
+  local mode = a.nvim_get_mode().mode
+  for _, v in pairs(action_utils.get_registered_mappings(prompt_bufnr)) do
+    -- holds true for registered keymaps
+    if type(v.func) == "table" then
+      local name = ""
+      for _, action in ipairs(v.func) do
+        if type(action) == "string" then
+          name = name == "" and action or name .. " + " .. action
+        end
+      end
+      if name and name ~= "which_key" then
+        if not opts.only_show_current_mode or mode == v.mode then
+          table.insert(mappings, { mode = v.mode, keybind = v.keybind, name = name })
+        end
+      end
+    end
+  end
+
+  table.sort(mappings, function(x, y)
+    if x.name < y.name then
+      return true
+    elseif x.name == y.name then
+      -- show normal mode as the standard mode first
+      if x.mode > y.mode then
+        return true
+      else
+        return false
+      end
+    else
+      return false
+    end
+  end)
+
+  local entry_width = #opts.column_padding
+    + opts.mode_width
+    + opts.keybind_width
+    + opts.name_width
+    + (3 * #opts.separator)
+  local num_total_columns = math.floor((vim.o.columns - #opts.column_indent) / entry_width)
+  opts.num_rows = math.min(
+    math.ceil(#mappings / num_total_columns),
+    resolver.resolve_height(opts.max_height)(_, _, vim.o.lines)
+  )
+  local total_available_entries = opts.num_rows * num_total_columns
+  local winheight = opts.num_rows + 2 * opts.line_padding
+
+  -- place hints at top or bottom relative to prompt
+  local picker = action_state.get_current_picker(prompt_bufnr)
+  local prompt_win = picker.prompt_win
+  local prompt_row = a.nvim_win_get_position(prompt_win)[1]
+  local prompt_pos = prompt_row <= 0.5 * vim.o.lines
+
+  local modes = { n = "Normal", i = "Insert" }
+  local title_mode = opts.only_show_current_mode and modes[mode] .. " Mode " or ""
+  local title_text = title_mode .. "Keymaps"
+  local popup_opts = {
+    relative = "editor",
+    enter = false,
+    minwidth = vim.o.columns,
+    maxwidth = vim.o.columns,
+    minheight = winheight,
+    maxheight = winheight,
+    line = prompt_pos == true and vim.o.lines - winheight or 0,
+    col = 1,
+    border = { prompt_pos and 1 or 0, 0, not prompt_pos and 1 or 0, 0 },
+    borderchars = { prompt_pos and "─" or " ", "", not prompt_pos and "─" or " ", "", "", "", "", "" },
+    noautocmd = true,
+    title = { { text = title_text, pos = prompt_pos and "N" or "S" } },
+  }
+  local km_win_id, km_opts = popup.create("", popup_opts)
+  local km_buf = a.nvim_win_get_buf(km_win_id)
+  a.nvim_buf_set_name(km_buf, "_TelescopeWhichKey")
+  a.nvim_buf_set_name(km_opts.border.bufnr, "_TelescopeTelescopeWhichKeyBorder")
+  a.nvim_win_set_option(km_win_id, "winhl", "Normal:" .. opts.normal_hl)
+  a.nvim_win_set_option(km_opts.border.win_id, "winhl", "Normal:" .. opts.border_hl)
+  a.nvim_win_set_option(km_win_id, "winblend", opts.winblend)
+
+  vim.cmd(
+    string.format(
+      "autocmd BufLeave <buffer> ++once lua pcall(vim.api.nvim_win_close, %s, true); pcall(vim.api.nvim_win_close, %s, true); require 'telescope.utils'.buf_delete(%s)",
+      km_win_id,
+      km_opts.border.win_id,
+      km_buf
+    )
+  )
+
+  a.nvim_buf_set_lines(
+    km_buf,
+    0,
+    -1,
+    false,
+    utils.repeated_table(opts.num_rows + 2 * opts.line_padding, opts.column_indent)
+  )
+
+  local keymap_highlights = a.nvim_create_namespace "telescope_whichkey"
+  local highlights = {}
+  for index, mapping in ipairs(mappings) do
+    local row = utils.cycle(index, opts.num_rows) - 1 + opts.line_padding
+    local prev_line = a.nvim_buf_get_lines(km_buf, row, row + 1, false)[1]
+    if index == total_available_entries and total_available_entries > #mappings then
+      local new_line = prev_line .. "..."
+      a.nvim_buf_set_lines(km_buf, row, row + 1, false, { new_line })
+      break
+    end
+    local display, display_hl = make_display(mapping)
+    local new_line = prev_line .. display .. opts.column_padding -- incl. padding
+    a.nvim_buf_set_lines(km_buf, row, row + 1, false, { new_line })
+    table.insert(highlights, { hl = display_hl, row = row, col = #prev_line })
+  end
+
+  -- highlighting only after line setting as vim.api.nvim_buf_set_lines removes hl otherwise
+  for _, highlight_tbl in pairs(highlights) do
+    local highlight = highlight_tbl.hl
+    local row_ = highlight_tbl.row
+    local col = highlight_tbl.col
+    for _, hl_block in ipairs(highlight) do
+      a.nvim_buf_add_highlight(km_buf, keymap_highlights, hl_block[2], row_, col + hl_block[1][1], col + hl_block[1][2])
+    end
+  end
+
+  -- only set up autocommand after showing preview completed
+  if opts.close_with_action then
+    vim.schedule(function()
+      vim.cmd(
+        string.format(
+          "autocmd User TelescopeKeymap ++once lua pcall(vim.api.nvim_win_close, %s, true); pcall(vim.api.nvim_win_close, %s, true); require 'telescope.utils'.buf_delete(%s)",
+          km_win_id,
+          km_opts.border.win_id,
+          km_buf
+        )
+      )
+    end)
+  end
+end
+
 -- ==================================================
--- Transforms modules and sets the corect metatables.
+-- Transforms modules and sets the correct metatables.
 -- ==================================================
 actions = transform_mod(actions)
 return actions
--- a/lua/telescope/actions/utils.lua
+++ b/lua/telescope/actions/utils.lua
@@ -78,4 +78,26 @@ function utils.map_selections(prompt_bufnr, f)
   end
 end
 
+local findnth = function(str, nth)
+  local array = {}
+  for i in string.gmatch(str, "%d+") do
+    table.insert(array, tonumber(i))
+  end
+  return array[nth]
+end
+
+--- Utility to collect mappings of prompt buffer in array of `{mode, keybind, name}`.
+---@param prompt_bufnr number: The prompt bufnr
+function utils.get_registered_mappings(prompt_bufnr)
+  local ret = {}
+  for _, mode in ipairs { "n", "i" } do
+    local mode_mappings = vim.api.nvim_buf_get_keymap(prompt_bufnr, mode)
+    for _, mapping in ipairs(mode_mappings) do
+      local funcid = findnth(mapping.rhs, 2)
+      table.insert(ret, { mode = mode, keybind = mapping.lhs, func = __TelescopeKeymapStore[prompt_bufnr][funcid] })
+    end
+  end
+  return ret
+end
+
 return utils
--- a/lua/telescope/mappings.lua
+++ b/lua/telescope/mappings.lua
@@ -30,6 +30,7 @@ mappings.default_mappings = config.values.default_mappings
       ["<C-q>"] = actions.send_to_qflist + actions.open_qflist,
       ["<M-q>"] = actions.send_selected_to_qflist + actions.open_qflist,
       ["<C-l>"] = actions.complete_tag,
+      ["<C-_>"] = actions.which_key, -- keys from pressing <C-/>
     },
 
     n = {
@@ -56,6 +57,7 @@ mappings.default_mappings = config.values.default_mappings
 
       ["<C-u>"] = actions.preview_scrolling_up,
       ["<C-d>"] = actions.preview_scrolling_down,
+      ["?"] = actions.which_key,
     },
   }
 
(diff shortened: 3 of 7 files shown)


l-kershaw:
I also had a go at making something similar (though never got around to getting it in shape to make a PR).

I haven't had a chance to test this but a few quick discussion points:
- sorting
  - I think it would be helpful to sort the mappings, as it makes it easier for people to find what they are looking for
  - my guess is that the best priority order for sorting would be name, keybind, mode
 - custom actions
   - while it is definitely easier to only include builtin actions, it would be good to give users a way of providing a name for their own actions that they attach to pickers, and then display it in this menu too
   - I thought a good way to do this would be to allow users to pass a table of the form `{func=...,name=...}` when adding a mapping, with the `func` being called as appropriate, and `name` getting passed to the help UI
 - visuals
   - I think it would be better to have the border match the others in telescope, maybe use the same borderchars and highlight as the prompt window? This might be easier when using the `popup` setup as with the other windows (although this may require a PR there to add `noautocmd` as an option)
   - I really like the idea of using multiple columns for this, it makes much better use of space than my attempt. Do we want a separator line between the different columns to aid with readability? I'm a bit confused by the fact that the top and bottom rows in your picture only have one mapping in, what's going on there?
   - Obviously this is WIP, so fine to have the size/position of the window hard coded for now, but we should probably think about the best way to calculate the size/position based on the layout given (e.g. `layout_strategy`, `prompt_position`, etc...).
   We could maybe just do it with two hardcoded options taking up the top/bottom third of the screen, depending on where the prompt window is? I don't know what's best here

I feel like I've ended up rambling a bit, sorry 😅 

I'm busy for the rest of the week, but will happily test this PR and give feedback as needed next week 🙂

fdschmidt93:
Many thanks for the comments :) of course all extremely valid and I pretty much agree on everything. Might have some time tonight or on the weekend to progress.

Sorry that this is still extremely raw; of course nothing will be hardcoded etc. I wrote this up quickly over lunch time and then pushed it before I went out as it was suitable to show an initial progress 😅

l-kershaw:
No worries about it being "raw" and having hardcoded stuff, it is always good to get feedback early in the process 👍
I just wanted to start a discussion about the best way to handle stuff, give some of my thoughts and see what your view was 🙂

fdschmidt93:
I was on mobile earlier, so commenting now in more depth:

> Sorting [...]

Fully agree :) we should probably also only show the current mode as in whether it's called from insert or normal mode to not clutter the hints with identical mappings across modes, but this can be decided later easily and made optionally.

> custom actions [...]  good to give users a way of providing a name for their own actions that they attach to pickers

Yes :) I've envisioned pretty much like you've laid it out here though this will be part of #1060. I've actually looked into this particular thing last night but I haven't come up with a really elegant solution to adequately amend `actions/mt.lua`. Happy to discuss it over there or on gitter.

> popup.nvim

I feel like popup.nvim could use some more love (I'd like to see docs :sweat_smile: ) as well but I see you are on the case :) 

> noautocmd

Yes, in the WIP it's surely needed, but maybe it can be avoided somehow? I'll try.

> separator line

I'd say optional; which-key as the main reference doesn't have it (at least with standard setup).

> [...] think about the best way to calculate the size/position, etc. [...]
We could maybe just do it with two hardcoded options taking up the top/bottom third of the screen, depending on where the prompt window is? I don't know what's best here

I actually had it at the bottom the entire time and just played around with it right before the screen shot :laughing: . Generally, I'd keep it as many of the things as simple as possible with sane defaults and maybe purposefully limited configuration as we then otherwise reinvent which-key, but I'll have to see how complex some of the things get.

l-kershaw:
> we should probably also only show the current mode as in whether it's called from insert or normal mode to not clutter the hints with identical mappings across modes, but this can be decided later easily and made optionally.

I really like this idea, it will make things much less overwhelming for users. Maybe we could make it clear what we are displaying within the title of the popup?

> > custom actions [...]  good to give users a way of providing a name for their own actions that they attach to pickers
> 
> Yes :) I've envisioned pretty much like you've laid it out here though this will be part of #1060. I've actually looked into this particular thing last night but I haven't come up with a really elegant solution to adequately amend `actions/mt.lua`. Happy to discuss it over there or on gitter.

I haven't really looked at the multi-select actions tbh. With my test setup (if I recall correctly 😅) it only required changes to `mappings.execute_keymap` to allow passing a table instead of a function. Although since I stupidly deleted that setup, that is only a recollection and may not work once the multi-select has been refactored.

> I feel like popup.nvim could use some more love (I'd like to see docs 😅 ) as well but I see you are on the case :)

I agree, there are a few things on my list for popup.nvim that should be relatively easy to implement. I'm trying to not do too many PRs at once to avoid conflicts between them. Slow and steady is better I think 🙂 

> > noautocmd
> 
> Yes, in the WIP it's surely needed, but maybe it can be avoided somehow? I'll try.

I think this will end up being needed, as else creating a window will set off the "leaving prompt buffer" autocmd, which closes the whole picker. It also really should be default for `popup.nvim` as that's how vim's popups work (it's on my todo list to sort a PR for this).

> > separator line
> 
> I'd say optional; which-key as the main reference doesn't have it (at least with standard setup).

That's fair, it's not particularly important as long as we get the spacing right 🙂 

> I actually had it at the bottom the entire time and just played around with it right before the screen shot 😆 (e: I remember why, because cursor prompt was showing through float haha) . Generally, I'd keep it as many of the things as simple as possible with sane defaults and maybe purposefully limited configuration as we then otherwise reinvent which-key, but I'll have to see how complex some of the things get.

I agree, this isn't something that really needs fine grained configuration. To me the simplest setup would be two layouts for the help window, a top one and a bottom one, each taking up maybe 30-40% of the height and the full width (maybe with padding). Then we check if the active prompt window is in the top/bottom half of the screen, and place the help window on the other side. This avoids your issue with the prompt cursor showing through the float.
One thing we should remain aware of is that no matter how we setup layout, there will be someone with a small enough terminal window to not fit all the mappings in. I don't think that dealing with this via scrolling or something is necessary for this PR, but we may end up having to sort this down the line.

fdschmidt93:
> I haven't really looked at the multi-select actions tbh. With my test setup (if I recall correctly sweat_smile) it only required changes to mappings.execute_keymap to allow passing a table instead of a function. Although since I stupidly deleted that setup, that is only a recollection and may not work once the multi-select has been refactored.

Off the top of my head, I must admit I'm a bit skeptical that it's just that (to solve everything all the issues around that) but I'll check then, thanks :)

> One thing we should remain aware of is that no matter how we setup layout, there will be someone with a small enough terminal window to not fit all the mappings in. I don't think that dealing with this via scrolling or something is necessary for this PR, but we may end up having to sort this down the line.

What about the following (now implemented): we try to fit all the keymaps in 40% (configurable) of the `vim.o.lines` (I think is the right thing to look here, please correct me if I'm wrong) which is placed as you've proposed and if not we replace the last entry with `...`? If it can be shown with less rows and the number of columns that fit in width, we do it with less.

Then we primarily only would have to migrate to popups, I think :)

lua/telescope/actions/init.lua
```diff
@@ -805,6 +805,213 @@ actions.cycle_previewers_prev = function(prompt_bufnr)
...
+    end
+  end)
+
+  local entry_width = #opts.column_padding
+    + opts.mode_width
+    + opts.keybind_width
+    + opts.name_width
+    + (3 * #opts.separator)
+  local num_total_columns = math.floor((vim.o.columns - #opts.column_indent) / entry_width)
+  opts.num_rows = math.ceil(math.min(#mappings / num_total_columns, opts.max_height * vim.o.lines))
```
l-kershaw, line 938: I think it would be better to use `resolver.resolve_height` here, as it offers more configurability for users, and does roughly the same thing.

lua/telescope/actions/init.lua
```diff
@@ -805,6 +805,213 @@ actions.cycle_previewers_prev = function(prompt_bufnr)
...
+  local total_available_entries = opts.num_rows * num_total_columns
+
+  km_buf = vim.api.nvim_create_buf(false, false)
+  vim.api.nvim_buf_set_name(km_buf, "_TelescopePickersKeymap")
+
+  --   place top or bottom
+  local prompt_win = action_state.get_current_picker(prompt_bufnr).prompt_win
+  local prompt_row = vim.api.nvim_win_get_position(prompt_win)[1]
+  local prompt_pos = prompt_row <= 0.4 * vim.o.lines
```
l-kershaw, line 947: I think it is best to use `0.5` instead of `0.4` here.

lua/telescope/actions/init.lua
```diff
@@ -805,6 +805,213 @@ actions.cycle_previewers_prev = function(prompt_bufnr)
...
+  local keymap_highlights = vim.api.nvim_create_namespace "telescope_keymaps"
+  local highlights = {}
+  for index, mapping in ipairs(mappings) do
+    local row = utils.cycle(index, opts.num_rows) - 1 + opts.line_padding
+    local prev_line = vim.api.nvim_buf_get_lines(km_buf, row, row + 1, false)[1]
+    if index == total_available_entries then
+      local new_line = prev_line .. "..."
+      vim.api.nvim_buf_set_lines(km_buf, row, row + 1, false, { new_line })
+      break
+    end
```
l-kershaw, line 977: With this implementation, if you have exactly the same number of mappings as available spaces, you don't see all of them. Maybe add a check here if it is the last mapping?

lua/telescope/mappings.lua
```diff
@@ -30,6 +30,7 @@ mappings.default_mappings = config.values.default_mappings
       ["<C-q>"] = actions.send_to_qflist + actions.open_qflist,
       ["<M-q>"] = actions.send_selected_to_qflist + actions.open_qflist,
       ["<C-l>"] = actions.complete_tag,
+      ["<C-z>"] = actions._toggle_registered_actions,
```
l-kershaw, line 33: I don't think that this is a good choice of mapping for this, as it is already used by neovim to "suspend". Could we use a mapping involving `?`?

lua/telescope/mappings.lua
```diff
@@ -56,6 +57,7 @@ mappings.default_mappings = config.values.default_mappings
 
       ["<C-u>"] = actions.preview_scrolling_up,
       ["<C-d>"] = actions.preview_scrolling_down,
+      ["<C-z>"] = actions._toggle_registered_actions,
```
l-kershaw, line 60: As above.

l-kershaw (commented):
Overall looks good!

- I will try to get a PR for popup `noautocmd` done this week.
- As you mentioned its hard to pass `opts` to configure this. I'm not sure whatthe best way to sort this would be, although the current setup seems reasonable. @Conni2461 any thoughts?

lua/telescope/mappings.lua
fdschmidt93: Default mapping will be removed once PR is final. `<C-?>` eg cannot be mapped because, if I'm not mistaken, the terminal doesn't send the keycodes you'd expect. If we'd want a default mapping, could you try if `<C-_>` does what you'd expect? Might be my German keyboard though which is set to US layout.

lua/telescope/actions/init.lua
fdschmidt93: :sweat_smile:

lua/telescope/actions/init.lua
fdschmidt93: Ok, no strong feelings here.

l-kershaw: My thinking is that if this is a check for the prompt being in the top half of the screen, it should check the top *half*. Probably wouldn't make a difference most of the time, but makes the functionality easier to understand IMO.

lua/telescope/actions/init.lua
fdschmidt93: Not so familiar with that code, would that be something like the below?

```lua
  opts.num_rows = math.ceil(
    math.min(#mappings / num_total_columns, resolver.resolve_height(opts.max_height)(_, _, vim.o.lines))
  )
```

l-kershaw: This looks good to me 👍

lua/telescope/actions/init.lua
fdschmidt93: My apologies, I'm too tired, 1am my current time zone, I shouldn't be replying 😅 I initially skimmed to quickly and thought you were referring to the default value of `opts.max_height`, which I would not have had strong feelings to change.

Here, it's just a typo, you are of course right. Thanks for reiterating :)

l-kershaw: No worries 🙂

fdschmidt93:
Changes beyond comments:
- For now using `<C-/>` as default keymap (which is closest thing to ? without having a common character)
- I've excluded the `_toggle_registered_actions` from the keymaps as it'll almost always first because of leading `_` and is not insightful information as you have to know the map to see the hint; happy to keep it if it's preferred that way

lua/telescope/mappings.lua
tjdevries: we can just make it normal mode "?" probably, right? That seems like a good mapping.

lua/telescope/actions/init.lua
```diff
@@ -805,6 +805,213 @@ actions.cycle_previewers_prev = function(prompt_bufnr)
   actions.get_current_picker(prompt_bufnr):cycle_previewers(-1)
 end
 
+--- Display the keymaps of registered actions similar to which-key.nvim.<br>
+--- - Notes:
+---   - Convenience function to provide configuration options at setup and not meant to be called directly.
+---   - See |actions._toggle_registered_actions| for full documentation.
+---@param opts table: options to pass to toggling registered actions
+actions.toggle_registered_actions = function(opts)
```
tjdevries, line 813: what is the purpose of this function?

I think it should be called something like `create_...` or some other name that shows you are returning a function, not doing the action.

tjdevries:
Thought I just had.

We can create a new `actions.generate` or `actions.bind` (or similar) module that can be used to generate action functions that have certain unique configuration (like the one in this module). They all take `opts` and generate an `action`.

fdschmidt93:
> Thought I just had.

Apologies, if I understand correctly what you are referring to I just wanted to elaborate on naming of the function.

Ok, will do.

tjdevries:
Sorry, I was typing fast :laughing: 

I'm saying we should create a new module `lua/telescope/actions/generate.lua` that every function in this module has the format of `action_generate.create_X(opts)`, which will return a new function that is compatible with every other action that you can get from `telescope.actions`

lua/telescope/mappings.lua
fdschmidt93: I guess discouraged since -
- we currently mirror normal/insert mode mappings for any built-in mapping (and insert mode is default for most users.. a bit unergonomic to switch to normal mode for a built-in action)
- we default to `only_show_current_mode = true` here as mappings are duplicated in float across modes b/c of mirroring; I guess we could also merge them in "n/i" -> "..." -> "...", which would look inconsistent with custom actions though

I'm only the devils advocate here though :) happy to just map it to "?" in normal mode

fdschmidt93:
I'll move this back to draft because other stuff needs to happen first -- outstanding items:
- [ ] Plenary popup needs `noautocmd` and https://github.com/nvim-lua/plenary.nvim/pull/195 (to have annotated border also at bottom of float, when hints are shown on top)
- [ ] #1093 such that (registered) user actions are also shown in float
- [ ] Do we want a default mapping?
- [ ] New `lua/telescope/actions.generate.lua` for actions that take `opts` due to more expansive config

lua/telescope/mappings.lua
```diff
@@ -75,9 +77,14 @@ local get_next_id = function()
...
 
-local assign_function = function(prompt_bufnr, func)
+local assign_function = function(prompt_bufnr, func, mode, keybind)
   local func_id = get_next_id()
 
+  if type(func) == "table" then
+    func = vim.deepcopy(func)
+    func.mode = mode
+    func.keybind = keybind
+  end
```
Conni2461, line 87: Hmmm stupid idea, no idea if that work, might not be even worth to pursuit.
How about we dont deepcopy the action but rather use `vim.api.nvim_buf_get_keymap(prompt_bufnr, "n")`

We then get rhs the action command as string. First int is `prompt_bufnr`, second is `func_id`. With `__TelescopeKeymapStore[prompt_bufnr][func_id]` we then can get the action.

Conni2461: Idk might be a dumb idea. Just poped in my head :laughing:

Conni2461:
Stupid question. But why not use the previewer. We can cycle between different preview functions. Which means we could also jump to a specific preview function. 

If this part was already discussed here, i am sorry. A lot of discussion here :laughing: and i only skimmed though it

fdschmidt93:
One reason not to use the previewer is to keep current view on results/previewer visible if that informs the action you'd chose from the hints popup. And I guess the hints popup feels familiar from whichkey-nvim.

lua/telescope/mappings.lua
fdschmidt93: deepcopying might not be needed but I wanted to make sure to avoid side effects. Will look into it.

Conni2461: might not be needed but we dont want to change the action reference and add this to the actions in `telescope.actions` module.

Another point is, and maybe tj needs to revert this. In #987 tj will remove the `vim.deepcopy` override which will probably break this.

https://github.com/nvim-telescope/telescope.nvim/pull/987/files#diff-0811ecf01c0cf6a5e4ed485cee93d1e3194003ff3794de1be96317d93e219906L1

maybe maybe not idk right now. gotta go.

Just things that popped in my head :laughing:

fdschmidt93: Yes, the first point is why I had it. Thanks for the important pointer, I'll look at this :) until dependencies are merged to plenary will take more time anyways.

fdschmidt93: Refactored to follow your approach. Let me know if `utils.get_registered_mappings` is ok for you like that. Func ref is untouched now. Wondering though whether upstream should incorporate Telescope's vim.deepcopy overwrite. Without it, yes, this would have broken.

fdschmidt93:
This is ready for review. Since I've changed how keymaps are collected, any `action_mt.transform_mod` mapping will also be picked up.

For the time being, default mappings are as follows trying to mimic our current default that insert and normal mode mappings mirror where possible. 
Default mappings -> insert mode: `<C-/>`, normal mode: `?`. 

Happy to go with something else. It's important to have insert and normal mode mappings though as we default to only showing mappings of current mode to not show many mappings twice/multiple times.

tami5:
great work, I'm going to test it out in a hour.

lua/telescope/actions/init.lua
```diff
@@ -805,8 +809,209 @@ actions.cycle_previewers_prev = function(prompt_bufnr)
...
+        if string.find(a.nvim_buf_get_name(buf), bufname) then
+          table.insert(ret, buf)
+        end
+      end
+    end
+    return ret
+  end
+
+  -- close on repeated keypress
+  local km_bufs = find_keymap_buf()
```
tami5, line 846: perhaps better than setting a variable that is going to be used once.
```suggestion
  local km_bufs = (function() 
    local ret = {}
    local bufs = a.nvim_list_bufs()
    for _, buf in ipairs(bufs) do
      for _, bufname in ipairs { "_TelescopeWhichKey", "_TelescopeWhichKeyBorder" } do
        if string.find(a.nvim_buf_get_name(buf), bufname) then
          table.insert(ret, buf)
        end
      end
    end
    return ret
  end)()
```

lua/telescope/mappings.lua
```diff
@@ -75,9 +77,14 @@ local get_next_id = function()
   return _mapping_key_id
 end
 
-local assign_function = function(prompt_bufnr, func)
+local assign_function = function(prompt_bufnr, func, mode, keybind)
   local func_id = get_next_id()
 
+  -- if type(func) == "table" then
```
tami5, line 83: No longer needed?

tami5, line 83: No longer needed?

tami5 (commented):
Great work @fdschmidt93 . It works and is written very well. I fully agree with
the default mappings. I think it makes a prefect sense.

if nothing else is pending I think this pr should be merge. I'm going to keep
testing it for the next hour, if anything comes up I live a comment.

Thanks @fdschmidt93, this help us explore the library of pickers we have :laugh:

lua/telescope/actions/init.lua
fdschmidt93: makes sense :) fixed

lua/telescope/mappings.lua
fdschmidt93: Thanks for catching that :sweat_smile: apologies :)

fdschmidt93:
Thanks @tami5 :)

lua/telescope/actions/init.lua
```diff
@@ -805,8 +809,207 @@ actions.cycle_previewers_prev = function(prompt_bufnr)
...
+  a.nvim_win_set_option(km_win_id, "winblend", opts.winblend)
+
+  vim.cmd(
+    "autocmd BufLeave <buffer> ++once lua pcall(vim.api.nvim_win_close, "
+      .. km_win_id
+      .. ", true); pcall(vim.api.nvim_win_close, "
+      .. km_opts.border.win_id
+      .. ", true); pcall(vim.api.nvim_buf_delete, "
+      .. km_buf
+      .. ", {force=true})"
```
Conni2461, line 958: same here. utils.buf_delete should be prefered. Also can we use string.format. Its kinda unreadable :laughing:

lua/telescope/actions/init.lua
```diff
@@ -805,8 +809,207 @@ actions.cycle_previewers_prev = function(prompt_bufnr)
...
+  if opts.close_with_action then
+    vim.schedule(function()
+      vim.cmd(
+        "autocmd User TelescopeKeymap ++once lua pcall(vim.api.nvim_win_close, "
+          .. km_win_id
+          .. ", true); pcall(vim.api.nvim_win_close, "
+          .. km_opts.border.win_id
+          .. ", true); pcall(vim.api.nvim_buf_delete, "
+          .. km_buf
+          .. ", {force=true})"
```
Conni2461, line 1005: See above

lua/telescope/actions/utils.lua
```diff
@@ -78,4 +78,26 @@ function utils.map_selections(prompt_bufnr, f)
...
+  local array = {}
+  for i in string.gmatch(str, "%d+") do
+    table.insert(array, tonumber(i))
+  end
+  return array[nth]
+end
+
+--- Utility to collect mappings of prompt buffer in array of `{mode, keybind, name}`.
+---@param prompt_bufnr number: The prompt bufnr
+function utils._get_registered_mappings(prompt_bufnr)
```
Conni2461, line 91: Why prefix with `_`? I am fine with having this as stable api without prefix or doing two `_` prefix so it doesnt get exported with docgen

lua/telescope/mappings.lua
```diff
@@ -75,7 +77,7 @@ local get_next_id = function()
   return _mapping_key_id
 end
 
-local assign_function = function(prompt_bufnr, func)
+local assign_function = function(prompt_bufnr, func, mode, keybind)
```
Conni2461, line 80: we no longer need to accept mode, keybind here. Also no longer need to pass it in down below

lua/telescope/utils.lua
```diff
@@ -22,6 +22,9 @@ end
 utils.get_default = function(x, default)
   return utils.if_nil(x, default, x)
 end
+utils.cycle = function(i, n)
```
Conni2461, line 25: empty line between functions

Conni2461 (commented):
LGTM. I just left some nits. But really good job here. I trust tami and you that this works. I havent tested it

Edit: Oh for the future. Can we only hide comments that are offtopic? Even if they are already "resolved". Sometimes it helps to even skim thought those. Idk how other maintainers feel about that just a thought right now :laughing:

lua/telescope/utils.lua
fdschmidt93: Second time, I'm close to raising an issue on stylua :laughing:

Conni2461: stylua never deleted a line for me :thinking:

fdschmidt93: I meant that I want stylua to add a line back like black for Python would to fix my twitching fingers

levouh:
@fdschmidt93 I swear I saw it before but don't see it anymore; isn't there a separate _PR_ to have this supported for any user-added mapping (per the `attach_mappings`) by making everything from there an `action`? As far as I can tell, this isn't supported by this _PR_ alone, right?

fdschmidt93:
There still are some issues around `action_mt.transform_mod` (custom actions defined like that cannot be combined with built-in actions, will eventually get fixed), though something like this this will be picked up:

```lua
local action_mt = require "telescope.actions.mt"

local custom_actions = action_mt.transform_mod {
  print_entry = function()
    print(vim.inspect(action_state.get_selected_entry()))
  end,
}

require "telescope".setup {
  defaults = {
    mappings = {
      i = {
        ["<C-x>"] = custom_actions["print_entry"],
      },
    },
  },
}
```

levouh:
Yet seemingly not through `attach_mappings` still, right? E.g.

```
local telefiles = {}

local action_mt = require("telescope.actions.mt")
local teleconfig = require("user.pack.config.telescope.config")
local userutil = require("user.lib.util")

local custom_action = action_mt.transform_mod({
  yank_entry = function(...)
    P({...})
  end,
})

telefiles.find_files = function(char)
  local opts = {
    hidden = true,
    search_dirs = { userutil.get_dir_from_char(char) },
    cwd = vim.fn.get_cwd(),
      teleconfig.add_map("all", map)

      map("i", "<C-y>", custom_action["yank_entry"])

      return false
    end,
  }

  require("telescope.builtin").find_files(opts)
end

return telefiles
```

The action itself still works, but doesn't show anything when pressing `<C-?>` for `<C-y>`.

fdschmidt93:
My man, read the documentation more carefully :) https://github.com/nvim-telescope/telescope.nvim#picker

```lua

function my_custom_picker(results)
  pickers.new(opts, {
    prompt_title = 'Custom Picker',
    finder = finders.new_table(results),
    sorter = sorters.fuzzy_with_index_bias(),
      -- Map "<cr>" in insert mode to the function, actions.set_command_line
      map('i', '<cr>', actions.set_command_line)

      -- If the return value of `attach_mappings` is true, then the other
      -- default mappings are still applies.
      --
      -- Return false if you don't want any other mappings applied.
      --
      -- A return value _must_ be returned. It is an error to not return anything.
      return true
    end,
  }):find()
end
```

because you `return false` in `attach_mappings`  the default mappings are not applied.

Btw, just kidding. I looked into the source code myself and couldn't find what's wrong until I realized you are returning `false` what I haven't seen before :laughing: You'd have to add the mapping back specifically.

Just to avoid ambiguity. Insert-mode mapping is `<C-/>` rather than `<C-?>` to not having to press shift.

levouh:
That mapping is added, read my comment more carefully.

>The action itself still works, but doesn't show anything when pressing <C-?> for <C-y>.

Pressing `<C-?>` (I understand it is `<C-/>`, but I am talking "phonetically") works just as intended, but `<C-y>` doesn't show up in that menu (which is my main question; how to get custom mappings to show up in the `whichkey` popup). I have a set of default mappings added via my own wrapping (the function call to `teleconfig.add_map("all", map)`).

fdschmidt93:
That's seems to be a bug with `actions.which_key` and `attach_mapping` where `return false`. This is something for our issue template (with a reproducible example to begin with) as opposed to providing convoluted examples that are neither easily grokked or reproduced. I was trying to make friendly banter after a long day of answering issues in hopes this was an easy oversight :facepalm:  Few people are taking their time to answer loads of issues/review PRs on a daily basis.

> The action itself still works, but doesn't show anything when pressing <C-?> for <C-y>.

Initially reads like `<C-y>` is working as intended when glancing over your `attach_mappings` with not knowing what `teleconfig` is supposed to do.

Anyways, please raise an issue with a reproducible example. The below for instance works for me.

```lua
local action_mt = require "telescope.actions.mt"

custom_actions = action_mt.transform_mod {
  print = function()
    print(vim.inspect(action_state.get_selected_entry()))
  end,
}

require "telescope.builtin".find_files {
    map("i", "<C-_>", actions.which_key)
    map("i", "<C-y>", custom_actions.print)
    return false
  end,
}
```

levouh:
I'll have to mess around and see what is up then, I agree that this isn't the place to troubleshoot something like this nor was the example particularly good to start with. I appreciate the time you spend working on features like this for _Telescope_ and I believe we were both being a bit too snarky on this fine Monday. Apologies.

levouh:
It actually works with something that _isn't_ `<C-y>`, as this is already mapped otherwise (shown via a call to `nvim_buf_get_keymap`):

```
...
{
  buffer = 11,
  expr = 1,
  lhs = "<C-Y>",
  lnum = 0,
  mode = "i",
  noremap = 1,
  nowait = 1,
  rhs = 'v:lua.cmp.utils.keymap.listen.expr("i", "<lt>C-y>")',
  script = 0,
  sid = -10,
  silent = 1
},
...
```

even though the designated mapping for `<C-y>` via _Telescope_ (using your example above) _does_ still work. The `nvim-cmp` mapping (which is returned above) is an `expr` mapping, hence why it still _works_ (returns `<C-y>` if the conditional criteria [mostly just `pumvisible() == 1`] fails). As for why it doesn't show up in the `whichkey` popup seems relatively straightforward; the `func` key for the return from `get_registered_mappings` ends up being `nil`, which fails to meet the conditional criteria for it to show up, etc. Not sure that there's a beautiful way around it, but just in case you ever run into this in the future @fdschmidt93.

fdschmidt93:
If my tests are correct (it's early in the morning so I can look at this before work ;) that happens because nvim-cmp attaches keymaps to a buffer on `InsertEnter`. That's why they also supersede telescope mappings because they are applied afterwards.

1. [Trigger autocmd](https://github.com/hrsh7th/nvim-cmp/blob/b6b15d5f6e46643462b5e62269e7babdab17331c/lua/cmp/init.lua#L124)
2. [Attach mappings on autocmd](https://github.com/hrsh7th/nvim-cmp/blob/b6b15d5f6e46643462b5e62269e7babdab17331c/lua/cmp/core.lua#L168)

because I've remembered not having this issue on nvim-compe when I wrote the feature (an looking often at outputs from `vim.api.nvim_buf_get_keymap`). In other words, this is an upstream issue. On nvim-compe, users just defined their on mappings globally themselves, which were easily superseded by buffer-local mappings.

PR #1094, opened 2021-08-08
fix: picker config

Conni2461:
doesn't work correctly if there is a `require("telescope.builtin")` before `telescope.setup`


p00f:
This works, thanks!

Issue #1095, opened 2021-08-09
Whether parameter query is supported ？


liuxian1990:
For example：

:Telescope find_files search_text='hello'

results：

![img](https://raw.githubusercontent.com/liuxian1990/webimages/2a36d7716169c7e71e9852cea8b93b9e60ee0d85/2021-08-09_20-04.png)

fdschmidt93:
Does `:Telescope find_files default_text=hello` work as intended for you?

liuxian1990:
> Does `:Telescope find_files default_text=hello` work as intended for you?
> 
> 如何: Telescope find _ files default _ text = hello work as intended for you？

Yes, thank you

liuxian1990:
I implemented a simple function

![](https://raw.githubusercontent.com/liuxian1990/webimages/master/Peek%202021-08-10%2010-22.gif)

Issue #1042, opened 2021-07-23
Search for file then method/line


johans:
First, thanks for this awsome project! I only miss this one feature described below.

Sublime Text and recently also VS Code has this really nice feature which lets you go directly to a function/method or line number via the quick open meny using the search patterns:
- "partOfFileName@function" and 
- "partOfFileName:linenumber"

What's also nice is that it supports previewing as you type which is really useful if you just want to look something up really quick. 

Is something like this possible to implement in Telescope?  Currently I first search and open the file then use `lsp_document_symbols({prompt_prefix=' ', default_text=' :method: '})` to filter on methods. This works but would be so cool if it could be improved into one action.

Attached a small video showing the feature:

![sublime](https://user-images.githubusercontent.com/1887/74606417-83084b80-50d0-11ea-9bf1-f02bc442f66d.gif)

fdschmidt93:
I'll admit its buried deep in `telescope/pickers.lua` but something like this is possible with a few workarounds.

```lua
local actions = require "telescope.actions"

file_and_symbol = function()
  require("telescope.builtin").find_files {
    attach_mappings = function(prompt_bufnr, _)
      actions.select_default:enhance {
        post = function()
          require("telescope.builtin").lsp_document_symbols()
          -- not sure why this is needed,
          -- but lsp_document_symbols otherwise doesn't properly start in insert mode ;)
          vim.cmd [[normal! A]]
        end,
      }
      return true
    end,
    -- on_input_filter_cb is ran before finding/sorting, so any char "should" be safe
    on_input_filter_cb = function(prompt)
      -- this hack is required since actions cannot be run from within here
      if prompt:sub(#prompt) == "@" then
        local key = vim.api.nvim_replace_termcodes("<CR>", true, false, true)
        vim.api.nvim_feedkeys(key, "i", false)
      end
    end,
  }
end
```

You can play around with it and see if you get a nicer solution with `on_input_filter_cb`, but the coroutine always complained when I wanted to call `actions` within `on_input_filter_cb`

johans:
@fdschmidt93 Amazing, it works really well! Thank you so much for this!

fdschmidt93:
@johans 

This came up again in gitter.

You can amend the above solution with something like this in your telescope- find_files picker setup to preview and goto line

```lua
local actions = require "telescope.actions"
local action_state = require "telescope.action.state"

require"telescope".setup{
  pickers =
    find_files = {
[...]

PR #1097, opened 2021-08-10
feat: api for clearing completion cb in userspace

fdschmidt93:
I guess it's orthogonal enough to #987 to be able to merge it right away. Hope there's nothing more to it :sweat_smile: 

More generally, some lower-lever picker functions (specifically, `default_text`, `on_complete`, or that some options can also be passed to pickers individually like `file_ignore_patterns`)  useful to end users are not yet documented though we should document these options at least after #987 or as part of #1024.

Examples: #1095, #1042, frecency's usage of `on_input_filter_cb`, and gitter discussion from which this PR evolved

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -970,6 +970,10 @@ function Picker:register_completion_callback(cb)
   table.insert(self._completion_callbacks, cb)
 end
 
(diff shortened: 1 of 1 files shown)


tjdevries:
Sweet, thanks :)

Issue #1068, opened 2021-08-01
Builtin LSP goto definition won't work in clojure_lsp


glyh:
### Description

I'm trying to use the commmand ":Telescope lsp_definitions", only to find it won't work. 

**Expected Behavior**

Cursor jump to the definition of current symbol. 

**Actual Behavior**

Nothing happens. 

### Details

<details><summary>Reproduce</summary>

1. 
```shell 
mkdir foo
cd foo
git init # This is for helping lspconfig to recognize the project root. 
touch test.clj
```
2. write the following to test.clj
```clj
(defn x [] 1)
(defn y [] (x))
```
3. `nvim -u init.lua`, where init.lua is: 
```lua
require('packer').startup(
{function()
  -- Packer can manage itself
  use 'wbthomason/packer.nvim'
  use 'neovim/nvim-lspconfig'
  use {
    'nvim-telescope/telescope.nvim',
    requires = {{'nvim-lua/popup.nvim'}, {'nvim-lua/plenary.nvim'}},
    commit = "b742c50bf119d7c442a553cc52e8afde0086ae02"
  }
end, config={git={default_url_format="https://hub.fastgit.org/%s"}}})

require('lspconfig').clojure_lsp.setup({})
```
4. type `:PackerSync`, after updating, restart with `nvim -u init.lua test.clj`

5. put cursor on x, type `:lua vim.lsp.buf.definition()`, it successfully jumped to definition of x. 

6. put cursor on x, type `:Telescope lsp_definitions`, nothing happens. 
</details>

<details><summary>Environment</summary>

- nvim --version output: 
```
NVIM v0.5.0
Build type: Release
LuaJIT 2.1.0-beta3
[...]

PR #1101, opened 2021-08-11
Builtin `list_or_jump` now handles non list replies from the server.

Linouth:
This PR fixes the bug described in issue #1068 

The `list_or_jump` function could only handle responses in a list format, however servers can also send a single result without surrounding table. This resulted in the jump function not working on some servers, such as zls and clojure_lsp.


Conni2461:
Thanks :)

PR #1112, opened 2021-08-13
Change the default sorter for live_grep to empty()

amenonsen:
Since live_grep passes its entire input text to rg (or well, the program
in vimgrep_arguments), there's nothing to fuzzy-search by, and the empty
sorter seems a more appropriate default, to present the results as-is.

Note that the default generic_sorter does in fact leave the results alone here, though perhaps more by coincidence than by design. I stumbled on this because the `fzy` native extension DOES change the order of the results—but in my opinion, live_grep is a special case and empty() is conceptually the correct default for it (which can, of course, still be overriden on request).

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -103,7 +103,7 @@ files.live_grep = function(opts)
     prompt_title = "Live Grep",
     finder = live_grepper,
     previewer = conf.grep_previewer(opts),
-    sorter = conf.generic_sorter(opts),
+    sorter = conf.empty_sorter(opts),
   }):find()
 end
 
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -360,6 +360,7 @@ local telescope_defaults = {
     ]],
   },
 
+  empty_sorter = { sorters.empty },
   generic_sorter = { sorters.get_generic_fuzzy_sorter },
   prefilter_sorter = { sorters.prefilter },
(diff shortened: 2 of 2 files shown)


fdschmidt93:
If we want `live_grep` to be essentially `rg` then I think we want to include a `opts.use_regex = true` then I think lines 47 and 83 in `files.lua` should be 
```lua
  opts.use_regex = utils.get_default(opts.use_regex, true)
```
```lua
prompt = opts.use_regex and prompt or escape_chars(prompt)
```

Is that what the user would expect? On the case to form an opinion on that ;)

clason:
I think @tjdevries async jobs rewrite of `live_grep` makes changes in this direction, too -- not sure he's pushed all of them yet? 

#987 at least does swap out the `generic_sorter` with the `highlighter_only` sorter, which may be more appropriate than `empty` since you (or at least I) still want the search match highlighted in the results list?

Conni2461:
What clason says here. #987 is what we want, introduction of highlighter_only + regex.
Also this PR is about to be merge anyways, so i would close this one. 

Sorry :)

Conni2461:
With about to be merged, i mean the plan was this week, so maybe even today :)

clason:
I'll believe it when I see it ;)

(Honestly excited, though!)

amenonsen:
Thank you for providing more context.

Conni2461:
You can already try 987. I am using it for the past ~2 weeks and besides the counter is out of screen, i haven't been running into issues

PR #1115, opened 2021-08-13
Feat: refine results action

fdschmidt93:
Putting this up for discussion. 

Actual use cases if someone has `telescope-fzf-native`:
- fuzzy refining live grep
- dynamic lsp workspacye symbols

TO DO/DECIDE:
- [ ] What `opts`, if not all, should be configurable? If all, results can also be refined via #1051 (done in stash) though with visible, unavoidable redraw from new picker
- [ ] Why is the prompt prefix not immediately updated?

Currently sorter, keeping prompt/prefix/multi selection & showing prev prompt in prefix are configurable. Would need `actions.generate` from #1062 to override defaults.

Preview: shows prompt is not immediately updated while actual refinement is instant

https://user-images.githubusercontent.com/39233597/129395845-a44278b3-2e92-4343-b653-5e5093d835b7.mp4


kylo252:
I'm really happy that would be available soon, much better than my naïve implementation that we briefly discussed over on gitter/matrix last week.

Some notes: 
I couldn't figure out how to add a highlight group for the previous queries. I'm guessing we can't keep the old one since it will become inconsistent once the results have been filtered, so it needs to be reapplied dynamically instead. Any ideas regarding that?

lua/telescope/actions/init.lua
```diff
@@ -805,6 +807,47 @@ actions.cycle_previewers_prev = function(prompt_bufnr)
...
+    current_picker.sorter:_init()
+  end
+
+  local new_finder = finders.new_table { results = results, entry_maker = entry_maker }
+
+  if not opts.reset_multi_selection then
+    opts.multi = current_picker._multi
+  end
+
+  if opts.prompt_to_prefix then
```
kylo252, line 842: This could be useful

```lua
if opts.prompt_to_prefix and action_state.get_current_line() ~= "" then
---
```

fdschmidt93: Yes, good point :) I was just always entering somthing in the prompt in my tests 😅

fdschmidt93:
> I couldn't figure out how to add a highlight group for the previous queries. I'm guessing we can't keep the old one since it will become inconsistent once the results have been filtered, so it needs to be reapplied dynamically instead. Any ideas regarding that?

I'm sorry, I don't entirely follow. The prompt can be augmented with the preceding queries and highlighted. Do you mean highlighting _each_ previous query individually to visually separate them? If so, yeah we could maybe do that, shouldnt be too hard but gets funny with hl groups I guess.

kylo252:
> Do you mean highlighting _each_ previous query individually to visually separate them? If so, yeah we could maybe do that, shouldnt be too hard but gets funny with hl groups I guess.

Yeah that's my intention. 
I was thinking we could test it with a different color and see if it still makes sense to show. My use-case is that it would be easier to spot which filter is doing better in terms of refining. As in: how *close* is the keyword to the original query, and so on.

fdschmidt93:
> I was thinking we could test it with a different color and see if it still makes sense to show.

As in, test whether there might be a previous query that has intermittently become "unnecessary" to achieve the same results and if so, hide/remove it or unify highlighting groups for it? 

> My use-case is that it would be easier to spot which filter is doing better in terms of refining.

I guess adds to the previous question; what solution would you envision? A full test, if I understand it correctly, would require testing all combinations of tokens (of course not completely.. but it's complex :) in they query (where token is the string that stems from pressing refine).

> how close is the keyword to the original query, and so on.

Can you please pinpoint closeness? 

I feel like I'm crazily overthinking what you want to achieve and keep re-reading it :sweat_smile:  I'm sorry, I end up thinking it's either trivial or very difficult. You can also ping me directly on Gitter on this, maybe this is easier if we chat directly (though travelling tomorrow til Monday).

Conni2461:
merging without docs for now, into dev branch

PR #1116, opened 2021-08-15
extend finders.new_table to support function returning result table

AckslD:
With this change `finders.new_table` now supports passing it a function which gets called each time to return a table of results which as before gets passed to a `entry_maker`. This is fully backwards compatible except for importing from `telescope.finders.async_static_finder`.


lua/telescope/finders/async_function_finder.lua
```diff
@@ -0,0 +1,53 @@
...
+      if entry then
+        entry.index = k
+        table.insert(results, entry)
+      end
+    end
+    return results
+  end
+
+  return setmetatable({
+    results = {'results'},  -- TODO
```
AckslD, line 35: Should this just be an empty table, or should maybe this key just not exist at all in this case? Not sure it it's actually accessed somewhere?

fdschmidt93, line 35: Not 100% sure, but could you please check what `results` in `:lua print(vim.inspect(TelescopeGlobalState[vim.api.nvim_get_current_buf()].picker.finder))` looks like? Likely just `{'results'}`. 

If so, would 
```lua
    results = {}
    close = function() end,
  }, {
    __call = void(async(function(_, _, process_result, process_complete)
      results = make_results()
```

Fix it? The idea is `finder.results` should hold the table, though probably _in practice_ doesn't matter as results are properly hand over to the `EntryManager`.

AckslD: I'll try it but would that actually work since is that actually the same `results`? I haven't worked much with metatables before but is there a notion of "self"? Ie if it should be something like `self.results = make_results()`?

AckslD: Ah just read now that the first argument is the table itself, fixed in latest commit

fdschmidt93: Yep, you are right. Elegant solve :)

fdschmidt93:
LGTM :) 

@Conni2461 happy to get a second set of eyes on this as it's lower level.

Conni2461:
technically it looks fine but it will delay #987 because tj needs to rebase. We move to `plenary.async` because `plenary.async_lib` is deprecated.
Also why the file rename?
Another thing is, you had a split before. We might actually wanna split it, similar to `job_finder` and `oneshot_job`. So that we have a clear separation. Idk i find it rather weird that the finder behaves completely differently if we pass in a function compared to a table.

Conni2461:
Also we have another different between this and `job_finder` we probably dont want. `job_finder` get s additionally the prompt passed in so it can generate the results based on the prompt. We might just also wanna do this here. For completion sake.

It would also be cool if you could share a justification or a snippet what you try to achieve :)

lua/telescope/finders/async_table_finder.lua
```diff
@@ -0,0 +1,64 @@
...
+
+local function make_results(results_maker, entry_maker)
+  local input_results = results_maker()
+  local results = {}
+  for k, v in ipairs(input_results) do
+    local entry = entry_maker(v)
+
+    if entry then
+      entry.index = k
+      table.insert(results, entry)
```
Conni2461, line 16: `results[k] = entry` is faster than `table.insert`

Conni2461:
Also also we currently already have something like this i think. Should have thought before writing :rofl: 

Maybe take a look at https://github.com/nvim-telescope/telescope.nvim/blob/f1a27baf279976845eb43c65e99a71d7f0f92d02/lua/telescope/builtin/lsp.lua#L312-L340

I also wrote this snippet right now to test it. Currently you need to put `async` before your function but that will also change in #987

```lua
local pickers = require "telescope.pickers"
local finders = require "telescope.finders"
local conf = require("telescope.config").values

local a = require "plenary.async_lib"
local async = a.async

pickers.new(opts, {
  prompt_title = "colors",
  finder = finders.new_dynamic {
    fn = async(function(prompt)
      return { prompt }
    end),
    entry_maker = require("telescope.make_entry").gen_from_string {},
  },
}):find()
```

AckslD:
Hey @Conni2461! Ah okay haha :) I think I missed `new_dynamic` since it didn't have a "doc-string". I think we can then close this. I'm anyway happy since I learned how metatables works in the process :)

PR #1117, opened 2021-08-15
fix: for telescope.symbols

### LINKED ISSUE #1072: Actions.insert doesn't behave as expected
I am playing around with `Telescope symbols` and noticed the following unexpected behavior when using it in insert mode (which is the most common use, I'd say):

1. `:inoremap <c-s> <cmd>Telescope symbols<cr>`
2. type `foobar`
3. place the cursor between `foo` and `bar`
4. trigger `<c-s>` and insert any symbol, say, `X`

**Expected**: `fooX|bar` in insert mode

**Actual**: `foob[X]ar` in normal mode

(Also, wouldn't a better place for sources be in `.local/share/nvim/telescope/symbols` rather than under `.config`? I feel there could be a general "data cleanup" for `telescope_history` and `file_frecency.sqlite3` as well...)

Conni2461:
Closes #1072

- keep insert
- also look in `.local/share/nvim/telescope/symbols/*.json` for symbols
  can be overriden with `symbol_path`

The only thing that is currently bothering me is that you end up in insert mode even if you started in normal mode. 
- Done this with actions `:enhance` good interface

CC @clason better late than never :rofl:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -560,12 +560,22 @@ builtin.planets({opts})                                    *builtin.planets()*
 
 
 builtin.symbols({opts})                                    *builtin.symbols()*
-    Lists symbols inside of data/telescope-sources/*.json found in your runtime
-    path. Check README for more info
+    Lists symbols inside of `data/telescope-sources/*.json` found in your
+    runtime path or found in `stdpath("data")/telescope/symbols/*.json`. The
+    second path can be customized. We provide a couple of default symbols which
+    can be found in https://github.com/nvim-telescope/telescope-symbols.nvim.
+    This repos README also provides more information about the format in which
+    the symbols have to be.
 
 
     Parameters: ~
         {opts} (table)  options to pass to the picker
+
+    Fields: ~
+        {symbol_path} (string)  specify the second path. Default:
+                                `stdpath("data")/telescope/symbols/*.json`
+        {sources}     (table)   specify a table of sources you want to load
+                                this time
 
 
 builtin.commands({opts})                                  *builtin.commands()*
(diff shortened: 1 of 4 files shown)


Conni2461:
clason should be ready for testing if you find some time :) 
We will end up in the mode we started. This is currently only happening for `symbols` but maybe we should actually write a interface for that at some point.
Symbols can also be put in `stdpath("data")/telescope/symbols/*.json` now, more information in the docs :)

clason:
Hmm, that still puts the symbol _before_ the cursor if it is at the end of a line?

clason:
Also, in normal mode the cursor moves, which I don't think is the expected behavior? Or maybe it is (behaving like `p`)?

Conni2461:
Late night programming is never good :rofl: 

This is kinda annoying we wanna do two different things based on if we started in normal mode or if we started in insert mode. No one is currently using this so i can technically do what i want: https://sourcegraph.com/search?q=context:global+insert_symbol+lang:lua&patternType=literal

You can try the lastest commit which should work, (`P` in normal). Maybe i do two actions one for when starting in normal mode, which is the implementation we have in master and a new action that should be triggered if started in insert mode.

clason:
Also, don't drink and develop! :)

That seems to work now, thank you! Yes, two different (internal) mappings sounds reasonable.

Conni2461:
only coffee is allowed while coding :)

yeah its `p`, i dont know  what i am even saying anymore :rofl: so this pr shouldn't change the current normal mode behavior but fixes insert mode.

clason:
@Conni2461 anything left you want to do here? Or should I go ahead and merge?

Conni2461:
I think this should be good.

PR #1120, opened 2021-08-18
fix: (terminal) buffer previewer

### LINKED ISSUE #714: Colorize terminal buffer preview

It would be awesome to colorize terminal buffer preview for more visibility.

Current previewer does not invoke buffer related autocommands maybe for performance perspective. Could we invoke some color related autocommands only when the buffer is terminal?

fdschmidt93:
Would close #714 by introducing a new previewer that reuses existing buffers.

@Conni2461 Idk, personally a bit torn on this. While `previewers.new_buffer_previewer` feels backward for `builtin.buffers` that we already have, though now ofc the buffer for `actions.delete` has to be replaced prior to deletion and the solution generally feels bloat. I'm not sure there are other side effects (that's why I abstained touching `new_buffer_previewer`, as we'd have conflicting logic ala creating buffers vs using existing ones).

I'm personally fine with terminal buffers not being highlighted ;)

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -846,7 +846,9 @@ internal.keymaps = function(opts)
       end)
       return true
     end,
-  }):find() end internal.filetypes = function(opts)
+  }):find()
+end
+internal.filetypes = function(opts)
   local filetypes = vim.fn.getcompletion("", "filetype")
 
   pickers.new(opts, {
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -644,7 +644,7 @@ internal.buffers = function(opts)
       action_set.select:enhance {
         post = function()
           local entry = action_state.get_selected_entry()
-          vim.api.nvim_win_set_cursor(0, { entry.lnum, entry.col })
+          vim.api.nvim_win_set_cursor(0, { entry.lnum, entry.col or 0 })
         end,
       }
     end,
@@ -834,6 +834,7 @@ internal.keymaps = function(opts)
     end,
   }):find()
 end
+
 internal.filetypes = function(opts)
   local filetypes = vim.fn.getcompletion("", "filetype")
 
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -578,6 +578,17 @@ internal.reloader = function(opts)
 end
 
 internal.buffers = function(opts)
+  -- TODO upstream issue? if chosen buffer was not yet attached to window, minimal style of previewer is used
+  local winopts = {
+    ["number"] = vim.api.nvim_win_get_option(0, "number"),
+    ["relativenumber"] = vim.api.nvim_win_get_option(0, "relativenumber"),
+    ["cursorline"] = vim.api.nvim_win_get_option(0, "cursorline"),
+    ["signcolumn"] = vim.api.nvim_win_get_option(0, "signcolumn"),
+    ["spell"] = vim.api.nvim_win_get_option(0, "spell"),
+    ["wrap"] = vim.api.nvim_win_get_option(0, "wrap"),
+    ["list"] = vim.api.nvim_win_get_option(0, "list"),
+    ["colorcolumn"] = vim.api.nvim_win_get_option(0, "colorcolumn"),
+  }
   local bufnrs = filter(function(b)
     if 1 ~= vim.fn.buflisted(b) then
       return false
@@ -640,6 +651,17 @@ internal.buffers = function(opts)
     previewer = previewers.buffers.new(opts),
     sorter = conf.generic_sorter(opts),
     default_selection_index = default_selection_idx,
+    attach_mappings = function(_, _)
+      action_set.select:enhance {
+        post = function()
+          local entry = action_state.get_selected_entry()
+          for opt, value in pairs(winopts) do
+            vim.api.nvim_win_set_option(0, opt, value)
+          end
+          vim.api.nvim_win_set_cursor(0, { entry.lnum, entry.col })
+        end,
+      }
+    end,
   }):find()
 end
 
@@ -824,10 +846,7 @@ internal.keymaps = function(opts)
       end)
       return true
     end,
-  }):find()
-end
-
-internal.filetypes = function(opts)
+  }):find() end internal.filetypes = function(opts)
   local filetypes = vim.fn.getcompletion("", "filetype")
 
   pickers.new(opts, {
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -578,17 +578,6 @@ internal.reloader = function(opts)
 end
 
 internal.buffers = function(opts)
-  -- TODO upstream issue? if chosen buffer was not yet attached to window, minimal style of previewer is used
-  local winopts = {
-    ["number"] = vim.api.nvim_win_get_option(0, "number"),
-    ["relativenumber"] = vim.api.nvim_win_get_option(0, "relativenumber"),
-    ["cursorline"] = vim.api.nvim_win_get_option(0, "cursorline"),
-    ["signcolumn"] = vim.api.nvim_win_get_option(0, "signcolumn"),
-    ["spell"] = vim.api.nvim_win_get_option(0, "spell"),
-    ["wrap"] = vim.api.nvim_win_get_option(0, "wrap"),
-    ["list"] = vim.api.nvim_win_get_option(0, "list"),
-    ["colorcolumn"] = vim.api.nvim_win_get_option(0, "colorcolumn"),
-  }
   local bufnrs = filter(function(b)
     if 1 ~= vim.fn.buflisted(b) then
       return false
@@ -655,9 +644,6 @@ internal.buffers = function(opts)
       action_set.select:enhance {
         post = function()
           local entry = action_state.get_selected_entry()
-          for opt, value in pairs(winopts) do
-            vim.api.nvim_win_set_option(0, opt, value)
-          end
           vim.api.nvim_win_set_cursor(0, { entry.lnum, entry.col })
         end,
       }
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -637,7 +637,7 @@ internal.buffers = function(opts)
       results = buffers,
       entry_maker = opts.entry_maker or make_entry.gen_from_buffer(opts),
     },
-    previewer = conf.grep_previewer(opts),
+    previewer = previewers.buffers.new(opts),
     sorter = conf.generic_sorter(opts),
     default_selection_index = default_selection_idx,
   }):find()
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -875,7 +875,60 @@ previewers.buffers = defaulter(function(opts)
       return "Buffers"
     end,
     setup = function(_, status)
-      return { ["winid"] = status.preview_win }
+      local win_id = status.picker.original_win_id
+      -- required because of see `:h local-options` as
+      -- buffers not yet attached to a current window take the options from the `minimal` popup ...
+      local state = {
+        ["previewed_buffers"] = {},
+        ["winid"] = status.preview_win,
+        ["win_options"] = {
+          ["colorcolumn"] = vim.api.nvim_win_get_option(win_id, "colorcolumn"),
+          ["cursorline"] = vim.api.nvim_win_get_option(win_id, "cursorline"),
+          ["foldlevel"] = vim.api.nvim_win_get_option(win_id, "foldlevel"),
+          ["list"] = vim.api.nvim_win_get_option(win_id, "list"),
+          ["number"] = vim.api.nvim_win_get_option(win_id, "number"),
+          ["relativenumber"] = vim.api.nvim_win_get_option(win_id, "relativenumber"),
+          ["signcolumn"] = vim.api.nvim_win_get_option(win_id, "signcolumn"),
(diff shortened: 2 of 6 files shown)


Conni2461:
I wouldn't touch `new_buffer_previewer` for this, i had something like this in mind when mentioning that i need to restructure it from in #714, then i ignored it for 4 and half months because i dont use previewer for `builtin.buffers` :rofl: 
Originally i was to lazy to write another base Previewer so i just used `vimgrep` and it worked till someone mentioned "special" buffers and i just introduced a silly workaround. We can remove those which probably will make PR smaller, so i dont actually have a problem doing this PR :+1: 
We can delete this https://github.com/nvim-telescope/telescope.nvim/blob/f1a27baf279976845eb43c65e99a71d7f0f92d02/lua/telescope/previewers/term_previewer.lua#L297-L300 and this https://github.com/nvim-telescope/telescope.nvim/blob/f1a27baf279976845eb43c65e99a71d7f0f92d02/lua/telescope/previewers/buffer_previewer.lua#L373-L378

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -872,4 +872,39 @@ previewers.display_content = defaulter(function(_)
...
+  return Previewer:new {
+    title = function()
+      return "Buffers"
+    end,
+    dyn_title = function(_, entry)
+      return Path:new(from_entry.path(entry, true)):normalize(cwd)
+    end,
+    preview_fn = function(self, entry, status)
+      if vim.api.nvim_buf_is_valid(entry.bufnr) then
+        vim.api.nvim_win_set_buf(status.preview_win, entry.bufnr)
```
Conni2461, line 887: I think you need to make sure that buffer is loaded. So `if vim.api.nvim_buf_loaded ==false then vim.fn.bufload` or so

For example if you load a session, all buffers are unloaded, so you will end up with no highlighting or something like that.

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -872,4 +872,39 @@ previewers.display_content = defaulter(function(_)
...
+      end
+
+      local input = direction > 0 and [[]] or [[]]
+      local count = math.abs(direction)
+
+      vim.api.nvim_buf_call(self.state.bufnr, function()
+        vim.cmd([[normal! ]] .. count .. input)
+      end)
+    end,
+  }
```
Conni2461, line 907: We should reuse this. So local function and then just setting it in `new_buffer_previewer` and here.

Conni2461:
otherwise LGTM thanks for picking up my slack :)

lua/telescope/previewers/buffer_previewer.lua
Conni2461: Okay just tested we dont need to. In the past in ran into this but maybe that was bug upstrem and `nvim_win_set_buf` now does ensure that. So better ignore that comment here :rofl:

lua/telescope/previewers/buffer_previewer.lua
fdschmidt93: This is a lot more funny than I've thought. It has the side effect that it scrolls the actual buffer in the background when pressing <c-u> for `builtin.buffers` :laughing: I've refactored `scroll_fn` to scroll on the window as opposed to the buffer, which has _another_ funny effect. This is relatively easy to make consistently.

fdschmidt93:
There is a crazy stupid issue with this I've so far just overlooked in my tests (please confirm if you have the time). If we use actual buffers for the preview, the fact that the popup style is `minimal` messes the buffer look once attached to the new window.

This does not happen if you select a buffer that is currently not opened in the window the picker was launched from but was opened in that window with the right settings before. Fade me.

lua/telescope/previewers/buffer_previewer.lua
Conni2461: `nvim_win_call` works for me

lua/telescope/builtin/internal.lua
```diff
@@ -637,9 +648,20 @@ internal.buffers = function(opts)
...
     sorter = conf.generic_sorter(opts),
     default_selection_index = default_selection_idx,
+    attach_mappings = function(_, _)
+      action_set.select:enhance {
+        post = function()
+          local entry = action_state.get_selected_entry()
+          for opt, value in pairs(winopts) do
+            vim.api.nvim_win_set_option(0, opt, value)
+          end
+          vim.api.nvim_win_set_cursor(0, { entry.lnum, entry.col })
```
Conni2461, line 661: this will break if `previewer = false` is set. i think

And then we dont actually need any to do anything

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -872,4 +867,35 @@ previewers.display_content = defaulter(function(_)
...
+previewers.buffers = defaulter(function(opts)
+  opts = opts or {}
+  local cwd = opts.cwd or vim.loop.cwd()
+  return Previewer:new {
+    title = function()
+      return "Buffers"
+    end,
+    setup = function(_, status)
+      return { ["winid"] = status.preview_win }
+    end,
```
Conni2461, line 879: We can just set `self.state.winid` in `preview_fn` every time. I dont really see a reason to give setup access to status

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -872,4 +867,35 @@ previewers.display_content = defaulter(function(_)
...
+    preview_fn = function(self, entry, status)
+      if vim.api.nvim_buf_is_valid(entry.bufnr) then
+        vim.api.nvim_win_set_buf(status.preview_win, entry.bufnr)
+        vim.api.nvim_win_set_option(status.preview_win, "winhl", "Normal:TelescopePreviewNormal")
+        vim.api.nvim_win_set_option(status.preview_win, "signcolumn", "no")
+        vim.api.nvim_win_set_option(status.preview_win, "foldlevel", 100)
+        vim.api.nvim_win_set_option(status.preview_win, "wrap", false)
+        self.state.bufnr = entry.bufnr
+        if not entry.col then
+          local _, col = unpack(vim.api.nvim_win_get_cursor(status.preview_win))
```
Conni2461, line 892: Current master sets `pcall(vim.api.nvim_buf_add_highlight, bufnr, ns_previewer, "TelescopePreviewLine", lnum - 1, 0, -1
` to indicate the current line, we should also do something similar here.

Conni2461:
> There is a crazy stupid issue with this I've so far just overlooked in my tests (please confirm if you have the time). If we use actual buffers for the preview, the fact that the popup style is minimal messes the buffer look once attached to the new window.

Good old weird bugs, i have not missed you

lua/telescope/previewers/buffer_previewer.lua
fdschmidt93: I've actually just refactored, such that `setup` and `teardown` take care of setting the window options (as opposed to in `builtin.buffers`) accordingly for which I would need status to get `original_win_id`. Fine to keep status here then?

lua/telescope/builtin/internal.lua
fdschmidt93: Yeah, I've refactored it to use `previewer.setup` and `previewer.teardown` already. See the other comment whether you are fine with this.

lua/telescope/previewers/buffer_previewer.lua
fdschmidt93: Agree we should, but I have to think more about how to. `nvim_buf_add_highlight` cannot be used, since it's highlighting the buffer of course in previewer and background :laughing:

lua/telescope/previewers/buffer_previewer.lua
Conni2461: then yeah

lua/telescope/previewers/buffer_previewer.lua
Conni2461: i am kinda glad i skipped this issue, seems like a lot of work for terminal highlighting

fdschmidt93: I think this is not impossible with `nvim_win_set_decoraton_provider` :laughing:

fdschmidt93:
Actually, it's intended behavior.. Just not really nice one as it's been probably set without floats in mind.

> When editing a buffer that has been edited before, the options from the window
that was last closed are used again.  If this buffer has been edited in this
window, the values from back then are used.  Otherwise the values from the
last closed window where the buffer was edited last are used.
It's possible to set a local window option specifically for a type of buffer.
When you edit another buffer in the same window, you don't want to keep
using these local window options.  Therefore Vim keeps a global value of the
local window options, which is used when editing another buffer.  Each window
has its own copy of these values.  Thus these are local to the window, but
global to all buffers in the window. 

`:h local-options`

lua/telescope/previewers/buffer_previewer.lua
fdschmidt93: Yeah, I've got it working with `nvim_win_set_decoration_provider`. I require Github Sponsors from anyone who upvoted the original issue :laughing:

fdschmidt93:
Well, that escalated quite a lot. It's done though it's quite complicated for a fix. While I cannot think of anything wrong with this right now, I'll test this more tomorrow or next week I suppose.

Complexity mostly stems from two things:

- Low-level api `vim.api.nvim_set_decoration_provider` needed to set buffer-window local extmarks -- it's unclear to me from docs whether this is adequately cleared but I think I cannot do more to assure this
- `:h local-options` -- if a buffer hasn't yet been attached to a window, it will carry over the minimal layout from the popup -- doing my best to recover from this, but maybe fillchars (which seem fine in my tests) or `EndOfBuffer` (doesn't seem to be cleared as per `:h vim.api.nvim_open_win`) are still acting up?

lua/telescope/builtin/internal.lua
Conni2461: Yeah its still broken if `previewer = false`. 
`E5108: Error executing lua ...lugged/telescope.nvim/lua/telescope/builtin/internal.lua:647: Argument "pos" must be a [row, col] array`

We do not have `col` by default just do `entry.col or 1`.Or dont enhance the action if `opts.previewer = false`

lua/telescope/builtin/internal.lua
```diff
@@ -826,7 +834,6 @@ internal.keymaps = function(opts)
     end,
   }):find()
 end
-
```
Conni2461, line 829: we are missing a newline :sob:

Conni2461 (changes requested):
LGTM but its still broken if `previewer = false` see one of the old conversations are opened it again.

Thanks :)

lua/telescope/builtin/internal.lua
fdschmidt93: What have I done? :scream: and why does neither luacheck or stylua check complain (and ofc my stylua not take care of this)

lua/telescope/builtin/internal.lua
fdschmidt93: Yeah, that was just a stupid oversight on my part, sorry; I've burnt a bit on this PR and overread in coop with the other issue :sweat_smile: 

Will fix later.

fdschmidt93: `column` in `nvim_win_set_cursor` is zero-indexed, so `entry.col or 0`; I also mix up zero and one indexing across functions more often than I'd like :laughing:

fdschmidt93:
Don't merge this yet.. if previewer is false, the window options are not applied :facepalm:

Conni2461:
> Don't merge this yet.. if previewer is false, the window options are not applied facepalm

Window options are applied just fine for me and i dont even see why this should be a problem. `previewer = false` hasn't changed in this PR really besides the additional action doing the a second jump. Works in master and this PR just fine

Conni2461:
Oh but buffer deletion is broken

fdschmidt93:
> Window options are applied just fine for me and i dont even see why this should be a problem. previewer = false hasn't changed in this PR really besides the additional action doing the a second jump. Works in master and this PR just fine

Yes, you're right. It cannot be an issue because the preview window is never created which could overwrite the buffer-window local options. But in one state I saw the options not applied and I cannot reproduce this now :laughing: I'm too fried.

fdschmidt93:
`action.delete_buffer` is now fine. Window options seem to be working as intended as well but I don't like having to wrap it in `builtin.buffers` itself.

fdschmidt93:
I've also tested whether `nvim_set_decoration_provider` is cleared by `nvim_buf_clear_namespace` and it's not. So if a user would ever create a window with the same id as preview window (s)he would have the cursorline highlighted with `TelescopePreviewLine`. Travelling this weekend, so I'll look into it thereafter.

fdschmidt93:
@Conni2461 could you please give this another whirl? `previewer=false` and `action.delete_buffer` are fixed. Sorry, couldn't test myself before you've took a crack at it :sweat_smile: 

I've asked on nvim gitter about clearing decoration provider and didn't get an answer (don't wanna ping bfredl, but I'm sure the circle of people to answer this definitively is very exclusive right now :laughing: )

I opened the buffer previewer 30 times in a session and virtually nothing happened in htop. I've minimized callbacks from `set_decoration_provider` as you'll see in [diff](https://github.com/nvim-telescope/telescope.nvim/pull/1120/commits/691be2a156d088d35baf7cf6aad8f554da7d1dc9) of final commit via returning `false` in `on_start` once previewer is closed.

Conni2461 (approved):
I think this is good. Thanks :)

~~And i also think this might fix https://github.com/nvim-telescope/telescope.nvim/issues/1145 which happens since today but not with this PR, no idea why or how this PR should affect this~~

Conni2461:
Oh wait it doesn't happen here because this branch is not yet rebased. So 1145 is a actual bug we need to address. Looking into it

PR #1124, opened 2021-08-18
fix(command parser): discard invalid lua expressions

l-kershaw:
Fix issue from [this](https://github.com/nvim-telescope/telescope.nvim/pull/1075#issuecomment-901371701) comment.

@gizmo385 could you let me know if this fixes the issue on your end?

PR #1128, opened 2021-08-19
fix(command parser): evaluate valid lua expressions

l-kershaw:
Found another mistake immediately after we merged #1124, sorry guys 😅
Currently `master` doesn't work if you use lua syntax for tables.

---

I feel like we should have some sort of tests for the command parser to avoid this happening in the future, but not really sure how to set it up. Is there a good way of testing if a picker has loaded correctly with the current testing framework?
@tjdevries / @Conni2461 any ideas?

### DIFF
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -104,7 +104,7 @@ local function convert_user_opts(user_opts)
           user_opts[key] = nil
         elseif select("#", assert(eval)()) == 1 and type(assert(eval)()) == "table" then
           -- allow if return a single table only
-          user_opts[key] = eval
+          user_opts[key] = assert(eval)()
         else
           -- otherwise return nil (allows split check later)
           user_opts[key] = nil


Conni2461:
What we could do is mock the actual function we run with a test function that checks if the values are set correctly. Idk see here https://github.com/nvim-lua/plenary.nvim/blob/master/TESTS_README.md#mocking-with-luassert never done mocking in lua tho and i think tj as well. Someone else wrote that file

Sorry for merging to early :rofl: you can merge this if its ready

l-kershaw:
@Conni2461 no worries, I was gonna merge straight away anyway, so would have still needed a second PR to fix 🤣

I tried to figure out what mocking is, but I don't really understand it. Hopefully there is a simpler way to test this.

We could maybe export the `convert_user_opts` function, and then test that this is working as expected.
This has the advantage that we don't need to worry about checking if the picker loaded correctly.
But it feels a bit weird to export such an internal facing function.

What are your thoughts?

jonathannerat:
@l-kershaw with lua syntax for tables you mean something like `find_command=fd,-t,f`? I tried to run this today and it fails because `loadstring` inside `convert_user_opts` parses `return fd,-t,f` successfully but eventually fails to run because it's invalid (fd, t and f are not variables).

Maybe we could wrap whatever `loadstring` returns with another pcall to check that?

PR #1134, opened 2021-08-20
fix(command_parser): pcall for the loadstring

l-kershaw:
Yet another fix for the command parser 😅 

Fixes [this comment](https://github.com/nvim-telescope/telescope.nvim/pull/1128#issuecomment-902682662)

@jonathannerat could you have a look/test this to make sure it works on your machine?

PR #1142, opened 2021-08-20
fix(command_definition): default to nil on empty <f-args>

### LINKED ISSUE #1132: Command :Telescope errors out since today.
### Description

After updating my plugins today when I run :Telescope I get the error
```
E5107: Error loading lua [string ":lua"]:1:1 unexpected symbol near ')'
```
And Telescope fails to open.
All `:Telescope whatever_finder` commands work.
This was working as expected before I updated today so I suspect has something to do with #1128 or #1123 ?

### Neovim version

v0.5.0

Manjaro XFCE

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. :Telescope
Screenshot_2021-08-20_14-34-25

### Expected behavior

Telescope should open.

### Actual behavior

Telescope fails to open and we get an error.

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

jonathannerat:
Should close #1132

When no arguments are passed to a command, `<f-args>` is simply removed from the definition, which causes `:Telescope` command to fail because of bad syntax (it calls `lua require('telescope.command').load_command(<line1>, <line2>, <count>, )`)

With `unpack({<f-args>})`, if no arguments are provided it returns `nil` by default, and then when arguments are provided, they are expanded correctly

### DIFF
--- a/plugin/telescope.vim
+++ b/plugin/telescope.vim
@@ -111,4 +111,4 @@ function! s:telescope_complete(arg,line,pos)
 endfunction
 
 " Telescope Commands with complete
-command! -nargs=* -range -complete=custom,s:telescope_complete Telescope    lua require('telescope.command').load_command(<line1>, <line2>, <count>, <f-args>)
+command! -nargs=* -range -complete=custom,s:telescope_complete Telescope    lua require('telescope.command').load_command(<line1>, <line2>, <count>, unpack({<f-args>}))


l-kershaw (approved):
Thanks, this is a very clever solution!
I came to the same conclusion as you about the problem but couldn't figure out a robust way of fixing it.

I have tested this alongside #1134, and all of the issues I am aware of seem to be fixed.
I will merge both in a moment.

PR #1093, opened 2021-08-08
fix: registering custom actions

fdschmidt93:
Closes #684 

This fixes registering custom actions by unifying the lookup (`mod` vs `registered_actions` in `action_mt`) and adding the custom action subsequently to `actions`, which allows nice things as illustrated in docs.

How does it work?
- Rather than keeping an inaccessible copy of actions in the function-local `mod` of `action_mt.create` in `actions_mt`, we now maintain an explicit copy of registered actions on which `run_replace_or_original` can rely and which new actions can be added to 
- `mt of actions/mt.lua` is now instantiated once and set as the metatable also for newly registered actions such that `actions._clear` works when a metatable is instantiated
- Newly registered actions are added to `actions` which already is the `redirect` of `transform_mod` in `actions_mt` (ie has the right metatables) once an action can be registered; this adds the nicety that actions can be called via `actions.my_cool_custom_action` anywhere after registration

**old tests regarding actions having to point to the same `mt` below**

-----

One thing I've run into and haven't fully wrapped my head around yet: edit -> That's because `actions._clear` does not affect `actions.print_entry` in the above example due to differing metatables, which is now fixed by instantiating `mt` in `actions/mt.lua` once and letting all actions have the same original metatable such that `actions._clear` works for newly instantiated actions just as well.

```lua
local actions = require "telescope.actions"
local action_state = require "telescope.actions.state"

local print_entry = actions.register_actions({
  print_entry = function()
    print(vim.inspect(action_state.get_selected_entry()))
  end,
  print_ordinal = function()
    print(vim.inspect(action_state.get_selected_entry()))
  end,
})[1]

require"telescope".setup {
    mappings = {
      i = {
        -- now fixed
        -- why does this work and the others below do not? E: this should not work; picker clears actions upon call
        ["<C-v>"] = actions.print_entry:enhance({post = actions.select_vertical}),
        -- of course maybe not super sensible, though neither of the below works (also not on master) -- should they?
        ["<C-v>"] = actions.toggle_selection:enhance({pre = actions.print_entry}),
        ["<C-v>"] = actions.toggle_selection:enhance({post = actions.select_vertical}),
        ["<C-v>"] = actions.toggle_selection:enhance({post = function() print(action_state.get_selected_entry() end}),
        -- addition variants all work ie actions.toggle_selection + actions.print_entry
      }
    }
  }
```

Update: these work as intended on both master (where applicable) and this branch

```lua
R("telescope.builtin").find_files {
[...]

PR #1143, opened 2021-08-20
fix: action mt so we can again concat actions from two different tables

### LINKED ISSUE #684: custom action example in README produces error

Conni2461:
Closes #684

- without actually changing the public interface
- without having a local table that keeps track of all actions

I could write more tests for this but its Friday evening and it was a long day ...

My take on https://github.com/nvim-telescope/telescope.nvim/pull/1093

### DIFF
--- a/lua/telescope/actions/mt.lua
+++ b/lua/telescope/actions/mt.lua
@@ -24,6 +24,9 @@ local append_action_copy = function(new, v, old)
   new._post[v] = old._post[v]
 end
 
+--TODO(conni2461): Not a fan of this solution/hack. Needs to be addressed
+local all_mts = {}
+
 --- an action is metatable which allows replacement(prepend or append) of the function
 ---@class Action
 ---@field _func table<string, function>: the original action function
@@ -143,6 +146,7 @@ action_mt.create = function()
     return self
   end
 
+  table.insert(all_mts, mt)
   return mt
 end
 
@@ -177,6 +181,12 @@ action_mt.transform_mod = function(mod)
   end
 
   return redirect
+end
+
+action_mt.clear_all = function()
+  for _, v in ipairs(all_mts) do
+    pcall(v.clear)
+  end
 end
 
 return action_mt
--- a/lua/telescope/actions/mt.lua
+++ b/lua/telescope/actions/mt.lua
@@ -14,6 +14,24 @@ local run_replace_or_original = function(replacements, original_func, ...)
   return original_func(...)
 end
 
+local append_action_copy = function(new, v, old)
+  table.insert(new, v)
+  new._func[v] = old._func[v]
+  new._static_pre[v] = old._static_pre[v]
+  new._pre[v] = old._pre[v]
+  new._replacements[v] = old._replacements[v]
+  new._static_post[v] = old._static_post[v]
+  new._post[v] = old._post[v]
+end
+
+--- an action is metatable which allows replacement(prepend or append) of the function
+---@class Action
+---@field _func table<string, function>: the original action function
+---@field _static_pre table<string, function>: will allways run before the function even if its replaced
+---@field _pre table<string, function>: the functions that will run before the action
+---@field _replacements table<string, function>: the function that replaces this action
+---@field _static_post table<string, function>: will allways run after the function even if its replaced
+---@field _post table<string, function>: the functions that will run after the action
 action_mt.create = function()
   local mt = {
     __call = function(t, ...)
@@ -47,23 +65,11 @@ action_mt.create = function()
     __add = function(lhs, rhs)
       local new_action = setmetatable({}, action_mt.create())
       for _, v in ipairs(lhs) do
-        table.insert(new_action, v)
-        new_action._func[v] = lhs._func[v]
-        new_action._static_pre[v] = lhs._static_pre[v]
-        new_action._pre[v] = lhs._pre[v]
-        new_action._replacements[v] = lhs._replacements[v]
-        new_action._static_post[v] = lhs._static_post[v]
-        new_action._post[v] = lhs._post[v]
+        append_action_copy(new_action, v, lhs)
       end
 
       for _, v in ipairs(rhs) do
-        table.insert(new_action, v)
-        new_action._func[v] = rhs._func[v]
-        new_action._static_pre[v] = rhs._static_pre[v]
-        new_action._pre[v] = rhs._pre[v]
-        new_action._replacements[v] = rhs._replacements[v]
-        new_action._static_post[v] = rhs._static_post[v]
-        new_action._post[v] = rhs._post[v]
+        append_action_copy(new_action, v, rhs)
       end
       new_action.clear = function()
         lhs.clear()
@@ -140,7 +146,7 @@ action_mt.create = function()
   return mt
 end
 
-action_mt.transform = function(k, mt, v)
+action_mt.transform = function(k, mt, _, v)
   local res = setmetatable({ k }, mt)
   if type(v) == "table" then
     res._static_pre[k] = v.pre
@@ -159,7 +165,7 @@ action_mt.transform_mod = function(mod)
 
   for k, v in pairs(mod) do
     local mt = action_mt.create()
-    redirect[k] = action_mt.transform(k, mt, v)
+    redirect[k] = action_mt.transform(k, mt, _, v)
   end
 
   redirect._clear = function()
--- a/lua/telescope/actions/mt.lua
+++ b/lua/telescope/actions/mt.lua
@@ -149,17 +149,22 @@ action_mt.transform = function(k, mt, v)
 end
 
 action_mt.transform_mod = function(mod)
-  local mt = action_mt.create()
-
   -- Pass the metatable of the module if applicable.
   --    This allows for custom errors, lookups, etc.
   local redirect = setmetatable({}, getmetatable(mod) or {})
 
   for k, v in pairs(mod) do
+    local mt = action_mt.create()
     redirect[k] = action_mt.transform(k, mt, v)
   end
 
-  redirect._clear = mt.clear
+  redirect._clear = function()
+    for k, v in pairs(redirect) do
+      if k ~= "_clear" then
+        pcall(v.clear)
+      end
+    end
+  end
 
   return redirect
 end
--- a/lua/telescope/actions/mt.lua
+++ b/lua/telescope/actions/mt.lua
@@ -45,28 +45,32 @@ action_mt.create = function()
     end,
 
     __add = function(lhs, rhs)
-      local new_actions = setmetatable({}, action_mt.create())
+      local new_action = setmetatable({}, action_mt.create())
       for _, v in ipairs(lhs) do
-        table.insert(new_actions, v)
-        new_actions._func[v] = lhs._func[v]
-        new_actions._static_pre[v] = lhs._static_pre[v]
-        new_actions._pre[v] = lhs._pre[v]
-        new_actions._replacements[v] = lhs._replacements[v]
-        new_actions._static_post[v] = lhs._static_post[v]
-        new_actions._post[v] = lhs._post[v]
+        table.insert(new_action, v)
+        new_action._func[v] = lhs._func[v]
+        new_action._static_pre[v] = lhs._static_pre[v]
+        new_action._pre[v] = lhs._pre[v]
+        new_action._replacements[v] = lhs._replacements[v]
+        new_action._static_post[v] = lhs._static_post[v]
+        new_action._post[v] = lhs._post[v]
       end
 
       for _, v in ipairs(rhs) do
-        table.insert(new_actions, v)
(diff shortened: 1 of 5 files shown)


fdschmidt93:
While I much prefer that this avoids `registered_actions` in my `actions_mt`, this has other problems as elaborated upon in my PR:
- You `transform_mod` a custom action and then add or enhance custom action somewhere; it's not cleared since it doesn't have the identical actions metatable for which `actions._clear` is called at picker instantiation (test at bottom)
- Custom action cannot be mapped via strings ie `[<CR>] = "my_cool_action"` or `[<CR>] = actions.my_cool_action + action.select_vertical` since it's not part of actions (see `actions.register_actions`) -- allows (a) inspecting all actions more easily and (b) more unified interface when registering multiple actions at once
- You suggested  `smart`, `multi`, `entries` versions of actions should just work, which I wanted to do via parsing `__index` of actions that automatically registers the right version of the action (see [here](https://github.com/nvim-telescope/telescope.nvim/blob/979be4a446320c649287b6514e724db8facd4df9/lua/telescope/actions/init.lua#L46), which I wanted to wrap in `actions.register_actions`)

The above two things are much more gently solved by registering custom actions from `actions` (since we are just inserting into `redirect`) combined with your approach. OK if I incorporate this in my branch to synthesize advantages? Changing the interface I think has clear advantages and `actions.transform_mod` cannot be used right now anyways. Thank you for the exemplifying :)

------
Test the first point:
```lua
local action_mt = require "telescope.actions.mt"
-- global for later inspection
custom_actions = action_mt.transform_mod({
  print_entry = function()
    print(vim.inspect(action_state.get_selected_entry()))
  end,
})
--

require 'telescope'.setup {
  defaults = {
    mappings = {
      i = {
        -- PR #1093
        ["<C-x>"] = custom_actions['print_entry'],
      },
    }
  }
}

-- later do something like thislocal actions = require "telescope.actions"
  attach_mappings = function()
    custom_actions['print_entry']:enhance { post = actions.select_vertical }
    return true
  end,
}
-- inspect custom_actions
print(vim.inspect(custom_actions))
```

Conni2461:
> elaborated upon in my PR

having read all of the conversation which probably was a mistake.

> You transform_mod a custom action and then add or enhance custom action somewhere; it's not cleared since it doesn't have the identical actions metatable for which actions._clear is called at picker instantiation (test at bottom)

i think i have a better way to solve this

> `[<CR>] = "my_cool_action"` or `[<CR>] = actions.my_cool_action + action.select_vertical`

if you have written an action i think you are able to map it via a table. the string part is only there for novice lua users who make the mistake that they forgot to require `actions`. i think this is nothing we need to support.

> Changing the interface I think has clear advantages

I still dont see the advantages and i prefer that `action_mt` is the lowers component in the stack. It allows us to easily test it in the spec file, especially if its stateless. Give me till Monday to address your concerns, in i think a better way. I do not want to block any of the other cool things :) but i already said it i have mixed feeling about your approach. I think it makes a lot more sense that the action object actually has a reference to the action function.

> Test the first point:

Btw for action_mt we talk spec file :laughing:

Conni2461:
The thing i saying is we dont need additional tables, storing the actions. We already have this.

I found a bug and this isnt tested but i am out of time for now and need to go. Its done monday

fdschmidt93:
> You suggested smart, multi, entries versions of actions should just work, which I wanted to do via parsing __index of actions that automatically registers the right version of the action (see [here](, which I wanted to wrap in actions.register_actions)

The only important thing I want to address is looking ahead to multi select actions. Could you please factor something in what I had in mind here if you agree with that? The other "advantages" would have been freebies in my suggestion. I dont wanna create new tables, just register into the right existing one. I hope it's clearer now, otherwise happy to discuss on Gitter.

Conni2461:
> The only important thing I want to address is looking ahead to multi select actions.

I am confused by this because i dont see whats needed for you. Ive read now some of your multiselection implementation and i dont see how you keep it in mind. All i see is that you plan that register forbides registering actions with the names `smart` `multi` or `entries` in the name. 

> The other "advantages" would have been freebies in my suggestion

which other advantages? I am also clearing custom actions now and i dont think how this makes even remotely sense. You have a table of custom actions but you can also say `actions.my_custom` ? why is this in the telescope action module, thats just confusing. I know whats happening here in your code but from a outside perspective i never see a reason to do that. 

> I dont wanna create new tables

I am not criticizing that you create new tables (they are cheap) i am criticizing that this would make our lowest building block stateful. So first time registering a mod will be different from any other registering mod. It is not deterministic what happens on the second, third, fourth register, unless you unload the module and rerequire it. For example i register a table with the keys `a`, `b`, `c` then i register a table with the keys `c`, `d`, and `f` then i only have 5 functions in the store because the first `c` was overriden. I have a similar problem on a smaller scale in my implementation. If i have `a.x + b.x` then we end up with a new action `<user_tbl>.x` one function from table `b`.

> otherwise happy to discuss on Gitter

I prefer to have this conversation here public on github, so if tj or someone else who is interested can come in here and have 100% of the information needed :)

fdschmidt93:
> All i see is that you plan that register forbides registering actions with the names smart multi or entries in the name.

The idea was to autogenerate these variants for any action, builtin or custom, on indexing. Registering is not in there yet but would have to be added. Of course that could be approached completely differently to begin with, eg all variants of an action are created upon creation/registering. Or we add them hardcoded.

> You have a table of custom actions but you can also say actions.my_custom

My idea was not to have a table of custom actions but unify all actions under `actions`. The freebies (which aren't necessary or anything) were "multi_custom_action" or `actions.custom_action". Again, can be viewed differently by separating builtin or custom actions. Ofc, what action can be multi etc also has to be considered.

> i am criticizing that this would make our lowest building block action_mt stateful.

Yes, understood and agree.

I've just clarified my original ideas and it doesn't really matter how we'll be doing that after all - what I wanted to propose is just one way of doing it I think is consistent.

I don't want to hold up anything here by discussing what we maybe should do, just go ahead :)

Conni2461:
I think i am done with my proposal. @tjdevries needs to make the decision now what he wants.

I've not added `actions.register_action` we could do this but i personally prefer to keep the actions module as a set of actions that can be mapped. I mentioned that on multiple occasions but i am also not against it if there is a good reason for it.

So basically can now pick bits he wants from this pr and from #1093 and i am kinda done and move on now. If you want this and you register part, fdschmidt93, feel free to integrate this in your PR. I am not really attached to it, i only really care about that bit that i explained a couple of times now. This PR mainly tries to solve #684 nothing more, nothing less.

A shortcoming here right now is that we only clear actions that are mapped. Which means replace or enhance a action that was never mapped will not be cleared. We could just always clear `actions` + `actions.set` so it doesnt happen for builtin things.

lua/telescope/actions/mt.lua
```diff
@@ -123,28 +140,35 @@ action_mt.create = function(mod)
   return mt
 end
 
-action_mt.transform = function(k, mt, mod, v)
+action_mt.transform = function(k, mt, v)
```
tjdevries, line 143: this is technically a breaking change, we could just use `_` for now if we wanted. Or just break it if we don't think anyone is really using it.

tjdevries:
From reading both I think I prefer this method.

Are there any outstanding bugs w/ this implementation as it is now?

lua/telescope/actions/mt.lua
```diff
@@ -45,18 +45,35 @@ action_mt.create = function(mod)
...
+      local new_action = setmetatable({}, action_mt.create())
       for _, v in ipairs(lhs) do
-        table.insert(new_actions, v)
+        table.insert(new_action, v)
+        new_action._func[v] = lhs._func[v]
+        new_action._static_pre[v] = lhs._static_pre[v]
+        new_action._pre[v] = lhs._pre[v]
+        new_action._replacements[v] = lhs._replacements[v]
+        new_action._static_post[v] = lhs._static_post[v]
+        new_action._post[v] = lhs._post[v]
```
tjdevries, line 56: can we write little helper function to copy these over? we do the same thing twice.

Also, have we documented what all the fields do? I don't remember if we did that or not. Could be helpful.

fdschmidt93:
@Conni2461 bumping the PR, would you please finish this when you get the chance? :) I guess it integrates extensions better with custom actions, e.g. I'll work on the file browser extension and https://github.com/nvim-neorg/neorg-telescope soon.

Conni2461:
I'll try to do it this weekend :) sorry for the slacking

lua/telescope/actions/mt.lua
```diff
@@ -14,7 +14,7 @@ local run_replace_or_original = function(replacements, original_func, ...)
   return original_func(...)
 end
 
-action_mt.create = function(mod)
+action_mt.create = function()
```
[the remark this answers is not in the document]
Conni2461: Yeah we can keep track of the mts we created but i was never really a big fan of that idea.

Also I am not really sure how likely it is that we will run into this issue. That someone replaced a action and never mapped it (this is the only action that we currently don't clear).

I wanna sleep one night about that. Worst case is i add it with a `FIXME` and we deal with it sometime else :laughing: I dont wanna block fd so i will merge it this weekend. I will run it and see if there are any issues :laughing:

Conni2461:
Will run today and merge this evening if there is no regression

Conni2461:
Should work again and i didnt run into any regression, the whole day.

We will see what happens :laughing:

PR #1146, opened 2021-08-21
fix: for default sorters how we insert data if #prompt == 0

Conni2461:
@mroavi does this fix your problems? This will make sure that initial sorting isnt touched for any sorter. 

@fdschmidt93 mentioned that he wants to work on a pr that makes element inserting more configurable so you can say always respect initial sorting and probably we can configure this correctly for some builtins.

But i would still push for this hotfix because it restores that initial default behavior for internal sorters we had prior to async_v2

### DIFF
--- a/lua/telescope/sorters.lua
+++ b/lua/telescope/sorters.lua
@@ -244,7 +244,7 @@ sorters.get_fuzzy_file = function(opts)
       if N == 0 or N < ngram_len then
         -- TODO: If the character is in the line,
         -- then it should get a point or somethin.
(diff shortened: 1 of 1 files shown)


mroavi:
Yes, I tested this fix and it indeed brings back the sorting behavior before https://github.com/nvim-telescope/telescope.nvim/commit/a97af306c4e9c9a6fa7c886c0ffe3079822c5203 was committed. Thanks a lot!

Conni2461:
i think this is good

PR #1149, opened 2021-08-22
[WIP] feat: async lsp

Conni2461:
i dont even pretend i know how `plenary.async` works but we should do this. This is probably not that good but it worked?!


lua/telescope/builtin/lsp.lua
```diff
@@ -259,44 +259,43 @@ lsp.range_code_actions = function(opts)
...
+        return
+      end
 
-  if vim.tbl_isempty(locations) then
-    print(
-      "No results from workspace/symbol. Maybe try a different query: "
-        .. "Telescope lsp_workspace_symbols query=example"
-    )
-    return
+      return locations
```
Conni2461, line 290: What we can also say that we can return is a Iterator rather than a table. That way we could return a iterator that iterates over a tag file rather than that we need to split the tag file and return a table. Open for suggestions

Conni2461:
not doing, because i actually dont care enough about it.

PR #1155, opened 2021-08-23
fix(pickers): attach_mappings on builtins.buffers needs return

ascandella:
In #1120, builtins.buffers added an `attach_mappings` function, but didn't return a value from it.

This causes an error in:
```
E5108: Error executing lua ...k/packer/start/telescope.nvim/lua/telescope/mappings.lua:178: Attach mappings must always return a value. `true` means use default mappings, `false` means only use atta
ched mappings  
```

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -665,6 +665,7 @@ internal.buffers = function(opts)
           vim.api.nvim_win_set_cursor(0, { entry.lnum, entry.col or 0 })
         end,
       }
+      return true
     end,
   }):find()
 end


Conni2461:
Thanks i missed that while testing because i have an additional mapping. Need to remember to start testing with miminal rc

fdschmidt93:
Thanks for the fix :) I'd have to check why that didn't popup in my case.

Conni2461:
@fdschmidt93 because we both have the delete buffer mapping. and we create a new attach_mappings function with the return value from this mapping.

https://github.com/nvim-telescope/telescope.nvim/blob/f45c170f2853c1c1492d3f30a950a99d30706ea2/lua/telescope/pickers.lua#L1104-L1110

at least thats my guess right now :laughing:

PR #1156, opened 2021-08-23
Buffers picker does not return true, must return something

eric-hansen:
Otherwise an error is thrown at lua/telescope/mappings.lua:178


Conni2461:
Sorry #1155 was a little bit faster

But still thanks :)

PR #1165, opened 2021-08-25
fix: initial normal mode

fdschmidt93:
`initial_mode=normal` currently has two side effects:
- `default_text` is not set, because `vim.api.nvim_feedkeys` is slower than needed (or from `initial_mode=insert` perspective, `vim.cmd[[startinsert!]]` was faster)  essentially. 
- the cursor is not at the right position: as is, the prompt is empty and the cursor is on the first offset, whereas with this PR, the prefix is shown and cursor in normal mode at first valid text position

I didn't want to meddle with `picker:set_prompt` as is, but of course the fix could be also implemented there.

Fixes last glitches before #1051 should be fine to be merged, but of course happy to get more testers/feedback on that PR :sweat_smile: .

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -370,13 +370,19 @@ function Picker:find()
 
     -- TODO(async): I wonder if this should actually happen _before_ we nvim_buf_attach.
     -- This way the buffer would always start with what we think it should when we start the loop.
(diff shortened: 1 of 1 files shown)


Conni2461 (approved):
LGTM

fdschmidt93:
thanks :)

PR #1168, opened 2021-08-26
wip: universal prefiltering

### LINKED ISSUE #1127: Feature Request: refactor prefiltering for more flexibility and use cases
As per this Reddit [comment](https://www.reddit.com/r/neovim/comments/p7cf4c/which_plugins_or_functionality_do_you_think_is/h9j0wvh?utm_source=share&utm_medium=web2x&context=3), users seek out more ways to refine and/or group their search results adequately. This issue is supposed to keep tabs on that.

Prefiltering should be refactored to enable users more customization on this. For instance, it should be easy

- to modify `cwd` of the entry in `builtin.live_grep` to filter on various levels of the path
- not only complete but be able to exclude tags
- hopefully have any easy interface to combine prefiltering sorters

our current implementation requires that the entry_maker provides the suitable `tag`. While I like this default

Other things:
- Documentation goes hand-in-hand with the improvements
- Improve efficiency. I think is it's comparably slow though unnoticed as it hasn't been stress tested enough.

**Additional context**
To that end, our current prefiltering implementation (see [here](https://github.com/nvim-telescope/telescope.nvim/blob/f67d3e883d7afafa93428a27cb1ba1c4144d9d4c/lua/telescope/sorters.lua#L583)) will have to be refactored and documentation be updated.

fdschmidt93:
Closes #1127 

Putting this wip up in case anyone already wants to chime in.

For now, there'll be  partially breaking changes with telescope-fzf-native because `clear_filter_fun` [here](https://github.com/nvim-telescope/telescope-fzf-native.nvim/blob/9fb0d2d2297f7e313abf33a80331fadf4df716a5/lua/telescope/_extensions/fzf.lua#L33) I think won't work as it is anymore :laughing: 

- Carve out filter functions in separate file
- `filter_functions` will be either function or metatables with `__call` where the latter may comprise `tags` on which can be filtered via completion. `metatable` itself now holds tags as filters can be stacked.

TODO:
- [x] Make sorter extensible, still required for fzf-native though
- [ ] Provide useful filtering functions
- [ ] Be able to combine tags
- [ ] Maybe provide way to exclude certain tags with separate delimiter.

Out-of-the-box filtering functions:
- [x] `lines`
- [x] `nvim-treesitter-textobjects` like `@function.outer`
- [ ] grouping paths with tags

### DIFF
--- a/lua/telescope/filters.lua
+++ b/lua/telescope/filters.lua
@@ -253,8 +253,8 @@ filters.treesitter = function(opts)
   end
 end
 
-filters.whitelist = function(opts)
-  assert(opts.whitelist, "Whitelist [array of string] is required!")
+filters.allowlist = function(opts)
+  assert(opts.allowlist, "Allowlist [array of string] is required!")
   opts.from_entry = utils.get_defaults(opts.from_entry, from_entry.path)
   if type(opts.from_entry) == "string" then
     local key = opts.from_entry
@@ -263,7 +263,7 @@ filters.whitelist = function(opts)
     end
   end
   return function(_, prompt, entry)
-    for _, v in ipairs(opts.whitelist) do
+    for _, v in ipairs(opts.allowlist) do
       if opts.from_entry(entry):find(v) then
         return 0, prompt
       end


tjdevries:
So this PR has two different things happening it looks like, if I'm correct. The first is tags (which we use with the `:tagname:` syntax) calculating and the second is filtering of results based on some items.

Am I correct on that?

Conni2461:
I would like to schedule this post release. I still have a couple of ideas i've been collecting in the past weeks.

For core we should focus on building an interface that can be easily extended, similar to sorter. And then we talk about what a filter needs. Does it need a init and destroy similar to sorter, because someone could write a filter in c that needs to do memory management. And then we almost a sorter.

So we could think about have multiple sorters like a chain. And a filter is just some kind of sorter that either says, yep we keep this entry or no we discard it.

And then we can have the complete stuff and can say. sorter 1 (the tag filter) owns the first part of the prompt and sorter 2 (fzf e.g.) the rest. And we do this owning stuff in the picker, no longer in the sorter. Like i had to add to make it work for fzf-native.

Just some raw thoughts for future reference.

pedromss:
Hey there. Felt the need to filter the `oldfiles` builtin by `cwd` and took that as an excuse to explore the picker by implementing the `cwd` myself (then reverted and used the `cwd_only`, obviously). I'm wondering if it would be feasible to add a `filter` option that would filter out or in anything in the results.

 As pointed out here https://github.com/nvim-telescope/telescope.nvim/pull/1168#issuecomment-916116436 this PR does seem to do filtering and more so:

1. Could the filter functionality be implemented separately?
2. Is anyone working on this?

Ideally I'd like to be able to do this with any picker:

```lua
require("telescope.builtin").find_files({
	filter = function(entry)
		return true
	end,
})
```

fdschmidt93:
> Is anyone working on this?

No, and at least I probably won't anytime soon. You're happy to have a go at it as per [Conni's Comment](https://github.com/nvim-telescope/telescope.nvim/pull/1168#issuecomment-986001845).

> Could the filter functionality be implemented separately?

Sure. You could always wrap a sorter into your own custom sorter that applies prefiltering beforehand. That's sort of what prefiltering right now is doing, albeit not that general.

Issue #1170, opened 2021-08-26
git_commits picker is extremely slow on large repositories


figsoda:
### Description

`Telescope git_commits` takes me 28 seconds to load on [nixpkgs](https://github.com/nixos/nixpkgs) master and freezes neovim completely in the mean time

### Neovim version

0.5.0

### Operating system and version

nixos unstable 5bc8b980b9178ef9a4bb622320cf34e59ea2ea10

### Steps to reproduce

```ShellSession
$ git clone https://github.com/nixos/nixpkgs
$ nvim nixpkgs
: Telescope git_commits
```

### Expected behavior

It should take less time to show the prompt and unfreeze neovim. `fzf.vim` starts up the prompt immediately with the latest commits and loads the rest in about 12 seconds for me.

### Actual behavior

see description

### ~~Minimal~~ config

```Lua
-- I don't think it's specific to config, but here's my config just in case

telescope = require("telescope")

telescope.setup({
  defaults = {
    borderchars = { "─", "│", "─", "│", "┌", "┐", "┘", "└" },
    layout_config = {
      height = 0.9,
      preview_width = 80,
      width = 0.9,
    },
    mappings = {
      i = {
        ["<c-s>"] = require("trouble.providers.telescope").open_with_trouble,
        ["<esc>"] = require("telescope.actions").close,
      },
    },
    prompt_prefix = " ",
    selection_caret = "❯ ",
  },
  pickers = {
    find_files = { hidden = true },
  },
})
telescope.load_extension("fzf")
```

PR #1176, opened 2021-08-27
fix: Should not block on git commands

tjdevries:
Closes: #1170

I didn't finish all of them yet because I didn't have time this morning, but this should make the one mentioned in the issue fast


Conni2461:
I'll merge this now. Should be good imo

PR #1183, opened 2021-08-31
Add visual selection to grep string

kdheepak:
This PR changes the default (when `opts.search` is not defined) to get the visual selection for grep_string if the user is in visual mode.

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -37,6 +37,31 @@ local escape_chars = function(string)
   })
 end
 
+local get_visual_selection = function()
+  local mode = vim.fn.mode()
+  if mode ~= "v" or mode ~= "V" or mode ~= "CTRL-V" then
+    return nil
+  end
+  vim.cmd [[visual]]
+  local _, start_row, start_col, _ = unpack(vim.fn.getpos "'<")
+  local _, end_row, end_col, _ = unpack(vim.fn.getpos "'>")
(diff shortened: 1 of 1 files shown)


fdschmidt93:
Hi there, thanks for your PR :)

We've discussed this before and the original verdict was to eventually have something like getting the visual selection in a function in core and then wrap the resulting selection string into `opts.search` for `builtin.grep_string`. See these two issues: 
* https://github.com/nvim-telescope/telescope.nvim/pull/494
* https://github.com/neovim/neovim/pull/13896

But maybe we should incorporate simple version into `builtin.grep_string` as imo users would expect that out-of-the-box.

As means of background, there are two primary issues around getting the visual selection that are difficult to solve universally:
1. multi-byte characters like German umlauts or . I apologize if not pasted correctly, I have to fix browser fonts...)

```lua
local a = "全ä"
local c = "ä a α"
```
Notice how the byte column in your vim statusline changes.

2. Virtual block mode

Not so relevant here, but a holistic solution for visual selection is particularly cumbersome because of that.

Every now and then I work on this, just to run into additional edge cases: https://github.com/fdschmidt93/dotfiles/blob/master/nvim/.config/nvim/lua/utils/visual/init.lua

I'll strive to try and pin down the last issues for nvim 0.6. Hopefully. :laughing:

kdheepak:
I should have realized this was already worked on! I saw the TODO and thought it might be low hanging fruit. Reading your PR and your comment makes me realize how wrong I was lol. 

Thanks for the detailed response! I’ll close this.

PR #1184, opened 2021-08-31
Add word to prompt title for grep string

kdheepak:
This PR adds the word to the prompt title for grep string. Example when `use` is searched for:

Screen Shot 2021-08-31 at 5 21 04 AM

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -143,7 +143,7 @@ files.grep_string = function(opts)
   end
 
   pickers.new(opts, {
-    prompt_title = "Find Word",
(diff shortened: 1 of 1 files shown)


Conni2461 (approved):
LGTM. i dont need it but maybe other core members feel differently about it. I leave it to them to merge it.

Thanks :)

fdschmidt93:
Do we want to account for the edge-case that a user is launching `builtin.grep_string` with `{search=vim.fn.input, use_regex=true}`? Then it should probably `prompt_title = "Find Word: " .. word` because parentheses then can have meaning.

I have no preference, just laying out for comprehensiveness.

kdheepak:
Thanks for merging!

PR #1191, opened 2021-09-01
perf: load builtin submodules when a picker is invoked

dhruvmanila:
This adds a layer when requiring the builtin submodules to be loaded
only when a picker is actually invoked.

Ref: https://github.com/tjdevries/lazy.nvim

Talked about this on gitter with @fdschmidt93 

This is a direct copy of the function from `lazy.nvim`, we can update the name if necessary but I don't think it should be as it is self explanatory.

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -71,10 +71,10 @@ local function require_on_exported_call(builtin_module)
   })
 end
 
-local files = require_on_exported_call("telescope.builtin.files")
-local git = require_on_exported_call("telescope.builtin.git")
-local internal = require_on_exported_call("telescope.builtin.internal")
-local lsp = require_on_exported_call("telescope.builtin.lsp")
+local files = require_on_exported_call "telescope.builtin.files"
+local git = require_on_exported_call "telescope.builtin.git"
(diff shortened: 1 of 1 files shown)


fdschmidt93 (approved):
LGTM thanks :) 

I have no feelings towards grouping modules now at the top of `builtin/init.lua` but maybe other members feel it would have been cleaner to just replace `require "telescope.builtin.{...}"` `require_on_exported_call "telescope.builtin.{...}` in place :sweat_smile:  Or is there something I'm missing?

dhruvmanila:
Oh, I did not think about that. It does make sense to keep the style similar. I will update it :)

Although, if any of the members feel that grouping the modules is better, I will revert it back

lua/telescope/builtin/init.lua
```diff
@@ -60,6 +60,17 @@ end
 
 local builtin = {}
 
+-- Ref: https://github.com/tjdevries/lazy.nvim
+local function require_on_exported_call(builtin_module)
```
tjdevries, line 64: can you change `builtin_module` to just `mod`. It's not specific to `buitin` type modules.

After that, we can merge. Thanks!

PR #1195, opened 2021-09-02
fix: should not use git log command in git stash picker

phelipetls:
This is a regression introduced not so long ago in https://github.com/nvim-telescope/telescope.nvim/commit/4d691fdc236006d6123560570af3e949a3b2c06e#diff-0b2350537ecf272d7adf81c80da8e63bd696f7b9d7f4c9f9de609a4df66107c5L82-R97.

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -87,11 +87,9 @@ git.stash = function(opts)
     finder = finders.new_oneshot_job(
       vim.tbl_flatten {
         "git",
-        "log",
-        "--pretty=oneline",
-        "--abbrev-commit",
-        "--",
-        ".",
+        "--no-pager",
+        "stash",
+        "list"
       },
       opts
     ),


lua/telescope/builtin/git.lua
```diff
@@ -87,11 +87,9 @@ git.stash = function(opts)
...
       vim.tbl_flatten {
         "git",
-        "log",
-        "--pretty=oneline",
-        "--abbrev-commit",
-        "--",
-        ".",
+        "--no-pager",
+        "stash",
+        "list"
```
tami5, line 92: fixes ci linting checking error
```suggestion
        "list",
```

tami5 (changes requested):
Oh nice catch, thanks @phelipetls.

lua/telescope/builtin/git.lua
phelipetls: I'll rebase this later

tami5 (approved):

Issue #1177, opened 2021-08-27
Suggestion for commands builtin copy command to register action


lbiaggi:
**Is your feature request related to a problem? Please describe.**
N/A
**Describe the solution you'd like**

I think the command pickers is awesome to search for something which you don't remember, especially when you are assigning a custom map in your neovim workflow, but I missed the ability to copy and paste it. I was amazed how easily I could make it knowing almost nothing about telescope stuff (just read the developer.md and project extension code) and thought that could be included or at least shared to others use it too.
**Describe alternatives you've considered**
I did this for myself, and thought that could be useful to others
```
custom_command_telescope.lua
local M = {}

local actions = require("telescope.actions")
local transform_mod = require('telescope.actions.mt').transform_mod

M.set_register = function(entry)
    local cmd = ":" .. entry.name
    vim.fn.setreg('+', cmd ,"c")
end

M.copy_to_reg = function(prompt_bufnr)
    local selected_entry = actions.get_selected_entry(prompt_bufnr)
    local value = selected_entry['value']
    M.set_register(value)
    actions.close(prompt_bufnr)
end

return transform_mod(M)
```
telescope command cfg
```
local custom_actions = require('custom_command_telescope')
...
  commands = {
    mappings = {
        n = {
            ["Y"] = custom_actions.copy_to_reg
        },
        i = {
            ["<C-Y>"] = custom_actions.copy_to_reg
        },
    },
...
```

**Additional context**

I think it could be extended to copy event definitions (pure lua bindings)

Regards,
Lucas

tami5:
Interesting 😍 , I like your solution. my lazy solution to the problem was to map `jk` to `<esc>` and then select and copy. Thanks for sharing

Conni2461:
Feel free to post a PR for that. I dont see a reason why we shouldnt ship this action :)

lbiaggi:
Okay, If you don't mind explain,  where I should do? (I'm quite lost), how am I supposed to insert a map to Y and C-Y  [here](https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/builtin/internal.lua#L278-L291) or isn't this the best place to do it?

tami5:
[...]

PR #1196, opened 2021-09-02
#1177  copy command entry to unnamedplus action

lbiaggi:
TODO:
- [x] customizable register
- [x] update builtin.commands to accept map

### DIFF
--- a/lua/telescope/actions/generate.lua
+++ b/lua/telescope/actions/generate.lua
@@ -54,7 +54,6 @@ action_generate.which_key = function(opts)
 end
 
 action_generate.copy_command_to_reg = function(opts)
-  dump(opts)
   return function(prompt_bufnr)
     actions.copy_command_to_reg(prompt_bufnr, opts)
   end
--- a/lua/telescope/actions/generate.lua
+++ b/lua/telescope/actions/generate.lua
@@ -53,4 +53,10 @@ action_generate.which_key = function(opts)
   end
 end
 
+action_generate.copy_command_to_reg = function(opts)
+  dump(opts)
+  return function(prompt_bufnr)
+    actions.copy_command_to_reg(prompt_bufnr, opts)
+  end
+end
 return action_generate
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -272,14 +272,13 @@ function actions.close_pum(_)
   end
 end
 
---- Copy current entry command to unnamed plus register
+--- Copy current entry command to a specific register or unnamed plus as default
 ---@param prompt_bufnr number: The prompt bufnr
 function actions.copy_command_to_reg(prompt_bufnr)
+    local selected = action_state.get_selected_entry()
+    local regtocopy = vim.fn.input("Register: ")
     -- concat ':' with command name entry
-    local function set_reg(entry) a.fn.setreg('+', ":" .. entry.name, "c") end
-    local selected = action_state.get_selected_entry(prompt_bufnr)
-    set_reg(selected['value'])
-    actions.close(prompt_bufnr)
+    a.fn.setreg(regtocopy ~= '' and regtocopy or "+", ":" .. selected['value'].name, "c")
 end
 
 actions._close = function(prompt_bufnr, keepinsert)
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -273,12 +273,17 @@ function actions.close_pum(_)
 end
 
 --- Copy current entry command to a specific register or unnamed plus as default
+--- Don't let user use  blackhole register (_), fallback to default
 ---@param prompt_bufnr number: The prompt bufnr
 ---@param opts table: options to pass to the action
 ---@field reg: the register which the content will be sent
 function actions.copy_command_to_reg(prompt_bufnr, opts)
   opts = opts or {} -- ensure opts is a table if nil
   opts.reg = utils.get_default(opts.reg, "+")
+  if opts.reg == _ then
+    print "Please do not use register _, falling back to default"
+    opts.reg = "+"
+  end
   local cmd = action_state.get_selected_entry()["name"]
   -- concat ':' with command name entry
   vim.fn.setreg(opts.reg, ":" .. cmd, "c")
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -278,16 +278,11 @@ end
 ---@param opts table: options to pass to the action
 ---@field reg: the register which the content will be sent
 function actions.copy_command_to_reg(prompt_bufnr, opts)
-  local validate_reg = function(reg)
-    while reg == '_' do
-      reg = vim.fn.input "Register > "
-    end
-    return reg
-  end
   opts = opts or {} -- ensure opts is a table if nil
   opts.reg = utils.get_default(opts.reg, "+")
-  opts.reg = validate_reg(opts.reg)
-
+  if opts.reg == "_" then
+    opts.reg = vim.fn.input "Register > "
+  end
   local cmd = action_state.get_selected_entry()["name"]
   -- concat ':' with command name entry
   vim.fn.setreg(opts.reg, ":" .. cmd, "c")
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -272,17 +272,15 @@ function actions.close_pum(_)
   end
 end
 
---- Copy current entry command to unnamed register
+--- Copy current entry command to unnamed plus register
 ---@param prompt_bufnr number: The prompt bufnr
 function actions.copy_command_to_reg(prompt_bufnr)
     -- concat ':' with command name entry
     local function set_reg(entry) a.fn.setreg('+', ":" .. entry.name, "c") end
-
     local selected = action_state.get_selected_entry(prompt_bufnr)
     set_reg(selected['value'])
     actions.close(prompt_bufnr)
 end
-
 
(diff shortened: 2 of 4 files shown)


fdschmidt93:
Hi there, thanks for your PR :) a couple of pointers and thoughts

- You might want to set up [stylua](https://github.com/JohnnyMorganz/StyLua) to not have to format manually (largely, stylua is not perfect). Should you have no formatter set up, I can recommend [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim)
- You can add a default mapping by changing the signature of `attach_mappings` in `builtin.commands` to `attach_mappings(prompt_bufnr, map)` and then eg `map("n", "Y", action.copy_command_to_reg)`
- Not 100% sure we want `action.close` in `action.copy_command_to_reg`; users could always compose `action.copy_command_to_reg + action.close` in `telescope.setup` for themselves. I'd personally agree it's a decent default, but it gets tedious if you _don't_ want this

lua/telescope/actions/init.lua
```diff
@@ -272,6 +272,16 @@ function actions.close_pum(_)
   end
 end
 
+--- Copy current entry command to unnamed plus register
+---@param prompt_bufnr number: The prompt bufnr
+function actions.copy_command_to_reg(prompt_bufnr)
+    -- concat ':' with command name entry
+    local function set_reg(entry) a.fn.setreg('+', ":" .. entry.name, "c") end
```
fdschmidt93, line 279: Is there a need for a closure here? I think it's cleaner to just write
```lua
a.fn.setreg('+', ":" .. entry.name, "c")
```
in line 281.

lua/telescope/actions/init.lua
```diff
@@ -272,6 +272,16 @@ function actions.close_pum(_)
...
+    local selected = action_state.get_selected_entry(prompt_bufnr)
```
fdschmidt93, line 280: `action_state.get_selected_entry()` suffices :)

lua/telescope/actions/init.lua
```diff
@@ -272,6 +272,16 @@ function actions.close_pum(_)
   end
 end
 
+--- Copy current entry command to unnamed plus register
+---@param prompt_bufnr number: The prompt bufnr
+function actions.copy_command_to_reg(prompt_bufnr)
```
fdschmidt93, line 277: Maybe we should make the register customizable by adding a `reg` parameter that defaults to `+` because (i) users might want to have it in another reg to begin with and (ii) [os differences](https://stackoverflow.com/questions/30691466/what-is-difference-between-vims-clipboard-unnamed-and-unnamedplus-settings)

fdschmidt93: One way of doing this is to put a configurable variant in `actions.generate` with the `which_key` example.

lbiaggi: I will try to understand what `actions.generate` do later, I did one attempt using fn.input, it's working.

lbiaggi: I still have a question, where would the user override it?

fdschmidt93: I think as it is, the user has to select a register via `vim.fn.input` (by either keeping it empty or not). What I had in mind is --

```lua
function actions.copy_command_to_reg(prompt_bufnr, opts)
  opts = opts or {} -- ensure opts is a table if nil
  opts.reg = utils.get_default(opts.reg, "+")
  local cmd = action_state.get_selected_entry()["name"]
  -- concat ':' with command name entry
  a.fn.setreg(opts.reg, ":" .. cmd, "c")
end
```
and then later in `actions.generate`

```lua
action_generate.copy_command_to_reg(opts)
  return function(prompt_bufnr)
     action.copy_command_to_reg(prompt_bufnr, opts)
  end
end
```

And then in `telescope.setup{}`, users would map `action_generate.copy` which returns the customized action.

lbiaggi: So in our idea if I wanted to change the reg from the default one, I have to do when mapping like:
`["C-y"] = actions.copy_command("a") ?`

lua/telescope/actions/generate.lua
```diff
@@ -53,4 +53,9 @@ action_generate.which_key = function(opts)
   end
 end
 
+action_generate.copy_command_to_reg = function(opts)
+  return function(prompt_bufnr)
+    actions.copy_command_to_reg(prompt_bufnr, opts)
+  end
+end
```
fdschmidt93, line 60: Small nit: maybe an extra line here before return :)

fdschmidt93:
Small nit, but otherwise LGTM.

If you want, you can add back `vim.fn.input` by introducing to call the function if `opts.reg` is `_` (with note in docs). `_` is the blackhole register that otherwise does nothing (see `:h registers`). That saves users from potentially awkwardly having to do that if they wanted some sort of on-the-fly register setting.

I wouldn't mind doing that as a default, though `+` as defaults makes just as much sense to me. I'd leave it up to you if you'd want to do it and to what you'd want to default then.

lua/telescope/actions/generate.lua
lbiaggi: The formatter removes the extra line :disappointed:, not sure how to deal with it yet.

lbiaggi:
I will just let it fall back to “+” if someone is crazy enough to try to use _, :rofl: .

fdschmidt93:
Apologies, just checking that my idea came across. I wanted to use `_` (because it's an unusable register) as a special character that then always calls `vim.fn.input` like

```lua
  if opts.reg == "_" then
    opts.reg = vim.fn.input("Register > ")
  end
```

and highlight that in the docs. Then, people can use `vim.fn.input` if they wanted to without having to wrap around the action.

lua/telescope/actions/init.lua
```diff
@@ -272,6 +272,24 @@ function actions.close_pum(_)
...
 
+--- Copy current entry command to a specific register or unnamed plus as default
+--- Don't let user use  blackhole register (_), fallback to default
+---@param prompt_bufnr number: The prompt bufnr
+---@param opts table: options to pass to the action
+---@field reg: the register which the content will be sent
+function actions.copy_command_to_reg(prompt_bufnr, opts)
+  opts = opts or {} -- ensure opts is a table if nil
+  opts.reg = utils.get_default(opts.reg, "+")
+  if opts.reg == _ then
```
fdschmidt93, line 283: Should be "_" I think in any case.

lbiaggi: Hum, not sure what I am doing wrong, but since .I started using efm to format lua code things like that started to happen (e.g removing "").

lbiaggi:
@fdschmidt93 to be more cautious with your suggestion, I created a closure that validate if the reg still as '_' and keep asking until they choose a valid one, sounds good right?

lua/telescope/actions/init.lua
```diff
@@ -272,6 +272,28 @@ function actions.close_pum(_)
...
 end
 
+--- Copy current entry command to a specific register or unnamed plus as default
+--- Don't let user use  blackhole register (_), let them choose instead
+---@param prompt_bufnr number: The prompt bufnr
+---@param opts table: options to pass to the action
+---@field reg: the register which the content will be sent
+function actions.copy_command_to_reg(prompt_bufnr, opts)
+  local validate_reg = function(reg)
+    while reg == '_' do
```
fdschmidt93, line 282: Why not the simple solution? I don't see why this would be more careful? Users have to confirm with `<CR>` on `vim.fn.input.

lbiaggi: Ah, I thought about someone who kept trying to use '_'. I went too far, sorry.

fdschmidt93: If you now add docs for `_`, I'll give it a whirl later today or tomorrow and then I think we can merge :)

lbiaggi: I think everything is done (sorry, I forgot to send a message).

lua/telescope/actions/init.lua
```diff
@@ -272,6 +272,23 @@ function actions.close_pum(_)
   end
 end
 
+--- Copy current entry command to a specific register or unnamed plus as default
+--- Don't let user use  blackhole register (_), let them choose instead
```
fdschmidt93, line 276: Sorry, should have caught this earlier: could you please clarify this to 
Passing `_` as a register will always trigger asking for user input.

lua/telescope/actions/init.lua
```diff
@@ -272,6 +272,23 @@ function actions.close_pum(_)
...
+---@param prompt_bufnr number: The prompt bufnr
+---@param opts table: options to pass to the action
+---@field reg: the register which the content will be sent
+function actions.copy_command_to_reg(prompt_bufnr, opts)
+  opts = opts or {} -- ensure opts is a table if nil
+  opts.reg = utils.get_default(opts.reg, "+")
+  if opts.reg == "_" then
+    opts.reg = vim.fn.input "Register > "
```
fdschmidt93, line 284: Also my bad, this should either validate that the user only has passed one char (get first char then or something) or use `vim.fn.getchar` instead.

Conni2461:
I said in #1177 I dont see an issue why we shouldn't get this action in core but looking at the conversation this was probably a mistake. I am sorry.

Maybe we should put a good enough version in https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes

What do you think?

Conni2461:
Closing since i've gotten an answer since my last comment.

Sorry for the bad suggestion. Feel free to add a configuration recipe :)

Issue #856, opened 2021-05-21
Pickers that do not change buffer change last insert mode in the original buffer


muniter:
### Description

Pickers that do not go to a different buffer change the last insert mode location ("^" mark) of the buffer it was called from. When the action select was performed while the prompt is in insert mode.

**Expected Behavior**

Telescope should not modify the buffer it was called from, last insert mode location ("^" mark) should not change to the line telescope was called from.

**Actual Behavior**

Telescope changes the last insert mode location which disrupt the flow of the user since now `gi` is the location telescope was called from. **This only happens if the picker does not change the buffer**

For some reason `stopinsert` is not working.
https://github.com/nvim-telescope/telescope.nvim/blob/1fefd0098e92315569b71a99725b63521594991e/lua/telescope/actions/init.lua#L209-L211

### Details

<details><summary>Reproduce</summary>

1. nvim -nu test.vim
2. enter insert mode
3. add a few lines to the file:
```
1
2
3
4
```
4. leave insert mode
5. check your last insert mode location `marks ^`
6. move to a diferent line that's not the last insert mode location
7. :Telescope current_buffer_fuzzy_find
8. pick any entry from the picker but do not leave insert mode in the picker/prompt using `<CR>`
9. check the last insert mode location `marks ^` it's now the line the picker was called from

This happens with all the pickers I've tried that do not change buffers (current_buffer_fuzzy_find, colorschemes, planets)

Small video demonstration

https://user-images.githubusercontent.com/9699804/119183571-e0a8e780-ba39-11eb-9f37-40fd7b2f46bd.mp4

</details>

<details><summary>Environment</summary>

- nvim --version output: NVIM v0.5.0-dev+1319-g7c9569702
- Operating system: Linux 5.8.0-53-generic #60~20.04.1-Ubuntu x86_64 GNU/Linux
- Telescope commit: 69eb5ea

</details>

<details><summary>Configuration</summary>
<p>

```viml
" Plug Configuration {{{
call plug#begin('~/.config/nvim/plugged')
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-telescope/telescope.nvim'
call plug#end()

lua require('telescope').setup()
```

l-kershaw:
I can reproduce this, but can't figure out what's causing it.

---

[...]

Issue #559, opened 2021-02-26
Opening files leads to problems with restoring views and treesitter


jghauser:
### Description
When I open a file from telescope, views aren't restored (e.g. cursor position, folds) and tree-sitter highlighting isn't used. Instead syntax highlighting is used. I can't create new folds either.

**Expected Behavior**
When I open a file from telescope, views should be restored in the same way as when I open a file directly. Treesitter should be enabled just as it is when normally opening a file.

**Actual Behavior**
Views aren't restored and treesitter isn't used. In order to get treesitter to work correctly and my views restored, I need to reload the buffer (`:e`) after opening the file. If I do just `:loadview` my view is restored, but treesitter highlighting is not. You can see that `:e` restores treesitter highlighting as doing `:syntax` before will give you the syntax items, whereas after `:e` it gives `No Syntax items defined for this buffer`.

### Details

<details><summary>Reproduce</summary>

1. open some file with folding and tree-sitter highlighting (e.g. a lua file) 
2. Move the cursor, open/close some folds
3. close nvim
4. run nvim
5. open telescope with e.g. `:Telescope find_files`
6. search for previous file and open it
</details>

<details><summary>Environment</summary>

- nvim --version: NVIM v0.5.0-dev+1109-g6b7cc45c4
- Operating system: Latest Arch Linux
- Telescope commit: e2696d6

</details>

<details><summary>Configuration</summary>
<p>

```viml
if !filereadable('/tmp/plug.vim')
  silent !curl --insecure -fLo /tmp/plug.vim
    \ https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
endif

source /tmp/plug.vim
call plug#begin('/tmp/plugged')
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-telescope/telescope.nvim'
Plug 'nvim-treesitter/nvim-treesitter', {'do': ':TSUpdate'} "treesitter
Plug 'zhimsel/vim-stay' "restore settings
call plug#end()

autocmd VimEnter * PlugClean! | PlugUpdate --sync | close

lua << EOF

require('telescope').setup{
  defaults = {
    vimgrep_arguments = {
      'rg',
      '--color=never',
      '--no-heading',
      '--with-filename',
      '--line-number',
      '--column',
      '--smart-case'
    },
    prompt_position = "bottom",
    prompt_prefix = ">",
    selection_strategy = "reset",
[...]

PR #1202, opened 2021-09-02
fix: exiting picker from insert mode

### LINKED ISSUE #1185: calling Telescope built-in picker moves cursor to the left by one
### Description

After mapping `<esc>` to exit Telescope picker and calling e.g. `:Telescope find_files` or any other built-in picker, the cursor will be shifted to the left by one.

### Neovim version

NVIM v0.6.0-dev+205-g9695691ee

Linux arch 5.13.13-arch1-1 #1 SMP PREEMPT Thu, 26 Aug 2021 19:14:36 +0000 x86_64 GNU/Linux

### Steps to reproduce

example having a line "foobar":
```
foobar
cursor at 'r'
:Telescope find_files
exit picker either by pressing <esc> or <c-[>
now cursor at 'a'
```

### Expected behavior

Cursor does not move.

### Actual behavior

Moves one to the left.

### Minimal config

```Lua
call plug#begin('$XDG_DATA_HOME/nvim/site/plugged')
	Plug 'nvim-lua/popup.nvim'
	Plug 'nvim-lua/plenary.nvim'
call plug#end()

lua << EOF
require('telescope').setup{
	defaults = { mappings = { i = { ["<esc>"] = require('telescope.actions').close, }, }, },
}
EOF
```

fdschmidt93:
Closes #1185 

This is quite similar to #1165. If I understand it correctly `vim.cmd [[stopinsert]]` is just too slow. So I guess what happens then is you actually land in the original window in insert mode and only leave it there, which, being on the original cursor position, means you exit insert mode to land in one column to the left. (Conversely, exitting in normal mode works just fine).

The other solution I supposed would be to await `vim.cmd` though this should surely feel faster (& not sure how I'd to that in new plenary async).

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -65,6 +65,8 @@ local action_utils = require "telescope.actions.utils"
 local action_set = require "telescope.actions.set"
 local entry_display = require "telescope.pickers.entry_display"
 local from_entry = require "telescope.from_entry"
+local async = require "plenary.async"
+local await_schedule = async.util.scheduler
 
 local transform_mod = require("telescope.actions.mt").transform_mod
 local resolver = require "telescope.config.resolve"
@@ -346,8 +348,14 @@ actions.close = function(prompt_bufnr)
   action_state.get_current_history():reset()
   local picker = action_state.get_current_picker(prompt_bufnr)
   local original_win_id = picker.original_win_id
+  local original_cursor = a.nvim_win_get_cursor(original_win_id)
 
   actions.close_pum(prompt_bufnr)
+  await_schedule(function()
+    if a.nvim_win_is_valid(original_win_id) then
+      a.nvim_win_set_cursor(original_win_id, original_cursor)
+    end
+  end)
 
   require("telescope.pickers").on_close_prompt(prompt_bufnr)
   pcall(a.nvim_set_current_win, original_win_id)
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -65,8 +65,6 @@ local action_utils = require "telescope.actions.utils"
 local action_set = require "telescope.actions.set"
 local entry_display = require "telescope.pickers.entry_display"
 local from_entry = require "telescope.from_entry"
-local async = require "plenary.async"
-local await_schedule = async.util.scheduler
 
 local transform_mod = require("telescope.actions.mt").transform_mod
 local resolver = require "telescope.config.resolve"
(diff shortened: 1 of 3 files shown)


lua/telescope/actions/init.lua
```diff
@@ -287,6 +290,11 @@ actions._close = function(prompt_bufnr, keepinsert)
   actions.close_pum(prompt_bufnr)
   if not keepinsert then
     vim.cmd [[stopinsert]]
+    await_schedule(function()
+      if a.nvim_win_is_valid(original_win_id) then
+        a.nvim_win_set_cursor(original_win_id, original_cursor)
+      end
+    end)
```
tjdevries, line 297: Can this not just get placed beneath the pcall on line 303 for set_current_win? I don't thin we need to do it this way.

Just `pcall(a.nvim_win_set_cursor, original_win_id, original_cursor)` or similar

fdschmidt93: No, tried it an desn't work. `stopinsert` for some reason only takes effect afterwards. So when you set the cursor position to where it's supposed to be while being in insert mode, you'll end up one to the left when exitting.

I tried a bunch of things to fix this 😅 it's not that I like this one all too much, it was just the only one that worked.

mortezadadgar:
Any update on this issue

muniter:
> Closes #1185 
> 
> This is quite similar to #1165. If I understand it correctly `vim.cmd [[stopinsert]]` is just too slow. So I guess what happens then is you actually land in the original window in insert mode and only leave it there

This would explain #856

darzok0914:
any update on this PR?

Conni2461:
rebased it.

If you are using vim-startify, my cursor e.g. is at `15:5`, i open a file. Its still there. It also happens if i am at file A line 100. Open telescope goto file B. I end up at line 100. 

So we basically only want this on `<C-c>` not on `<CR>`

fdschmidt93:
@Conni2461 ok to move history reset to `on_close_prompt`? Seems like the most appropriate place, since we have to avoid meddling with the original window if we eg to select_{vertical, horizontal, ...} as otherwise if we'd set the cursor, going back to the original window would place the cursor one too far to the right (if that makes sense).

For some reason, I cannot leave insert mode fast enough in `actions.close` so well, we just adjust the cursor.

Conni2461:
> Ok to move history reset to on_close_prompt?

Yes. Works for me. Fix also works for me. Thanks. Merging to master because its a ffix

> For some reason, I cannot leave insert mode fast enough in actions.close so well, we just adjust the cursor.

Would be great if we could leave insert before doing `:edit` (...) also because of this https://github.com/nvim-telescope/telescope.nvim/issues/559#issuecomment-1002803413

`:help stopinsert`
> Stop Insert mode or |Terminal-mode| as soon as possible.

Conni2461 (approved):

ajitid:
Do we have a different ticket to stop cursor moving to the left when Enter key is pressed?

What I'm doing here:
1. Open `:Telescope buffers`
2. Hit Enter key
3. Watch cursor moving to the left
4. Repeat 

https://user-images.githubusercontent.com/26769888/169488940-fe10be4e-8358-471d-876f-ed1ebd300c3c.mp4

(Can confirm the PR works just fine for <c-c> and <esc> keys.)

clason:
I'm afraid this broke `actions.insert_symbol_i` again...

(I.e., triggering it on `foo|bar` puts the symbol after the `b` instead of the cursor position.)

fdschmidt93:
I'm very sure I'm misunderstanding something here, but looks as intended for me..?

https://user-images.githubusercontent.com/39233597/171050067-e45851c3-13ca-4337-8ac3-c7a1a9a772db.mp4

clason:
That doesn't look like the symbol inserter to me?

(on mobile, will share config when I get home)

clason:
Make sure to have trailing characters! If you insert at the end of a line, you obviously won't see the issue.

PR #1210, opened 2021-09-04
Add builtin.jumplist back into README.md

thornycrackers:
Looks like 618e0e6075b4215e43c6a848daa37ef4e354b5dc dropped this entry, just adding it back.


fdschmidt93:
Thanks :) Some things fall through the cracks in bigger projects 😅

PR #1211, opened 2021-09-04
Fix: Escape live grep prompt characters

chase:
When using `:Telescope live_grep` using the default settings for `rg`, searching for `-c` leads to Telescope failing and a warning being thrown because `-c` is not being escaped.

To fix this and potentially other issues, the prompt will now use `escape_chars` for live_grep

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -93,7 +93,7 @@ files.live_grep = function(opts)
       search_list = filelist
     end
 
-    return flatten { vimgrep_arguments, additional_args, prompt, search_list }
+    return flatten { vimgrep_arguments, additional_args, escape_chars(prompt), search_list }
   end, opts.entry_maker or make_entry.gen_from_vimgrep(
     opts
   ), opts.max_results, opts.cwd)


fdschmidt93:
We actually only recently defaulted away from `escaping_chars` as part of #987 such that `live_grep` now essentially acts as a front-end to ripgrep (very similar to [FZF](https://github.com/junegunn/fzf#3-interactive-ripgrep-integration). However, I agree that we maybe should allow to optionally escape characters similar to `grep_string` via `use_regex=false`.

chase:
> We actually only recently defaulted away from `escaping_chars` as part of #987 such that `live_grep` now essentially acts as a front-end to ripgrep (very similar to [FZF](https://github.com/junegunn/fzf#3-interactive-ripgrep-integration). However, I agree that we maybe should allow to optionally escape characters similar to `grep_string` via `use_regex=false`.

Ah, that explains why there was a regression in behavior.

The problem here is that it _is a valid regex_, but ripgrep is receiving it as a flag because it doesn't handle argument escaping correctly. I'm not sure it's the best UX to expect to be able to pass arguments directly to ripgrep, FZF uses `--`, the no more flags delimiter to avoid such problems.

I will close this PR and make a PR to use `--` instead.

There is another problem that occurs with the live_grep source simply returns an empty string, it crashes Telescope and it shouldn't, but that's a problem for another PR.

PR #1215, opened 2021-09-06
fix: Missing documentation double quotes around the 'prompt_prefix' character

miguelmoraperea:
This commit fixes an incorrect documentation. The  'prompt_prefix' character should be wrap in double quotes for the change to take effect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -732,11 +732,11 @@ tab completions and setting options.
 :Telescope find_files
 
 " Setting options
-:Telescope find_files prompt_prefix=🔍
+:Telescope find_files prompt_prefix="🔍"
 
 " If option is table type in Lua code, you can use `,` to connect each command string, e.g.:
 " find_command,vimgrep_arguments are both table type. So configure it on command-line like so:
-:Telescope find_files find_command=rg,--ignore,--hidden,--files prompt_prefix=🔍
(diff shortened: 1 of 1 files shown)


Conni2461:
Closing. We do not want to have double quotes in the prompt prefix

image

miguelmoraperea:
Using no quotes around the prefix character:
`nnoremap <leader>fg :lua R('telescope.builtin').live_grep({ prompt_prefix=🔍 })<cr>`

The change does not take effect:
image

But using quotes:
`nnoremap <leader>fg :lua R('telescope.builtin').live_grep({ prompt_prefix="🔍 " })<cr>`

It works (no quotes showing on the prompt):
image


If I use the Telescope command instead:
`:Telescope find_files find_command=rg,--ignore,--hidden,--files prompt_prefix="🔍"`

Then yes, the quotes appear, we shouldn't do that, I fixed that in my commit.

The only line modified now is the one that calls the built-in functions and passes the prompt_prefix as an option in the function call.

Should we re-open this ticket or should I create a new ticket/pull request?

Conni2461:
We definitely need quotes for lua strings. I overlooked that there is also a lua command in that diff. Sorry

Thanks for checking :) happy to merge it (cant reopen this pr)

miguelmoraperea:
No worries @Conni2461, I created a new pull request: #1238

PR #1217, opened 2021-09-06
[Fix] Add no more flags delimiter for ripgrep

chase:
This matches the behavior of fzf, if a user needs to pass additional
arguments they can use additional_args or change vimgrep_arguments.

See #1211 for discussion of this solution, this fixes bugs that were introduced by #987


Conni2461:
yeah for now we should do this

Thanks :)

PR #1219, opened 2021-09-06
Exclude unlisted buffers in oldfiles picker

thraizz:
By excluding unlisted buffers in oldfiles picker when `include_current_session` is set, we exclude files opened by LSPs.

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -344,7 +344,7 @@ internal.oldfiles = function(opts)
   local results = {}
 
   if opts.include_current_session then
-    for _, buffer in ipairs(vim.split(vim.fn.execute ":buffers! t", "\n")) do
+    for _, buffer in ipairs(vim.split(vim.fn.execute ":buffers t", "\n")) do
       local match = tonumber(string.match(buffer, "%s*(%d+)"))
       if match then
         local file = vim.api.nvim_buf_get_name(match)


Conni2461:
discussion why `!` was added: https://github.com/nvim-telescope/telescope.nvim/pull/657#discussion_r594353918

thraizz:
> discussion why `!` was added: [#657 (comment)](https://github.com/nvim-telescope/telescope.nvim/pull/657#discussion_r594353918)

Ah I see. But with the unlisted buffers included, we also include every file used by the LSP.
Compare these screenshots from the same directory:
with-unlisted-buffers
without-unlisted-buffers

thraizz:
Lets take this as an example output for the `buffers! t` command:
```
:buffers! t                                                                                                                                                
  1 %a   "user/admin.py"                1 second ago                                                                                                       
 15u#    "~/remote/development/lakur-kanban-backend/boards/models.py" 10 seconds ago                                                                       
 10u     "~/remote/development/lakur-kanban-backend/manage.py" line 0                                                                                      
 11u     "~/remote/development/lakur-kanban-backend/setup.py" line 0                                                                                       
 12u     "~/remote/development/lakur-kanban-backend/boards/__init__.py" line 0
 ```
 Here, I dropped the buffer for the boards/models.py file, so it is included in the unlisted buffer but the id is 15u#, which is not the same as the files opened by the LSP, which would match [0-9].*u, so maybe we can just use the unlisted buffers list and filter out these created by the LSP.

**Comment by Conni2461** [comment]
I see what you are saying i just wanted to give context :laughing: please talk about this with @jesseleite you both given good arguments for keeping and removing it. Its hard for me to make a decision because i am more of a [frecency](https://github.com/nvim-telescope/telescope-frecency.nvim) guy :rofl: 

Other maintainers are welcome too :laughing:

**Comment by fdschmidt93** [comment]
I'm also not an `oldfiles` guy myself but as affects LSPs in general (tested on Pyright & Sumneko, and I guess it makes sense) something should be done about it.

I see

> Because if you manually open a file buffer, then :bdelete to close that buffer, it'll get removed from the finder until you restart Vim and see it in v:oldfiles, which doesn't make any sense. It's the same as my remove_buf() argument from our other thread here 433 here

but I think the created problem weighs more heavily than the original solution.

Wouldn't it be much cleaner for any user desiring that feature to have a `BufDelete` autocommand that adds a buffer that is loaded and listed and to be deleted to the users `v:oldfiles`? See also `:h v:oldfiles`

@tami5 @jesseleite

**Comment by jesseleite** [comment]
I agree this needs to be fixed. As you can see in the other thread though, there are clear problems both with and without the `!`. I believe the fix needs to be storing each buffer in a table when the _user_ opens files (not _lsp_ crawling files). I don't think the magic I introduced with `:buffers` will work for all use cases with or without that `!`.

**Comment by jesseleite** [comment]
@fdschmidt93 Ah yes, that's what I was saying in my above comment. I think a `BufEnter` might be better than `BufDelete` though. I'm pretty sure `v:oldfiles` will still populate with results if you close vim (without deleting each buffer), so I'd expect the same of `:Telescope oldfiles`, hence `BufEnter`. Thoughts?

**Comment by jesseleite** [comment]
@fdschmidt93 Ah yes, that's basically what I was saying in my above comment. I think a `BufEnter` might be better than `BufDelete` though. I really don't think the user should have to delete a buffer for it to appear in in the oldfiles list. In my opinion, if you just switch to another buffer, the previously open buffer should now be added to the oldfiles list. At least that's how `:Telescope oldfiles` works now, and also how `:History` worked in fzf.vim. Using a `BufEnter` event should fix the LSP crawling issue though 👍

**Comment by fdschmidt93** [comment]
Agreed, `BufEnter` probably is better.

FZFs relevant function as a reference.

```viml
function! fzf#vim#_recent_files()
  return fzf#vim#_uniq(map(
    \ filter([expand('%')], 'len(v:val)')
    \   + filter(map(fzf#vim#_buflisted_sorted(), 'bufname(v:val)'), 'len(v:val)')
    \   + filter(copy(v:oldfiles), "filereadable(fnamemodify(v:val, ':p'))"),
    \ 'fnamemodify(v:val, ":~:.")'))
endfunction
```

I see three approaches:
1. Map FZF 1-1: they solve it a bit more cleanly I suppose. Not sure about sorting though
2. User maintains his own autocommand
3. Telescope maintains an autocommand

If 1. allows for adequate sorting, I think this is the clear favorite. I'm ok with 3 if we (a) copy the oldfiles in a telescope internal list and (b) have autocommands that manage that internal list. We shouldn't meddle with `v:oldfiles`. Otherwise 2.

jesseleite:
I had planned to do #3 in a bugfix PR since I originally implemented the `oldfiles` command; Just don't have much time lately.

You can see I tried to get clever with `:buffers! t` in the current implementation, but it ended up causing problems with LSP crawling, etc. At this point, I see no reason to re-invent the wheel if what fzf.vim did worked so wonderfully. For that reason I still say option #3 there, where Telescope includes a `BufEnter` autocommand, merging into an internal list copied from `v:oldfiles`, rather than injecting into `v:oldfiles` itself. (I honestly didn't even know you could change `v:oldfiles` at runtime, but I think we should be treating the native `v:oldfiles` table as if it's immutable, in case other plugins interact with it.)

Another advantage to #3 is that it's less configuration for Telescope users. Anyone should be able to just install Telescope and run `:Telescope oldfiles` IMO.

jesseleite:
PS. To clarify, the reason why I don't think we can just drop the `!` from `:buffers! t` is because the `!` shows unlisted buffers after they are closed. For example, if you open a vim `:help` file, it wouldn't get listed until you restart vim and that file gets populated into the `v:oldfiles` list. The `!` fixes that, but ends up being over-zealous when it comes to LSP crawled files.

PPS. We obviously don't want to list temporary scratch buffers, terminal buffers, etc., but help files are valid files that are normally included in `v:oldfiles`, so I think they should be handled by `:Telecope oldfiles` as well. In my eyes, the whole point of `:Telescope oldfiles` is to show a MRU-sorted `v:oldfiles` list that is updated in realtime as you open new file buffers ❤️

jesseleite:
@thraizz Oh super interesting find! I'm curious as well.

@Conni2461 Curious on your thoughts; Is it safe to depend on that `line 0` for LSP-opened buffers?

PR #1221, opened 2021-09-07
Fix whichkey overriding opts

levouh:
Fixing overriding of numbers passed as `opts` being overridden with the result of a repeated string (causing errors on subsequent calls)

---

When using `telescope.actions.generate`, the same table reference is passed. From this, if we override a value in that table (originally a number) with a string, we get errors later. To replicate this, setup a mapping that uses `require("telescope.actions.generate").which_key({ only_show_current_mode = false })`, toggle the `whichkey` popup on, then try to toggle it off again.

CC: @fdschmidt93

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -1002,13 +1002,7 @@ actions.which_key = function(prompt_bufnr, opts)
     }, ";")
   ))
 
-  a.nvim_buf_set_lines(
-    km_buf,
-    0,
-    -1,
-    false,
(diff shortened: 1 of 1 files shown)


fdschmidt93:
I've tested it locally and this only affects `column_indent`? Just two nits: I'd prefer to keep `opts.column_padding` for consistency with other `opts` and have a small comment for why `column_indent` is an outlier.

Otherwise, LGTM - thanks :)

levouh:
Let me know how that looks.

fdschmidt93:
Thanks :)

PR #1224, opened 2021-09-07
feat: lsp type definitions

Rodrigodd:
Adds `lsp_type_definitions`, similar to `lsp_definitions` and `lsp_implementations`.


Conni2461:
Thanks :)

Issue #1126, opened 2021-08-19
current_buffer_tags: No write since last change


ViViDboarder:
### Description

When trying to Telescope to a buffer tag using `vim.api.nvim_set_keymap("n", "<leader>t", "<cmd>Telescope current_buffer_tags<CR>", {silent=true, noremap=true})`
` when a buffer hasn't been written, I am given the error `No write since last change`. This seems to suggest that the plugin is trying to open a new buffer rather than jumping to the location.

This works as expected with lsp and `lsp_document_symbols`.

### Neovim version

NVIM v0.5.0

### Operating system and version

macOS Big Sur

### Steps to reproduce

1. `nvim -nu minimal.lua minimal.lua`
2. Make an edit anywhere, but do not write the buffer
3. Generate tags for the file `:! ctags %`
4. Try to jump to a tag with `:Telescope current_buffer_tags`

### Expected behavior

Jump to tag within current, unwritten buffer.

### Actual behavior

Error message: `No write since last change`

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          { 'nvim-telescope/telescope-fzf-native.nvim', run = 'make' },
        },
      },
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
      display = { non_interactive = true },
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
  require('telescope').load_extension('fzf')
  -- ADD INIT.LUA SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing Telescope and dependencies.")
  vim.fn.system { 'git', 'clone', '--depth=1', 'https://github.com/wbthomason/packer.nvim', install_path }
end
load_plugins()
require('packer').sync()
[...]

PR #1230, opened 2021-09-09
fix: jump to tag if buffer has unsaved changes

Conni2461:
fix https://github.com/nvim-telescope/telescope.nvim/issues/1126

Comparing absolute path vs relative path in action/set.lua caused :edit to run anyway. So `:set hidden` didnt fix it. So we do something similar to files where we get the absolute path on `<cr>` with `.path`. But we should only do this once so metatable.

@ViViDboarder can you confirm that fixes the issue? Thanks

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -883,6 +883,17 @@ function make_entry.gen_from_ctags(opts)
     end
   end
 
+  local mt = {}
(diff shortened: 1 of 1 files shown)


ViViDboarder:
👍 this fixes the issue for me! Thanks!

Conni2461:
I think this is good enough. Will merge and see what happens :rofl:

Issue #857, opened 2021-05-21
Feat: Add option to easily toggle tree-sitter based highlighting 


shaunsingh:
**Is your feature request related to a problem? Please describe.**
There have recently been a few bugs in treesitter (mainly https://github.com/tree-sitter/tree-sitter/issues/973#issuecomment-810480327, that greatly increases the loading times for certain filetypes 
 
For example, kotlin files can take upto two seconds to load. While this is somewhat bearable when opening singular files, when browsing through 10's of files it can take quite a while. 

**Describe the solution you'd like**
At the moment, the README discusses that treesitter is an optional dependency and can be used for syntax highlighting, but doesn't mention how to disable it. Adding either an option or more information to the README would be a good change 

**Additional context**
When searching for certain filetypes (e.g. php and kotlin) neovim will freeze for 1-2 seconds while the file loads.

johnobla:
I respectfully disagree with this feature request.

Since Treesitter is an optional dependency and isn't included with Telescope, Telescope doesn't use Treesitter if it's not installed.

Similar to the ripgrep optional dependency, Telescope feels amazing when it "just works" with Treesitter.

It would be disappointing if this feature request removed that amazing feeling for newcomers to Telescope.

shaunsingh:
I understand your stance on the issue, but shouldn't we favor form over function? As much as I like the beautiful look of treesitter, a fuzzy finders first and foremost job should be fast file searching, the nice colors should come second.

johnobla:
I agree completely. I also understand the frustration that comes with laggy loading, after all it's why we're here and not on VSCode 🙂

Our differing opinions come from changing the default functionality, rather than choosing form over function.

Right now the default functionality says:
1. If you don't already have Treesitter installed, I won't touch it
2. If you already have Treesitter installed, I'll use it to make your previews prettier

With the feature request, we'd be  changing the default functionality to cater to those who
1. Have Treesitter installed and benefit from its syntax highlighting
2. But don't want the syntax highlighting applied when searching for files

This is a group that should be catered for (just like any other), just not as the default option for Telescope.

shaunsingh:
You're right. I think an option would be better than changing the defaults.

Conni2461:
Its already possible. Not as user friendly, but way more flexible. You can do different things like highlight only certain filetypes with regex and other default.

Related threads (first):
- https://github.com/nvim-telescope/telescope.nvim/issues/623#issuecomment-792233601
[...]

Issue #623, opened 2021-03-07
Faster preview


YodaEmbedding:
Telescope freezes when attempting to preview large files. For instance, previewing a 100000 line CSV file causes a 10 second long freeze.

Possible solutions, in order of easiest to most difficult to implement (I presume):

 - Disable preview for files >1MB.
 - `cat file.csv | head -c 1000000 | head -n 1000 > file.csv.preview` so that a preview does not need to be generated for too much data. Here, `file.csv.preview` is a smaller file (<1MB and <1000 lines) that can read by telescope.
 - Async preview. The preview will still take time and CPU to generate, but at least telescope will not freeze.

Conni2461:
We have considered all the above and we found a pretty good solution, (in my opinion, but i also designed them sooooooo :rofl: ) already.
In fact previewers are already async :) File reading happens off thread, but neovim does not allow to write to the buffer async as well as highlighting which still happens in main thread. And the last thing is mostly what makes it slow. But i have never displayed a 100000 line CSV file :laughing: 

So we decided to make `buffer_previewer_maker` overridable. Its the heart of the buffer previewer and if you want some information about how previewers work `:help telescope.previewers`.

I wanted to make an extension that lets you not have to override the function but i never found the interest in doing that. So here are some snippets:

Disable highlighting for certain files:
```lua
local previewers = require('telescope.previewers')

local _bad = { '.*%.csv', '.*%.lua' } -- Put all filetypes that slow you down in this array
local bad_files = function(filepath)
  for _, v in ipairs(_bad) do
    if filepath:match(v) then
      return false
    end
  end

  return true
end

local new_maker = function(filepath, bufnr, opts)
  opts = opts or {}
  if opts.use_ft_detect == nil then opts.use_ft_detect = true end
  opts.use_ft_detect = opts.use_ft_detect == false and false or bad_files(filepath)
  previewers.buffer_previewer_maker(filepath, bufnr, opts)
end

require('telescope').setup {
  defaults = {
    buffer_previewer_maker = new_maker,
  }
}
```

Ignore files bigger than a threshold. Make sure you are on current master, i pushed a fix to make this work:
```lua
local previewers = require('telescope.previewers')

local new_maker = function(filepath, bufnr, opts)
  opts = opts or {}

  vim.loop.fs_stat(vim.fn.expand(filepath), function(_, stat)
    if not stat then return end
    if stat.size > 100000 then
      return
    else
[...]

PR #1231, opened 2021-09-10
feat: skip/timeout preview if file cannot be easily previewed

fdschmidt93:
Addresses #623 

Allows premature skipping previewing when filesize is too large or reading file + parsing of lines exceeds timeout.

What's primarily missing is to make timeout and filesize limit configurable.

timeout

### DIFF
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -399,8 +399,11 @@ append(
     your picker.
 
     Fields:
-      - check_mime_type:  Use `file` to try to infer whether the file to preview
-                          is a binary if plenary's filetype detection fails.
+      - check_mime_type:  Use `file` if available to try to infer whether the
+                          file to preview is a binary if plenary's
+                          filetype detection fails.
+                          Windows users get `file` from:
+                          https://github.com/julian-r/file-windows
                           Set to false to attempt to preview any mime type.
                           Default: true
       - filesize_limit:   The maximum file size in MB attempted to be previewed.
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -395,7 +395,8 @@ append(
     By default it is a table, with default values (more below).
     To disable previewing, set it to false. If you have disabled previewers
     globally, but want to opt in to previewing for single pickers, you will have to
-    pass `show_preview = true` to the `opts` of your picker.
+    pass `preview = true` or `preview = {...}` (your config) to the `opts` of
+    your picker.
 
     Fields:
       - check_mime_type:  Use `file` to try to infer whether the file to preview
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -383,6 +383,52 @@ append(
 )
 
 append(
+  "preview",
+  {
+    check_mime_type = true,
+    filesize_limit = 25,
+    timeout = 250,
+    treesitter = true,
+  },
+  [[
+    This field handles the global configuration for previewers.
+    By default it is a table, with default values (more below).
+    To disable previewing, set it to false. If you have disabled previewers
+    globally, but want to opt in to previewing for single pickers, you will have to
+    pass `show_preview = true` to the `opts` of your picker.
+
+    Fields:
+      - check_mime_type:  Use `file` to try to infer whether the file to preview
+                          is a binary if plenary's filetype detection fails.
+                          Set to false to attempt to preview any mime type.
+                          Default: true
+      - filesize_limit:   The maximum file size in MB attempted to be previewed.
+                          Set to false to attempt to preview any file size.
+                          Default: 25
+      - timeout:          Timeout the previewer if the preview did not
+                          complete within `timeout` milliseconds.
+                          Set to false to not timeout preview.
+                          Default: 250
+      - hook(s):          Function(s) that takes `(filepath, bufnr, opts)`
+                          to be run if the buffer previewer was not shown due to
+                          the respective test.
+                          Available hooks are: {mime, filesize, timeout}_hook, e.g.
+                          preview = {
+                            mime_hook = function(filepath, bufnr, opts) ... end
+                          }
+                          See `telescope/previewers/*.lua` for relevant examples.
+                          Default: nil
+      - treesitter:       Determines whether the previewer performs treesitter
+                          highlighting, which falls back to regex-based highlighting.
+                          `true`: treesitter highlighting for all available filetypes
+                          `false`: regex-based highlighting for all filetypes
+                          `table`: table of filetypes for which to attach treesitter
+                          highlighting
+                          Default: true
+    ]]
+)
+
+append(
   "vimgrep_arguments",
   { "rg", "--color=never", "--no-heading", "--with-filename", "--line-number", "--column", "--smart-case" },
   [[
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -34,10 +34,10 @@ local function defaulter(f, default_opts)
   default_opts = default_opts or {}
   return {
     new = function(opts)
-      opts.preview = opts.preview or {}
-      if conf.preview == false and not opts.show_preview == true then
+      if conf.preview == false and not opts.preview then
         return false
       end
+      opts.preview = type(opts.preview) ~= "table" and {} or opts.preview
       if type(conf.preview) == "table" then
         for k, v in pairs(conf.preview) do
           opts.preview[k] = vim.F.if_nil(opts.preview[k], v)
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -17,6 +17,8 @@ local ns_previewer = vim.api.nvim_create_namespace "telescope.previewers"
 
 local has_file = 1 == vim.fn.executable "file"
 
+-- TODO(fdschmidt93) switch to Job once file_maker callbacks get cleaned up with plenary async
+-- avoids SIGABRT from utils.get_os_command_output due to vim.time in fs_stat cb
 local function capture(cmd, raw)
   local f = assert(io.popen(cmd, "r"))
   local s = assert(f:read "*a")
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -1,6 +1,7 @@
 local from_entry = require "telescope.from_entry"
 local Path = require "plenary.path"
 local utils = require "telescope.utils"
+local strings = require "plenary.strings"
 local putils = require "telescope.previewers.utils"
 local Previewer = require "telescope.previewers.previewer"
 local conf = require("telescope.config").values
@@ -9,11 +10,96 @@ local pfiletype = require "plenary.filetype"
 local pscan = require "plenary.scandir"
 
 local buf_delete = utils.buf_delete
-local defaulter = utils.make_default_callable
 
 local previewers = {}
 
 local ns_previewer = vim.api.nvim_create_namespace "telescope.previewers"
+
+local has_file = 1 == vim.fn.executable "file"
+
+local function capture(cmd, raw)
+  local f = assert(io.popen(cmd, "r"))
+  local s = assert(f:read "*a")
+  f:close()
+  if raw then
+    return s
+  end
+  s = string.gsub(s, "^%s+", "")
+  s = string.gsub(s, "%s+$", "")
+  s = string.gsub(s, "[\n\r]+", " ")
+  return s
+end
+
+local function defaulter(f, default_opts)
+  default_opts = default_opts or {}
+  return {
+    new = function(opts)
+      opts.preview = opts.preview or {}
+      if conf.preview == false and not opts.show_preview == true then
+        return false
+      end
+      if type(conf.preview) == "table" then
+        for k, v in pairs(conf.preview) do
+          opts.preview[k] = vim.F.if_nil(opts.preview[k], v)
+        end
+      end
+      return f(opts)
+    end,
+    __call = function()
+      local ok, err = pcall(f(default_opts))
+      if not ok then
+        error(debug.traceback(err))
+      end
+    end,
+  }
+end
+
+local function set_timeout_message(bufnr, winid, message)
+  local height = vim.api.nvim_win_get_height(winid)
+  local width = vim.api.nvim_win_get_width(winid)
+  vim.api.nvim_buf_set_lines(
+    bufnr,
+    0,
+    -1,
+    false,
+    utils.repeated_table(height, table.concat(utils.repeated_table(width, "╱"), ""))
+  )
+  local anon_ns = vim.api.nvim_create_namespace ""
+  local padding = table.concat(utils.repeated_table(#message + 4, " "), "")
+  local lines = {
+    padding,
+    "  " .. message .. "  ",
+    padding,
+  }
+
+  local col = math.floor((width - strings.strdisplaywidth(lines[2])) / 2)
+  for i, line in ipairs(lines) do
+    vim.api.nvim_buf_set_extmark(
+      bufnr,
+      anon_ns,
+      math.floor(height / 2) - 1 + i,
+      0,
+      { virt_text = { { line, "Normal" } }, virt_text_pos = "overlay", virt_text_win_col = col }
+    )
+  end
+end
+
+-- modified vim.split to incorporate a timer
+local function split(s, sep, plain, opts)
+  opts = opts or {}
+  local t = {}
+  for c in vim.gsplit(s, sep, plain) do
+    table.insert(t, c)
+    if opts.preview.timeout then
+      local diff_time = (vim.loop.hrtime() - opts.start_time) / 1e6
+      if diff_time > opts.preview.timeout then
+        return
+      end
+    end
+  end
+  return t
+end
+local bytes_to_megabytes = math.pow(1024, 2)
 
 local color_hash = {
   ["p"] = "TelescopePreviewPipe",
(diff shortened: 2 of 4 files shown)


tjdevries:
I like this idea -- it would be good to have previewers not lock things up so much.

Conni2461:
I love the plenary.benchmark module (i just merged it, if you wanna try it you might need to pull). Lets us compare functions easily. Benchmark with a file with 250k lines

```lua
local fd = assert(vim.loop.fs_open("files", "r", 438))
local stat = assert(vim.loop.fs_fstat(fd))
local data = assert(vim.loop.fs_read(fd, stat.size, 0))
assert(vim.loop.fs_close(fd))

local bench = require "plenary.benchmark"

bench("gmatch vs vim.split", {
  warmup = 3,
  runs = 80,
  fun = {
    {
      "gmatch",
      function()
        local result = {}
        for line in data:gmatch "[^\n]+" do
          table.insert(result, line)
        end
        return result
      end,
    },
    {
      "vim.split",
      function()
        return vim.split(data, "\n")
      end,
    },
  },
})
```

image

fdschmidt93:
I've actually switched back to vim's internal implementation because there were other issues anyways :sweat_smile:

Conni2461:
funny that we just commented at the same time :rofl: i am just comparing functions now :laughing:

Conni2461:
We should probably also incorporate [dont preview biniaries](https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes#dont-preview-binaries) otherwise i like the stuff happening here. Good job :)

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -15,6 +18,47 @@ local previewers = {}
...
+    )
+  end
+end
+
+-- modified vim.split to incorporate a timer
+local function split(s, sep, plain, opts)
+  opts = opts or {}
+  local t = {}
+  for c in vim.gsplit(s, sep, plain) do
+    table.insert(t, c)
```
Conni2461, line 50: I was just about to comment lets not use table.insert but thats a lie too. I have been brainwashed for so long.

```lua
local fd = assert(vim.loop.fs_open("files", "r", 438))
local stat = assert(vim.loop.fs_fstat(fd))
local data = assert(vim.loop.fs_read(fd, stat.size, 0))
assert(vim.loop.fs_close(fd))

local bench = require "plenary.benchmark"

bench("table.insert vs index", {
  warmup = 10,
  runs = 80,
  fun = {
    {
      "table.insert",
      function()
        local result = {}
        for line in vim.gsplit(data, "\n") do
          table.insert(result, line)
        end
        return result
      end,
    },
    {
      "index",
      function()
        local result = {}
        local i = 1
        for line in vim.gsplit(data, "\n") do
          result[i] = line
          i = i + 1
        end
        return result
      end,
    },
  },
})
```
image

Conni2461: Okay index is slower for big files(250k), faster for small files (100)

fdschmidt93: Noticeably so that we should change?

Conni2461: 8.1 μs (index) vs 18.0 μs (table.insert) for a 100 line file

Akeboshiwind:
Cool stuff!

It might be beyond the scope of this PR, but would it be possible to allow users to provide a default previewer to use if the timeout is met?

As a case for this, personally I'd want to see a truncated view of the file to get an idea of what's in it. But I can see that might not be for everyone.

fdschmidt93:
Makes sense, let me know if that work for you.

Akeboshiwind:
Looks great thanks! I don't have the time to test it right now, but I will tomorrow.

As another note: I presume you disable the timeout by setting `timeout` to false like with `filesize_limit`? A note in the docs would be great :)
Also the second sentence of the description of `timeout` seems incomplete?

fdschmidt93:
Thanks for the comments :) keep'em coming :D I've pushed this through in the late evening, sorry that a couple of things slipped.

lua/telescope/config.lua
```diff
@@ -323,6 +323,33 @@ local telescope_defaults = {
                           Default: 1000
     ]],
   },
+  previewers = {
+    {
+      filesize_limit = 25,
+      timeout = 250,
```
Conni2461, line 329: another thing we can do (maybe out of scope for this PR) is a treesitter field. Where we can then say `treesitter = false` to disable treesitter highlighting and `treesitter = { "lua", "c" }` to enable treesitter only for a subset of languages.

Would resolve https://github.com/nvim-telescope/telescope.nvim/issues/857

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -122,20 +197,68 @@ previewers.file_maker = function(filepath, bufnr, opts)
...
+          :new({
+            command = "file",
+            args = { "--mime-type", "-b", filepath },
+            on_exit = function(j)
+              local mime_type = vim.split(j:result()[1], "/")[1]
+              if mime_type ~= "text" and type(ft) == "string" and ft == "" then
+                vim.schedule(function()
+                  set_timeout_message(bufnr, opts.winid, "Binary cannot be previewed")
+                end)
+                return
```
Conni2461, line 210: I would only ask the `mime-type` oracle if ft is an empty string. Otherwise we kinda already know that its not a binary. I dont really know if its suitable to realize. We already are in callback hell here, even before your PR (so its not your fault)

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -376,13 +503,17 @@ previewers.vimgrep = defaulter(function(opts)
...
+        preview_timeout = opts.preview_timeout,
+        preview_timeout_hook = opts.preview_timeout_hook,
+        preview_filesize_limit = opts.preview_filesize_limit,
         callback = function(bufnr)
           jump_to_line(self, bufnr, entry.lnum)
         end,
       })
     end,
   }
-end, {})
```
Conni2461, line 385: You changed `defaulter` and removed `, {})` for the two important onces. But dont you have to adjust this for all other previewers as well?

lua/telescope/config.lua
```diff
@@ -323,6 +323,33 @@ local telescope_defaults = {
                           Default: 1000
     ]],
   },
+  previewers = {
```
Conni2461, line 326: If it is possible i would want to merge this after docs refactor. I dont have the strength to rebase config.lua again :laughing:

lua/telescope/previewers/buffer_previewer.lua
fdschmidt93: Ah sorry, I first removed them and then thought maybe there are cases I'm not seeing and wanted to revert all of them to as it was befote. Thought I did, I'll check once more.

fdschmidt93: Ah sorry, I first removed them and then thought maybe there are cases I'm not seeing and wanted to revert all of them to as it was befote. Thought I did, I'll check once more.

lua/telescope/previewers/buffer_previewer.lua
fdschmidt93: Makes sense. One thing I've realized, `file` is not a command on Windows and would introduce a dependency for any Windows user. From 10 minutes Google search, I'm not sure there is a cool workaround. So mime type detection only for Linux/Mac users? :sweat_smile:

lua/telescope/config.lua
fdschmidt93: Yeah, I'll look into this. I'd say we should do this as part of this PR.

lua/telescope/config.lua
fdschmidt93: Totally understand. I _shouldn't_ have meaningful time to work on this before Saturday again anyways? Otherwise I'm getting to distracted from supposedly more relevant stuff :sweat_smile:

lua/telescope/config.lua
fdschmidt93: Added. I've also changed the configuration from `previewer` to `preview` as I kind of prefer to have these two disentangled. Let me know if you prefer `previewer` at setup though.

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -1,20 +1,94 @@
...
+    __call = function()
+      local ok, err = pcall(f(default_opts))
+      if not ok then
+        error(debug.traceback(err))
+      end
+    end,
+  }
+end
+
+local function set_timeout_message(bufnr, winid, message)
```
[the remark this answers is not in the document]
Akeboshiwind: Ok, it looks like it works sometimes, but seemingly randomly 🤔

Akeboshiwind (commented):
I had a chance to try it out, it works great other than a few notes

lua/telescope/previewers/buffer_previewer.lua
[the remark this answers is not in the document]
fdschmidt93: > You might want to disable scrolling on these buffers.

I don't think that's possible, how would I do that generally? For instance, you could just as well scroll beyond any other previewed buffer right now apart from in the buffer picker. But I think I know how we can limit scrolling.

> Also, for some reason locally the text isn't centered (I love how this display looks btw)

Thanks! Can you reproduce this somehow, I c

lua/telescope/config.lua
Conni2461: i would prefer `previewer` just because we currently can already disable with `Telescope find_files previewer=false`. I dont know tho, because maybe i wanna do `:lua require("telescope.builtin").find_files { previewer = { timeout = 1000 } }` and that would override the previewer handle.

Maybe `preview` is fine. I dont wanna write this in stone yet, i need to think about it.

lua/telescope/previewers/buffer_previewer.lua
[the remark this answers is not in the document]
fdschmidt93: I looked into scrolling. The way you have to do scrolling so it works as expected always allows you to have the bottom line as the top line. And it seems like we have a bug in there that the last line is always an empty string, that's why can get seemingly outside the buffer. So outside of scope of this PR.

I'll gladly fix centering though if you can give me a reproducible example :)

fdschmidt93:
Yeah, mime type probably has to be replaced anyways because it's not working on Windows.

lua/telescope/previewers/buffer_previewer.lua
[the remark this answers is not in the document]
Akeboshiwind: I've managed to find a minimal re-production, but it's kind of flakey as it only works 50% of the time. Also, once it's rendered one way (either right or wrong) it stays that way until you reopen nvim.


1. Create a test file:
```bash
$ yes "This is just a sample line appended  to create a big file." | head -n 1000000 > out.txt
$ ls -lah
$ ls -lah
total 116872
drwxr-xr-x   4 user  wheel   128B 14 Sep 16:31 .
drwxrwxrwt  16 root  wheel   512B 14 Sep 16:24 ..
-rw-r--r--   1 user  wheel   1.4K 14 Sep 16:23 minimal.lua
-rw-r--r--   1 user  wheel    56M 14 Sep 16:18 out.txt
```
2. Open nvim using the minimal config `nvim -nu minimal.lua`
  I used the below minimal config, the main differences are:
    - Using your branch
    - Setting the `filesize_limit`

<details><summary>minimal.lua</summary>

```lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'fdschmidt93/telescope.nvim',
        branch = 'feat/preview-timeout',
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup({
      defaults = {
          previewers = {
              filesize_limit = 150
          }
      }
  })
end
end
load_plugins()
require('packer').sync()
```
</details>

3. Run `:Telescope find_files` and navigate to the file
4. Try it a couple times until it works 🤷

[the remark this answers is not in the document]
fdschmidt93: Cannot reproduce :(

https://user-images.githubusercontent.com/39233597/133324669-c9c91bf3-d3c2-4a97-881d-07da3f7e05d8.mp4

Akeboshiwind:
Ah, you have to completely close nvim and then re-try.
Once nvim is open it will be consistent.

I'll try and get a video when I'm home.

lua/telescope/previewers/buffer_previewer.lua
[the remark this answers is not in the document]
fdschmidt93: I have closed and re-opened nvim with minimal init opening `out.txt` and in several other folders with large files (txt files, binaries, etc). Still doesn't happen for me. :/

Akeboshiwind:
Really strange 🤔.
Given that it's intermittent, I'm assuming it's some kind of race
condition? But I'm not sure how that would cause this issue.

lua/telescope/previewers/buffer_previewer.lua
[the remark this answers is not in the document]
fdschmidt93: Ah, I've also realized `filesize_limit` was set wrongly for me with your minimal init lua because we've changed to preview (for the time being). Now it just times out (also as intended though).

Race condition sounds very unlikely, the functions as a whole are adequately scheduled and if we set the message, nothing else should affect the buffer. We need `vim.api.nvim_win_get_width` to adequately fill background of preview and then just lay over the message with an extmark on top of that.

Sorry, out of my depth here as well. Hopefully someone else can also repro to help pin this down?

lua/telescope/previewers/buffer_previewer.lua
fdschmidt93: @Conni2461 

I've looked into mime type detection on Windows for too long and didn't find a satisfactory solution that introduces no dependencies (apart from maybe looking into the registry for the file extension which requires admin rights? :laughing: ). It would apparently easiest for Windows users to download `file` :man_shrugging: 

Do you then want to do mime-type detection for Linux/macOS users only or remove it entirely?

Conni2461: we have mime type detection for `cat` since an eternity. We are doing this: https://github.com/nvim-telescope/telescope.nvim/blob/2c71ffec076dffab56333c0b66f95f0d38024cd2/lua/telescope/previewers/term_previewer.lua#L72-L78

not good because its sync, but we are only doing it if we have the dependency, otherwise we dont support it. In my opinion we dont actually have to support any platform that exists.

fdschmidt93: I guess since it's `utils.get_os_command_output` doesn't work in this context (unless I'd await it, or something? Though would induce another callback I suppose). `capture` seems to work very well here. I can try to make it work nicely with `Job` if you prefer that though.

lua/telescope/previewers/buffer_previewer.lua
fdschmidt93: I think I got them all reverted.

fdschmidt93:
@Conni2461 done here, happy to get your eyes on this once you're available for hopefully final remarks :)

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -9,12 +10,97 @@ local pfiletype = require "plenary.filetype"
...
+  local s = assert(f:read "*a")
+  f:close()
+  if raw then
+    return s
+  end
+  s = string.gsub(s, "^%s+", "")
+  s = string.gsub(s, "%s+$", "")
+  s = string.gsub(s, "[\n\r]+", " ")
+  return s
+end
```
Conni2461, line 31: This is the only thing that looks kinda unexpected. Everything else looks good to me.

I dont get this capture function. First i dont get the reasoning here. https://github.com/nvim-telescope/telescope.nvim/pull/1231#discussion_r710400211 Why wouldnt it work. It executes a command synchronous similar to this function but it uses plenary.job which is a wrapper around [vim.loop.spawn](https://github.com/luvit/luv/blob/master/docs.md#uvspawnpath-options-on_exit). plenary.job has a sync interface. 

Second i would prefer if we use a async version, that isnt blocking because afaik, io.popen is blocking. Plenary.job just abstracts the pain away from using `uv.spawn` but gives us the callback hell. We havent really done a good uv.spawn abstraction that uses plenary.async that we could use here. Also i guess that would require us to rewrite everything :laughing: 

Otherwise looks amazing :) Thanks you very much

fdschmidt93: As soon as `utils.get_os_command_output` is within `vim.loop.fs_stat`, it triggers `SIGABRT` because `Job:sync` relies on `vim.wait` which causes the error signal. I'm not sure, I guess that wouldn't happen if plenary used uv timers?

> Second i would prefer if we use a async version, that isnt blocking because afaik, io.popen is blocking

Yeah, it's blocking. Takes ~ 8ms (`job:sync` is c. 20% faster) for a simple lua file.

> prefer if we use a async version [...] Also i guess that would require us to rewrite everything

Just to check, you mean wrap essentially everything in Job:start for `file` that runs the entire logic inside it? That gets a bit funny though if the user doesn't have file and induces ever more callback hell. Wouldn't it be a good compromise to use `utils.get_os_command_output` outside `vim.loop.fs_stat`. We would only be "blocking" on filetypes plenary doesn't catch anyways.

Conni2461: i forgot we are also in a `vim.loop.fs_stat`. Yeah lets merge. Its good enough. We can fix it later. By getting rid of the callback hell and replacing it with `plenary.async`

clason:
Sorry for not commenting on this earlier: Blindly defaulting to treesitter is dangerous since not everybody even has treesitter installed. As it's still experimental, it should be opt-in, not opt-out.

fdschmidt93:
The default experience should be now the same as before with #1249. As such, now the only thing that has changed is that we allow to more easily opt-out partly or entirely. We default to treesitter if available and otherwise do syntax highlighting.

In doing so, we also test whether ts highlighting is activated for the filetype in question and fallback to syntax highlighting otherwise. Granted, this could be rewritten to probably save maybe 5ms (really only guesstimate) per preview on average via caching, but it should be already fast enough.

So I'd personally just leave it as is. Treesitter opt-in by default will lead to a lot more issues I'd say.

clason:
That was my bad; I didn't see that the check was there and just not working for me due to a (now fixed) bug. Installing `nvim-treesitter` _is_ opting in, so that is fine. 5ms is nothing to sneeze at, but if someone cares about this they're presumably fine with tuning their config and explicitly disabling everything they don't need.

Issue #1038, opened 2021-07-21
Previous location is not appended to the jump list when using lsp_references.


kcivkulis:
### Description

When jumping to a reference in the same file using `:Telescope lsp_references` the previous location is not appended to jump list, therefore ctrl-o doesn't work as expected.

**Expected Behavior**

Previous location is appended to jump list.

**Actual Behavior**

Jump list is not changed.

### Details

<details><summary>Reproduce</summary>

1. Move to some symbol
2. :clearjumps
3. :Telescope lsp_references
4. Select any reference in the same file
5. Observe that jump list is still empty
</details>

<details><summary>Environment</summary>

- nvim --version output: 
```
NVIM v0.6.0-dev+48-g5377b2b00
Build type: Release
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/cc -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -DNDEBUG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto
-DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/kristaps/neovim/build/config -I/home/kristaps/neovim/src -I/home/kristaps/neovim/.deps/usr/include -I/usr/include -I/home/kristaps/neovim/build/src/nvim/auto -I/home/kristaps/neovim/build/include
Compiled by kristaps

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/share/nvim"

Run :checkhealth for more info
```
- Operating system: 
Arch linux
- Telescope commit: 
6646900

</details>

kcivkulis:
This _could_ be fixed by adding `vim.cmd "normal! m'"` to `action_set.edit` function, although this is definitely a hack.

[...]

PR #1234, opened 2021-09-11
Save position into jumplist before 'edit' action.

### LINKED ISSUE #1520: The builtin marks picker does not save the cursor's position to the jump list before jumping to local marks
### Description

The builtin marks picker does not push locations to the jump list when jumping to local marks.

Here is the relevant code:

```lua
  -- action_set.edit in telescope/actions/set.lua
  if row and col then
    local ok, err_msg = pcall(a.nvim_win_set_cursor, 0, { row, col })
    if not ok then
      log.debug("Failed to move to cursor:", err_msg, row, col)
    end
  end
```

`nvim_win_set_cursor` does not save the previous location to the jump list.

### Neovim version

```markdown
NVIM v0.6.0-dev+558-g3ba800f15
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

Manjaro Linux x86_64 -  Kernel: 5.10.79-1-MANJARO

### checkhealth telescope

```markdown
    - OK: plenary installed.

    - OK: fd: found fd 8.2.1
```

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. Insert the character "a" in the first line and add a local marker with `ma`.
3. Insert the character "b" in the first line and add a local marker with `mb`.
4. Insert the character "c" in the first line and add a local marker with `mc`.
5. Start Telescope's mark builtin picker with `:Telescope marks` and pick the entry for the local marker `a`.
6. Start Telescope's mark builtin picker with `:Telescope marks` and pick the entry for the local marker `b`.
7. Start Telescope's mark builtin picker with `:Telescope marks` and pick the entry for the local marker `c`.
8. Open the jump list with `:jump` and note the cursor positions were not added.

### Expected behavior

The jump list should be populated with the positions of the cursor after every jump to a local marker using Telescope's builtin marks picker.

### Actual behavior

The jump list is not populated with the positions of the cursor after every jump to a local marker using Telescope's builtin marks picker.

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

caojoshua:
fixes https://github.com/nvim-telescope/telescope.nvim/issues/1038
fixes #1520

Whenever telescope does not sort of edit/selection action, it calls `nvim_win_set_cursor()`, which does not add anything to the jumplist. As noted in the issue, this affects `lsp_references` picker. This PR adds the current location to the jumplist, as [neovim builtin LSP does when jumping to location](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/util.lua#L1015-L1021).

Builtin LSP also adds the current word to the tagstack. I didn't add it in this PR because I don't want to copy and paste code, but if we want it I can quickly add it in.

This change affects all edit/selection actions, and I'm not sure if its desirable in all cases and pickers. Open to discussion.

### DIFF
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -129,7 +129,12 @@ action_set.edit = function(prompt_bufnr, command)
 
   local entry_bufnr = entry.bufnr
 
+  local picker = action_state.get_current_picker(prompt_bufnr)
   require("telescope.actions").close(prompt_bufnr)
+
+  if picker.push_cursor_on_edit then
+    vim.cmd "normal! m'"
+  end
 
   if entry_bufnr then
     if not vim.api.nvim_buf_get_option(entry_bufnr, "buflisted") then
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -933,6 +933,7 @@ internal.marks = function(opts)
     },
     previewer = conf.grep_previewer(opts),
     sorter = conf.generic_sorter(opts),
+    push_cursor_on_edit = true,
   }):find()
 end
 
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -45,6 +45,7 @@ lsp.references = function(opts)
       },
       previewer = conf.qflist_previewer(opts),
       sorter = conf.generic_sorter(opts),
+      push_cursor_on_edit = true,
     }):find()
   end)
 end
@@ -151,6 +152,7 @@ lsp.document_symbols = function(opts)
         tag = "symbol_type",
         sorter = conf.generic_sorter(opts),
       },
+      push_cursor_on_edit = true,
     }):find()
   end)
 end
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -110,6 +110,8 @@ function Picker:new(opts)
     tiebreak = get_default(opts.tiebreak, config.values.tiebreak),
     selection_strategy = get_default(opts.selection_strategy, config.values.selection_strategy),
 
+    push_cursor_on_edit = get_default(opts.push_cursor_on_edit, config.values.push_cursor_on_edit),
+
     layout_strategy = layout_strategy,
     layout_config = config.smarter_depth_2_extend(opts.layout_config or {}, config.values.layout_config or {}),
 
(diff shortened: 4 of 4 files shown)


Conni2461:
I am not sure if we want to do it for every picker. Give me a day to think about it and maybe test it (i actually dont know if thats an issue for find_files, grep_string, etc)

caojoshua:
I get you, I'm not sure if I would want to do it for every picker either. But it would also be tedious to leave a mark for some pickers and not others. Not sure if there is a cleaner way to fix the linked issue.

Massolari:
I don't know if it makes sense for every picker, but I came across this issue because I want this for the document_symbols(), so, limit this only for references() maybe won't be good enough

caojoshua:
Seems people have started to care about this, so I started thinking about it again. Here I added a picker option `push_cursor_on_edit` (very open to better naming) to add a jumplist entry when using an `edit` action. With this, you could do:

```
lua require('telescope.builtin').lsp_document_symbols({ push_cursor_on_edit = true})
```

I don't provide `push_cursor_on_edit = true` by default for any pickers, and leave it open to individual configs. Although if users feel strongly about particular pickers having the action on by default, I can make the change in this or another PR.

 Other options I considered:

* Separate actions for `edit_and_push_cursor`. Users can override actions mappings as needed. Disadvantage is that separate mappings are required for each action to edit regular, tab, split, and vspit.
* Push cursor when instantiating the picker. Don't think users want this though, because it will also push to jump list when user exits out of telescope without an edit action.
* Support callbacks on closing Telescope window. Seems unnecessary as I can't think of other use cases for it.

caojoshua:
If we're going to make it opts, then why not also make it config? Users can override opts, so if we leave it out of default config, wouldn't just be an undocumented config without a default?

Conni2461:
Yes thats the point. Its can be an undocumented internal thing until we introduce some real event system where we && people can hook stuff into. Because i dont like the solution but we should still fix this issue. I think it doesnt make sense to configure that on a global level but should rather be configured by us here for example: https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/builtin/lsp.lua#L79

caojoshua:
Only change is I removed `push_cursor_on_edit` from default config.

> but should rather be configured by us here for example: https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/builtin/lsp.lua#L79

Is this something we want in this PR? This change would require a change to every picker that wants the new behavior, and I don't want to be the one making those choices.

Conni2461:
Sorry for the late response. I added the new option to the mentioned builtins and tested it.

Thanks for the PR and your patients :)

PR #1237, opened 2021-09-12
fix: make qf commands possible from any picker

### LINKED ISSUE #1218: Items added to Quickfix window by `actions.smart_send_to_qflist` don't allow `cfdo` commands
### Description

If I grep via Telescope and add the four found files to the Quickfix with `<C-q>` then the quickfix works as expected. I am able to do something like `:cfdo :call append(4, '"status": "not started"')` to add a line to each file. Beautiful!

However, if I find files with Telescope, then select the files (using Tab) I want and run `["<C-q>"] = actions.smart_send_to_qflist + actions.open_qflist,`, which I used on the basis of info here: https://github.com/nvim-telescope/telescope.nvim/issues/634 then the Quickfix window appears but the results don't allow commands like `:cfdo :call append(4, '"status": "not started"')` – they have no effect on them.

### Neovim version

NVIM v0.5.0-dev+1384-gf2906a466
Build type: Release
LuaJIT 2.1.0-beta3

macOS 11.5.2

### Steps to reproduce

1. Open Telescope
2. Find Files and search for a string
3. From the list, select 2-3 files with Tab
4. Use the `actions.smart_send_to_qflist + actions.open_qflist` (think the default is `<A-q>`)
5. Try and run a cfdo command on those files such as adding a line with `:cfdo :call append(4, '"status": "not started"')`

### Expected behavior

I would expect the Quickfix window here to behave the same way as it works if using Find files and <C-q>.

### Actual behavior

Nothing occurs when running `cfdo` operations.

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

fdschmidt93:
Should close #1218

@benfrain -- could you please a) test this and b) give me confirmation this should not make anything worse (I cannot think of anything)? 

I'm not really a quickfix user yet (though I admittedly should acquire some mastery ;).

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -619,8 +619,8 @@ local entry_to_qf = function(entry)
   return {
     bufnr = entry.bufnr,
     filename = from_entry.path(entry, false),
-    lnum = utils.get_default(entry.lnum, 1),
-    col = utils.get_default(entry.col, 1),
+    lnum = vim.F.if_nil(entry.lnum, 1),
+    col = vim.F.if_nil(entry.col, 1),
     text = text,
   }
 end
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -619,8 +619,8 @@ local entry_to_qf = function(entry)
   return {
     bufnr = entry.bufnr,
     filename = from_entry.path(entry, false),
-    lnum = entry.lnum,
-    col = entry.col,
+    lnum = utils.get_default(entry.lnum, 1),
+    col = utils.get_default(entry.col, 1),
     text = text,
   }
 end


lua/telescope/actions/init.lua
```diff
@@ -619,8 +619,8 @@ local entry_to_qf = function(entry)
   return {
     bufnr = entry.bufnr,
     filename = from_entry.path(entry, false),
-    lnum = entry.lnum,
-    col = entry.col,
+    lnum = utils.get_default(entry.lnum, 1),
```
Conni2461, line 622: I've was reading quite a lot of `builtin/`, `make_entry.lua` today (for readme, missing opts) and saw that we still have quite a lot of usage from these old `utils`. We should move towards `vim.F.if_nil`. And maybe plan other refactors. Clean up old code.

I might do this after the readme rewrite

fdschmidt93: Yeah, I've stuck with `utils.get_default` to remain consistent with rest of telescope.  I'll quickly change this to `vim.F.if_nil` then.

Conni2461: Yeah i wanna move towards utils if they are available in vim and remove them here.

I know that you know about `vim.F.if_nil`, i saw you use it, i just wanted to mention it :laughing:

benfrain:
@fdschmidt93 this works in that the resultant items from a find files in the quickfix list are 'navigatable' (<CR> on them and you go there). Which is the original issue.

However, there is still a related issue:

1. I use Telescope and see a bunch of files
2. I 'select' the files I want to see in the quickfix with tab and press <C-q>
3. All the files from that list go to the Quickfix, not just the selected ones (tried this without the  `smart_send_to_qflist` too with same problem)

In case relevant, here is that piece of my config:

```lua
  pickers = {
    buffers = {
      sort_lastused = true,
      mappings = {
        i = {
          ["<C-w>"] = "delete_buffer",
          ["<C-q>"] = actions.smart_send_to_qflist + actions.open_qflist,
        },
        n = {
          ["<C-w>"] = "delete_buffer",
        },
      },
    },

```

If this simply needs a different issue opening, let me know.

fdschmidt93:
I've quickly tested your scenario and `actions.smart_send_to_qflist` and it works as intended for me (selections if selected, entries if no selection).

Please open a new issue with repro steps with `minimal_init.lua`.

Conni2461:
Its a bug in this configuration (i was just about to comment :laughing:)

You only set it for the buffers builtin. You want it here

```lua
telescope.setup {
  defaults = {
    mappings = {
      i = {
        ["<C-q>"] = actions.smart_send_to_qflist + actions.open_qflist,
      },
    },
  },
  pickers = {
  }
}
```

on default we map `send_to_qflist` to `<c-q>`.

fdschmidt93:
Thanks Conni, sometimes detail slip with work on one monitor and browser/telescope on the other :laughing:

benfrain:
@fdschmidt93 @Conni2461 thanks so much and sorry for the inconvenience!

fdschmidt93:
I'm optimistic this PR is otherwise harmless (since it should only add factually correct info), so I'll merge.

PR #1238, opened 2021-09-13
fix: Docs missing quotes around 'prompt_prefix'

miguelmoraperea:
The 'prompt_prefix' must be wrap in quotes when passed as part of lua
table options when calling a Telescope built-in function.

This commit fixes the missing quotes on the documentation example.


Conni2461:
Thanks :)

Issue #1241, opened 2021-09-14
 a new option to disable the change  of order item by string length


windwp:
**Is your feature request related to a problem? Please describe.**
https://github.com/nvim-telescope/telescope-fzf-native.nvim/issues/30#issuecomment-918873675

 Sample I have a list of files and  I already sorted that list.  I want the telescope to display and sort with fzf-native  but
that sort doesn't need to be reorder by string length.

Run that code with my PR.
```lua

--  setup telescope with extennsion fzf native and run that code

local finders = require('telescope.finders')
local pickers = require('telescope.pickers')
local actions = require('telescope.actions')

local conf = require('telescope.config').values

--- a list of items I already sorted I only want to filter without effect by string length.

local items = {
    'lua/abcdef/init.lua',
    'init.lua',
    'lua/abcdef/ok/init.lua',
    'lua/fedsadsa/ok.lua',
}

local list = {}

for _, value in pairs(items) do
    table.insert(list, {
        line = 0,
        display = value,
        key = value,
        text = value,
    })
end

local make_entry = function()
    return function(entry)
        return {
            valid = true,
            value = entry.text,
            ordinal = entry.key,
            display = entry.display,
        }
    end
end

local opts = require('telescope.themes').get_dropdown()

-- CHANGE a new  OPTION HERE
opts.reorder_by_length = false

pickers.new(opts, {
    prompt_title = 'Test reorder',
    finder = finders.new_table({
        results = list,
        entry_maker = make_entry(),
    }),
    sorter = conf.file_sorter(),
    attach_mappings = function()
        actions.select_default:replace(function(prompt_bufnr)
            actions.close(prompt_bufnr)
        end)
        return true
    end,
}):find()

```
run that code and input `init`

windwp:
https://github.com/nvim-telescope/telescope.nvim/pull/1401

PR #1242, opened 2021-09-14
feat: add new option to change order of result

windwp:
check issue #1241


Conni2461:
Not exactly what i had in mind. I had more of a function in mind that does the sorting (comparision, whatever) can then be configured by an extension. So fzf can have a value like reorder_by_length_for_values_with_same_score. Or keep_order, idk. That would be better and more flexible in my opinion.

fdschmidt93:
Agreed, I think this should be a user configurable function that defaults to sorting by length of the ordinal. This is essentially custom tie breaking that applies to several use cases. This should easily be overridden for any picker by passing the function, e.g. `opts.tiebreak = function(listed_entry, current_entry)` or something.

See also https://github.com/nvim-telescope/telescope.nvim/pull/987#discussion_r686623272

windwp:
i agree that option is too simple.  that code should move to a sorter. 

function add_entry should only do the task to add an entry. we don't need to change the order again on that function.

fdschmidt93:
> that code should move to a sorter.

Not sure, depends on what we want to do. It's challenging to 
* achieve full flexibility in breaking ties from within the sorter: how you break ties always has to be linked _unambiguously_ to the score in principled fashion
* extend your own tie breaker: coming up with the right logic for your criterion can be hard

But it would be the fastest option for sure.

> function add_entry should only do the task to add an entry.

I don't see a conflict here, breaking ties can be seen as an entry in order. Having the function in add entry makes it very easy to extend how to break ties, but of course gets more expensive if many entries have the same score.

I'd personally lean towards just having it within add entry as in practical use cases that suffices at acceptable speeds.

PR #1243, opened 2021-09-14
Fix #1182: pickers use entry.filename as default and fallback to entry.value

### LINKED ISSUE #1182: rg: file_ignore_patterns will be ignored in searches
### Description

This is my setup:
```
local telescope = require("telescope")
telescope.setup {
    defaults = {
        file_sorter = require "telescope.sorters".get_fzy_sorter,
        generic_sorter = require "telescope.sorters".get_fzy_sorter,
        vimgrep_arguments = {
            "rg",
            "--color=never",
            "--no-heading",
            "--with-filename",
            "--line-number",
            "--column",
            "--smart-case",
            "--hidden"
        },
        file_ignore_patterns = {
            "node_modules",
            ".work/.*",
            ".cache/.*",
            ".idea/.*",
            "dist/.*",
            ".git/.*"
        },
       ... some other stuff ...
    }
}
```

When I use `live_grep` and search for `work.build` nothing is found. I have a lot of files including the string `.work/build/...`. When I remove the pattern `".work/.*"` from `file_ignore_patterns` I find all the matches and the output matches that of `rg` run in a shell.

Perhaps I am doing something wrong here.

### Neovim version

NVIM v0.6.0-dev+nightly-353-gff0833cb4

macOS 11.5

### Steps to reproduce

See above.

### Expected behavior

`file_ignore_patterns` should only affect the files that are searched.

### Actual behavior

See above.

### Minimal config

```Lua
See above.
```

TheLeoP:
Hi again @fdschmidt93 .

So, the diff is just two lines and it's pretty much exactly what you suggest on #1182 issue. I'm ready to make any changes if needed.

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -1,7 +1,6 @@
 require "telescope"
 
 local a = vim.api
-local if_nil = vim.F.if_nil
 
 local async = require "plenary.async"
 local await_schedule = async.util.scheduler
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -1,6 +1,7 @@
 require "telescope"
 
 local a = vim.api
+local if_nil = vim.F.if_nil
 
 local async = require "plenary.async"
 local await_schedule = async.util.scheduler
@@ -1049,7 +1050,7 @@ function Picker:get_result_processor(find_id, prompt, status_updater)
     -- a ton of time on large results.
     log.trace("Processing result... ", entry)
     for _, v in ipairs(self.file_ignore_patterns or {}) do
-      local file = type(entry.value) == "string" and entry.value or entry.filename
+      local file = if_nil(entry.filename, type(entry.value) == "string" and entry.value) -- false if none is true
       if file then
         if string.find(file, v) then
           log.trace("SKIPPING", entry.value, "because", v)
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -1049,7 +1049,7 @@ function Picker:get_result_processor(find_id, prompt, status_updater)
     -- a ton of time on large results.
     log.trace("Processing result... ", entry)
     for _, v in ipairs(self.file_ignore_patterns or {}) do
-      local file = if_nil(entry.filename, type(entry.value) == "string" and entry.value) -- false if none is true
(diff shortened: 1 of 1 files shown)


lua/telescope/pickers.lua
```diff
@@ -1,6 +1,7 @@
 require "telescope"
 
 local a = vim.api
+local if_nil = vim.F.if_nil
```
fdschmidt93, line 4: Small nit, just leave `vim.F.if_nil` for now. We'll migrate from `utils.get_default` to internal functions eventually. Therefore better not to introduce shorthands for now here.

fdschmidt93:
Did you test that this works well with the `builtin.{live_grep, grep_string, find_files}` pickers as I'd hoped? I'll have time to test drive this latest on Saturday myself.

TheLeoP (commented):
Undesrtood

TheLeoP:
Yes, I tested the change for `builtin.{live_grep, grep_string, find_files}` and it worked as expected (or at least it seemed to work as expected).

TheLeoP:
I correct myself: the change work as expected on `builtin.{live_grep, grep_string}`. But, not mattering the config, find_files always shows files like `work/should_not_show` and never shows files like `.work/should_not_show`

Conni2461:
whats the actual config? i briefly tested it and it works for me

TheLeoP:
This is the config:
```lua
telescope.setup {
    defaults = {
        vimgrep_arguments = {
            "rg",
            "--color=never",
            "--no-heading",
            "--with-filename",
            "--line-number",
            "--column",
            "--smart-case",
            "--hidden"
        },
        file_ignore_patterns = {
            ".work\\.*",
            ".work/.*",
        },
    }
}
```

I guess it's because of the program I'm using with find_files and the fact that I'm not passing any arguments to it (so it doesn't show the files starting with a dot because they are hidden). If that's the case, the change should be working as intended.

Conni2461:
It works for me i just tested it. Thanks for the fix.

The file ignore pattern is not a 100% correct lua regex. You should read `:help telescope.defaults.file_ignore_patterns`.

If you want to ignore all files under the directory work. Then it would be `work/.*` if you only want to ignore all files under the directory .work it would be `%.work/.*`. etc

Issue #445, opened 2021-01-19
Loading swap files errors telescope/actions/init.lua:127: Vim(edit):E325: ATTENTION


rockerBOO:
### Description

Opening files under swap, error, when recovering from telescope. The error does not seem to prevent me from opening the file after pressing enter. Lower priority but annoying for anyone using swap files.

**Expected Behavior**
After recovering the file from swap, it should open the file without erroring.

**Actual Behavior**
```
Found a swap file by the name "~/.config/local/share/nvim/swap//%home%rockerboo%.dotfiles%config%nvim%lua%plugin%tel
escope.lua.swp"
          owned by: rockerboo   dated: Tue Jan 19 14:39:52 2021
         file name: ~rockerboo/.dotfiles/config/nvim/lua/plugin/telescope.lua
          modified: no
         user name: rockerboo   host name: Monolith
        process ID: 32451 (STILL RUNNING)
While opening file "/home/rockerboo/.config/nvim/lua/plugin/telescope.lua"
             dated: Tue Jan 19 13:57:13 2021

(1) Another program may be editing the same file.  If this is the case,
    be careful not to end up with two different instances of the same
    file when making changes.  Quit, or continue with caution.
(2) An edit session for this file crashed.
    If this is the case, use ":recover" or "vim -r /home/rockerboo/.config/nvim/lua/plugin/telescope.lua"
    to recover the changes (see ":help recovery").
    If you did this already, delete the swap file "/home/rockerboo/.config/local/share/nvim/swap//%home%rockerboo%.d
otfiles%config%nvim%lua%plugin%telescope.lua.swp"
    to avoid this message.

Swap file "~/.config/local/share/nvim/swap//%home%rockerboo%.dotfiles%config%nvim%lua%plugin%telescope.lua.swp" alre
ady exists!
[O]pen Read-Only, (E)dit anyway, (R)ecover, (Q)uit, (A)bort:
```

```
"~/.config/nvim/lua/plugin/telescope.lua" 78 lines, 2120 characters
Using swap file "~/.config/local/share/nvim/swap//%home%rockerboo%.dotfiles%config%nvim%lua%plugin%telescope.lua.swp
"
Original file "~/.dotfiles/config/nvim/lua/plugin/telescope.lua"
Recovery completed. Buffer contents equals file contents.
You may want to delete the .swp file now.


[...]

PR #1247, opened 2021-09-15
fix: wrap vim.cmd in pcall to avoid error

gavynriebau:
When opening a file under swap from the "fd" finder results an error was
printed after the file was opened, wrapping the call to "vim.cmd" in a
pcall avoids the error.

Issue: #445


Conni2461 (approved):
LGTM thanks :)

PR #1248, opened 2021-09-15
feat: support <pattern> for git-for-each-ref

aenichols:
Wanted to add the support for using \<pattern\> for [git-for-each-ref](https://git-scm.com/docs/git-for-each-ref)

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -807,6 +807,9 @@ builtin.git_branches({opts})                          *builtin.git_branches()*
 
     Parameters: ~
         {opts} (table)  options to pass to the picker
+        
+    Fields: ~
+        {pattern} (string)  specify the pattern to match all refs
 
     Options: ~
         {cwd}          (string)   specify the path of the repo
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -807,14 +807,12 @@ builtin.git_branches({opts})                          *builtin.git_branches()*
 
     Parameters: ~
         {opts} (table)  options to pass to the picker
-        
-    Fields: ~
-        {pattern} (string)  specify the pattern to match all refs
 
     Options: ~
         {cwd}          (string)   specify the path of the repo
         {use_git_root} (boolean)  if we should use git root as cwd or the cwd
                                   (important for submodule) (default: true)
+        {pattern}      (string)   specify the pattern to match all refs
 
 
 builtin.git_status({opts})                              *builtin.git_status()*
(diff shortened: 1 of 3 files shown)


lua/telescope/builtin/git.lua
```diff
@@ -195,7 +195,14 @@ git.branches = function(opts)
     .. "%(authorname)"
     .. "%(upstream:lstrip=2)"
     .. "%(committerdate:format-local:%Y/%m/%d %H:%M:%S)"
-  local output = utils.get_os_command_output({ "git", "for-each-ref", "--perl", "--format", format }, opts.cwd)
+
+  local cmd = { "git", "for-each-ref", "--perl", "--format", format }
```
fdschmidt93, line 199: Nitty remark, this can be simplified as follows.

```lua
 local output = utils.get_os_command_output({  "git", "for-each-ref", "--perl", "--format", format, opts.pattern }, opts.cwd)
```

since lua does not insert `nil` into tables, e.g.`print(vim.inspect({1, nil}))` which prints `{1}`

doc/telescope.txt
```diff
@@ -592,6 +592,9 @@ builtin.git_branches({opts})                          *builtin.git_branches()*
 
     Parameters: ~
         {opts} (table)  options to pass to the picker
+        
+    Fields: ~
```
fdschmidt93, line 596: Typically, we add docs via EmmyLua annotations which are then converted into docs via https://github.com/tjdevries/tree-sitter-lua through CI. 

Here, you can fix it by just adding the identical annotation analogously to `lua/telescope/builtin/init.lua` right below L190 as per other annotations in the file with `@field` argument.

lua/telescope/builtin/git.lua
aenichols: I actually tried that but I was getting an error when I didn't provide a pattern. I will double check!

doc/telescope.txt
aenichols: I think I got it.

lua/telescope/builtin/git.lua
fdschmidt93: Ok :) maybe it doesn't work for some reason in this context? Not sure

aenichols: No you were right, just tried it. I pushed up a change. idk

Conni2461:
LGTM thanks :)

I took the liberty to rebase it and rerun docgen + stylua. I just merged in a huge docs pr thats why there was a conflict and i felt guilty for not doing it the other way around :laughing:

PR #1249, opened 2021-09-16
fix: fallback to syntax hl if treesitter fails

fdschmidt93:
Slight logical error that syntax highlighting fallback doesn't kick in in case treesitter test returns false while having `preview = { treesitter = true}}`.

@clason I think that should fix your issues if I understood them correctly. Apologies, my bad.


clason:
Yes, that fixes it. Thanks!

PR #1250, opened 2021-09-17
feat: migrate to vim.diagnostic

fdschmidt93:
I need my `lsp_document_diagnostics` :laughing: I'll finish this tomorrow.

It's not super clean yet but this essentially makes `lsp_{document, workspace}_diagnostic` use `vim.diagnostics` if found.

TODO:
- [ ] clean up sign linking
- [ ] clean up entry maker for diagnostics

Among the changes:
- `vim.diagnostic.severity` has everything in uppercase vs `vim.lsp.diagnostic.DiagnosticSeverity` which doesn't line up as nicely when linking signs (TODO).. but why?
- diagnostics are now returned more consistent, though different than before: the bufnr is now in every diagnostic (as opposed to being the key of a table), regardless of whether for current buffer or not, which makes loops incompatible.

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -3,6 +3,10 @@ local Job = require "plenary.job"
 
 local log = require "telescope.log"
 
+local has_vim_diagnostic = vim.diagnostic ~= nil
+-- TODO why is vim.diagnostic all uppercase..
+local type_diagnostic = vim.lsp.protocol.DiagnosticSeverity
+
 local utils = {}
 
 utils.get_separator = function()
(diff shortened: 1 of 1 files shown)


tami5:
😟  < praying it will not break for 0.5 users, because he is currently one 😆

lua/telescope/utils.lua
```diff
@@ -3,6 +3,10 @@ local Job = require "plenary.job"
 
 local log = require "telescope.log"
 
+local has_vim_diagnostic = vim.diagnostic ~= nil
+-- TODO why is vim.diagnostic all uppercase..
+local type_diagnostic = vim.lsp.protocol.DiagnosticSeverity
```
tami5, line 8: I guess to match the protocol naming convention

fdschmidt93: Am I missing something? https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#diagnosticSeverity

Conni2461 (commented):
What was the reasoning for putting all the lsp stuff in utils?

Can we not just move it to local functions inside lsp? I dont like the idea that other people may build stuff that relies on these interfaces. They should not do that so we should hide it imo

fdschmidt93:
No reason, I was just new at this at the time :sweat_smile: I'll actually wait for `vim.diagnostic` to become more stable as upstream issues with `lsp_document_diagnostics` are now fixed, we might also clean up a bit more broadly then (e.g. severity sorting and filtering via core I think).

tjdevries:
I don't know if we should call it `lsp_*` anymore, since it is not `lsp_*` anything.

Perhaps leave the `lsp_*` functions for now, but warn users on 0.6 they are they are not what they should use. 0.5 users can not use vim.diagnostic yet because it is not merged for them.

lua/telescope/utils.lua
fdschmidt93: To follow up

> We use our own severity levels to decouple vim.diagnostic from LSP. These are designed similarly to vim.log.levels and currently include: `vim.diagnostic.severity.{ERROR, WARN, INFO, HINT}. In practice, these match the LSP diagnostic severity levels exactly, but we should treat this as an interface and not assume that they are the same. The "translation" between the two severity types is handled transparently in vim.lsp.diagnostic.

As the main post of the original [PR](https://github.com/neovim/neovim/pull/15585) has been edited to explain this. Makes sense, it's just convenient when it lines up with the signs as well :sweat_smile:

Conni2461:
@fdschmidt93 Can i finish `Telescope diagnostics_*` ? i want this in first release. `Telescope lsp_*_diagnostics` will probably be an alias with deprecation warning. Or do you have the time and want to finish it?

fdschmidt93:
No, you're very happy to take over, thanks :)

Conni2461:
https://github.com/nvim-telescope/telescope.nvim/pull/1553

Conni2461:
Implemented by #1553

Thanks for your initial work :)

Issue #1167, opened 2021-08-26
Show path endings in file previewer


minhduc0711:
**Is your feature request related to a problem? Please describe.**
Some of my paths are too long and I want to see the actual filenames, but I don't want to use `path_display= "shorten"`.

**Describe the solution you'd like**
An option to display the paths from their endings.

**Additional context**
By default, the paths are "left-aligned". I want them to be "right-aligned" instead.
![image](https://user-images.githubusercontent.com/34294427/130900628-f4d96e38-fef7-4492-961d-04f09bab2ed8.png)

Conni2461:
we have `help telescope.defaults.dynamic_preview_title` but that is also doesnt truncate on the left side so i dont think it will help you for java.

Should be fairly easy to extend `dynamic_preview_title` but our strings library doesnt support truncate on the left side. https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/strings.lua#L95-L116

So this would be a double PR to plenary and telescope. PRs welcome

minhduc0711:
But `dynamic_preview_title` only changes the previewer's title. I prefer the paths to be left truncated in the "Results" window as well, is it possible to add a new `path_display` option for this?

joeycozza:
I know this is NOT a perfect solution, but until someone smarter than me can implement a solution that takes into account the view width, you can use something like this. This works for me because I almost always have the same sized width. So you could just tweak that maxLength variable to whatever you want and this will cut off the start of the strings that are too long and add "..." to them.

```lua
 path_display = function (opts, path)
      local pathLength = string.len(path)
      local maxLength = 41
      if pathLength > maxLength then
        path = '...'..string.sub(path, pathLength - maxLength - 3)
      end
      return path
    end,
```
![Screen Shot 2021-09-16 at 3 29 33 PM](https://user-images.githubusercontent.com/3885959/133688379-174d8b3e-3734-4a14-a1bd-083d65c59184.png)

l-kershaw:
@minhduc0711 I am closing this, as there is now a `truncate` option for `path_display` (see `:h telescope.defaults.path_display` for more details)
If this doesn't fix your issue, feel free to reopen 🙂

PR #1254, opened 2021-09-17
feat: `truncate` option for `path_display`

l-kershaw:
**WIP**

Use the new left truncation of strings from plenary to truncate paths.
Currently just a minimal implementation that doesn't take stuff like "are there icons" or "how long is the selection_caret" into account.
Also think it _could_ be a bit of a performance hit with this implementation, but haven't thought about it properly yet.

Demo:
image

Could be useful for [this issue](https://github.com/nvim-telescope/telescope.nvim/issues/1167) and similar to implementation in [this comment](https://github.com/nvim-telescope/telescope.nvim/issues/1167#issuecomment-921268416).

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -1,8 +1,10 @@
-local truncate = require("plenary.strings").truncate
 local Path = require "plenary.path"
 local Job = require "plenary.job"
 
 local log = require "telescope.log"
+
+local truncate = require("plenary.strings").truncate
+local get_status = require("telescope.state").get_status
 
 local utils = {}
 
@@ -363,7 +365,7 @@ utils.transform_path = function(opts, path)
         transformed_path = Path:new(transformed_path):shorten(path_display["shorten"])
       end
       if vim.tbl_contains(path_display, "truncate") or path_display.truncate then
-        local status = require("telescope.state").get_status(vim.api.nvim_get_current_buf())
+        local status = get_status(vim.api.nvim_get_current_buf())
         local width = vim.api.nvim_win_get_width(status.results_win) - 5
         transformed_path = truncate(transformed_path, width, nil, -1)
       end
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -366,7 +366,7 @@ utils.transform_path = function(opts, path)
       end
       if vim.tbl_contains(path_display, "truncate") or path_display.truncate then
         local status = get_status(vim.api.nvim_get_current_buf())
-        local width = vim.api.nvim_win_get_width(status.results_win) - 5
+        local width = vim.api.nvim_win_get_width(status.results_win) - status.picker.selection_caret:len() - 2
         transformed_path = truncate(transformed_path, width, nil, -1)
       end
     end
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -1,3 +1,4 @@
+local truncate = require("plenary.strings").truncate
 local Path = require "plenary.path"
 local Job = require "plenary.job"
 
@@ -360,6 +361,11 @@ utils.transform_path = function(opts, path)
 
       if vim.tbl_contains(path_display, "shorten") or path_display["shorten"] ~= nil then
         transformed_path = Path:new(transformed_path):shorten(path_display["shorten"])
+      end
+      if vim.tbl_contains(path_display, "truncate") or path_display.truncate then
+        local status = require("telescope.state").get_status(vim.api.nvim_get_current_buf())
+        local width = vim.api.nvim_win_get_width(status.results_win) - 5
+        transformed_path = truncate(transformed_path, width, nil, -1)
       end
     end
 
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -365,9 +365,10 @@ utils.transform_path = function(opts, path)
         transformed_path = Path:new(transformed_path):shorten(path_display["shorten"])
       end
       if vim.tbl_contains(path_display, "truncate") or path_display.truncate then
-        local status = get_status(vim.api.nvim_get_current_buf())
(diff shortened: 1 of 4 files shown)


lua/telescope/utils.lua
```diff
@@ -361,6 +364,11 @@ utils.transform_path = function(opts, path)
       if vim.tbl_contains(path_display, "shorten") or path_display["shorten"] ~= nil then
         transformed_path = Path:new(transformed_path):shorten(path_display["shorten"])
       end
+      if vim.tbl_contains(path_display, "truncate") or path_display.truncate then
+        local status = get_status(vim.api.nvim_get_current_buf())
+        local width = vim.api.nvim_win_get_width(status.results_win) - status.picker.selection_caret:len() - 2
```
Conni2461, line 369: We could hide/cache the width in status or opts, depending how long it takes to get the status. prefixed with `__` or so.

We could benchmark it ;)

l-kershaw: Couldn't figure out an easy way to benchmark it since there are lots of moving pieces.
Have stored the length in `opts` as `__length`.

l-kershaw:
I think the "nightly" release isn't available to download once again.
I will rerun the tests when it is back up.

clason:
As is usually the case, just the publish script failed; in this case you can always download the assets directly from the Actions page: https://github.com/neovim/neovim/actions/workflows/release.yml

lua/telescope/utils.lua
```diff
@@ -370,6 +379,11 @@ utils.transform_path = function(opts, path)
   end
 end
 
+utils.calc_result_length = function()
```
Conni2461, line 382: lets make this local

Conni2461 (approved):
Besides nit LGTM thanks :)

lua/telescope/utils.lua
l-kershaw: Agreed, was originally trying to calculate the length elsewhere but couldn't get it to work properly.
Forgot to make it local again 🤦

Conni2461:
Oh another thing. Some docs in config.lua :)

l-kershaw:
This is a good minimal implementation now, and should work reasonably in all circumstances.

One thing I would still like to do is take into account if icons have been disabled.
In this case the paths are truncated slightly shorter than they need to be, as it is allowing space for the icon (which is not present).

Also need to add docs (thanks for the reminder @Conni2461 😅)

Conni2461:
> Also need to add docs (thanks for the reminder @Conni2461 :sweat_smile:)

I also just remembered :laughing: 

> One thing I would still like to do is take into account if icons have been disabled.

Right now it should always be opts (later we might need to check config as well). strdisplaywidth should be 1 and 1 space. We could also just do `strings.strdisplaywidth` with a random icon.

l-kershaw:
I have added docs now

---

@Conni2461 are you thinking that we just hard code that we expect the icon to take up 2 columns of space (at least for now)?
That could work, as only downside is not quite using the full width of the results window for those not using icons.

Conni2461:
I would say, right now we can make this assumption. We (as in telescope builtins) do not use more space i think. If someone has its own icon set in devicons with a space after each it would be 3. That said we could just ask devicons for an icon (lua, c, ...) and use that as length. Then if someone has a different icon set, we would be right again. Under the assumption that they use the same display width for all icons.

We will never be able to catch all cases but if we are ~ 99.9% correct, its good enough for me. They can still provide their own path_display function if its they want something slightly different.

l-kershaw:
I think supporting different icon sets isn't really required, so if we're happy to have a bit of extra space on the right when icons are disabled, then I'm happy to merge.

Conni2461 (approved):
:+1: LGTM Thanks :)

tami5:
First of all thanks @l-kershaw  this is something I wanted for a long time.

I have an issue where the dropdown them is broken and the text is hidden.
Screen Shot 2021-09-19 at 10 30 50 PM
 
With default theme, no text
Screen Shot 2021-09-19 at 10 32 24 PM

relevant part of my config 

```lua
  df.layout_config.horizontal = { prompt_position = "top", width = 0.8, height = 0.8, preview_width = 0.6 }
  df.layout_config.vertical = { height = 0.8, preview_height = 0.5 }
  df.set_env = { COLORTERM = "truecolor", LESS = "iMRS" }
  df.results_title = false
  df.scroll_strategy = "cycle"
  df.selection_strategy = "reset"
  df.sorting_strategy = "ascending"
  df.layout_strategy = "flex"
  df.borderchars = { "─", "│", "─", "│", "┌", "┐", "┘", "└" } -- { "─", "│", "─", "│", "╭", "╮", "╯", "╰" }
  df.mappings = ma
  df.history = { path = "~/.local/share/nvim/databases/telescope_history.sqlite3", limit = 100 }
  df.path_display = "truncate"

  require("telescope").setup {
    defaults = df,
    pickers = require "user.pack.telescope.overrides",
    extensions = ext.config,
  }


```

The wrong positioning of the dropdown maybe caused by another pr. I'll check but text doesn't show when I have path_display = "truncate"

Conni2461:
:laughing: `:help telescope.defaults.path_display` its a table not a string. `df.path_display = { "truncate" }`

tami5:
😆 thanks @Conni2461 ❤️ . 

AMAZING TRULY. I'd love to have extra space on the right and maybe adjustable too. but perhaps this can be looked over later.

Screen Shot 2021-09-19 at 10 49 00 PM

Conni2461:
we could also accept `path_display = { truncate = 2 }` meaning 2 additional spaces. Similar to shorten.

Glad it works for you now :)

tami5:
hmmm unfortunately, it didn't add additional spaces

Conni2461:
> we could also

COULD, not implemented yet. Just an idea

tami5:
I need to change my glasses 😆 . Yah this is exactly what I need todo

PR #1256, opened 2021-09-18
fix: turns out tests weren't really asserting anything

tjdevries:
@l-kershaw I fixed it I think


tjdevries:
can wait til nightly is downloadable again though before we push anything. May be other tests that are failing for other reasons.

Conni2461:
Seems to work for everything that isnt mac os :rofl: 

We can also remove the macos job again.

Conni2461:
done in #1491

PR #1257, opened 2021-09-18
feat: expand file browser functionality

### LINKED ISSUE #584: Renaming files and folder on Telescope file_browser

fdschmidt93:
Supersedes #613 

Closes #584  #1337

Large share of the creds of course to @elianiva and everyone helping on the original PR

The `builtin.file_browser` now

* can toggle between a `{file, folder}_browser` (where `file_browser` shows the currently picked folder) which keep multi selections in sync via caching entries by absolute path
* Enables copying(+), deleting*, moving(+), renaming, creating, and opening (with system default app) files (+: via multi selections, *: supports multi selections)
* Actions are now fully remappable and combinable with other built-in actions

Note on usage:
* Copying/Moving files: multi select and then `<C-y>/y` or `m` (moving only normal mode for now) in target directory

Demo: Moving around files across folders into another folder

https://user-images.githubusercontent.com/39233597/133908731-4097f20a-4a8b-4ba4-99de-429bc24e3ef1.mp4


fdschmidt93:
@Conni2461 what do you think about `builtin.files.{__browse_files, __browse_folders}`? Something like it would be required to implement the demo and I guess broader actions (bc of refreshing), but I'm not crazy about it. Would you prefer a separate file for internal finders or something else? thanks in advance :) Then I can clean this up.

Conni2461:
I dont know. I need to think about this but (for a change) i kinda wanna do fun telescope things this week, so i dont feel like even thinking about it. I am also not that interested about these fs operations anymore. So if you wanna finish this sooner rather than later, you are on your own :)

Originally we planed a simple `file_browser` and an extension for the fs operations. We talked about that on gitter but you cant really search for on messages (that was idk last year beginning this year). I have one mention here on github https://github.com/nvim-telescope/telescope.nvim/pull/290#issuecomment-788234337

If you can wait, we can talk about this next week, because right now i dont know.

fdschmidt93:
Yeah, no worries -- more than earned :) thanks for the pointer. I'll try and think about something I'd happily get merged.

fdschmidt93:
Ok, I think I found a satisfactory solution for the finders switch. I'm pretty happy with this now :)

@tami5 may I please ask you to review / test drive this PR whenever you have too much time this week before Conni takes a look (he's understandably too busy with more important things :))? :) Ideally in conjunction with https://github.com/nvim-lua/plenary.nvim/pull/241 I'll probably keep cleaning up the plenary PR a bit further as mentioned over there.

lua/telescope/actions/init.lua
```diff
@@ -1055,6 +1057,260 @@ actions.which_key = function(prompt_bufnr, opts)
...
+        .. os_sep
+        .. "a"
+        .. os_sep
+        .. "new_file.lua"
+    )
+    return
+  end
+  local fpath = Path:new(finder.path .. os_sep .. file)
+  if fpath:exists() then
+    print "File or folder already exists."
```
tami5, line 1103: I believe in that case it might be better to use `vim.api.nvim_echo` with Error highlight.

lua/telescope/actions/init.lua
```diff
@@ -1055,6 +1057,260 @@ actions.which_key = function(prompt_bufnr, opts)
...
+  -- "../" more common so test first
+  if old_name.filename == "../" or old_name.filename == "./" then
+    print "Please select a file!"
+    return
+  end
+  entry[1] = vim.fn.input("Insert a new name: ", old_name:make_relative())
+
+  -- avoid error on refresh
+  if old_name:is_dir() and entry[1]:sub(-1, -1) ~= os_sep then
+    entry[1] = entry[1] .. os_sep
```
tami5, line 1150: I didn't like the result of this. Before rename I see filename under selection, after rename I see fullpath? it should just update the filename and that's it.

Also, I believe a print message from old_filename to new_filename would be important here.

lua/telescope/builtin/files.lua
```diff
@@ -289,134 +289,120 @@ local function prepare_match(entry, kind)
...
+  opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or cwd
+  opts.files = vim.F.if_nil(opts.files, true)
 
   pickers.new(opts, {
-    prompt_title = "File Browser",
-    finder = opts.new_finder(opts.cwd),
+    prompt_title = opts.files and "File Browser" or "Folder Browser",
+    results_title = opts.files and Path:new(opts.cwd):make_relative(cwd) .. os_sep or "Results",
+    finder = fb_finder(opts),
     previewer = conf.file_previewer(opts),
```
tami5, line 375: got the following error while trying to preview a root directory with no user access to read.

```lua
Error executing vim.schedule lua callback: .../pack/packer/start/plenary.nvim/lua/plen
ary/filetype.lua:103: bad argument #1 to 'ipairs' (table expected, got nil)
```

lua/telescope/builtin/files.lua
```diff
@@ -289,134 +289,120 @@ local function prepare_match(entry, kind)
...
+      map("i", "<C-e>", actions.create_file)
+      map("n", "<C-e>", actions.create_file)
+      map("i", "<C-r>", actions.rename_file)
+      map("n", "<C-r>", actions.rename_file)
+      map("i", "<C-y>", actions.copy_file)
+      map("n", "<y>", actions.copy_file)
+      map("i", "<C-h>", actions.toggle_hidden)
+      map("n", "<h>", actions.toggle_hidden)
+      map("i", "<C-g>", actions.goto_prev_dir)
+      map("n", "g", actions.goto_prev_dir)
```
tami5, line 399: this is a bit slow when the user has g.. mappings in normal mode, it would be cool if somehow `<nowait>` is given (not sure that would fix it though)

lua/telescope/actions/init.lua
```diff
@@ -1055,6 +1057,260 @@ actions.which_key = function(prompt_bufnr, opts)
...
+  end
+
+  old_name:rename { new_name = entry[1] }
```
tami5, line 1153: should check if `old_name == new_name` and if so skip

lua/telescope/builtin/files.lua
```diff
@@ -289,134 +289,120 @@ local function prepare_match(entry, kind)
...
-      map("n", "<C-e>", create_new_file)
+      map("i", "<C-e>", actions.create_file)
+      map("n", "<C-e>", actions.create_file)
+      map("i", "<C-r>", actions.rename_file)
+      map("n", "<C-r>", actions.rename_file)
+      map("i", "<C-y>", actions.copy_file)
+      map("n", "<y>", actions.copy_file)
+      map("i", "<C-h>", actions.toggle_hidden)
+      map("n", "<h>", actions.toggle_hidden)
+      map("i", "<C-g>", actions.goto_prev_dir)
```
tami5, line 398: `<C-h>` make more sense then `<C-g>` for goto_prev_dir, no?

lua/telescope/actions/init.lua
```diff
@@ -1055,6 +1057,260 @@ actions.which_key = function(prompt_bufnr, opts)
...
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local finder = current_picker.finder
+  local parent_dir = Path:new(finder.path):parent()
+
+  if not bypass then
+    if vim.loop.cwd() == finder.path then
+      print "You can't go up any further!"
+      return
+    end
+  end
```
tami5, line 1279: I vote for bypass to be the default, because I expected to freely naviagte fs without much attention to current dir. At first I thought it was a bug, because it prints that then when I got into a dir, then do goto_prev_dir multiple times, it managed to get me out.

Either ways, an expected bug here

tami5 (commented):
Great work @fdschmidt93, I was going to test more but my dinner is ready 😭

lua/telescope/actions/init.lua
fdschmidt93: Ahh, yeah. Thanks for spotting this :) I haven't yet renamed the right file :laughing:  It's probably better done with picker refresh again.

fdschmidt93:
Thanks already for the comments so far @tami5 :)

I'll try to incorporate them some time tomorrow (evening, probably). How do you like the UX otherwise with toggling between the file and folders picker? Something like that for navigation was what I always am missing in nvim-tree or the like, so I've built it in :) Happy to get more ideas if you have some here.

lua/telescope/builtin/files.lua
tami5: 😆 hit c-g until I got to my real home 😆

tami5:
> Thanks already for the comments so far @tami5 :)
> 
> I'll try to incorporate them some time tomorrow (evening, probably). How do you like the UX otherwise with toggling between the file and folders picker? Something like that for navigation was what I always am missing in nvim-tree or the like, so I've built it in :) Happy to get more ideas if you have some here.

I don't use nvim-tree so I'm not sure.
so far I really loved the 'filename<c-e>' it really makes sense as well toggling between folders/files. One ux advantage might be to make folders auto toggle when the user types for example in telescope.nvim dir `lua/te` this shows 0 results and has `/` maybe here should auto-toggle folder picker. Don't you think?

fdschmidt93:
Update for anyone subscribed to the PR as file browser expansion seems to be a rather anticipated feature. As a beacon of hope nvim-lua/plenary.nvim#241 now only needs to undergo (most likely) Conni's review. Then I can clean up #1257 which should hopefully be more straightforward given that the branch is rather advanced.

ikerurda:
Great work @fdschmidt93!
I've noticed that the cwd flag when running :Telescope file_browser cwd=... or require("telescope.builtin").file_browser({ cwd = ... }) doesn't work. Could you check it out?

fdschmidt93:
> I've noticed that the cwd flag when running :Telescope file_browser cwd=... or require("telescope.builtin").file_browser({ cwd = ... }) doesn't work. Could you check it out?

Ah, right now that's `path`. It gets a bit funky to think about the current directory of the file browser versus the directory of the folder browser (which is nvim cwd). I have to revisit that and clean that up.

> And actions.toggle_browser freezes nvim and doesn't open anything (in nvim 0.5.1 MacOS).

Works fine for me on nvim master. Is there any sort of error log? Maybe plenary's scandir relies on some lua version your nvim is not linked against or something? Should be easily tested with

```lua
local pscan = require "plenary.scandir"

-- pscan.ls_async("/", {
--   hidden = true,
--   group_directories_first = true,
--   on_exit = vim.schedule_wrap(function(data, sections)
--     P(data)
--   end),
-- })

print(vim.inspect(pscan.scan_dir(vim.loop.cwd(), {
  hidden = true,
  add_dirs = true,
  depth = 1,
})))
```
Does this work for you and give you files / folders of nvim `cwd`?

ikerurda:
It works now, thanks! :)
But.. why does it fail when mapping actions.move_file to C-m?

fdschmidt93:
> But.. why does it fail when mapping actions.move_file to C-m?

<C-m> sends the same key codes as <CR>, that's why it wasn't mapped by default.

> Also, I think that previously, the mapping C-d was used to scroll down the preview window. You are overwriting it now.

For the time being the mappings are mostly for me to test the file browser. However, the file picker is going to be a very "active" picker and therefore I'd personally lean to have all these actions as defaults because I'd say that is what users want from the file browser (otherwise you can also use find files). In any case, you can always remap that mapping again.

lua/telescope/builtin/files.lua
```diff
@@ -289,134 +289,153 @@ local function prepare_match(entry, kind)
...
-          end
-
-          return rawget(t, rawget({ value = 1 }, k))
-        end
-
-        return function(line)
-          local tbl = { line }
-          tbl.ordinal = Path:new(line):make_relative(opts.cwd)
-          return setmetatable(tbl, mt)
+local fb_finder = function(opts)
```
fdschmidt93, line 336: There's an annoying issue I'll have to look into and is probably not so easily fixed.

1. Multi select
2. Search
3. Find multi selection not highlighted again

This currently happens, because this PR creates new finders eagerly in `_call`. This affects `master` in a different form.

1. Launch `file_browser`
2. Multi select
3. Go to different folder
4. Return & find multi selections gone

Ultimately, the multi selection is inadequately preserved since the entry is created anew from a new finder resulting in a new metatable for the entry, such that `picker:is_multi_selected` doesn't result in `true` for the same entry. I'll have to think about this some more to come up with an hopefully easy fix (maybe add selection as a table w/o `metatable`, if easily possible?).

fdschmidt93: Now the entries are cached by their absolute path to transparently remain multi selections across file and folder browsers.

fdschmidt93:
I'm pretty happy with this PR as a basic replacement for plugins like `nvim-tree.lua`. Basic FS operations seem to work well and fast.

I'd be happy to get testers on this branch, which I'll keep in sync with `master`. In case you want to recursively copy folders, this depends on https://github.com/nvim-lua/plenary.nvim/pull/241 . While it works well for me, please note that this is not "battle-tested" ;)

I haven't followed the discussion around `dev` branch (is it supposed to be just a temp branch or `stable/unstable`-like?), but the revamped folder browser seems to me like a good candidate for `dev/unstable`.

PS Don't get hung up on mappings just yet; I trust anyone who's a beta tester to be able to change them on their own. You're happy to share thoughts on better default mappings though.

l-kershaw:
I'll try to test this when I get a chance.

Re: `dev` branch, the `layout_actions` one is only temporary as there were two PRs that related to the same thing, and changed the configuration method between the two of them. So its just to avoid users having to reconfigure twice between them.

ikerurda:
Thanks @fdschmidt93 !
Have you thought about implementing bulk renaming?
For example as implemented in ranger:
selected files are listed in a nvim buffer and when closed, files are renamed.

I'm don't know much lua (so don't see myself doing this) but really appreciate your work :)

fdschmidt93:
Basic version working.

https://user-images.githubusercontent.com/39233597/140835190-41cab41a-1b30-4b14-9dd1-490eeb882fa1.mp4

fdschmidt93:
@Conni2461 I guess if users are willing to contribute, the `file_browser` will grow into something that's a bit beyond of the scope of `core`. 

Do you prefer spinning it into it's own extension?

ikerurda (commented):
I think there's an error on the lines 1176 and 1177:
Wouldn't it be something like this?

`set_bkm(buf, "n", "<CR>", "<cmd>**lua _G.**__TelescopeBatchRename()<CR>", opts)`
`set_bkm(buf, "i", "<CR>", "<cmd>**lua _G.**__TelescopeBatchRename()<CR>", opts)`

fdschmidt93:
No, it should work all the same with or without `_G.` prefix. I've also checked it by hand just now, the function is properly cleared.

ikerurda:
I was refering also to the missing `<cmd>lua` and the `__TelescopeBatchRename` (before `__TelescopeBulkRename`)

Otherwise from your last commit I'm getting this error:
`E492: Not an editor command: __TelescopeBulkRename()`

fdschmidt93:
Ah sorry, I'm super sleepy :facepalm: apologies,forget entirely what I said. I feel like I read something totally different about clearing the function. Of course you're right, must have gotten it wrong when I renamed it.

Conni2461:
Yeah I think it shouldnt be here. Always should have been an extension. This was my mistake.

fdschmidt93:
Closing the PR accordingly and will release a beta version over at the soon-to-be made public `nvim-telescope/telescope-file-browser.nvim`.

Once the extension is released, we'll probably do a `file-browser-ectomy` of core I suppose.

fdschmidt93:
The extension has now made been made accessible: https://github.com/nvim-telescope/telescope-file-browser.nvim

Please consider it a beta for the time being and see the associated [roadmap](https://github.com/nvim-telescope/telescope-file-browser.nvim/issues/3).

benfrain:
Woohoo! Can’t wait to try this. Thanks for all the hard work 👍👍👍👍

PR #1258, opened 2021-09-18
fix(builtin): internal.help_tags not finding all tags

muniter:
After changes on &rtp in neovim/neovim it's neccessary to use `nvim_get_runtime_file` instead https://github.com/neovim/neovim/pull/15632 

Checked the rest of the codebase and the issue pattern `globpath(&rtp ...)` is not present anywhere else.


Conni2461:
Thank you Founding Father. It a honor to merge it Sir :)

PR #1260, opened 2021-09-19
fix: adjust layout positioning for new indexing

l-kershaw:
Adjustment of positioning indexing for after [this plenary PR](https://github.com/nvim-lua/plenary.nvim/pull/227) is merged.


l-kershaw:
Have now implemented using the new `col=0` ⇒ "horizontally centered".
Note that we aren't able to use the vertical centering due to our tabline adjustments.

PR #1264, opened 2021-09-19
feat(defaults) :enable customization of path_display.truncate

tami5:
followup to #1254 and implementation of[@Conni2461's idea](https://github.com/nvim-telescope/telescope.nvim/pull/1254#issuecomment-922527270)

- [x] Fix the empty of empty lines.
- [x] edit docs
- [x] fix lint

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -330,9 +330,8 @@ local is_uri = function(filename)
 end
 
 local calc_result_length = function(truncate_len)
-  ins(truncate_len)
   local status = get_status(vim.api.nvim_get_current_buf())
-  local len=vim.api.nvim_win_get_width(status.results_win) - status.picker.selection_caret:len() - 2
+  local len = vim.api.nvim_win_get_width(status.results_win) - status.picker.selection_caret:len() - 2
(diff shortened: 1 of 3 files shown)


tami5:
does make sense to make default to 2?

Screen Shot 2021-09-19 at 11 50 39 PM

lua/telescope/utils.lua
```diff
@@ -372,6 +372,11 @@ utils.transform_path = function(opts, path)
       if vim.tbl_contains(path_display, "truncate") or path_display.truncate then
         if opts.__length == nil then
           opts.__length = calc_result_length()
+          if type(path_display.truncate) == "number" then
```
Conni2461, line 375: can we actually do the calculation in the `calc_result_length` local function?

lua/telescope/utils.lua
```diff
@@ -372,6 +372,11 @@ utils.transform_path = function(opts, path)
...
+            opts.__length = opts.__length - path_display.truncate
+          else
+            opts.__length = opts.__length - 2
```
Conni2461, line 378: I think we dont need a default padding

l-kershaw (commented):
I agree with @Conni2461's points, and don't have anything further to suggest.

PR #1267, opened 2021-09-20
docs: make_entry and entry_display

TC72:
Made a start on text to help explain entry_display as there's nothing right now.

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1,3 +1,25 @@
+---@tag telescope.pickers.entry_display
+
+---@brief [[
+---
+--- Each picker has a finder made up of two parts, the results which are the data to be displayed,
+--- and the entry_maker. These entry_makers are functions returned from make_entry functions in 
+--- this document. These will be referrd to as entry_makers.
+---
+--- Every entry maker returns a function which accepts the data to be used for an entry. This 
+--- function will return the final text to be displayed for the entry's result in the result 
+--- panel and an optional table with the color highlights.
+---
+---
+---
+---
+---
+---@brief ]]
+
+
+
+
+
 local entry_display = require "telescope.pickers.entry_display"
 local utils = require "telescope.utils"
 local strings = require "plenary.strings"
--- a/lua/telescope/pickers/entry_display.lua
+++ b/lua/telescope/pickers/entry_display.lua
@@ -6,19 +6,16 @@ entry_display.truncate = strings.truncate
 ---@tag telescope.pickers.entry_display
 
 ---@brief [[
---- Entry Display takes care of formatting the text displayed for each entry.
---- This includes adding highlight colors.
+--- Entry Display is used to format each entry shown in the result panel.
 ---
---- Most pickers will call the create function only once.
---- The configuration.items passed in will provide the column
---- widths which will normally be the same for every entry.
---- This is why a single call is made to create.
+--- Entry Display create() will give us a function based on the configuration
+--- of column widths we pass into to it. We then can use this function n times 
+--- to return a string based on structured input.
 ---
---- The create function will use the column widths to create a generator table.
---- Each item in the table is a function used to format the text in each column,
---- in terms of width and how it's justified.
---- It's also possible for the final column not to have a fixed width, this will
---- be shown in the configuartion as 'remaining = true'.
+--- The create function will use the column widths passed to it in
+--- configaration.items. Each item in that table is the number of characters in
+--- the column. It's also possible for the final column to not have a fixed width,
+--- this will be shown in the configuartion as 'remaining = true'.
 ---
 --- An example of this configuration is shown for the buffers picker
 ---
@@ -38,13 +35,13 @@ entry_display.truncate = strings.truncate
 --- Finally we have also defined the seperator between each column will be the space " ".
 ---
 ---
---- Most pickers store the function returned by create() in a reference called display.
---- This function is called for every entry which is displayed and passes in the actual text
---- which should be used to create the string displayed for the entry.
+--- It's best practice for pickers to store the function returned by create() in a reference
+--- called display. This function is called for every entry which is displayed and passes in the
+--- actual text which should be used to create the string displayed for the entry.
 --- The text is passed in as a table, the number of items will match the number of columns.
 --- Each item can be a string or a table containing the text along with the highlight color.
 ---
---- An example of how the display reference will be used is shown, again for the buffers pickerfunction self:
+--- An example of how the display reference will be used is shown, again for the buffers picker:
 ---
 --- return displayer {
 ---   { entry.bufnr, "TelescopeResultsNumber" },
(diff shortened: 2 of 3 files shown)


Conni2461:
As promised I am helping you on this. You get a real review later today. But i am thinking we should export this to viml with docgen and make annotations for the functions.

For export you need to wrap it in a `brief` block, like here https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/builtin/init.lua#L5

After documenting displayer we should do something similar for make_entry and the purpose for `gen_from_...`, etc

TC72:
This is probably too long winded to be at the top of this section of code.
README.md needs text for Finders which would include entry_maker, some of this could be used there.

Conni2461:
You also need to add the file here: https://github.com/nvim-telescope/telescope.nvim/blob/master/scripts/gendocs.lua#L11

and give it a tag similar to here https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/builtin/init.lua#L1

lua/telescope/pickers/entry_display.lua
```diff
@@ -3,6 +3,76 @@ local strings = require "plenary.strings"
 local entry_display = {}
 entry_display.truncate = strings.truncate
 
+---@tag telescope.pickers.entry_display
+
+---@brief [[
+--- Entry Display takes care of formatting the text displayed for each entry.
+--- This includes adding highlight colors.
```
Conni2461, line 10: I dont know. I dont 100% like the way this is formulated. Entry Display doesnt take care of it, first `create` will create a function based on the input and returns that function. We then can use this function n times to return a string based on structured input.

And the highlighting part is just the second return value

Conni2461: Its basically a `string.format` :laughing:

lua/telescope/pickers/entry_display.lua
```diff
@@ -3,6 +3,76 @@ local strings = require "plenary.strings"
...
+---
+--- Most pickers will call the create function only once.
+--- The configuration.items passed in will provide the column
+--- widths which will normally be the same for every entry.
+--- This is why a single call is made to create.
```
Conni2461, line 15: See above but basically we generate a custom function and we only need to do it once. You can do it n times of course but thats just waist of performance :laughing:

lua/telescope/pickers/entry_display.lua
```diff
@@ -3,6 +3,76 @@ local strings = require "plenary.strings"
...
+---
+--- The create function will use the column widths to create a generator table.
+--- Each item in the table is a function used to format the text in each column,
+--- in terms of width and how it's justified.
```
Conni2461, line 19: I dont think we need to mention the generator table, because thats an internal thing. What comes out is a function that uses this generator tables (array) of functions

lua/telescope/pickers/entry_display.lua
```diff
@@ -3,6 +3,76 @@ local strings = require "plenary.strings"
...
+---     },
+---   }
+---
+--- This shows 4 columns, the first is defined in the opts as the width we'll use when display_string
+--- the number of the buffer. The second has a fixed width of 4 and the 3rd column's widht will be 
+--- decided by the width of the icons we use. The fourth column will use the remaining space.
+--- Finally we have also defined the seperator between each column will be the space " ".
+---
+---
+--- Most pickers store the function returned by create() in a reference called display.
```
Conni2461, line 41: Rather than saying "Most pickers" we can just say its best practice.

lua/telescope/pickers/entry_display.lua
```diff
@@ -3,6 +3,76 @@ local strings = require "plenary.strings"
...
+--- The final layer above this is the make_entry function.
+--- This is called with the opts for the picker and it also returns a function.
+--- That function accepts the entry and on eof it's returns is the make_display function
+--- from above.
+---
+--- At the very end of this process when the picker calls make_entry with an entry to be displayed
+--- the final return 2 values, final_str and highlights.
+--- final_str is all the text to be displayed for this entry as a single string. If parts of the
+--- string are to be highlighted they will be described in the highlights table. Each highlight
+--- has a table containing the highlight color, along with the start and end position in final_str.
```
Conni2461, line 73: I know you moved that documentation away from `make_entry.lua` but lets talk about how `make_entry` works over at that file. It would be awesome if we could document both things :)

Thanks :)

Conni2461:
No idea why docgen ci isnt running right now :thinking:

TC72:
I've made some changes and made a very rough start on using some of this documentation  in `make_entry.lua`.

tjdevries:
Btw, I updated some stuff for entry maker in #1473 as an unrelated commit :laughing:

lua/telescope/make_entry.lua
```diff
@@ -1,3 +1,25 @@
+---@tag telescope.pickers.entry_display
```
Conni2461, line 1: tag is wrong

lua/telescope/make_entry.lua
```diff
@@ -1,3 +1,25 @@
...
+---@brief [[
+---
+--- Each picker has a finder made up of two parts, the results which are the data to be displayed,
+--- and the entry_maker. These entry_makers are functions returned from make_entry functions in 
+--- this document. These will be referrd to as entry_makers.
+---
+--- Every entry maker returns a function which accepts the data to be used for an entry. This 
+--- function will return the final text to be displayed for the entry's result in the result 
+--- panel and an optional table with the color highlights.
+---
```
Conni2461, line 12: Not just display. `ordinal`, `lnum`, `col`. We are basically taking (most of the time) a string and making it in a structured table that the picker can use to do things.

lua/telescope/pickers/entry_display.lua
```diff
@@ -3,6 +3,62 @@ local strings = require "plenary.strings"
 local entry_display = {}
 entry_display.truncate = strings.truncate
 
+---@tag telescope.pickers.entry_display
+
+---@brief [[
+--- Entry Display is used to format each entry shown in the result panel.
+---
+--- Entry Display create() will give us a function based on the configuration
+--- of column widths we pass into to it. We then can use this function n times
```
Conni2461, line 12: "into to it"

lua/telescope/pickers/entry_display.lua
```diff
@@ -3,6 +3,62 @@ local strings = require "plenary.strings"
...
+---   }
+---
+--- This shows 4 columns, the first is defined in the opts as the width we'll use when display_string
+--- the number of the buffer. The second has a fixed width of 4 and the 3rd column's widht will be 
+--- decided by the width of the icons we use. The fourth column will use the remaining space.
+--- Finally we have also defined the seperator between each column will be the space " ".
+---
+---
+--- It's best practice for pickers to store the function returned by create() in a reference
+--- called display. This function is called for every entry which is displayed and passes in the
```
Conni2461, line 39: We dont care how the reference is named. The important part is that its stored outside of the function reference which is passed to `display`, because `display` is called basically inside a loop. And creating it multiple time is a waste

lua/telescope/make_entry.lua
```diff
@@ -1,3 +1,25 @@
...
+---
+---
```
Conni2461, line 14: We also wanna mention here that `display` can be a string or a function reference. And we should mention how often `display` will be called.

Maybe we wanna mention tjs new approach on overriding only parts of a entry_maker. You might wanna ask him for a example because i already forgot. :grimacing:

Conni2461:
We should try to include these documentation additions in the first release, how much time do you have this month? I'll try to have a first release at the end of the month /year :laughing:

TC72:
It depends, supposed to be going to Italy for 11 days over Xmas.
My office party in London just got cancelled and I'm expecting that trip to be next.

Conni2461:
Documentation move into #1850 

Rebase was not possible. I reworded and cleaned it up a little bit. Thanks for the initial work :)

PR #1270, opened 2021-09-20
break: c rewrite

Conni2461:
Nobody intends to rewrite telescope in c.

There is stuff missing (truncate) and we can play around with different data structures. And other ideas. This is basically a WIP / proof of concept branch.

benchmark script shows for max sorting 10k items (thats where this should help == scrolling):
```
Benchmark Group: 'lua vs c ffi' -----------------------
Benchmark #1: 'c ffi'
  Time(mean ± σ):    226.7 ms ±  27.7 ms
  Range(min … max):  187.3 ms … 262.0 ms  10 runs
Benchmark #2: 'lua'
  Time(mean ± σ):    1.3 s ± 181.6 ms
  Range(min … max):  1.1 s … 1.6 s  10 runs
Summary
  'c ffi' ran
  5.8 ± 1.1 times faster than 'lua'
  ```

Edit: Anyone who wants to test it. You need to run `make` in telescope directory.


tami5:
PURE SPEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEDDDDDDDDDDD

fdschmidt93:
I've actually started converting this to rust as part of my learning the language and my progress might be of interest to some:

https://github.com/fdschmidt93/telescope.nvim/blob/feat/rs-manager/src/lib.rs

The (super simple) implementation is pretty much as fast as the current linked list (same mean, much lower standard deviation, ie lua linked list might be faster/slower). What's interesting is that it scales super well to having _a lot_ more entries (think 10K+ rather than 1K). But yeah, maybe a heap or linked list is better.

PR #1273, opened 2021-09-21
feat: filetype_hook & improved docs; fix preview partial override

### LINKED ISSUE #1288: Add option to customize the character used in the background of the "buffer_previewer" in function "set_timeout_message"
___First of all, thank you all very much for the great work, this plugin is awesome!___ 🔭🚀

The is no problem whatsoever. I just want more options for customization.

I would like to have an option in the function ```require('telescope').setup()``` to customize the character used to make the pattern of the preview background when the function [```set_timeout_message```](https://github.com/nvim-telescope/telescope.nvim/blob/4816a27d7680feed293d21a945c476947af20d29/lua/telescope/previewers/buffer_previewer.lua#L59) is called. Currently the character is hardcoded as "/" [right here](https://github.com/nvim-telescope/telescope.nvim/blob/4816a27d7680feed293d21a945c476947af20d29/lua/telescope/previewers/buffer_previewer.lua#L67).

For now, I'm using my fork with a personal branch, where I hardcoded the character to a more pleasant one (in my opinion): "·".

**Additional context**
I'm only opening this here because I was not sure about what would be the best place to put such option in order to submit a PR.

___Again, thanks everyone for the great work on this plugin.___

fdschmidt93:
~~I think some users will appreciate this `head` example using plenary (which I used b/c I wasn't sure whether `head` command is fully OS-agnostic, whereas plenary implementation looks like it is).~~

PR got quite a bit broader:
- Improve docs
- Introduce filetype_hook
- Allow setting highlight groups for `TelescopePreviewMessage` and `TelescopePreviewMessageFillchar`
- Allow setting `msg_bg_fillchar` via `preview` in `telescope.setup`
- factor out `set_timeout_message` to `set_preview_message`

And closes #1288

### DIFF
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -402,6 +402,7 @@ append(
     filesize_limit = 25,
     timeout = 250,
     treesitter = true,
+    msg_bg_fillchar = "╱",
   },
   [[
     This field handles the global configuration for previewers.
@@ -426,14 +427,52 @@ append(
                           complete within `timeout` milliseconds.
                           Set to false to not timeout preview.
                           Default: 250
-      - hook(s):          Function(s) that takes `(filepath, bufnr, opts)`
-                          to be run if the buffer previewer was not shown due to
-                          the respective test.
-                          Available hooks are: {mime, filesize, timeout}_hook, e.g.
-                          preview = {
-                            mime_hook = function(filepath, bufnr, opts) ... end
-                          }
-                          See `telescope/previewers/*.lua` for relevant examples.
+      - hook(s):          Function(s) that takes `(filepath, bufnr, opts)`, where opts
+                          exposes winid and ft (filetype).
+                          Available hooks (in order of priority):
+                          {filetype, mime, filesize, timeout}_hook
+                          Important: the filetype_hook must return true or false
+                          to indicate whether to continue (true) previewing or not (false),
+                          respectively.
+                          Two examples:
+                          local putils = require("telescope.previewers.utils")
+                          ... -- preview is called in telescope.setup { ... }
+                            preview = {
+                              -- 1) Do not show previewer for certain files
+                              filetype_hook = function(filepath, bufnr, opts)
+                                -- you could analogously check opts.ft for filetypes
+                                local excluded = vim.tbl_filter(function(ending)
+                                  return filepath:match(ending)
+                                end, {
+                                  ".*%.csv",
+                                  ".*%.toml",
+                                })
+                                if not vim.tbl_isempty(excluded) then
+                                  putils.set_preview_message(
+                                    bufnr,
+                                    opts.winid,
+                                    string.format("I don't like %s files!",
+                                    excluded[1]:sub(5, -1))
+                                  )
+                                  return false
+                                end
+                                return true
+                              end,
+                              -- 2) Truncate lines to preview window for too large files
+                              filesize_hook = function(filepath, bufnr, opts)
+                                local path = require("plenary.path"):new(filepath)
+                                -- opts exposes winid
+                                local height = vim.api.nvim_win_get_height(opts.winid)
+                                local lines = vim.split(path:head(height), "[\r]?\n")
+                                vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, lines)
+                              end,
+                            }
+                          The configuration recipes for relevant examples.
+                          Note: if plenary does not recognize your filetype yet --
+                          1) Please consider contributing to:
+                             $PLENARY_REPO/data/plenary/filetypes/builtin.lua
+                          2) Register your filetype locally as per link
+                             https://github.com/nvim-lua/plenary.nvim#plenaryfiletype
                           Default: nil
       - treesitter:       Determines whether the previewer performs treesitter
                           highlighting, which falls back to regex-based highlighting.
@@ -442,6 +481,8 @@ append(
                           `table`: table of filetypes for which to attach treesitter
                           highlighting
                           Default: true
+      - msg_bg_fillchar:  Character to fill background of unpreviewable buffers with
+                          Default: "╱"
     ]]
 )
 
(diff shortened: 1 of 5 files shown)


lua/telescope/config.lua
```diff
@@ -428,9 +428,18 @@ append(
       - hook(s):          Function(s) that takes `(filepath, bufnr, opts)`
                           to be run if the buffer previewer was not shown due to
                           the respective test.
-                          Available hooks are: {mime, filesize, timeout}_hook, e.g.
+                          Available hooks are: {mime, filesize, timeout}_hook;
+                          for instance, for too text large files you can get the first
```
l-kershaw, line 432: "for too text large files" -> "for too large text files"

fdschmidt93: haha, thanks :)

l-kershaw:
This looks good to me! It's definitely helpful when there is a functional example in the docs 👍

---

On an unrelated to this PR note, is there currently a way to override enable/disable the previewer purely based on filetype (e.g. pdf)?
Might be nice to have a table somewhere like:
```
{ allow = { 'pdf', ... }, block = { 'csv', ... } }
```
this would be particularly helpful for Windows users, as they can implement it in a way that means they don't need to install another dependency.

_Not asking you to implement this if you don't feel like it, just wanted to double check it wasn't currently possible, and get your thoughts_

fdschmidt93:
> On an unrelated to this PR note, is there currently a way to override enable/disable the previewer purely based on filetype (e.g. pdf)?

If it's just based on filetypes plenary detects then it's not yet implemented but would be very easy and would go here.

https://github.com/nvim-telescope/telescope.nvim/blob/60660334c70d9d81dccc10a563e01920b9455e76/lua/telescope/previewers/buffer_previewer.lua#L185-L284

L261 would essentially be have to wrapped in an allow/blocklist (probably blocklist? Is there a registered filetype for which we cannot preview something?). Maybe even a filetype hook that takes `(bufnr, ft, opts)` and then users can just do what they want with it (eg test for their filetypes and otherwise show `putils.highlighter` which I guess is fine if adequately documented). For instance, I might want to pipe csv through something that displays columns nicer. What do you think?

fdschmidt93:
I've started addressing this, but it gets very funny as depending on what you want to do, you have to `vim.schedule` the previewing segment of the `filetype_hook` but the entire function may not be scheduled, so it returns in time whether to preview or not.

I'm hoping https://github.com/neovim/neovim/issues/15655 will make this a lot more elegant.

Conni2461:
Whats the scope of this PR? Update some documentation or refactor the code?

What does this new hook offer that the previous way of changing this didnt already offer? Rather than just make the code even more complicated.

Ref here:
https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes#previewers

Disabling preview/highlighting for certain filetypes/endings was always possible. The `preview` config value should have been for normies, because to admit overriding the whole function isnt user friendly.

fdschmidt93:
> Whats the scope of this PR? Update some documentation or refactor the code?

It turned from updating docs into a bit more :laughing: 

> Disabling preview/highlighting for certain filetypes/endings was always possible. The preview config value should have been for normies, because to admit overriding the whole function isnt user friendly.

Yeah, I guess I was trying to integrate the recipe (which I didn't think of) even further into a dedicated hook, though it's probably better done as per the recipe to avoid callback hell. I don't have a preference; very happy to revert this to just a better docs PR.

l-kershaw:
Ah, my apologies, I hadn't seen the configuration recipe when I brought this idea up 😔

Although it would be nice to implement a user friendly way of doing this, if it is going to make code harder to maintain in the long term, then I agree that it's best to just point to the configuration recipe. If we do this, it might be nice to still export the `putils.set_preview_message` (as has been changed in this PR) for use in that configuration recipe and in other custom user preview stuff 🙂

Conni2461:
I am not against making `set_preview_message` public. I just asked for the scope of this the PR, because title and description said something different than the diff.

I am also not against adding interfaces for novice users, i already accepted that. But that filetype hook is not that. So i asked what we gain over the current way. If the current way is no longer desired, then we need to deprecate it.

fdschmidt93:
> I am not against making set_preview_message public. I just asked for the scope of this the PR, because title and description said something different than the diff.

Yeah, I just haven't updated them yet, I also hadn't fully decided what to do with this PR myself; that's why I've set this to WIP ;)

> I am also not against adding interfaces for novice users, i already accepted that. But that filetype hook is not that. So i asked what we gain over the current way. If the current way is no longer desired, then we need to deprecate it.

See below as mentioned above.

> I've started addressing this, but it gets very funny as depending on what you want to do, you have to vim.schedule the previewing segment of the filetype_hook but the entire function must not be scheduled, so it returns in time whether to preview or not.
I'm hoping neovim/neovim#15655 will make this a lot more elegant.

That's why I was torn last night what to do with this.

fdschmidt93:
Ok, I think I'm happy with this, we now:
- have much better docs, incl. important words on plenary filetype detection (which doesn't seem to detect eg csv?)
- `vim.schedule_wrap` the relevant hooks for the user
- lay out that `filetype_hook` has to return true by default (we could also default to that ourselves.. but I prefer that the user be specific)
- expose set_preview_message

I don't think test failures are related.

Conni2461:
I already said everything about this PR that needed to be set. I am not really hyped about another hook.

> I'm hoping neovim/neovim#15655 will make this a lot more elegant.

I dont know why you are waiting for this. This will hit nightly first (it didnt hit yesterdays 0.5.1 release and who knows if it will be in any 0.5.* or if we have to wait till 0.6), which makes it useless for us for the time being, because having two versions of that function (just for that) is just unnecessary. And i dont think we have enough people to maintain two versions/branches, just for that.

fdschmidt93:
> I dont know why you are waiting for this. 

As mentioned above, I've worked around this now as I'm aware this unfortunately isn't maintainable with 0.5.1 users.

> I already said everything about this PR that needed to be set.  I am not really hyped about another hook.

That's fine, though the implementation has changed and it's a easier to use the hook now.

> I already said everything about this PR that needed to be set. I am not really hyped about another hook.

That's also fine, I just wasn't going to merge this though without final remarks or being passed the ball back from your end. If you didn't want to review it or anything, that would have been completely fine.

tjdevries:
I haven't checked everything, but why isn't this just one callback that gets passed both the filetype and the size information? Why two different functions? 

Or why can't it take a list of functions or something? I'm not sure I want to enumerate every possible way you might want to filter something out. If that makes sense.

PR #1274, opened 2021-09-21
fix(lsp): check if action.command.arguments is nil

jose-elias-alvarez:
The code added in #738 attempts to access `action.command.arguments[1]` unconditionally, but the [LSP specification](https://microsoft.github.io/language-server-protocol/specification#command) defines `arguments` as optional. This PR adds a check to make sure `action.command.arguments` is not `nil` before accessing it.


Conni2461:
Thanks :)

PR #1275, opened 2021-09-21
Fix option name for builtin.oldfiles

gbrlsnchs:


tami5:
I'd rather you change docs then change actual code, b/c others ppl might have already setup their picker using `cwd_only`

gbrlsnchs:
> I'd rather you change docs then change actual code, b/c others ppl might have already setup their picker using `cwd_only`

I could do that, but then it would be inconsistent with `builtin.buffers`, which correctly uses `only_cwd`. What to do?

Conni2461:
Yeah we are aware that the naming doesnt align. I most likely overlooked that. And i already said that we could make `only_cwd` and `cwd_only` for as synonym and accept both for the time being.

Issue #1278, opened 2021-09-22
help_tags breaks when there is `.doc` in `file_ignore_patters`


rafamadriz:
### Description

Since commit 02bc88e6d61bd10272277796ca8ea3ad96a388dd, `Telescope help_tags` is broken when there is `.doc` in `file_ignore_patterns`

Here's the error:

```
Error executing vim.schedule lua callback: ...te/pack/packer/opt/plenary.nvim/lua/plenary/filetype.lua:121: attempt to index local 'filepath' (a nil value)
```

### Neovim version

NVIM v0.5.0
Build type: Release
LuaJIT 2.0.5

### Operating system and version

Arch Linux 5.10.67-1-lts

### checkhealth telescope

```markdown
health#telescope#check
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - OK: nvim-treesitter installed.

## Checking external dependencies
  - OK: rg: found ripgrep 13.0.0
  - OK: fd: found fd 8.2.1

## ===== Installed extensions =====

## Telescope Extension: `projects`
  - INFO: No healthcheck provided
```

### Steps to reproduce

1. `nvim -nu minimal_init.lua`
2. `:Telescope help_tags`

### Expected behavior

`:Telescope help_tags` doesn't give any error and works as it normally does.

### Actual behavior

Outputs the following error: 

```
Error executing vim.schedule lua callback: ...te/pack/packer/opt/plenary.nvim/lua/plenary/filetype.lua:121: attempt to index local 'filepath' (a nil value)
```

Gives the `help_tags` as only option:

![Screenshot_20210922_105802](https://user-images.githubusercontent.com/67771985/134357707-6019e0ba-5187-4fba-9dc7-a036d4159c2f.png)

And when selected it just opens the help tags file:

![Screenshot_20210922_105817](https://user-images.githubusercontent.com/67771985/134357816-97da48d5-f219-4979-9ff2-499c1d58ea48.png)

### Minimal config

```Lua
-- This file is a minimal `init.lua` example for debuggin plugins.
-- Open with `nvim -nu debug.lua`

local fn = vim.fn
local install_path = "/tmp/nvim/site/pack/packer/start/packer.nvim"
local compile_path = install_path .. "/plugin/packer_compiled.lua"

vim.cmd([[set packpath=/tmp/nvim/site]])

local function load_plugins()
	local pack = require("packer")
	local use = pack.use
	pack.init({ compile_path = compile_path })
[...]

PR #1279, opened 2021-09-22
fix(#1278): set `opts.file_ignore_patterns = {}`

rafamadriz:


Conni2461:
Thanks :) (CI failure is unrelated)

Issue #1240, opened 2021-09-13
Colored output on grep_string?


fsouza:
**Is your feature request related to a problem? Please describe.**

Hi there, finally came around to adopt telescope and one thing that I miss from my fzf days is the ability to filter colored output from from rg.

If I pass `--color=always` to rg, telescope doesn't display any results, and I assume that filtering on ascii codes isn't really supported?

**Describe the solution you'd like**

I'd like to be able to invoke `grep_string()` and process colored output from rg.

**Describe alternatives you've considered**

No colors? 🙈 

**Additional context**

If this is something yall think makes sense, I can try to send a PR.

fdschmidt93:
Just super quick comments because it's way too late in my timezone :sweat_smile: 

This is something we want and a PR would be very welcome :) as noted here

https://github.com/nvim-telescope/telescope.nvim/blob/e7aa63db8fea4983658d03b5000d59f75259638b/lua/telescope/builtin/files.lua#L105-L109

I haven't thought about this myself all too much, but I think https://github.com/neovim/neovim/pull/14871 will be really cool for this (hopefully by end of this week :).

Conni2461:
PR that wants to do that: https://github.com/nvim-telescope/telescope.nvim/pull/670

currently its not supported because then we would need to parse the shell color codes. I send @jesseleite a code snippet (a term color code parser) that a wrote some time ago, with that we could realize that. Maybe i should share this snippet and someone else can pick it up, i know he is very busy lately with a newborn :)

fdschmidt93:
Unsolicited and uninformed comment: if that requires `rg --json` https://github.com/neovim/neovim/pull/14871 would be great.

TC72:
Hey @Conni2461, I've been speaking to @jesseleite and wanted to try to help out with this. Could you please post the snippet for the color parser?

Conni2461:
json would be easier to implement no doubt. But we need to see how fast it will be in the end, when its done. We need to be able to run it in `:help api-fast` otherwise its useless for us. From what i've seen from this PR this should be possible because we will get a `vim.json.*` interface. None the less for now i think parsing color codes would be a good enough solution. We later can benchmark that version with a json version and decide which one we want to keep.

I share both the gist + my conversation because it contains my thoughts on how to solve this issue.
https://gist.github.com/Conni2461/6f0c9353b83ace6a98b2793d0b6fd6d1

[...]

PR #1281, opened 2021-09-24
Issue #1240 - Added highliting of grep results

TC72:
This is linked to issue #1240.

Added @Conni2461 color convertor code to parse colored results from `rg` and convert them to highlights.

This will hopefully be replaced in the future by a better version which uses JSON results from `rg` to handle this.

The reason json would be better is this code is creating the highlights, but then adjusts them based on dev icons being added and the file path being shortened. This is because the filename including path can't be fetched until the color codes are stripped as part of color conversion.

A kind of chicken and egg problem, but this could be refactored to do a 1st run through that just strips the color codes, but going to json will be better.

The code could also be optimised to stop after a certain length as there's no point highlighting text that's not going to be displayed due to the width of the results panel.

It would be great to do this in `rg` using `--max-columns` and `max-columns-preview` as it would reduce the amount of text we get back.
Also look at using `--trim` to remove the white space from the result.

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -192,7 +192,8 @@ telescope.setup({opts})                                    *telescope.setup()*
         - "shorten"   only display the first character of each directory in
                       the path
         - "truncate"  truncates the start of the path when the whole path will
-                      not fit
+                      not fit. To increase the the gap between the path and the edge.
+                      set truncate to number `truncate = 3`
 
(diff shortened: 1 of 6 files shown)


Conni2461:
Should be moved to nvim-telescope/telescope-rg.nvim and / or replaced by `vim.json`.

Sorry

PR #1286, opened 2021-09-26
feat: add highlight group: TelescopePreviewMessage

saviocmc:
Add a highlight group to be used when displaying a message in the ```buffer_previewer``` with the function [```set_timeout_message```](https://github.com/nvim-telescope/telescope.nvim/blob/4816a27d7680feed293d21a945c476947af20d29/lua/telescope/previewers/buffer_previewer.lua#L59).

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -66,6 +66,9 @@ local function set_timeout_message(bufnr, winid, message)
     false,
     utils.repeated_table(height, table.concat(utils.repeated_table(width, "╱"), ""))
   )
+  for linenr = 0, height do
+      vim.api.nvim_buf_add_highlight(bufnr, -1, "TelescopePreviewMessage", linenr, 0, -1)
+  end
   local anon_ns = vim.api.nvim_create_namespace ""
   local padding = table.concat(utils.repeated_table(#message + 4, " "), "")
   local lines = {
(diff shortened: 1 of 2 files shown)


fdschmidt93:
Thanks for the PR! :) I've incorporated these changes also now in #1273 as otherwise there'd be a conflict and allow to setting the background fillchar as per your FR. As a tip on performance, `nvim_buf_add_highlight` uses extmarks in the background, which in turn can be used to set the highlighting for the entire buffer at once :) cf example below

```lua
  vim.api.nvim_buf_set_extmark(
    bufnr,
    anon_ns,
    0,
    0,
    { end_line = height, hl_group = "TelescopePreviewMessageFillchar" }
  )
```
So if you don't mind, I think we can close this.

Conni2461:
????

I think we should not take a PR away from a First-time contributor! I am fine with merging it

fdschmidt93:
Up to you. In my PR I've set hl group for entire buffer at once as well as split hl groups between message and background fillchar to preempt future requests. To avoid a conflict between this and my PR & to address #1288 in one go, I've incorporated it into my PR to avoid extra work. 

But happy to merge this instead if that's preferred.

Conni2461:
Sorry, but i honestly think he deserves the credit for this. Even thought it will be slightly changed.

Thanks for the PR :)

saviocmc:
Thank you for the welcoming messages! I'm very happy to give a first contribution :)

I'm sorry I didn't see your PR first @fdschmidt93, it does cover my end goal.

Thank you all for the great work! 🔭🚀

Issue #1283, opened 2021-09-26
Pressing enter without available items in `help_tags` leads to nil-value error


jonasstr:
### Description

When no items are available in `Telescope help_tags` and the user presses Enter, an error is shown instead of doing nothing or alternatively informing the user that there are no items available.

### Neovim version

NVIM v0.5.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

### Operating system and version

Linux Mint 20.2

### checkhealth telescope

```markdown
health#telescope#check
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - OK: nvim-treesitter installed.

## Checking external dependencies
  - OK: rg: found ripgrep 11.0.2
  - WARNING: fd: not found. Install [sharkdp/fd](https://github.com/sharkdp/fd) for extended capabilities

## ===== Installed extensions =====
```

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. Type the command `Telescope help_tags`
3. Enter a search term that leads to no items showing up, e.g. `aaaaaaaaaaaaa`
4. Press Enter.

### Expected behavior

Either don't close the window and show an info message
or close the window and show an info message.

In any case there should not be a `nil value` error.

### Actual behavior

The following error is shown:
```
E5108: Error executing lua .../start/telescope.nvim/lua/telescope/builtin/internal.lua:588: 
attempt to index local 'selection' (a nil value)
```
and the Telescope window is closed.

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'

local function load_plugins()
    require('packer').startup {
        {
            'wbthomason/packer.nvim',
                {
                    'nvim-telescope/telescope.nvim',
                    requires = {
                        'nvim-lua/plenary.nvim',
                        { 'nvim-telescope/telescope-fzf-native.nvim', run = 'make' },
                    },
                },
                -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
        },
            config = {
                package_root = package_root,
                compile_path = install_path .. '/plugin/packer_compiled.lua',
                display = { non_interactive = true },
            },
    }
    end

_G.load_config = function()
    require('telescope').setup()
[...]

PR #1289, opened 2021-09-27
fix: show error message when nothing is selected

jonasstr:
This continues the fixes done in 4816a27d7680feed293d21a945c476947af20d29 (ref #1283).

The only picker that shows this error now (of those I tested) is `Telescope buffers` - I did not fix this one because there is already a message saying "[telescope] Nothing currently selected" above the nil error message (not sure why it's shown twice).

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -301,13 +301,20 @@ end
 
 actions.edit_command_line = function(prompt_bufnr)
   local entry = action_state.get_selected_entry()
+  if selection == nil then
+    print "[telescope] Nothing currently selected"
+    return
+  end
   actions.close(prompt_bufnr)
   a.nvim_feedkeys(a.nvim_replace_termcodes(":" .. entry.value, true, false, true), "t", true)
 end
 
 actions.set_command_line = function(prompt_bufnr)
   local entry = action_state.get_selected_entry()
-
+  if selection == nil then
+    print "[telescope] Nothing currently selected"
+    return
+  end
   actions.close(prompt_bufnr)
   vim.fn.histadd("cmd", entry.value)
   vim.cmd(entry.value)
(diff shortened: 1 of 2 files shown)


l-kershaw:
Thanks for the PR!

There are a few cases where the relevant variable is called `entry` rather than `selection`, which means that `selection` is not defined. This is what is causing the `luacheck` to fail.
In these cases, I would suggest you rename the `entry` variable to `selection` as it is a bit clearer about what the variable represents (make sure you change all of the references to `entry` to `selection`, e.g. `entry.value` -> `selection.value`).

Also, I think it would be better to use `telescope`'s logging system rather than just a `print` to inform the user that nothing is selected. Maybe use `log.warn`?

Feel free to ping me if you have any questions 🙂

jonasstr:
Yeah sorry I am not sure why I forgot to change this! :) Ok I will look into `log.warn`. Should I just add more commits here (I assume they will be squashed when merging) or force-push to the existing branch? Thanks :smile:

l-kershaw:
Just adding more commits is fine (I normally end up with a load of extra ones trying to keep the linting happy 😅).
We always squash when merging, so its better to keep the commit history in the PR (where possible), in case we need to go back and look at it 🙂

Conni2461:
Just saying i dont think we need `log.*` here, we can just `print` because there is no real reason to write it in the log file. Also i decided for `print` yesterday because we already have that exact same "code block" in actions.set with a print

https://github.com/nvim-telescope/telescope.nvim/blob/a6c7498bdcd661e86ec0c1181ed50024beb540f7/lua/telescope/actions/set.lua#L90-L93

l-kershaw:
> Just saying i dont think we need log.* here, we can just print because there is no real reason to write it in the log file. Also i decided for print yesterday because we already have that exact same "code block" in actions.set with a print

Fair enough. It shouldn't make much of a difference either way, so happy to leave it as `print` 👍

jonasstr:
> Just adding more commits is fine (I normally end up with a load of extra ones trying to keep the linting happy).

Well, I'm glad I'm not the only one :sweat_smile:. Sorry for the extra commits..

l-kershaw:
LGTM, thanks for the PR  👍

No worries about the extra commits 🙂

PR #1290, opened 2021-09-27
fix highlight declaration of 'TelescopePreviewMessageFillchar'

saviocmc:
I was testing the plugin with the lastest merge in the master and noted the highlight 'TelescopePreviewMessageFillchar' was incorrectly declared as 'TelescopePreviewFillchar'.

I also think it should by default link to 'TelescopePreviewMessage', so the user can choose to set 'TelescopePreviewMessage' to change both or to set each one individually.


fdschmidt93:
LGTM :) Thanks for cleaning up, not sure why the other PR got too quickly merged. I had it correctly prior to rebasing, was too quick :laughing:

PR #1280, opened 2021-09-22
feat: more prompt position strategies

l-kershaw:
This PR adds the `prompt_position` option for the following layout stratgies:
- `vertical`
- `bottom_pane`
- `center`

`vertical` was noted as being "planned" in the code
`bottom_pane` was requested in [this comment](https://github.com/nvim-telescope/telescope.nvim/issues/765#issuecomment-851753342)

---

EDIT: I have now also added some more flexibility to the `get_dropdown` and `get_ivy` themes if the user customises the `prompt_position`

PR #1293, opened 2021-09-28
fix: which key float positioning

fdschmidt93:
@l-kershaw could it be that one of the recent plenary popup PRs might have broken positioning of `actions.which_key`? :sweat_smile:  Judging by one of the PR titles alone I tried this fix and seems to work. If you could please confirm this should be fine, that'd be great :) thanks

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -1029,7 +1029,9 @@ actions.which_key = function(prompt_bufnr, opts)
   local picker = action_state.get_current_picker(prompt_bufnr)
   local prompt_win = picker.prompt_win
   local prompt_row = a.nvim_win_get_position(prompt_win)[1]
-  local prompt_pos = prompt_row <= 0.5 * vim.o.lines
+  -- TODO(fdschmidt93): resolve side in more principled fashion
+  -- ivy theme right around 0.5 * vim.o.lines; simple heuristic to circumvent
+  local prompt_pos = prompt_row < 0.45 * vim.o.lines
 
   local modes = { n = "Normal", i = "Insert" }
   local title_mode = opts.only_show_current_mode and modes[mode] .. " Mode " or ""
@@ -1041,8 +1043,8 @@ actions.which_key = function(prompt_bufnr, opts)
     maxwidth = vim.o.columns,
     minheight = winheight,
     maxheight = winheight,
-    line = prompt_pos == true and vim.o.lines - winheight or 1,
-    col = 1,
+    line = prompt_pos == true and vim.o.lines - winheight + 1 or 1,
+    col = 0,
     border = { prompt_pos and 1 or 0, 0, not prompt_pos and 1 or 0, 0 },
     borderchars = { prompt_pos and "─" or " ", "", not prompt_pos and "─" or " ", "", "", "", "", "" },
     noautocmd = true,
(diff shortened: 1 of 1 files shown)


l-kershaw:
Ah, I forgot about the mappings popup when migrating stuff to the new indexing, so yes I would expect there to be incorrect positioning for this at the moment.

As a general rule, if you don't want something centered (either vertically or horizontally) you should increase the old index by 1.
If you do want it centered, set the index to 0.

From a very quick glance at this, I think the lower position `line` index still needs adjusting (just add 1), and so does the `col` index (could either add 1 or just set to 0 to center).

I won't have time to have a proper look at this for the rest of the day, but will try to look at it this evening 🙂

l-kershaw:
@fdschmidt93 I've tested the adjustments I mentioned and that seems to fix the positioning in all of the scenarios I've tested.

This is what I used:
```
    line = prompt_pos == true and vim.o.lines - winheight + 1 or 1,
    col = 0,
```
Note that `col=2` would work the same, but `col=0` seems simpler to me.

fdschmidt93:
Thanks a lot :)

One thing where this still seems to fail is the ivy theme, as that now has become effectively heigher? I've implemented a small heuristic, which I'm reasonably certain will file with more variety via #1280.

Would you say it's ok to leave it for the time being and accept this quick fix? I haven't yet been able to come up with a comprehensively general solution yet and it's quite tedious for something I'm not really using myself.

l-kershaw:
@fdschmidt93 this is what I came up with:
```lua
  local win_central_row = function(win_nr)
    return a.nvim_win_get_position(win_nr) + 0.5 * a.nvim_win_get_height(win_nr)
  end
  -- place hints at top or bottom relative to prompt
  local prompt_row = win_central_row(picker.prompt_win)
  local results_row = win_central_row(picker.results_win)
  local preview_row = win_central_row(picker.preview_win)
  local prompt_pos = prompt_row < 0.45 * vim.o.lines or prompt_row < 0.55 * vim.o.lines and results_row + preview_row < vim.o.lines
```

Its not particularly pretty, but it seems to catch the idea of "if the prompt is close to the middle, look at the result and preview windows too". Obviously the numbers in there can be tweaked if need be

What do you think of this implementation?

fdschmidt93:
I wouldn't be opposed because it appears like a practical. I was thinking about checking on what side the most free space after accounting for results and preview, but won't have the time before next week to tackle this.  Does it work for you with the ivy? Maybe I patched my code wrong locally.

l-kershaw:
I had a small mistake in the implementation I put above (now edited to fix), and have tweaked the numerical constants to give slightly better experience with `get_ivy` on my screen size.

This is definitely an approximation of the sort of behaviour that I would like in an ideal world, but it is good enough for the vast majority of situations. If you want to have a go at making a more robust setup for this, then I'd be happy to look over it (I'm away next week, but can look when I get back). But also, since this isn't a feature that you are really using, then its fine to leave it as a "works 95% of the time" solution, and if someone else wants to implement something more intricate, they can submit a PR.

fdschmidt93:
Agreed, thanks :) I'll quickly test your revision and then ideally merge.

fdschmidt93:
Works great for standard layout positions with top and bottom prompt as well as ivy and dropdown themes. Thanks a lot :)

But just in case, we'll best revisit this with #1280  as well :sweat_smile: if you don't mind rebasing that PR I'll try to also test that

Issue #1228, opened 2021-09-09
Move out ALL builtins


Conni2461:
Disclaimer: This is something TJ just brought up on Discord.

TJ and I think we should move out all builtins to own repositories. like `telescope-files.nvim` `telescope-git.nvim` `telescope-lsp.nvim` etc, ....

Neither TJ nor I enjoy maintaining a lot of the issues that currently are piling up. I only use 5-6 pickers so I am also the wrong person to maintain most of these issues (because i dont have any strong feeling for them) and from my conversation with TJ i think he feels the same.

We can still have them under `lua/telescope/builtin/` and we can still support them with `:Telescope ...` and configuration with `telescope.setup{ pickers = { ... } }` but with moving them we could gain more flexibility. People can pick whatever picker they need and we (core team and other maintainers) can focus more on the "engine" telescope.

We then can assign more people to these individual repositories.

Thoughts are welcome :)

clason:
It's a completely understandable endeavor, but let me just note for the sake of completeness that this would add friction to users:

1. For one, the "one stop shop" nature of Telescope is a big selling point.
2. And I just _know_ there will be a constant stream of issues where people forgot to install one of the add-ons and wonder why `Telescope find_files` isn't working... 
3. It also feels wrong to have Telescope without even basic FZF equivalents. 
4. And are you sure you want to go full CoC-like "plugin ecosystem"?

Unless you want to pull in the external repos as submodules (including fzf-native)? That might be a compromise here.

Conni2461:
TJ just mentioned it "out of the blue" and i put it here because to collect more thoughts. So thank you very much :) these are good points.

I am torn apart what i wanna do, we have ~50 pickers and they are a lot to maintain (you know that) and most of the issues that were opened lately didn't interested me.

I dont know if moving them out even has a benefit if no one wants to maintain them in the end we have the same issue.

tjdevries:
Submodules would be an acceptable solution in my mind. I do want to try and continue to provide out-of-the-box support for a bunch of stuff when you download telescope.

In contrast, `fzf` the engine (or executable) lives primarily apart from `fzf.vim`.

I will need to think about it a bit more, but sometimes I get demotivated looking at issues about pickers that I've never used :laughing: or trying to weigh in and decide on those (and then some aspect of them leaks into the "engine" or "core" part of telescope, which makes things complicated later).

I will think some more as time goes on :)

clason:
[...]

PR #1294, opened 2021-09-28
fix: stacking highlights in buffer previewer

### LINKED ISSUE #1213: Buffers picker :: Odd colorization of highlighted lines
### Description

Using the `buffers` picker, colorization of the highlighted line in the preview window recently seems to have changed. Whereas previously the highlighted line in the Telescope preview window looked the same as highlighting it in a regular buffer.

Depending on the colorscheme, characters like parentheses are sometimes difficult to read. Furthermore this behaviour cannot be replicated when using a different picker that highlights lines in the preview window, for instance the `current_buffer_fuzzy_find`.

This change got introduced by commit [79dc995f820150d5de880c08e814af327ff7e965](https://github.com/nvim-telescope/telescope.nvim/commit/79dc995f820150d5de880c08e814af327ff7e965). Reading through a few other issues like https://github.com/nvim-telescope/telescope.nvim/issues/1151, disabling Treesitter resolves the issue, but as said already in that thread; it's a workaround and rather obvious that the problem disappears when disabling Treesitter.

### Neovim version

NVIM v0.5.0
Build type: Release
LuaJIT 2.1.0-beta3

MacOS 11.5.2 + Ubuntu 20.04

### Steps to reproduce

1. Neovim (+)0.5.0 loaded up with Treesitter, Telescope, and (optionally) a custom colorscheme (for instance, solarized :: ishan9299/nvim-solarized-lua)
2. Open any file (for instance a Lua file) and navigate to a line with text. Bonus points if navigating to a line that contains parentheses or curly braces
3. Open the picker : : Telescope buffers
4. Boom 💥

### Expected behavior

I would expect highlighted lines in a Telescope preview window to look the same as they do in a regular buffer as right right now colorization is a bit off in preview windows for the `buffers` picker.

### Actual behavior

Prior to commit `79dc995f820150d5de880c08e814af327ff7e965`
telescope-good-00

With commit `79dc995f820150d5de880c08e814af327ff7e965`
telescope-bad-00

How a selected line looks like in a regular buffer (latest, so post `79dc995f820150d5de880c08e814af327ff7e965` commit)
telescope-good-01

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
      { "nvim-treesitter/nvim-treesitter" },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
  require("nvim-treesitter.configs").setup {
    ensure_installed = { "lua" },
    highlight = { enable = true },
  }
end
end
load_plugins()
require('packer').sync()
```

fdschmidt93:
Closes #1213 

This is a slightly (but totally ok) hacky solution because we'd otherwise need `hl_mode` to affect `hl_group` in `nvim_buf_set_extmark`.

@lcrockett does it now work as expected for you? For me it seems to work as expected with your minimal config.

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -1124,8 +1124,9 @@ previewers.buffers = defaulter(function(opts)
           -- only set if winid and rows are matching
           pcall(vim.api.nvim_buf_set_extmark, bufnr, ns_previewer, lnum - 1, 0, {
             end_col = #line,
+            virt_text = {{line, "TelescopePreviewLine"}},
             virt_text_pos = "overlay",
-            hl_group = "TelescopePreviewLine",
+            hl_mode = "combine",
             ephemeral = true,
             priority = 101, -- 1 higher than treesitter
           })
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -1124,7 +1124,7 @@ previewers.buffers = defaulter(function(opts)
           -- only set if winid and rows are matching
           pcall(vim.api.nvim_buf_set_extmark, bufnr, ns_previewer, lnum - 1, 0, {
             end_col = #line,
-            virt_text = {{line, "TelescopePreviewLine"}},
+            virt_text = { { line, "TelescopePreviewLine" } },
             virt_text_pos = "overlay",
             hl_mode = "combine",
             ephemeral = true,


fdschmidt93:
Sorry, deleted / turned off all telescope notifications the past days as I'm quite busy atm and missed style failure :sweat_smile: 

I looked at it only earlier today and wanted to revisit it some other time hoping to find some good solution (99% sure there's not). I'd say let's just move back and make "actual buffers" previewer opt-in while documenting this issue. I fully agree mru is more important than syntax highlighting, though at the same time a to me surprising amount of users where "upvoting" the original issue https://github.com/nvim-telescope/telescope.nvim/issues/714 Any future issues with the "actual buffers" previewer are then left to fix for users that have issue with it, because I'm kind of done with it :sweat_smile: Further, I'm sure you already know, but we then have to revert some stuff with the special buffers for the "cloning buffers" previewer. I'll sit on the train Wednesday morning in all likelihood; I'll do my very best (honestly, no hard promises because it's just sucking motivation) to prepare the PR then if you agree with this plan.

Conni2461:
No worries :) I also am currently not motivated to do anything. I dont like the idea of doing it opt-in i would just do more opinionated things rather than make everyone happy. Because this just leads to more bugs and we will never make everyone happy. People can still overwrite it by passing in a different function pointer (to `previewer = ...`).

I've been thinking quite a lot where we should go with telescope and ^^^ is the reason. I havent found a solution for the issues we currently face. #1228 is obv related to this.

lcrockett:
Disregarding the current discussion here; @fdschmidt93 this PR fixes #1213 indeed and the original behaviour has returned with it. Not too sure if based on the discussion here this might change or not in the future. However, cheers on this and other efforts regarding `nvim-telescope` !

fdschmidt93:
> I dont like the idea of doing it opt-in i would just do more opinionated things rather than make everyone happy. Because this just leads to more bugs and we will never make everyone happy. 

Yeah, you're totally right. Probably then just off to the wiki with it or something eventually.

PR #1298, opened 2021-10-01
chore: highlight which_key on README

fdschmidt93:
This came up on Gitter. You can make the argument `telescope.setup` per README is not "clean" anymore, but it's not intended as a COPYME. Nevertheless, it's an outlier. I don't really have an opinion, whatever works.

~~(Created the PR right from GH, there's a slight error that `<C-h>` needs to be a string, don't behead me just yet ;)~~

cc @clason


fdschmidt93:
Thanks :)

I'll give others the chance to chime in, in case they have some feelings about this (mainly setup I guess). If there's no more comments by end of day, I'll merge.

Issue #365, opened 2020-12-26
Feature Request: Open in float


tami5:
It would be cool if there is an action and mapping to open an entry in float

sunjon:
what "properties" would the window have once it's no longer a telescope window? I guess this is related to persistent windows.

tami5:
@sunjon Yes, I believe  it is related to #310 . For properties, it should simply inherits the same window properties that of the telescope instance .

TC72:
I've just added a PR for this #1299

PR #1299, opened 2021-10-02
Added the option to open selection in a popup - Issue #365

TC72:
Added an option to open the selection in a popup window. I've added a default mapping of <C-f>, f for floating window.

It's sized to fill the space used by all telescope windows so it gives the impression the prompt and results windows have been closed to expand the Preview window.

An autocmd is added to make sure the border of the popup is closed when the popup gets closed.

When this window opens we are completely outside of Telescope, but this causes issues if you try to use Telescope while this popup is open. In general Telescope should probably check if it's being opened in a popup.
If this is fixed then It might be possible to use `Telescope resume` to go back to the previous state.

This code could be made simpler if we used the `opts.keep_last_buf` which keeps alive the preview buffer.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -249,6 +249,18 @@ actions.select_tab = {
 
 -- TODO: consider adding float!
 -- https://github.com/nvim-telescope/telescope.nvim/issues/365
+actions.select_popup = {
+  pre = function(prompt_bufnr)
+    action_state.get_current_history():append(
+      action_state.get_current_line(),
+      action_state.get_current_picker(prompt_bufnr)
+    )
+  end,
+  action = function(prompt_bufnr)
+    --actions.file_popup(prompt_bufnr)
+    return action_set.select(prompt_bufnr, "popup")
+  end,
+}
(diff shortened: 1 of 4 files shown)


johnybx:
This is really nice :+1: . It would be cool if it was supported also [here](https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/builtin/lsp.lua#L47) for lsp builtin action so that it can be used for example for preview definition.

TC72:
I'll have a look at that, help is another one I'd like to be able to open in the popup.
Long term it'll still be nicer if we could keep this within telescope by just hiding the prompt and the results, but for now it's a start.

cagdassalur:
Is this planned to be merged?

TC72:
> Is this planned to be merged?

I doubt it, the code I wrote isn't great and it's more like a hack of how to get this working.

cagdassalur:
ah too bad, i was looking into how to implement something like this.. So maybe we should close the pr for less pollution?

TC72:
Yes, I think that’s a good idea.

PR #1304, opened 2021-10-04
feat: add find_command to the list of defaults

kylo252:
This allows setting the option once as a default in `setup()` and so
skipping the check every time you don't call `find_files` with
arguments.

This also allows setting certain desired flags to be on all the time for
the finder, similar to `vimgrep_arguments`, e,g. `--hidden`.

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -16,6 +16,7 @@ local os_sep = Path.path.sep
 
 local flatten = vim.tbl_flatten
 local filter = vim.tbl_filter
+local get_default = utils.get_default
 
 local files = {}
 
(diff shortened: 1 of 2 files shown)


fdschmidt93:
Would it make sense to do the `find_command` resolution in `builtin.find_files` just once at startup to which `conf.find_command` defaults to? I'm assuming it's not really a startup hog, since `rg` essentially is a hard dependency.

Not manipulating the `find_command` myself, but from a quick glance of the phone (and if memory serves me well enough) adding a type annotation (ie a table of strings) and an example would be helpful, I think? :) please ignore if I'm wrong.

Otherwise LGTM thanks :)

Conni2461:
Do we really need this? We can already do this like that:

```lua
telescope.setup {
  defaults = {
    -- ....
  },
  pickers = {
    find_files = {
      find_command = { "fd", "--type", "f", "--hidden", "--no-ignore" }
    },
  }
}
```

kylo252:
> Would it make sense to do the find_command resolution in builtin.find_files just once at startup to which conf.find_command defaults to?

I was also going for that idea but decided to wait until I can discuss it with you guys first. The logic can also simplified quite a bit, if ripgrep is a hard requirement then there's no need to add all those other lookups afterwards. What's even better is that `fd` and `rg` share quite a few common flags, so that will make the code also shorter.

> Do we really need this? We can already do this like that:
> 
> ```lua
> telescope.setup {
>   defaults = {
>     -- ....
>   },
>   pickers = {
>     find_files = {
>       find_command = { "fd", "--type", "f", "--hidden", "--no-ignore" }
>     },
>   }
> }
> ```

Nice! how can we document this to make it more obvious?

fdschmidt93:
> Nice! how can we document this to make it more obvious?

Technically speaking, it is right there at `:h telescope.setup`, but I agree we probably should highlight this once more at top of `telescope.builtin` and maybe even have a separate section or something more elaborate in `telescope.setup`.

PR #1306, opened 2021-10-05
entry_to_qf should use the entry.path

malkomalko:
## Summary
I noticed when using the quickfix and location list actions, the file path is escaped using `fnameescape`. This is no good because when navigating the list you'll see file names with spaces are escaped with `\ `. This is not needed as it will cause navigating the list to not work properly.

## Screenshots

- Before
image
- After
image
- Using vim's builtin `:lv favorite **/*92*`
image

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -671,7 +671,7 @@ local entry_to_qf = function(entry)
 
   return {
     bufnr = entry.bufnr,
-    filename = from_entry.path(entry, false),
+    filename = entry.path,
     lnum = vim.F.if_nil(entry.lnum, 1),
     col = vim.F.if_nil(entry.col, 1),
     text = text,
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -20,6 +20,7 @@ local action_state = require "telescope.actions.state"
 local action_utils = require "telescope.actions.utils"
 local action_set = require "telescope.actions.set"
 local entry_display = require "telescope.pickers.entry_display"
+local from_entry = require "telescope.from_entry"
 
 local transform_mod = require("telescope.actions.mt").transform_mod
 local resolver = require "telescope.config.resolve"
@@ -670,7 +671,7 @@ local entry_to_qf = function(entry)
 
   return {
     bufnr = entry.bufnr,
-    filename = entry.path,
+    filename = from_entry.path(entry, false, true),
     lnum = vim.F.if_nil(entry.lnum, 1),
     col = vim.F.if_nil(entry.col, 1),
     text = text,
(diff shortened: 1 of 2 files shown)


lua/telescope/actions/init.lua
```diff
@@ -671,7 +671,7 @@ local entry_to_qf = function(entry)
 
   return {
     bufnr = entry.bufnr,
-    filename = from_entry.path(entry, false),
+    filename = entry.path,
```
malkomalko, line 674: `from_entry.path()` would fall back to `entry.filename` and `entry.value` in that order, but I don't believe the quickfix/loclist actions need that as well.

fdschmidt93: I can reproduce that `vim.fn.fnameescape` does not work well with the quickfix list (eg as mentioned space in the fname or something like `test%.lua`). 

However, this breaks sending outputs from other pickers to the quickfix list, e.g. `lsp_{document, workspace}_{diagnostics, symbols}` as `from_entry` resolves the filename attribute.

Maybe we should add an option to not escape the path to `from_entry` instead? Not sure.

lua/telescope/from_entry.lua
```diff
@@ -10,8 +10,12 @@ This will provide standard mechanism for accessing information from an entry.
 
 local from_entry = {}
 
-function from_entry.path(entry, validate)
+function from_entry.path(entry, validate, skip_escaping)
   local path = entry.path and vim.fn.fnameescape(entry.path) or nil
+  if skip_escaping then
+    path = entry.path
+  end
```
malkomalko, line 17: Does this keep things working in tact now for lsp etc @fdschmidt93?

fdschmidt93: This should work :)

Apologies for the nitpick, can we do it like as per below to be a bit more structurally in line with the remainder of `from_entry`? Right now, the function signature is also a bit misleading, as escaping currently does not apply to `entry.filename` or `entry.value`. I'm also wondering whether we should just expose `fnameescape` as an option we default to (because the qf list is the more out of the ordinary case here, I guess). 

```lua
function from_entry.path(entry, validate, fnameescape)
  fnameescape = vim.F.if_nil(fnameescape, true
  local path
  if entry.path then
    path = entry.path
  end
  ...
  path = fnameescape and vim.fn.fnameescape(path) or path
```
What do you think? `vim.fn.fnameescape` should typically work just fine (on Linux, macOS), but not sure if it behaves as expected on Windows.

Conni2461:
poke @malkomalko @fdschmidt93 Whats missing here? I never got an answer here: https://github.com/nvim-telescope/telescope.nvim/pull/1306#discussion_r735074527

Conni2461:
There is a new PR that tries to fix the issue: #1712

Conni2461:
Not gotten an answer since Oct 24 so i decided to merge the other PR.

Still thanks for your work :)

PR #1307, opened 2021-10-05
feat(pickers): hide cursor in normal mode

tami5:
### Purpose

Make cursor hidden when `initial_mode == "normal"`.

### Todos 

- [x] Hide/Unhide cursor with `intial_mode == "normal"`
- [ ] Toggle back cursor when going back to insert mode.
  - Maybe autocmd? ++once?
- [ ] update docs for initial_mode

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -1108,6 +1108,14 @@ actions.which_key = function(prompt_bufnr, opts)
   end
 end
 
+actions._toggle_cursor = function()
+  if vim.o.termguicolors and vim.o.guicursor ~= "" and vim.o.guicursor ~= "a:HiddenCursor" then
+    state.set_global_key("guicursor", vim.o.guicursor)
+    vim.o.guicursor = "a:HiddenCursor"
+  else
+    vim.o.guicursor = state.get_global_key "guicursor"
+  end
+end
 -- ==================================================
 -- Transforms modules and sets the correct metatables.
 -- ==================================================
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -390,6 +390,7 @@ function Picker:find()
         -- otherwise (i) insert mode exitted faster than `picker:set_prompt`; (ii) cursor on wrong pos
         await_schedule(function()
           vim.cmd [[stopinsert]]
+          actions._toggle_cursor()
         end)
       end
     else
@@ -641,6 +642,10 @@ function Picker.close_windows(status)
   end, 10)
 
   state.clear_status(status.prompt_bufnr)
+
+  if status.picker.initial_mode == "normal" then
+    actions._toggle_cursor()
+  end
 end
 
 function Picker:get_selection()
--- a/plugin/telescope.vim
+++ b/plugin/telescope.vim
@@ -78,6 +78,10 @@ highlight default link TelescopeResultsDiffAdd DiffAdd
 highlight default link TelescopeResultsDiffDelete DiffDelete
 highlight default link TelescopeResultsDiffUntracked NonText
 
+augroup TelescopeAutocmds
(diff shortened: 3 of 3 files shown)


lua/telescope/actions/init.lua
```diff
@@ -1108,6 +1108,18 @@ actions.which_key = function(prompt_bufnr, opts)
   end
 end
 
+actions._hide_cursor = function(hide)
+  if not hide then
+    return
+  end
+  state.set_global_key("guicursor", vim.o.guicursor)
```
Conni2461, line 1115: Why write this into the global state? You can just write this into the picker (hidden with a `__`), there is no reason to have this around after the picker was closed. (Thats the reason why global state exist, so we can still have access to current selection or prompt input after the picker was closed)

lua/telescope/actions/init.lua
```diff
@@ -1108,6 +1108,18 @@ actions.which_key = function(prompt_bufnr, opts)
...
+  vim.o.guicursor = "a:HiddenCursor"
+  local toggle_back = ':silent lua vim.o.guicursor = require "telescope.state".get_global_key "guicursor"'
+  vim.cmd([[ autocmd! BufLeave <buffer> ++nested ++once ]] .. toggle_back)
+  vim.cmd([[ autocmd! InsertEnter <buffer> ++nested ]] .. toggle_back)
+  vim.cmd [[ autocmd! InsertLeave <buffer> ++nested :silent lua vim.o.guicursor = "a:HiddenCursor" ]]
```
Conni2461, line 1120: Dot use `lua vim.o` just use viml, we are already in viml.

lua/telescope/actions/init.lua
```diff
@@ -1108,6 +1108,18 @@ actions.which_key = function(prompt_bufnr, opts)
...
 
+actions._hide_cursor = function(hide)
+  if not hide then
+    return
+  end
+  state.set_global_key("guicursor", vim.o.guicursor)
+  vim.o.guicursor = "a:HiddenCursor"
+  local toggle_back = ':silent lua vim.o.guicursor = require "telescope.state".get_global_key "guicursor"'
+  vim.cmd([[ autocmd! BufLeave <buffer> ++nested ++once ]] .. toggle_back)
+  vim.cmd([[ autocmd! InsertEnter <buffer> ++nested ]] .. toggle_back)
```
Conni2461, line 1119: Why `set_` and `get_` `global_key`. Why is that not just a local var and then `:silent set guicursor=...`

tami5: it can be set using local var but idk will try

tjdevries:
Can this be implemented using something else? Just like something for open and close and you can pass functions that can be executed on open and close? I'm not sure I like how it's implemented right now. I would rather not use the autocmds to do this.

tami5:
Closing .. If anyone want to hide cursor here's a personal hack to create autocmd for

```lua
autocmds.telescope_prompt = function() -- run FileType TelescopePrompt
  vim.cmd "setlocal nocursorline"
  local current = require("telescope.actions").get_current_picker(vim.api.nvim_get_current_buf())

  if current.initial_mode == "normal" then
    _G.__BACKUPCURSOR = vim.o.guicursor
    vim.cmd "highlight HiddenCursor gui=reverse blend=100" -- redundant to be called multiple times
    vim.o.guicursor = "a:HiddenCursor"
    vim.cmd [[ autocmd! BufLeave <buffer> ++nested ++once lua vim.o.guicursor = __BACKUPCURSOR ]]
    vim.cmd [[ autocmd! InsertEnter <buffer> ++nested lua vim.o.guicursor = __BACKUPCURSOR ]]
    vim.cmd [[ autocmd! InsertLeave <buffer> ++nested lua vim.o.guicursor = "a:HiddenCursor" ]]
  end
end
```

PR #1310, opened 2021-10-06
feat(healthcheck): migrate to new lua pattern

muniter:
New healthcheck pattern in master requires the renaming of the function :) https://github.com/neovim/neovim/pull/15259 this of course keeps compatibility with current stable 0.5.1

### DIFF
--- a/autoload/health/telescope.vim
+++ b/autoload/health/telescope.vim
@@ -1,3 +1,3 @@
 function! health#telescope#check()
-  lua require 'telescope.health'.check_health()
+  lua require 'telescope.health'.check()
 endfunction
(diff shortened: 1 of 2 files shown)


muniter:
No, It's backwards compatible now, I addressed the issue.

muniter:
In master it preferes the Lua one and ignores the vim one, If you are in 0.5 you won't find the Lua one.

muniter:
> But the vim one calls the Lua one (with the old name), doesn't it?
> 
> https://github.com/nvim-telescope/telescope.nvim/blob/ec487779909ebc8bffbbe054a2e8a4fbe05c0e51/autoload/health/telescope.vim#L2

That's why on this PR I rename it, therefor there's support for both.

PR #1312, opened 2021-10-06
Changed the results window highlight to always be full width

TC72:
Thought it was easier to submit this to show the effect working than describe it.  
Making the selection movement easier on the eye by staying full width, not sticking to the content width.  

Normal Width

Full Width

The code I've written is a very quick hack,  it just adds spaces to the end of a line to pad it out.  
This can probably be done in a much nicer way using virtual text which could allow it to happen on the previewer as well.  
I've also had to add an extra 2 to the width for the math to work correctly, still looking at why.

I'm looking into what's going on with the `= ` not being highlighted, don't think it's related to this code.

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -234,6 +234,13 @@ telescope.setup({opts})                                    *telescope.setup()*
 
         Default: function that shows current count / all
 
+                                         *telescope.defaults.hl_result_eol*
+    hl_result_eol: ~
+        Changes if the highlight for the selected item in the results
+        window is always the full width of the window
+
+        Default: true
+
                                  *telescope.defaults.dynamic_preview_title*
     dynamic_preview_title: ~
         Will change the title of the preview window dynamically, where it


SwenChan:
> Can you update to neovim 0.5.1 (if that segfault is still happening there. I will revert).
> 
> Edit: I know the segfault is fixed in nightly.

Thx. The issue was gone when I use neovim@0.5.1

Conni2461:
Sorry tc. I have to revert this :| I am not happy about this decision.

We should try to implement this feature without relying on `nvim_buf_set_extmark` and `hl_eol`

TC72:
@Conni2461 no problem, it has to be done right.

fdschmidt93:
> no problem, it has to be done right.

One way or another, this is the right solution. Extmarks are a critical API and segfaults were fixed super swiftly once they occurred upstream.

Maybe a working compromise would be to set `hl_results_eol=false` (or `nil`) via a version check, assuming the extmark api otherwise works. Probably much fewer headaches to just bump minimum version to 0.5.1 though and re-merge then. Saves low leverage of resources trying to make this work.

clason:
FWIW, 0.6.0 is planned for the end of the month (if all goes according to plan, which it never does...)

And Neovim itself supports only two versions:
1. the last stable (0.5.1 currently, 0.6.0 when that is released)
2. the _latest_ commit on master
Plugins should do the same unless they have good reason.

Conni2461:
> Fwiw, this is the right solution.

i know that. Just segfaulting 0.5.0

I didnt bump the version because `VimR` is still on 0.5.0: https://github.com/nvim-telescope/telescope.nvim/issues/1391#issuecomment-955739752

clason:
>  didnt bump the version because VimR is still on 0.5.0

Yeah, but there's no indication that it will be updated anytime soon (great talk at vimconf notwithstanding). An unmaintained (or at least on hiatus) GUI should not hold back telescope development...

TC72:
> > no problem, it has to be done right.
> 
> Fwiw, this is the right solution. Extmarks are a critical API and segfaults were fixed super swiftly once they occurred upstream.
> 

@fdschmidt93 just to be clear, I meant the way it gets released has to be right. Something so vital to telescope can't be causing problems for people running 0.5.0.

clason:
Disagree; people should be using 0.5.1; otherwise you have to keep supporting _every_ release until eternity.

The right thing to do for a plugin would be to tag a commit before using 0.5.1 specifics so people who can't update can freeze the plugin at the last compatible state.

Conni2461:
> people should be using 0.5.1

Yeah i agree, people should do that. If you guys wanna bump it for this feature then feel free. But we have to expect a lot of tickets. We still get tickets from people that have issues with the recent `plenary.popup` additions because they are using a 0.5.0-dev version (https://github.com/nvim-telescope/telescope.nvim/issues/1394, https://github.com/nvim-lua/plenary.nvim/issues/219 and more)

I just didnt wanna deal with it

clason:
But a clear notice at the top of the readme and be done with it :) I'll be happy to step in and close any issues from people on unsupported versions :]

Conni2461:
@TC72 wanna do a new PR with a version bump ?

TC72:
@Conni2461 #1403 
Is that OK?

PR #1325, opened 2021-10-09
feat: add `gg` and `G` to default mappings

goolord:
since we have scrolling now, I think these are sensible default mappings :)


goolord:
ah, I didn't realize that the behavior of those keybindings even changed with the scrolling PR. I can try to fix them, though this is getting mildly out of the scope of this pr imo

Conni2461:
This is out of scope of this PR. Just wanted to mention it. Updating readme is enough. After that i would merge it. Thanks for the PR :)

If you wanna write these new functions that would be awesome :)

PR #1340, opened 2021-10-12
fix: Bug with nil action error

rafaeldelboni:
After https://github.com/nvim-telescope/telescope.nvim/pull/1311 being merged I started to get the following error while trying to run a code action with clojure-lsp:
```
Error executing vim.schedule lua callback: ...acker/start/telescope.nvim/lua/telescope/builtin/lsp.lua:264: attempt to index local 'action' (a nil value)
```
So to fix this I added this extra check to use the right `action` when `resolved_action` is null.


Conni2461:
i dont see why this is happening. Your lsp saying it has resolve capabilities + error not returning anything + `resolved_action` being nil.

I have a hunch whats wrong, so what neovim version are you running (`nvim --version` output).

PR #1359, opened 2021-10-19
fix(diagnostic): set new highlight groups

ChrisAmelia:
LspDiagnosticsDefault* -> Diagnostic*
LspDiagnosticsSign* -> DiagnosticSign

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -935,6 +935,8 @@ function make_entry.gen_from_lsp_diagnostics(opts)
   opts = opts or {}
 
   local lsp_type_diagnostic = vim.lsp.protocol.DiagnosticSeverity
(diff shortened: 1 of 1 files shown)


ChrisAmelia:
After having tested it, I noticed there's an issue with the type `Warn` that is not properly handled.
Simple step to reproduce is opening a lua file and put:
```lua
notCorrect
```
Then open the Telescope's diagnostics with `:lua require'telescope.builtin'.lsp_document_diagnostics({previewer = false})`. This is the result:
image

Conni2461:
https://github.com/nvim-telescope/telescope.nvim/pull/1553

Conni2461:
Implemented by https://github.com/nvim-telescope/telescope.nvim/pull/1553 

Thanks for your initial work :)

PR #757, opened 2021-04-14
fix checking for git dir in a bare repo

polarmutex:
fixes #756

PR #1370, opened 2021-10-23
feat(worktrees): added support for bare repo git operations

pranavrao145:
This PR introduces one tiny new function to cover a git corner case. It's somewhat of an extension of #757.

If you work with `git-worktree` a lot, along with ThePrimeagen's [git-worktree.nvim](https://github.com/ThePrimeagen/git-worktree.nvim) there's one corner case you run into a lot: when you're in a bare repo, but NOT in an actual working tree. If you're in this situation, `git` actually still supports some operations, such as `git log` and `git branch`.  

The way Telescope worked before, it checked solely for the working tree, and if it didn't exist, it complained the directory was not a git directory, when it in fact was, but just a bare repo. This also invalidated any operations that you could conduct in the bare repo (like the ones talked about above). This PR changes that.

Instead of just checking for the working tree, this PR adds a second check for if you're in a bare repo (in which case you're still "in a git directory" and able to perform certain git operations). It also adds in one function to check if the current directory is a working tree, and runs that function before every operation that needs a working tree. This way, operations like `git_commits` and `git_branches` can still be run when not in a working tree, but anything else that needs the working tree like `git_status` can't.

The initial intent of this was to fix a really buggy flow in [git-worktree.nvim](https://github.com/ThePrimeagen/git-worktree.nvim), but the more I worked on it, the more it seemed like a common logic sort of thing.

Please take a look and let me know what you think! Apologies in advance if there I missed anything, I'm quite new to `lua` and the `nvim` API. If you find anything that needs fixing, let me know and I can fix it!

Thanks!

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -14,6 +14,10 @@ local conf = require("telescope.config").values
 local git = {}
 
 git.files = function(opts)
+  if opts.is_bare then
+    error "This operation must be run in a work tree"
+  end
+
   local show_untracked = utils.get_default(opts.show_untracked, true)
   local recurse_submodules = utils.get_default(opts.recurse_submodules, false)
   if show_untracked and recurse_submodules then
@@ -294,6 +298,10 @@ git.branches = function(opts)
 end
 
 git.status = function(opts)
+  if opts.is_bare then
+    error "This operation must be run in a work tree"
+  end
+
   local gen_new_finder = function()
     local expand_dir = utils.if_nil(opts.expand_dir, true, opts.expand_dir)
     local git_cmd = { "git", "status", "-s", "--", "." }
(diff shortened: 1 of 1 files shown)


lua/telescope/builtin/git.lua
```diff
@@ -351,8 +364,10 @@ local set_opts_cwd = function(opts)
   local use_git_root = utils.get_default(opts.use_git_root, true)
 
   if ret ~= 0 then
-    local output = utils.get_os_command_output({ "git", "rev-parse", "--is-inside-work-tree" }, opts.cwd)
-    if output[1] ~= "true" then
+    local in_worktree = utils.get_os_command_output({ "git", "rev-parse", "--is-inside-work-tree" }, opts.cwd)
+    local in_bare = utils.get_os_command_output({ "git", "rev-parse", "--is-bare-repository" }, opts.cwd)
+
+    if in_worktree[1] ~= "true" and in_bare[1] ~= "true" then
```
Conni2461, line 370: I have an idea. We do this check once (the two commands) and of we are in a bare and not in a worktree (like if we cant execute it `find_files`, `status`) we set a flag in `opts` (`opts.is_bare`).

The builtins we want to disable (files, status) then have a function call at the top of the function (`validate_worktree(opts)`) that errors if its forbidden.

Does this make sense for you? That way we dont call `git` that often

pranavrao145: I think this is a much cleaner approach. I've refactored it to only call the git operations once, and then do a simple check for the `find_files` and `status`. Take a look and let me know what you think!

Conni2461: Why do we need the `set_opts_bare` function. We could just do this here. after the if in 365. Otherwise we execute `git rev-pare --is-inside-work-tree` and `git rev-parse --is-bare-repository` twice again

pranavrao145: Yep sorry, didn't paying attention to the code I already wrote :laughing: - fixed now!

Conni2461 (approved):
code LGTM. Thanks :)

I'll test it and merge it later today (if i dont forget)

pranavrao145:
Sounds good - thanks!

Conni2461:
Sorry. was busy

pranavrao145:
No worries, thank you!

PR #1382, opened 2021-10-26
I've added some new functionalities to file_broser

ikerurda:
C-a to clear prompt
C-x to delete a file/folder
C-t to change directory
C-h to toggle hidden files/folders
C-r to rename a file/folder
C-o to open file with system default app


fdschmidt93:
I'm now more dedicated again to get #1257 merged in the near future. 

As an update, nvim-lua/plenary.nvim#241 now only needs to undergo (most likely) Conni's review. Then I can clean up #1257 which should hopefully be more straightforward given that the branch is rather advanced.

I'll do my best that we can include an action on opening with the system default app (have to check whether it behaves well), thanks for that :)

Therefore, I'd close this as #1257 will supersede this PR.

Once again thanks for your PR!

ikerurda:
Thanks! good work :)
Hope to see this implemented

PR #1386, opened 2021-10-28
fix: actions.paste_register selctions content

alexander-born:
I got the following error when using the `actions.paste_register`.
```
E5108: Error executing lua ...cker/start/telescope.nvim/lua/telescope/actions/init.lua:377: Expected lua string 
```
I inspected the `selection` variable selection and there is no `content` member, see
```lua
{ "doc-requirements.txt",                                                                                                                                                                                                                                             
  <metatable> = {                                                                                                                                                                                                                                                     
    __index = <function 1>,                                                                                                                                                                                                                                           
    cwd = "/home/",                                                                                                                                                                                                                                      
    display = <function 2>                                                                                                                                                                                                                                            
  }                                                                                                                                                                                                                                                                   
}                                                                                                                                                                                                                                                                     
```

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -374,7 +374,7 @@ actions.paste_register = function(prompt_bufnr)
   -- ensure that the buffer can be written to
   if vim.api.nvim_buf_get_option(vim.api.nvim_get_current_buf(), "modifiable") then
     print "Paste!"
-    vim.api.nvim_paste(selection.content, true, -1)
+    vim.api.nvim_paste(selection[1], true, -1)
   end
 end
 


Conni2461:
The table looks like this:

https://github.com/nvim-telescope/telescope.nvim/blob/0caec3d6e4d3c3c71339eb18a9aae7ed0f24badc/lua/telescope/make_entry.lua#L641-L649

How do you use this function? In a custom picker or with `Telescope registers` (works with that one)

alexander-born:
Interesting. Maybe I am using it wrong.

I mapped this action in my telescope setup.
```lua
    require'telescope'.setup({
        defaults = {
            mappings = {
              i = {
                ["<C-v>"] = actions.paste_register
              }
            }
          },
...
```
Then I run `find_files` (`:lua require('telescope.builtin').find_files({previewer = false})` and try to paste the current selection which throws the mentioned error.

I can confirm it works with `Telescope registers`.

alexander-born:
I just looked into to `make_entry.lua` file and saw that the entry table looks different for the different generation methods.
In order to make my usecase work I just wrote a custom action to paste the content of a file entry.
```lua
local paste = function(prompt_bufnr)
  local selection = require "telescope.actions.state".get_selected_entry()
  if selection == nil then return end

  require'telescope.actions'.close(prompt_bufnr)

  -- ensure that the buffer can be written to
  if vim.api.nvim_buf_get_option(vim.api.nvim_get_current_buf(), "modifiable") then
    vim.api.nvim_paste(selection[1], true, -1)
  end
end
```

I willl close this then.

PR #1401, opened 2021-11-01
Added an opts parameter to be used when entry scores are tied

TC72:
`opts.fzf_tiebreak()` can be set to a function which is passed the ordinal of the entry we're checking and the ordinal of the existing result when they have identical scores.

The default value sorts them based on the length of the ordinal, this is the default for fzf.

If you want to sort them in the order they appear in the file then you could use this:
```
opts.fzf_tiebreak = function( current_entry, existing_entry ) return false end
```

FZF allows a tiebreak to choose based on where the matched pattern is in the string. I need to check if we have access tot he search word or if we need to pass it as a 3rd parameter to this function.

### DIFF
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -154,6 +154,20 @@ append(
 )
 
 append(
+  "fzf_tiebreak",
+  function(current_entry, existing_entry)
+    return #current_entry < #existing_entry
+  end,
+  [[
+  fzf_tiebreak can be set to a function
+
+  Signature: function(current_entry.ordinal,existing_entry.ordinal) -> boolean
+
+  Default: function that breaks the tie based on entry length
+  ]]
+)
+
+append(
   "selection_strategy",
   "reset",
   [[
@@ -274,20 +288,6 @@ append(
   Boolean defining if borders are added to Telescope windows.
 
   Default: true]]
-)
-
-append(
-  "fzf_tiebreak",
-  function(current_entry, existing_entry)
-    return #current_entry < #existing_entry
-  end,
-  [[
-  fzf_tiebreak can be set to a function
-
-  Signature: function(entry2,entry1) -> boolean
-
-  Default: function that breaks the tie based on entry length
-  ]]
 )
 
 append(
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -276,11 +276,9 @@ append(
   Default: true]]
 )
 
-
-
 append(
   "fzf_tiebreak",
-  function( current_entry, existing_entry )
+  function(current_entry, existing_entry)
     return #current_entry < #existing_entry
   end,
   [[
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -159,13 +159,17 @@ append(
     return #current_entry.ordinal < #existing_entry.ordinal
   end,
   [[
-  tiebreak can be set to a function which decides how to break
-  a tie when 2 entries have the same score
+  A function that determines how to break a tie when two entries have
+  the same score.
+  Having a function that always returns false would keep the entries in
+  the order they are found, so existing_entry before current_entry.
+  Vice versa always returning true would place the current_entry
+  before the existing_entry.
 
-  Signature: function(current_entry,existing_entry, prompt) -> boolean
+  Signature: function(current_entry, existing_entry, prompt) -> boolean
 
-  Default: function that breaks the tie based on the length of the entry's ordinal
-  ]]
+  Default: function that breaks the tie based on the length of the
+           entry's ordinal]]
 )
 
 append(
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -154,16 +154,17 @@ append(
 )
 
 append(
-  "fzf_tiebreak",
-  function(current_entry, existing_entry)
-    return #current_entry < #existing_entry
+  "tiebreak",
+  function(current_entry, existing_entry, _)
+    return #current_entry.ordinal < #existing_entry.ordinal
   end,
   [[
-  fzf_tiebreak can be set to a function
+  tiebreak can be set to a function which decides how to break
+  a tie when 2 entries have the same score
 
-  Signature: function(current_entry.ordinal,existing_entry.ordinal) -> boolean
+  Signature: function(current_entry,existing_entry, prompt) -> boolean
 
-  Default: function that breaks the tie based on entry length
+  Default: function that breaks the tie based on the length of the entry's ordinal
   ]]
 )
 
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -276,6 +276,22 @@ append(
   Default: true]]
 )
 
+
+
+append(
+  "fzf_tiebreak",
+  function( current_entry, existing_entry )
+    return #current_entry < #existing_entry
+  end,
+  [[
+  fzf_tiebreak can be set to a function
+
+  Signature: function(entry2,entry1) -> boolean
+
+  Default: function that breaks the tie based on entry length
+  ]]
+)
+
 append(
   "path_display",
   {},
--- a/lua/telescope/entry_manager.lua
+++ b/lua/telescope/entry_manager.lua
@@ -1,5 +1,4 @@
 local log = require "telescope.log"
-local config = require "telescope.config"
 
 local LinkedList = require "telescope.algos.linked_list"
 
--- a/lua/telescope/entry_manager.lua
+++ b/lua/telescope/entry_manager.lua
@@ -130,6 +130,8 @@ function EntryManager:add_entry(picker, score, entry)
     return
   end
 
+  -- may need the prompt for tiebreak
+  local prompt = picker:_get_prompt()
   for index, container, node in self.linked_states:ipairs() do
     info.looped = info.looped + 1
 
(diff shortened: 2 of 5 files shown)


fdschmidt93 (approved):
Generally LGTM, thanks :)

Some thoughts:

* I would just name it `opts.tiebreak` or `opts.break_ties` though. The function itself doesn't have to have any relation with `fzf`.
* We probably would want to override this, for instance, in `current_buffer_fuzzy_find`, to default to the line number with the example you gave.
* Are there scenarios in which we'd prefer the signature to be `(current_entry, existing_entry)` rather than passing the ordinal directly? Stuff like proximity-based tie breaking for find files becomes much easier when you can access `entry.path` right away

TC72:
Thanks for the feedback, I'll change it to just opts.tiebreak.

Not sure what you mean for `current_buffer_fuzzy_find`. I did a quick test on a mapping I have for that and changed the order as I'd expected.

```
  opts.fzf_tiebreak = function( entry1, entry2 ) return false end
```

BUT, I would still like us to have an `opts` to switch off sorting completely. This would be the same as `--no-sort` in fzf.
So an obvious name would be `opts.no-sort` for that.

I'll change it to pass the whole entry rather than just the ordinal.
I know you're busy but is there an obvious way for me to grab the value of `prompt`? I see these 2 functions used within the pickers code, `self:_get_next_filtered_prompt()` and `self:_get_prompt()`.
Does the _ at the start of the function signify this should only be called inside the pickers code?

fdschmidt93:
> Not sure what you mean for current_buffer_fuzzy_find. I did a quick test on a mapping I have for that and it changed the order as I'd expected.

Apologies, I didn't properly remember something that Conni fixed after the large async PR https://github.com/nvim-telescope/telescope.nvim/pull/1146

> Does the _ at the start of the function signify this should only be called inside the pickers code?

It mostly signifies that these functions are not necessarily APIs for userland. I'm not 100% sure, I think somewhere at https://github.com/tjdevries/tree-sitter-lua there is also logic to not generate docs for functions like that (of the top of my head, I think `__` prefix)? Don't quote me on that, TJ and Conni would know by heart :)

> obvious way for me to grab the value of prompt

```lua
local prompt = current_picker:_get_prompt()
```

The other function is logic for input callbacks.

> Finally, I would still like us to have an opts to switch off sorting completely.

Maybe I'm misunderstanding what you are trying to achieve, but we already have that.

https://github.com/nvim-telescope/telescope.nvim/blob/0caec3d6e4d3c3c71339eb18a9aae7ed0f24badc/lua/telescope/sorters.lua#L496-L502

Conni2461:
> Does the _ at the start of the function signify this should only be called inside the pickers code?

One underscore means its a internal function and shouldnt be used by someone outside of this repo, extension, because we dont guarantee anything for those functions. We might just change (or remove) them without warning.

Two underscore function will not be exported by docgen: https://github.com/tjdevries/tree-sitter-lua/blob/db1478e96cbe8b0272e87567510d84363f07767b/lua/docgen/help.lua#L104-L114

> Conni needs our support

I am here. Its just new semester with a lot of classes + new part time job (20h a week). Ill try to make it work.

 I agree with schmidts comments about the code.

Thanks for the PR :)

TC72:
I've changed it to `opts.tiebreak`.
Pass the whole `entry` object, not just the `ordinal`.
Also passing the `prompt`.

I was able to use this `opts` to order based on the search term being closer to the start of the line:
```
  opts.tiebreak = function( entry1, entry2, prompt )
        local start_pos1,_ = entry1.ordinal:find(prompt)
        if start_pos1 then
            local start_pos2,_ = entry2.ordinal:find(prompt)
            if start_pos2 then
              return start_pos1 < start_pos2
            end
        end
        return false
    end
```

But, that example only handles cases where there was a full match, not a fzy match.

TC72:
Instead of doing this:

```
local prompt = current_picker:_get_prompt()
```

I tried getting the prompt inside `EntryManager:add_entry` just using
```
local prompt = picker:_get_prompt()
```

That's causing tests to fail, I can't see why as `add_entry` is only called in one place with `picker` passed in as `self` so it can't be `nil`.
But before spending time fixing this I'll wait for a response to see if you're happy with me changing the `tiebreak()` to send it the `prompt` as a parameter.

lua/telescope/entry_manager.lua
```diff
@@ -130,14 +130,16 @@ function EntryManager:add_entry(picker, score, entry)
...
+  local prompt = picker:_get_prompt()
   for index, container, node in self.linked_states:ipairs() do
     info.looped = info.looped + 1
 
     if container[2] > score then
       return self:_insert_container_before(picker, index, node, new_container)
     end
 
-    if score < 1 and container[2] == score and #entry.ordinal < #container[1].ordinal then
+    if score < 1 and container[2] == score and picker.tiebreak(entry, container[1], prompt) then
```
fdschmidt93, line 142: I would change it to `picker:tiebreak(entry, container[1])`. This preserves maximum flexibility at little to no cost with a transparent signature.

Your `prompt` use case could then with minimal overhead as follows:

```lua
local opts = {}
opts.tiebreak = setmetatable({}, {
  __call = function(self, picker, current_entry, existing_entry)
    if not self.prompt then
      self.prompt = picker:_get_prompt()
    end
    -- .. use prompt in comparison
  end,
})
```

Not that I could come up on the spot why anyone would use more than `prompt` of picker, but I'm sure people will manage :sweat_smile:

lua/telescope/config.lua
```diff
@@ -151,6 +151,21 @@ append(
...
+append(
+  "tiebreak",
+  function(current_entry, existing_entry, _)
+    return #current_entry.ordinal < #existing_entry.ordinal
+  end,
+  [[
+  tiebreak can be set to a function which decides how to break
+  a tie when 2 entries have the same score
+
+  Signature: function(current_entry,existing_entry, prompt) -> boolean
```
fdschmidt93, line 163: space after comma

lua/telescope/config.lua
```diff
@@ -151,6 +151,21 @@ append(
   - "ascending"]]
 )
 
+append(
+  "tiebreak",
+  function(current_entry, existing_entry, _)
+    return #current_entry.ordinal < #existing_entry.ordinal
+  end,
+  [[
+  tiebreak can be set to a function which decides how to break
```
fdschmidt93, line 160: \`tiebreak\`
I think we typically have full sentences ending with a full stop in comparable docs here.

when two entries (sorry, OCD-level nit).

Let's clarify that returning `true` means the current entry is placed before the existing entry (IIRC).

TC72:
I've made those changes and in the description explained how simple `return false` and `return true` tiebreak functions would work.

darzok0914:
is there any documentation added for this feature? how to set it and use it?

darzok0914:
hi, I wanted to try this feature but it doesn't seem to work, 

this is what I did:


local M = {}

M.curr_buf = function()
    local opts = {}
    opts.previewer=false
    opts.tiebreak = function(picker,current_entry,existing_entry)
        return false
    end
    require("telescope.builtin").current_buffer_fuzzy_find(opts)
end

did I miss anything? my goal is to have results appear in the same order as they are in the buffer

TC72:
@darzok0914 this PR hasn't been merged so it won't work.

@fdschmidt93 where are we on this, do I need to do anything else?

darzok0914:
@TC72 thanks for your response, so the issue was due to an error in my opts that I was fixed thanks to @Conni2461 in https://github.com/nvim-telescope/telescope-fzf-native.nvim/issues/45 

now it is all working well, I forked the latest telescope and just merged your changes. Can't wait to have this merged :)

windwp:
@Conni2461  Can we merge it?

lua/telescope/config.lua
```diff
@@ -151,6 +151,26 @@ append(
...
+  a tie when two entries have the same score.
+  A simple function which always returns false would keep the
+  entries in the order they are found, so existing_entry
+  before current_entry.
+  Vice versa always returning true would place the current_entry
+  before the existing_entry.
+
+  Signature: function(picker,current_entry,existing_entry) -> boolean
```
Conni2461, line 168: I dont like `picker,current_entry` please put a space in

lua/telescope/config.lua
```diff
@@ -151,6 +151,26 @@ append(
...
+
+  Default: function that breaks the tie based on the length of the entry's ordinal
```
Conni2461, line 170: We dont have wrapping for documentation here. So you need to manually make sure that the line isnt longer than 78.
And i think this line is too long

Conni2461 (commented):
Otherwise LGTM. @TC72 please resolve conversation after you addressed them. Otherwise i am wasting time double checking, because i havent maintained this PR

darzok0914:
@windwp what's the status on that PR?

windwp:
@darzok0914 i am not create this PR.
maybe @TC72 is busy

lua/telescope/entry_manager.lua
```diff
@@ -130,14 +130,16 @@ function EntryManager:add_entry(picker, score, entry)
     return
   end
 
+  -- may need the prompt for tiebreak
+  local prompt = picker:_get_prompt()
```
Conni2461, line 134: `_get_prompt` is not cheep. because `add_entry` will be called a lot of times. Will pass in the prompt as argument

TC72:
Thanks for merging this. I'm having real problems getting notifications from github or would have been involved in December.

PR #1403, opened 2021-11-01
Full width select

TC72:
New PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,14 @@ require('telescope').setup{
   defaults = {
     -- Default configuration for telescope goes here:
     -- config_key = value,
-    -- ..
+    mappings = {
+      i = {
+        -- map actions.which_key to <C-h> (default: <C-/>)
+        -- actions.which_key shows the mappings for your picker,
+        -- e.g. git_{create, delete, ...}_branch for the git_branches picker
+        ["<C-h>"] = "which_key"
+      }
+    }
   },
   pickers = {
     -- Default configuration for builtin pickers goes here:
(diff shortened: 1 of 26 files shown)


fdschmidt93:
Thanks taking another crack at this :)

I think what Conni meant with version bump entails:
* grep for 'nvim-0.5' and replace `has("nvim-0.5")` with `has("nvim-0.5.1")`, essentially. Mentioning it like that since, I think quotation marks are a bit inconsistent between `.vim` and `.lua`
* updating https://github.com/nvim-telescope/telescope.nvim#getting-started

Conni2461:
We can also do 1 PR with 2 commits and we dont squash

fdschmidt93:
Ah yes, simple solutions :laughing: :sweat_smile:

TC72:
Do I need to do another commit where I explicitly say "Update plugin requirement to nvim 0.5.1"?

I'll understand how I'm supposed to do this one day.

fdschmidt93:
> Also from Conni's note, should I have done a commit where I explicitly say "Update plugin requirement to nvim 0.5.1"?

Yes, two commits (ideally not but not necessarily in order):
1. BREAKING: bump minimum nvim version to 0.5.1 (then e.g. Packer highlights these)
2. feat: full-width hl of selection

I realize that requires a bit of git-gymnastics; that's why I thought it's just suggested to bump in a separate PR.

Not sure, but we might want to have a sticky issue with breaking changes as well.

TC72:
OK, I removed the version bump from this PR and crated this one: #1405

Conni2461:
Thanks :)

FWIW the "git-gymnastics" should have been the easier route. Less branch switching, etc. :laughing:

PR #1405, opened 2021-11-02
Bumped version requirement to Neovim 0.5.1

TC72:
This is required by #1403 or it will cause a segfault.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,14 @@ require('telescope').setup{
   defaults = {
     -- Default configuration for telescope goes here:
     -- config_key = value,
-    -- ..
+    mappings = {
+      i = {
+        -- map actions.which_key to <C-h> (default: <C-/>)
+        -- actions.which_key shows the mappings for your picker,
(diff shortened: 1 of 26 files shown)


fdschmidt93:
Why is the diff interleaved with #1403? Wouldn't it have been easier to make a clean branch off of current master?

https://github.com/nvim-telescope/telescope.nvim/blob/587a10d1494d8ffa1229246228f0655db2f0a48a/plugin/telescope.vim#L1

https://github.com/nvim-telescope/telescope.nvim/blob/587a10d1494d8ffa1229246228f0655db2f0a48a/README.md?plain=1#L43

Also need bumping.

TC72:
> Why is the diff interleaved with #1403? Wouldn't it have been easier to make a clean branch off of current master?

I'm afraid the only answer for that is I'm an idiot.
Hopefully fixed with this PR including the other required changes: #1406

fdschmidt93:
I guess we can close this to keep everything tidy. Otherwise, let me know and we reopen this.

PR #1406, opened 2021-11-02
breaking: bump nvim version requirement to 0.5.1

TC72:
Neovim 0.5.1 is required for #1403 or it will cause a segfault.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
             manager: sudo apt-get
             packages: -y fd-find
           - os: ubuntu-20.04
-            url: https://github.com/neovim/neovim/releases/download/v0.5.0/nvim-linux64.tar.gz
+            url: https://github.com/neovim/neovim/releases/download/v0.5.1/nvim-linux64.tar.gz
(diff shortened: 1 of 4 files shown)


fdschmidt93:
As to your question earlier, I briefly checked, you can replace `0.5.0` with `0.5.1` and then it should work (dl works in browser). Then I think we have it :)

https://github.com/nvim-telescope/telescope.nvim/blob/587a10d1494d8ffa1229246228f0655db2f0a48a/.github/workflows/ci.yml#L12-L28

TC72:
I tried but maybe I don't have permission to change a workflow?
```
! [remote rejected] version_bump_0.5.1 -> version_bump_0.5.1 (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/ci.yml` without `workflow` scope)
```

TC72:
Had to jump through one extra hoop as my github token wasn't set-up for workflows.
Looks like it's working now.

fdschmidt93 (approved):
LGTM thanks :) I'd want to have Conni's sign-off though here as well.

Conni2461 (approved):
LGTM thanks :)

PR #1408, opened 2021-11-02
feat: extend functionality of `entry_display.create`

### LINKED ISSUE #1380: Enhance lsp reference
1. add caller information  in lsp reference result view
2. result view is too short to show reference information completely  

1. both lsp reference and document symbol include range field, we can check range and find out caller

2. change width from 50 to 100
function make_entry.gen_from_quickfix(opts)
  opts = opts or {}
  -- print(vim.inspect(opts))
    separator = "▏",
    items = {
      { width = 8 },
      { width = 100 },
      { remaining = true },
    },
  }
截屏2021-10-26 下午5 48 15

l-kershaw:
This PR extends `entry_display.create` to allow passing functions or fractional values to `width` option.
It also adjusts one of the widths in `make_entry.gen_from_quickfix`.

The widths in other `make_entry` functions can also be adjusted in the same way, but I am less familiar with them, so don't know what good choices would be for them.

This PR is aimed at reducing issues like #1380, where the hardcoded values for widths aren't optimal for some screen/picker sizes. That being said I'm not 100% sure if this resolves #1380. @wellcomez does this (or something along these lines) fix your issue?

### DIFF
--- a/lua/telescope/pickers/entry_display.lua
+++ b/lua/telescope/pickers/entry_display.lua
@@ -10,15 +10,15 @@ entry_display.create = function(configuration)
   for _, v in ipairs(configuration.items) do
     if v.width then
       local justify = v.right_justify
-      local s
+      local width
       table.insert(generator, function(item)
-        if s == nil then
+        if width == nil then
           local status = state.get_status(vim.api.nvim_get_current_buf())
-          s = {}
+          local s = {}
           s[1] = vim.api.nvim_win_get_width(status.results_win) - #status.picker.selection_caret
           s[2] = vim.api.nvim_win_get_height(status.results_win)
+          width = resolve.resolve_width(v.width)(nil, s[1], s[2])
         end
-        local width = resolve.resolve_width(v.width)(nil, s[1], s[2])
         if type(item) == "table" then
           return strings.align_str(entry_display.truncate(item[1], width), width, justify), item[2]
         else
--- a/lua/telescope/pickers/entry_display.lua
+++ b/lua/telescope/pickers/entry_display.lua
@@ -1,4 +1,6 @@
 local strings = require "plenary.strings"
+local state = require "telescope.state"
+local resolve = require "telescope.config.resolve"
 
 local entry_display = {}
 entry_display.truncate = strings.truncate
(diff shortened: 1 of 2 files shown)


lua/telescope/pickers/entry_display.lua
```diff
@@ -8,11 +10,19 @@ entry_display.create = function(configuration)
...
       local justify = v.right_justify
+      local s
       table.insert(generator, function(item)
+        if s == nil then
+          local status = state.get_status(vim.api.nvim_get_current_buf())
+          s = {}
+          s[1] = vim.api.nvim_win_get_width(status.results_win) - #status.picker.selection_caret
+          s[2] = vim.api.nvim_win_get_height(status.results_win)
+        end
+        local width = resolve.resolve_width(v.width)(nil, s[1], s[2])
```
Conni2461, line 21: Cant we cache `[v.width] = resolve.resolve_width`? Because this is the generator function that will be called for every entry that will be displayed

lua/telescope/pickers/entry_display.lua
```diff
@@ -8,11 +10,19 @@ entry_display.create = function(configuration)
...
       local justify = v.right_justify
+      local width
       table.insert(generator, function(item)
+        if width == nil then
+          local status = state.get_status(vim.api.nvim_get_current_buf())
+          local s = {}
+          s[1] = vim.api.nvim_win_get_width(status.results_win) - #status.picker.selection_caret
+          s[2] = vim.api.nvim_win_get_height(status.results_win)
+          width = resolve.resolve_width(v.width)(nil, s[1], s[2])
+        end
```
Conni2461, line 21: this should not work. we need a table because multiple functions point to the same reference. So the width for the second column will always be the same as the width for the first column. (Under the assumption that the second one has a width set)

l-kershaw: I didn't think that this was the case, since `width` is scoped inside the for loop.
In my tests things seem to work as expected, although I might not be testing widely enough.

Conni2461: You are right

Conni2461 (approved):

l-kershaw:
@Conni2461 thanks for reviewing this 🙂

Are there any other places where `entry_display.create` is used that you think we could make use of this functionality?
The `gen_from_autocommands`, `gen_from_commands` and `gen_from_vimoptions` seem like the most reasonable candidates to me, but I'm not sure what would be good choices for fractional widths there, as I don't really use them in my workflow.

l-kershaw:
I've had a go at coming up with good defaults for fractional widths in cases where it seemed reasonable, but being honest I don't use most of them, so may have chosen a suboptimal value in places.

Unless there are any complaints, I will probably merge this in a day or two. If any issues come up after that they should be relatively easy to fix

Issue #1409, opened 2021-11-02
The `show_line` option is ignored in lsp_document_symbols


YorickPeterse:
### Description

The `show_line` option is used to disable showing the line of an LSP symbol. According to the documentation, this option is disabled by default. However, in reality it seems this option is enabled. For example:

![Screenshot from 2021-11-02 22-32-50](https://user-images.githubusercontent.com/86065/139954791-4cb3ff69-5140-4c52-a2c2-e27dcd23a293.png)

One part of the problem seems to be the code found here:

https://github.com/nvim-telescope/telescope.nvim/blob/9cad3a4a5d0e36b07b25c4be1db1c1306fcec945/lua/telescope/make_entry.lua#L356-L367

It seems that when `ignore_filename` is set to `true` (which is the default), the `show_line` option is ignored and the line is always displayed.

Another problem appears to this:

https://github.com/nvim-telescope/telescope.nvim/blob/f31ef362931907bbdd3b46cb880b34493c2d1882/lua/telescope/builtin/lsp.lua#L134

If `ignore_filename` is set to `false`, the expression evaluates to `opts.ignore_filename = false or true`, resulting in `opts.ignore_filename = true`. Unless I'm mistaken this basically makes the option useless, as it's always set to `true`.

### Neovim version

NVIM v0.6.0-dev+547-gfa97d3485
Build type: RelWithDebInfo
LuaJIT 2.0.5

### Operating system and version

Arch Linux, kernel 5.14

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - OK: nvim-treesitter installed.

## Checking external dependencies
  - OK: rg: found ripgrep 13.0.0
  - OK: fd: found fd 8.2.1

## ===== Installed extensions =====

## Telescope Extension: `fzf`
  - INFO: No healthcheck provided
```

### Steps to reproduce

Configure Telescope along the lines of the following:

```lua
require('telescope').setup({
  pickers = { lsp_document_symbols = { show_line = false } }
})
```

Next, open a document that has a LSP client attached, and run `Telescope lsp_document_symbols`.

### Expected behavior

The line the symbol is defined on isn't displayed.

### Actual behavior

The line is always displayed

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])

[...]

PR #1410, opened 2021-11-02
Fix option handling for lsp_document_symbols

YorickPeterse:
As discussed in https://github.com/nvim-telescope/telescope.nvim/issues/1409, the lsp_document_symbols picker wasn't handling its options properly. The `ignore_filename` option was always set to `true`, and the `show_line` option was effectively ignored. In addition, if `show_line` did work, it would result in the line and column number being displayed, rather than the actual source line (which is what other pickers do, except for the treesitter picker).

This commit fixes both issues. When `show_line` is enabled, the source line is displayed. Line and column numbers are not shown, mostly to keep the behaviour the same as other pickers. If this information should be shown (which didn't work before anyway), it should probably use a separate option.

This fixes https://github.com/nvim-telescope/telescope.nvim/issues/1409

### DIFF
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -144,7 +144,6 @@ lsp.document_symbols = function(opts)
       return
     end
 
-    opts.ignore_filename = opts.ignore_filename or true
     pickers.new(opts, {
       prompt_title = "LSP Document Symbols",
       finder = finders.new_table {
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -144,6 +144,7 @@ lsp.document_symbols = function(opts)
       return
     end
 
+    opts.path_display = { "hidden" }
     pickers.new(opts, {
       prompt_title = "LSP Document Symbols",
       finder = finders.new_table {
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1195,9 +1195,11 @@ builtin.quickfix({opts})                        *telescope.builtin.quickfix()*
         {opts} (table)  options to pass to the picker
 
     Options: ~
-        {ignore_filename} (boolean)  dont show filenames (default: true)
-        {trim_text}       (boolean)  trim results text (default: false)
-        {nr}              (number)   specify the quickfix list number
+        {show_line}   (boolean)  show results text (default: true)
+        {trim_text}   (boolean)  trim results text (default: false)
+        {fname_width} (number)   defines the width of the filename section
+                                 (default: 30)
+        {nr}          (number)   specify the quickfix list number
 
 
 builtin.quickfixhistory({opts})          *telescope.builtin.quickfixhistory()*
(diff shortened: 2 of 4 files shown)


lua/telescope/builtin/lsp.lua
```diff
@@ -131,7 +131,7 @@ lsp.document_symbols = function(opts)
     return
   end
 
-  opts.ignore_filename = opts.ignore_filename or true
+  opts.ignore_filename = utils.get_default(opts.ignore_filename, true)
```
YorickPeterse, line 134: I'll be honest: the way these options work is a bit wack. Basically you have two different options (`show_line` and `ignore_filename`), but only one of these can be enabled at a time. I decided not to refactor that as I'm not familiar enough with the Telescope code for that. Ideally this option is also renamed to `show_filename` with the default being `false`, as inverted options (`ignore_x` defaulting to `true`) are a bit harder to wrap your head around.

YorickPeterse:
With these fixes, the default is to only show the symbol name and type. If `show_line = true`, then the source line is also included. If `ignore_filename = false`, the filename (and _just_ the filename, not including the line/column) is displayed. The latter is to match other pickers, which only show the filename.

lua/telescope/builtin/lsp.lua
Conni2461: Yeah i agree, but we kinda have this problem of changing option names that it would break things.
It happened and we have this option for 11 different pickers which makes it annoying.

`vim.F.if_nil` over `utils.get_default`. we wanted to move all of these over to `vim.F.if_nil` but havent found the time.

About options being a little bit wack i also agree. I think we did a way better job implementing `make_entry` for ctags. where you can have `show_line` and `filepath` which is there solely handled with `path_display`. So you can hide it with `path_display = { "hidden" }` maybe we should do this. for all other pickers (like the lsp once)

Conni2461:
hmm maybe we actually never need the filename for document symbols. Its always the same file. I think this is the intention behind treesitter and document_symbols and thats why it displayed the line number and col. Its technically more useful.

YorickPeterse:
> hmm maybe we actually never need the filename for document symbols. Its always the same file. I think this is the intention behind treesitter and document_symbols and thats why it displayed the line number and col. Its technically more useful.

Good point, it probably would also simplify the code a bit, as there's one less option to worry about.

Conni2461:
Do you have an idea on how to proceed here? I dont have any strong preferences. But we should fix it up that either the current documentation reflects a possible implementation or the current implementation reflects a possible documentation.

I think we can just remove filename for these two pickers (document_symbols and treesitter). We dont need them there.

You also need to rebase. Sorry thats my bad. Merged something else before :grimacing:

YorickPeterse:
@Conni2461 Provided the filename isn't really useful here, I think we should just not support it and update the documentation for `lsp_document_symbols` to clarify this.

Conni2461:
I am so sorry for not responding for 6 months (LUL). I rebased this to the current dev branch (see #1938) and wanted to merge but i just realized that both `lsp_workspace_symbols` as well as `lsp_dynamic_workspace_symbols` are using the same `make_entry` function.

So we should not drop `ignore_filename` just set it to true for `lsp_document_symbols` and update the documentation. I will keep the changes you've done to show_line because i think they are good.

Ill finish it. Thanks for your work here and sorry again for the really really late response

Conni2461:
okay there is more stuff that we can clean up. Like we can drop `ignore_filename` from everywhere in favor of `path_display={"hidden"}` to make everything consistent. that would be a breaking change but thats fine.

I try to finish this in the next couple of days :)

YorickPeterse:
@Conni2461 Thanks for looking into this, and no worries about it taking a bit longer than expected :smile:

PR #1421, opened 2021-11-05
fix: Border window also need to be winblend due to window overlap in multigrid UI

akiyosi:
In a GUI that follows the multigrid UI, `winblend` must be set for all overlapping float windows in order to make them transparent. 
Since the current telescope.nvim does not have `winblend` set for border windows, the winblend setting in the telescope is disabled in the multigrid GUI. Thus, there is a border window that is not set to be transparent behind each float window.

This PR also sets `winblend` for the border float window generated by telescope.nvim, which fixes this problem.

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -291,6 +291,9 @@ function Picker:_create_window(bufnr, popup_opts, nowrap)
     a.nvim_win_set_option(win, "wrap", false)
   end
   local border_win = opts and opts.border and opts.border.win_id
+  if border_win then
+    a.nvim_win_set_option(border_win, "winblend", self.window.winblend)
+  end
   return win, opts, border_win
 end
 


Conni2461 (approved):
Makes sense to me. Thanks

Can you rebase? I think the docs changes are already in master

akiyosi:
@Conni2461 
Thanks for the confirmation, I've rebased it.

l-kershaw:
Do we need a check for whether or not we actually have a border?
It doesn't seem to break when I disable borders, but I would expect it to.
How does `nvim_win_set_option` handle being passed `nil` for the window?

akiyosi:
@l-kershaw 
Thanks for the reply.

I followed neovim's code; [helpers.c:L439](https://github.com/neovim/neovim/blob/caa6992a1071a2ac373bec21085685da4a1790d6/src/nvim/api/private/helpers.c#L439), and my understanding is that the following code is called, which will output an "invalid win" error.
If my understanding is correct, then if nil is passed, it will only cause an error.
Also, if I run telescope.nvim with `border = false`, I don't get any errors, and it doesn't seem to cause any incorrect behavior.

l-kershaw:
@akiyosi I've done a couple more tests, and I think that if `nil` is passed for the window to `nvim_win_set_option`, then it defaults to the current window.

I don't think that with the way things currently work in telescope we actually require a check (I think the current window at the relevant time is just the prompt window).
However, just to avoid confusing issues coming up if things change in the future, it does no harm to add a check like:
```
if border_win then
  a.nvim_win_set_option(border_win, "winblend", self.window.winblend)
end
```

akiyosi:
@l-kershaw 
Thanks for the appropriate guidance.
As you pointed out, I thought I should implement the code to anticipate future possibilities, so I fixed and rebased the code.

l-kershaw (approved):
@akiyosi Thanks for the quick changes 🙂

Merging now

PR #1429, opened 2021-11-08
feat: `exclude` option for `shorten` in `path_display`

### LINKED ISSUE #880: shorten_path: Support length and option to exlcude positions from shortening

l-kershaw:
Requires nvim-lua/plenary.nvim#268

Users can specify which parts of the path to exclude from shortening.
Still defaults to shortening everything except the tail of the path.
Will probably mostly be used as `exclude = {1,-1}`, which shortens everything except the first folder, and the tail.

Resolves #880

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -207,6 +207,20 @@ telescope.setup({opts})                                    *telescope.setup()*
           Similarly, `path_display.shorten = 2` will give a path like:
             `al/be/ga/delta.txt`
 
+        You can also further customise the shortening behaviour by
+        setting `path_display.shorten = { len = num, exclude = list }`,
+        where `len` acts as above, and `exclude` is a list of positions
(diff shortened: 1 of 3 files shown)


Conni2461 (approved):
Looks good.
Thanks ;)

PR #1431, opened 2021-11-09
Fix typo in telescope.txt

dsychin:
There seems to be a typo in the themes section of the help text.

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1487,8 +1487,8 @@ builtin.lsp_workspace_diagnostics({opts}) *builtin.lsp_workspace_diagnostics()*
 
 Themes are ways to combine several elements of styling together.
 
-They are helpful for managing the several differnt UI aspects for telescope and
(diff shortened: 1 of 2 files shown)


fdschmidt93:
Thanks!

However, our vimdocs are auto-generated from EmmyLua annotations via https://github.com/tjdevries/tree-sitter-lua

Therefore, the typo would actually need to be fixed here

https://github.com/nvim-telescope/telescope.nvim/blob/e4ab6d28448747e7b0f922f5b8d2dbd63c24c176/lua/telescope/themes.lua#L11

and then CI will commit an update to reflect that change in the docs in your PR.

dsychin:
Hi @fdschmidt93, I've updated the change.

fdschmidt93:
docgen CI doesn't seem to be firing. I _think_ the fix is to go to `Settings->Actions->Check [] Allow all actions`.

dsychin:
Hmm, it is already set to allow all actions.

image

PR #1436, opened 2021-11-09
feat: results scrolling actions

### LINKED ISSUE #1326: [Mapping feature] results_scrolling_up and results_scrolling_down hooks
Hello Telescope crew,

Excellent work with regards to the recent merge of #1232 (results scrolling, up to 1,000). Many thanks for that.

With that foundation in place, I would like to be able to page up/down the results list quickly via custom mappings, similar to the existing `results_scrolling_up` and `results_scrolling_down` hooks listed in [mappings.lua](https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/mappings.lua#L25). But for the results list.

I could not seem to find an existing request for this, hence this new issue (apologies if I am mistaken).

Currently, I only see `<Up>` and `<Down>` single line results scrolling. Like preview scrolling, ideally I would like half page result scrolling hooks.

Possible?
Or already available?

Cheers.

### LINKED ISSUE #1430: PageUp and PageDown bindings

Please add PageUp / PageDown actions that would scroll the result window (not the preview window).

I think the above sentence is self-descriptive. That would be useful to page and scroll through big results (e.g., grep or rg). PageUp/PageDown (similar to Ctrl-U/Ctrl-D/Ctrl-B/Ctrl-F in vim) should work by default, or through a customizable keybindings.

Note that, fzf supports this by default.

l-kershaw:
WIP

Doesn't look quite right to me when scrolling upwards as the selection looks like it is alternating between the top and the middle of the results window. Any suggestions to fix this would be appreciated 🙂

Fixes #1430
Fixes #1326

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -2198,6 +2198,15 @@ action_set.scroll_previewer({prompt_bufnr}, {direction}) *action_set.scroll_prev
         {direction}    (number)  The direction of the scrolling
 
 
+action_set.scroll_results({prompt_bufnr}, {direction}) *action_set.scroll_results()*
+    Scrolls the results up or down
+
+
+    Parameters: ~
+        {prompt_bufnr} (number)  The prompt bufnr
+        {direction}    (number)  The direction of the scrolling
+
+
 
 ================================================================================
                                                        *telescope.actions.utils*
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -2190,7 +2190,9 @@ action_set.edit({prompt_bufnr}, {command})                 *action_set.edit()*
 
 
 action_set.scroll_previewer({prompt_bufnr}, {direction}) *action_set.scroll_previewer()*
-    Scrolls the previewer up or down
+    Scrolls the previewer up or down. Defaults to a half page scroll, but can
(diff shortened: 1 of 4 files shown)


l-kershaw:
I have fixed the "selection alternation" issue, so I think I'm happy with this now.

Demo:
results_scrolling

@fdschmidt93 or @Conni2461, could you test this if you have time?

fdschmidt93:
LGTM, just tried it and seems to work very well across the `{default, ivy, dropdown}` theme and both `sorting_strategy` `asencding` and `descending`. Code also looks great :)

Tiny nit on docs, while we are at it. As a person who virtually never scrolls, I was slightly confused the first time around how fast it is supposed to scroll (of course, half a page is figured out right away though it's too easily clarified).

`Scrolls the results half a page up or down`

and I think we should do the same for `--- Scrolls the previewer up or down`

Then you can merge from my end. Conni seems to be on (well-deserved) hiatus. We should merge what we can with reasonable confidence I suppose :) (at least so long as we don't have fixed versions :sweat_smile:)

l-kershaw:
Updated docs and merging now.

Didn't realise that Conni was on hiatus, will try not to bother them with as much stuff then 👍

Conni2461:
Not on hiatus

PR #1439, opened 2021-11-10
Fix actions.git_checkout_current_buffer calling wrong method

hungpt2702:
`actions.git_checkout_current_buffer` should call `action_state.get_selected_entry` instead of `actions.get_selected_entry`.


fdschmidt93:
LGTM thanks :)

PR #1444, opened 2021-11-11
docs: add a changelog entry for updated version requirements

### LINKED ISSUE #1443: Possibility of adding versioning to the project
Thanks for building this plugin, it's been extremely useful. Recently I ran `:PackerSync` and started receiving notifications about the recent requirement bump to neovim 0.5.1. I noticed there is no versioning around the project so I had to manually pin to an old commit in my packer config to:
```
use {
    commit = "3b9ac8e",
```
So that the error would go away. I couldn't find any version or CHANGELOG information on the change and had to jump through the issues to find the commit I could pin to. This would probably be a frustrating experience for some users, have there been talks of adding more information around releases so that breaking changes can be mitigated? The messaging around:
```
Telescope.nvim requires at least nvim-0.5.1. Please update or uninstall
```
Could be a little more explicit on what versions require the update?

l-kershaw:
Adds a changelog explaining the version bump and points users to it when they don't have the right version.

I can't think of a way of checking if someone has an old version of nightly to give them the error on launch as well.
If anyone can think of one, then that would be helpful.

Fixes #1443

### DIFF
--- a/doc/telescope_changelog.txt
+++ b/doc/telescope_changelog.txt
@@ -127,8 +127,8 @@ PR: https://github.com/nvim-telescope/telescope.nvim/pull/1406
 Telescope requires Neovim release 0.5.1 or a recent nightly
 
 Due to making use of newly implemented extmark features, Telescope now
-requires users to be on Neovim 0.5.1 (the most recent stable version) or on a
-recent version of Neovim nightly (at most 1 week old).
+requires users to be on Neovim 0.5.1 (the most recent stable version) or on
+the LATEST version of Neovim nightly.
 
 
  vim:tw=78:ts=8:ft=help:norl:
--- a/doc/telescope_changelog.txt
+++ b/doc/telescope_changelog.txt
@@ -119,5 +119,16 @@ Guide to switch over to plenary.path
         before: require("telescope.path").read_file_async(filepath, callback)
         now:    require("plenary.path"):new(filepath):read(callback)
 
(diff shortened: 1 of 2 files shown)


doc/telescope_changelog.txt
```diff
@@ -119,5 +119,16 @@ Guide to switch over to plenary.path
...
+                                                    *telescope.changelog-1406*
+
+Date: November 4, 2021
+PR: https://github.com/nvim-telescope/telescope.nvim/pull/1406
+
+Telescope requires Neovim release 0.5.1 or a recent nightly
+
+Due to making use of newly implemented extmark features, Telescope now
+requires users to be on Neovim 0.5.1 (the most recent stable version) or on a
+recent version of Neovim nightly (at most 1 week old).
```
fdschmidt93, line 131: I'd like to keep this consistent with the README:

> Neovim (v0.5.1) or the latest neovim nighly commit is required for telescope.nvim to work.

which avoids the slippery slope that "at most 1 week old" may be.

l-kershaw: I see your point, but I also think that it is easy for a user to miss the word "latest", which then causes with confusion.
Being explicit about a timeframe helps to avoid that.

Maybe we could make the changelog say:

> ... or on the LATEST version of Neovim nightly.

and then leave the README as is?

fdschmidt93: Sounds good :)

PR #1448, opened 2021-11-11
feat(draft): experimental parsing colors from rg

fdschmidt93:
experimental branch to parse colors from `rg` directly.

It sort of works:

- offset computation for highlighting requires mores sophistication (why +4, proper devicons handling)
- combining with `fzy` highlighting required for eg `grep_string`
- integrating highlight from `entry_maker` into `pickers.highlight_one_row`

Not sure if I'll finish this this but seems quite a bit easier than #1281 -- very happy for someone who desires this feature to take over.

Tagging #1240

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -2367,7 +2367,7 @@ work and make it really easy create new previewers.
 
 Furthermore, there are a collection of previewers already defined which can be
 used for every picker, as long as the entries of the picker provide the
-necessary fields. The more important ones are
+necessary fields. The more important once are
   - `previewers.cat`
   - `previewers.vimgrep`
   - `previewers.qflist`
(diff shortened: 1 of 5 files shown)


Conni2461:
`vim.json` is not part of 0.5.1 so we need to wait a little longer. But the json will probably be faster than the other approach. 

What i am thinking is that we can play around with the highlighing here: https://github.com/nvim-telescope/telescope-live-grep-raw.nvim doesnt matter if this is with this pr or with ansi parsing

fdschmidt93:
Makes sense, I like the idea: @weeman1337 would you want to take a crack at this as part of https://github.com/nvim-telescope/telescope-live-grep-raw.nvim ?

weeman1337:
> Makes sense, I like the idea: @weeman1337 would you want to take a crack at this as part of https://github.com/nvim-telescope/telescope-live-grep-raw.nvim ?

Yes, why not . Keep an eye on https://github.com/nvim-telescope/telescope-live-grep-raw.nvim/issues/1

Conni2461:
Closing. People will play around with it in telescope-rg. Which they can do because 0.6 is out :tada:

PR #1449, opened 2021-11-12
feat(lap_references): exclude current line

tami5:
### Purpose

~I don't know why this isn't default, but I added an option `include_declaration` in case someone, for some reason want to have the current line from which the lsp-references was called be the first item in the picker.~

Recent changes will just ignore current line from the result by default. Which is sufficient. `include_declaration` will still be an option though default to true.

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1342,10 +1342,12 @@ builtin.lsp_type_definitions({opts})          *builtin.lsp_type_definitions()*
         {opts} (table)  options to pass to the picker
 
     Options: ~
-        {jump_type}       (string)   how to goto definition if there is only
-                                     one, values: "tab", "split", "vsplit",
-                                     "never"
-        {ignore_filename} (boolean)  dont show filenames (default: true)
+        {jump_type}           (string)   how to goto definition if there is
+                                         only one, values: "tab", "split",
+                                         "vsplit", "never"
+        {ignore_filename}     (boolean)  dont show filenames (default: true)
+        {include_declaration} (boolean)  include symbol declaration in the lsp
+                                         references (default: true)
 
 
 builtin.lsp_implementations({opts})            *builtin.lsp_implementations()*
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -13,8 +13,10 @@ local utils = require "telescope.utils"
 local lsp = {}
 
 lsp.references = function(opts)
+  local filepath = vim.api.nvim_buf_get_name(opts.bufnr)
+  local lnum = vim.api.nvim_win_get_cursor(opts.winnr)[1]
   local params = vim.lsp.util.make_position_params(opts.winnr)
-  params.context = { includeDeclaration = true }
+  params.context = { includeDeclaration = vim.F.if_nil(opts.include_declaration, true) }
 
   vim.lsp.buf_request(opts.bufnr, "textDocument/references", params, function(err, result, ctx, _config)
     if err then
@@ -24,10 +26,11 @@ lsp.references = function(opts)
 
     local locations = {}
     if result then
-      vim.list_extend(
-        locations,
-        vim.lsp.util.locations_to_items(result, vim.lsp.get_client_by_id(ctx.client_id).offset_encoding) or {}
-      )
+      local results = vim.lsp.util.locations_to_items(result, vim.lsp.get_client_by_id(ctx.client_id).offset_encoding)
+      locations = vim.tbl_filter(function(v)
+        -- Remove current line from result
+        return not (v.filename == filepath and v.lnum == lnum)
+      end, results or {})
     end
 
     if vim.tbl_isempty(locations) then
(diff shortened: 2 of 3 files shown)


thuan1412:
It's really great PR.

tami5:
@fdschmidt93 in last commit I added `ignore_symbols`, and not sure if this is the same as `opts.symbol`. in ignore_symbol case, if I have `{"function"}` it will ignore it `Function` kind

doc/telescope.txt
```diff
@@ -2367,7 +2373,7 @@ work and make it really easy create new previewers.
 
 Furthermore, there are a collection of previewers already defined which can be
 used for every picker, as long as the entries of the picker provide the
-necessary fields. The more important ones are
+necessary fields. The more important once are
```
disrupted, line 2376: this seems like a mistake

disrupted (changes requested):

fdschmidt93:
Quick question. `ignore_symbols` doesn't relate to the original intent of the PR right? That got me a bit confused :sweat_smile: 

The code looks good and for that reason I didn't include it, as the user can quite simply do `opts.ignore_symbols` via `opts.symbols` I suppose. I'm ok with including it, but then we should raise an error if both `opts.ignore_symbols` and `opts.symbols` are set at the same time.

Conni2461:
@tami5 Are you still working on this? As fd said please split up the PRs. And you might wanna rebase

Conni2461:
poke @tami5

doc/telescope.txt
tami5: 😅 I made it worst, thanks @disrupted

tami5:
> poke @tami5

:| oh sorry @Conni2461,  I don't know why I didn't received mentions, or notifications. I look into it, Anyways, yes, I will hopefully close this PR today

> Quick question. ignore_symbols doesn't relate to the original intent of the PR right? That got me a bit confused 😅

Yes, indeed, it doesn't relate to the original intent of the PR, I guess while creating the PR I fixed another issue I had. I will extract to another PR.

> The code looks good (haven't tried it, but I trust you :) and I didn't include it, as the user can quite simply do opts.ignore_symbols via opts.symbols. I'm ok with including it, but then we should raise an error if both opts.ignore_symbols and opts.symbols are set at the same time.

😆 trust is big word for me 🙈. Yes, I felt it does, but again one makes you explicitly mention what you want and the other let you mention what you don't want. An error would make sense indeed.

tami5:
Okay it seems that having ignore declaration by default has some drawbacks, for example, calling lsp_references on a REFERENCE doesn't include the declaration by default. 

I'm currently thinking of a workaround this .... if non found, I think I'd rather deal with the pain of having the declaration shown as first result and not having the declaration while checking out a reference .. tbc

tami5:
hmmm maybe I could instead remove current line from the matches 🤔

tami5:
ready to be merged if anyone interested in reviewing it

lua/telescope/builtin/lsp.lua
```diff
@@ -36,10 +46,7 @@ lsp.references = function(opts)
 
     pickers.new(opts, {
       prompt_title = "LSP References",
-      finder = finders.new_table {
-        results = locations,
-        entry_maker = opts.entry_maker or make_entry.gen_from_quickfix(opts),
-      },
+      finder = finders.new_table { results = locations, entry_maker = entry_maker },
```
Conni2461, line 49: why was entry_maker moved up?

lua/telescope/builtin/lsp.lua
```diff
@@ -13,21 +13,31 @@ local utils = require "telescope.utils"
...
-  params.context = { includeDeclaration = true }
+  local filepath = vim.fn.expand "%:p"
+  local lnum = vim.fn.line "."
+  local entry_maker = opts.entry_maker or make_entry.gen_from_quickfix(opts)
+  local action = "textDocument/references"
+  local params = vim.tbl_extend("keep", vim.lsp.util.make_position_params(), {
+    context = {
+      includeDeclaration = vim.F.if_nil(opts.include_declaration, true),
+    },
+  })
```
Conni2461, line 24: the previous solution is more elegant imo

lua/telescope/builtin/lsp.lua
```diff
@@ -13,21 +13,31 @@ local utils = require "telescope.utils"
...
-        locations,
-        vim.lsp.util.locations_to_items(result, vim.lsp.get_client_by_id(ctx.client_id).offset_encoding) or {}
-      )
+      local offset_encoding = vim.lsp.get_client_by_id(ctx.client_id).offset_encoding
+      local fixed_result = vim.lsp.util.locations_to_items(result, offset_encoding)
+      local found = false -- break early :)
+      locations = vim.tbl_filter(function(v)
+        -- Remove current line from result
+        return found or not (v.filename == filepath and v.lnum == lnum)
+      end, fixed_result or {})
```
Conni2461, line 40: i dont get this `found` var. You never change it. so ?

lua/telescope/builtin/lsp.lua
tami5: readability no special reason

lua/telescope/builtin/lsp.lua
tami5: oh yah nice catch

PR #1450, opened 2021-11-13
fix typo in previewers documentation

zklinger:
While trying out generating documentation on my local machine I have noticed PR #1433 has modified the documentation file but not the source file the content is being generated from.


fdschmidt93:
Thanks for cleaning up my slight oversight on that PR :)

PR #1451, opened 2021-11-13
feat: allow changing results/prompt title & persisting multi on picker:refresh

fdschmidt93:
This PR exposes the `{prompt, results}_border` to the picker, such that the title can be changed on the fly with `picker.prompt_border:change_title` which is a nicety of the upcoming file browser extension.

In addition, it allows possibly maintaining multi selections when a finder is refreshed.

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -335,13 +335,15 @@ function Picker:find()
   end
   -- This is needed for updating the title
   local preview_border = preview_opts and preview_opts.border
+  self.preview_win = preview_win
   self.preview_border = preview_border
 
   local prompt_win, prompt_opts, prompt_border_win = self:_create_window("", popup_opts.prompt)
   local prompt_bufnr = a.nvim_win_get_buf(prompt_win)
 
-  self.prompt_border = prompt_opts and prompt_opts.border
   self.prompt_bufnr = prompt_bufnr
+  self.prompt_win = prompt_win
+  self.prompt_border = prompt_opts and prompt_opts.border
 
   -- Prompt prefix
   local prompt_prefix = self.prompt_prefix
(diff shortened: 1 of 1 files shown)


lua/telescope/pickers.lua
```diff
@@ -336,8 +337,10 @@ function Picker:find()
   local preview_border = preview_opts and preview_opts.border
   self.preview_border = preview_border
 
-  local prompt_win, _, prompt_border_win = self:_create_window("", popup_opts.prompt)
+  local prompt_win, prompt_opts, prompt_border_win = self:_create_window("", popup_opts.prompt)
   local prompt_bufnr = a.nvim_win_get_buf(prompt_win)
+
+  self.prompt_border = prompt_opts and prompt_opts.border
   self.prompt_bufnr = prompt_bufnr
```
l-kershaw, line 344: Might be good to standardise this so things match between results/preview/prompt. The layout you have for results seems like the best way. We can also add `self.preview_win` and `self.prompt_win` at the same time (maybe not for `self.preview_bufnr`, as it changes right?)

lua/telescope/pickers.lua
```diff
@@ -787,7 +790,7 @@ function Picker:refresh(finder, opts)
   if finder then
     self.finder:close()
     self.finder = finder
-    self._multi = MultiSelect:new()
+    self._multi = vim.F.if_nil(opts.multi, MultiSelect:new())
```
l-kershaw, line 793: Probably a good idea to put this in the docs bit above like `opts.reset_prompt`. Can also explain what sort of object to be passed here.

l-kershaw (commented):
This seems like a good idea, as in some cases (like `file_browser`) its good to give a bit more context on where the results are from.

lua/telescope/pickers.lua
fdschmidt93: Good idea :) I streamlined the docs. Side note: probably eventually good to push for broader docs again (eg get #927 merged).

lua/telescope/pickers.lua
fdschmidt93: Agreed, but I suppose in big picture it's more convoluted as it practically makes `telescope.state` a bit vain I suppose. I'd have to grep the code, but my first hunch is we are inconsistent between getting buffers / windows from picker or the (global) state.

fdschmidt93:
Thanks for the review :)

fdschmidt93:
I'd merge this tomorrow noon unless there's further comments since I've realized forwarding the multi selections is quintessential to the forthcoming `telescope-file-browser` extension 😅

PR #1452, opened 2021-11-14
fix: git stash entry formatting

farbodsz:
Currently entry items in git stash aren't displayed properly. Here's an example:

_Before_

screenshot_20211114_181906

Note that:

1. There is a space before the entry item text
2. The whole entry is not displayed (because this commit message contains a colon)

Here is the equivalent output from `git stash --no-pager list`

screenshot_20211114_182316

This commit fixes this issue by changing the split delimiter and including everything after the second '`:`' in the displayed entry text.

_After_

screenshot_20211114_181921

As an aside, I do think it would be useful to also show the branch name, as you get with the command line output. Happy to add this into the PR if you agree.

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -72,7 +72,8 @@ git.commits = function(opts)
 end
 
 git.stash = function(opts)
-  opts.entry_maker = vim.F.if_nil(opts.entry_maker, make_entry.gen_from_git_stash())
+  opts.show_branch = vim.F.if_nil(opts.show_branch, true)
+  opts.entry_maker = vim.F.if_nil(opts.entry_maker, make_entry.gen_from_git_stash(opts))
 
   pickers.new(opts, {
     prompt_title = "Git Stash",
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -971,6 +971,8 @@ builtin.git_stash({opts})                                *builtin.git_stash()*
         {cwd}          (string)   specify the path of the repo
         {use_git_root} (boolean)  if we should use git root as cwd or the cwd
                                   (important for submodule) (default: true)
+        {show_branch}  (boolean)  if we should display the branch name for git
+                                  stash entries (default: true)
 
 
 builtin.builtin({opts})                                    *builtin.builtin()*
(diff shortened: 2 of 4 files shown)


l-kershaw:
This seems like a good idea, thanks 👍
Definitely think that it is worth showing the branch name (maybe add an option for it with `git.stash` in `telescope/builtin/git.lua`?).

Depending on the final implementation, it may be worth using the `--format=...` option in the git command to specify the way we want git to print things. _(Happy to wait and see on this if you'd rather not do this right now)_

Feel free to ping me if you need any help/feedback on this 🙂

farbodsz:
Thanks for the feedback, @l-kershaw , and apologies for the long delay - I've been fairly busy of the last couple weeks so only had a chance to look at this again just today.

I've just pushed a commit that includes the branch name and adds highlights for the different components like this:

screenshot_20211203_213447

This corresponds to the following `git stash list` output:

```
stash@{0}: WIP on master: f27742c chore: update todo
stash@{1}: On feat/codegen: Move params out of its own node
stash@{2}: WIP on feat/codegen: 1cbc829 build: add `--parse-only` flag for mccomp
```

Let me know what you think, and if all good I'll start adding the option for it in `telescope/builtin/git.lua`.

l-kershaw:
The picture looks good to me, thanks 🙂

One thing I would change is the `ordinal` (the thing the prompt is compared against when searching) should be the `commit_info` as that is longer and contains more unique info.

farbodsz:
Good point about `ordinal`, @l-kershaw .

Both that, and the show branch option are done now. Anything else that needs to be done before merging?

lua/telescope/builtin/git.lua
```diff
@@ -72,7 +72,8 @@ git.commits = function(opts)
 end
 
 git.stash = function(opts)
-  opts.entry_maker = vim.F.if_nil(opts.entry_maker, make_entry.gen_from_git_stash())
+  opts.show_branch = vim.F.if_nil(opts.show_branch, true)
```
Conni2461, line 75: we need to document this option here: https://github.com/nvim-telescope/telescope.nvim/blob/27294d73e4562cbc83c9c488b4618f012deade7e/lua/telescope/builtin/init.lua#L204-L207
with `@field`

Conni2461:
Thanks :)

I executed docgen because ci is docgen sometimes/always broken

PR #1453, opened 2021-11-14
breaking: remove file browser

fdschmidt93:
The `builtin.file_browser` will be spun off into its own extension: https://github.com/nvim-telescope/telescope-file-browser.nvim (not yet accessible to public) The extension will incorporate a lot of demanded features and essentially constitutes #1257.

Hence, it makes sense to cut out the picker from `telescope-core`. Note that the PR should not yet be merged but is a prep for the initial release of the extension, which will happen throughout the upcoming week.

(Cutting out is also necessary, such that `:Telescope file_browser` adequately forwards to the extension.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,6 @@ Built-in functions. Ready to be bound to any key you like.
 | `builtin.git_files`                 | Fuzzy search through the output of `git ls-files` command, respects .gitignore, optionally ignores untracked files                |
 | `builtin.grep_string`               | Searches for the string under your cursor in your current working directory                                                       |
 | `builtin.live_grep`                 | Search for a string in your current working directory and get results live as you type (respecting .gitignore)                    |
-| `builtin.file_browser`              | Lists files and folders in your current working directory, open files, navigate your filesystem, and create new files and folders |
 
 ### Vim Pickers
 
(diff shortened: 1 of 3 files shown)


tjdevries:
Sweet, I was just checking to make sure this was getting deleted. The only thing I was wondering was if there was a way that we could let people know that it's in an extension now when they try and do `:Telescope file_browser` or something and it (in their minds) just randomly doesn't work anymore.

Thanks for doing this :) new extension looks good.

fdschmidt93:
I'd say let's have 
* a sticky issue for breaking changes and say in 2 or 4 weeks the picker will be removed
* announce breaking change via packer (I think `break:` in commit message makes it light up)
* one week ahead of removal add deprecation notice when the picker is launched

Thoughts?

tjdevries:
:+1:

doc/telescope.txt
```diff
@@ -2367,7 +2339,7 @@ work and make it really easy create new previewers.
 
 Furthermore, there are a collection of previewers already defined which can be
 used for every picker, as long as the entries of the picker provide the
-necessary fields. The more important ones are
+necessary fields. The more important once are
```
ineol, line 2342: I don't think this was a typo

fdschmidt93: This is not a change I've made, but I guess must have got intertwined with the state of `master` this branch is based off of.

Thanks for pointing it out! I'll merge sure this is untouched when I rebase prior to merging this PR (hopefully soon :)

fdschmidt93:
Apologies this took quite a bit longer than I originally planned.

PR #1383, opened 2021-10-26
feat: layout actions

l-kershaw:
Add action to cycle a picker through predetermined layouts.

Currently a little rough around the edges, but a decent proof of concept.

**Demo**
cycle_layout

PR #1459, opened 2021-11-15
feat: layout actions (including `toggle_preview`)

l-kershaw:
This is a combination of #1305 and #1383.
Big thanks to @johnybx for the implementation of the preview toggle.

Demo:
`toggle_preview`
toggle_preview

`toggle_prompt_position`
toggle_prompt_position2

`toggle_mirror`
toggle_mirror2

`cycle_layout_next`/`cycle_layout_prev`
cycle_layout

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -143,6 +143,20 @@ telescope.setup({opts})                                    *telescope.setup()*
         }
 
 
+                                     *telescope.defaults.cycle_layout_list*
+    cycle_layout_list: ~
+        Determines the layouts to cycle through when using `actions.cycle_layout_next`
+        and `actions.cycle_layout_prev`.
+        Should be a list of "layout setups".
+        Each "layout setup" can take one of two forms:
+        1. string <br>
+            This is interpreted as the name of a `layout_strategy`
+        2. table <br>
+            A table with possible keys `layout_strategy`, `layout_config` and `previewer`
+
+        Default: TODO
+
+
                                               *telescope.defaults.winblend*
     winblend: ~
         Configure winblend for telescope floating windows. See |winblend| for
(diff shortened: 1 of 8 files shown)


sunjon:
Great work. Looking forward to this.

In the first screenshot, the flash as the float window is removed/redrawn is quite noticeable. I'd suggest drawing a temporary blank window behind it (lower z-indez) before the transition, and removing it when done. What do you think?

l-kershaw:
I can't think of a robust way of doing this that isn't going to result in some level of "flash" in certain circumstances.
Should we account for text in popup menus disappearing on popup close? Do we match the colour of the popup or the background with the blank window? What if the popup that starts there is a different colour to what ends there?

That being said, I think the severity of the "flash" in the gif is larger than I have noticed when using this feature personally (probably a result of the framerate I had my screen recorder set to). However, obviously it would be better if we could remove the "flash" completely.

---

One other way around this is delaying part of the rearrangement until after another bit is complete. For example:
- when preview is being disabled, wait for the results and prompt to finish moving, before removing the preview.
- when preview is being enabled, wait for the preview to be created before moving the results and prompt.
- when preview is just moving, don't do any waiting.

The problem is, I'm not sure how to correctly wait for these events within the lua api. All of the lua functions seem slightly out of sync when I try it (e.g. `nvim_win_is_valid` returns true before the window is actually visible, `nvim_win_get_height` returns the new height before it has been visually updated) Any ideas?

---

Another way around this is temporarily pausing the drawing of the whole application, but I have no idea how to do that or if it is even possible.

l-kershaw:
I have come up with a workaround for the "flash" that just defers some parts of the rearrangement to the next step in the event loop. The code looks a little hacky, but it is a lot simpler than anything else I could think of.

This fixes the "flash" for me, @sunjon could you check that it also fixes it for you?

l-kershaw:
Updated demo for `toggle_preview`:
toggle_preview2

If anyone else can confirm that they also don't get the "flash" when toggling preview with this updated PR, then I'm happy for this to be merged.
@fdschmidt93 or @Conni2461, could one of you test if you have time?

Conni2461:
It looks good. I ran it today. Thanks for the good work :)

We should squash this PR so johnybx PR is a real commit on master. So can you squash the last 4 commits locally? so we end up with:
- feat: add hide / unhide preview (#1305)
- feat: layout actions (#1383)
- clean: ...

After that you can merge with "rebase and merge". If you need help with the git gymnastics. Feel free to ask. btw i should probably mention that i plan to maintain telescope saturday and sunday exclusively :grimacing:

l-kershaw:
Thanks for the git guidance, Conni 😊

That's fine, no pressure at all to put more time in than you want to 🙂

tami5:
@l-kershaw I just got around to test and integrate toggle preview, but I got the following error, not sure if something todo with how I mapped it but

```lua
E5108: Error executing lua ...ck/packer/start/telescope.nvim/lua/telescope/pickers.lua:604: attempt to call field 'move' (a nil value)
stack traceback:
        ...ck/packer/start/telescope.nvim/lua/telescope/pickers.lua:604: in function 'recalculate_layout'
        ...ck/packer/start/telescope.nvim/lua/telescope/pickers.lua:647: in function 'full_layout_update'
        ...er/start/telescope.nvim/lua/telescope/actions/layout.lua:37: in function 'run_replace_or_original'
        ...packer/start/telescope.nvim/lua/telescope/actions/mt.lua:30: in function 'key_func'
        ...k/packer/start/telescope.nvim/lua/telescope/mappings.lua:236: in function 'execute_keymap'
```

my config `  n["<C-]>"] = require("telescope.actions.layout").toggle_preview`

l-kershaw:
@tami5 are you using a recent version of `plenary`?
The `popup.move` function was implemented last year, so should be available for you.

tami5:
:| indeed this odd let me check

PR #1465, opened 2021-11-18
fix: use `string.format` on `builtin.resume` error message

l-kershaw:
Little fix for error message on `builtin.resume` when trying to resume a picker past the cache size


fdschmidt93:
:laughing: 

Thanks :)

PR #1469, opened 2021-11-19
fix: enable unmapping default attach_mappings

fdschmidt93:
I think this fixes https://github.com/nvim-telescope/telescope-file-browser.nvim/issues/5#issuecomment-973917663

Pasting example:

```lua
require "telescope".setup {
  pickers = {
    git_branches = {
      mappings = {
        i = {
          ["<C-a>"] = false, -- should unmap create branch, does not
          ["<C-a>"] = function() print("test") end, -- successfully overrides
        },
      },
    },
  },
}
```

This could probably be more elegantly implemented, but between default mappings, picker mappings & attach mappings, attach mappings on picker call it's quite convoluted.

`map(mode, keybind, false)` will always be passed somewhere by the user and therefore should definitively disallow any mapping to that key.


tjdevries:
I think we can close in favor of: #1473

PR #1471, opened 2021-11-19
fix: error when opts is nil

disrupted:
Fixes a small error I was getting after updating today.

```
packer.nvim: Error running config for telescope.nvim: ...pack/packer/opt/telescope.nvim/lua/telescope/pickers.lua:1244: attempt to index local 'opts' (a nil value)
```


fdschmidt93 (commented):
I think this is a bit more cleanly and future-proof fixed if we 

```lua
opts = opts or {}
defaults = defaults or {}
```

right below `pickers.new`.

Thoughts?

disrupted:
@fdschmidt93 I agree, this is much nicer and solid for the future.

fdschmidt93:
LGTM thanks :)

PR #1492, opened 2021-11-25
fix: check if buf is loaded before edit_buffer

ddcien:


Conni2461:
why ? Current version works already fine for unloaded buffers. If the buffer is unloaded, it will be loaded and opened. Just checked.

So what are you trying to fix?

ddcien:
When using `Telescope quickfix ` or `Telescope loclist`, the unloaded buf will not be listed.  This will fix that.

Conni2461:
Hmm i cant reproduce. Also i dont see how this patch would fix it. Can you give me some real reproduce steps?

ddcien:
## Prepare

```sh
touch a.c b.c
nvim a.c
```
## 
```vim
:call setqflist([{"filename": "b.c"}])
:Telescope quickfix
```
Select `b.c`, then use `:ls!` command to show all buffers. The current buffer `b.c` is found NOT listed!

Conni2461:
Okay thanks for the reproduce. I still dont like the solution because i am not sure what it will break.

I think a more sane solution could be something like this. What do you think
```diff
diff --git a/lua/telescope/actions/set.lua b/lua/telescope/actions/set.lua
index b7170db..81f9bc5 100644
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -125,6 +125,9 @@ action_set.edit = function(prompt_bufnr, command)
   require("telescope.actions").close(prompt_bufnr)
 
   if entry_bufnr then
+    if not vim.api.nvim_buf_get_option(entry_bufnr, "buflisted") then
+      vim.api.nvim_buf_set_option(entry_bufnr, "buflisted", true)
+    end
     edit_buffer(command, entry_bufnr)
   else
     -- check if we didn't pick a different buffer
```

ddcien:
> Okay thanks for the reproduce. I still dont like the solution because i am not sure what it will break.
> 
> I think a more sane solution could be something like this. What do you think
> 
> ```diff
> diff --git a/lua/telescope/actions/set.lua b/lua/telescope/actions/set.lua
> index b7170db..81f9bc5 100644
> --- a/lua/telescope/actions/set.lua
> +++ b/lua/telescope/actions/set.lua
> @@ -125,6 +125,9 @@ action_set.edit = function(prompt_bufnr, command)
>    require("telescope.actions").close(prompt_bufnr)
>  
>    if entry_bufnr then
> +    if not vim.api.nvim_buf_get_option(entry_bufnr, "buflisted") then
> +      vim.api.nvim_buf_set_option(entry_bufnr, "buflisted", true)
> +    end
>      edit_buffer(command, entry_bufnr)
>    else
>      -- check if we didn't pick a different buffer
> ```

Yes, your solution is even better. And I think of that every entry should have its own optional selection method, for example, `cc {index} ` for qflist, `ll {index}`  for loclist. Please think about that.

Conni2461:
feel free to apply these changes to this PR, then i can merge it. You deserve to get credit, you found the issue :)

We already have a way to substitute, the default selection with `action_set.select:replace`, we could define this for quickfix and loclist. I just dont know if its worth the effort.

Conni2461 (approved):
Thanks :)

PR #1494, opened 2021-11-25
fix: check for Windows drive letter when parsing vimgrep

### LINKED ISSUE #1396: [Bug]: builtin `live_grep` don't work properly in windows
### Description

When calling telescope with following lua script, there have two problem only happend in windows.
```lua
  cwd = vim.fn.stdpath "data",
  search_dirs = { vim.fn.stdpath "data" }
}
```
1. the `cwd` current working directory doesn't begin with what I set it to, as showing in the following first picture.
2. when press `enter` on any item of the result, it opens a blank buffer with wrong path, as showing in the following second picture where I tryed to save the opened buffer.

### Neovim version
```
NVIM v0.5.1
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe /DWIN32 /D_WINDOWS /W3 -DNVIM_TS_HAS_SET_MATCH_LIMIT /MD /Zi /O2 /Ob1 /DNDEBUG /W3 -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_DEPRECATE -DWIN32 -D_WIN32_WINNT=0x0600 -DINCLUDE_GENERATED_DECLARATIONS -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -ID:/a/neovim/neovim/build/config -ID:/a/neovim/neovim/src -ID:/a/neovim/neovim/nvim-deps/usr/include -ID:/a/neovim/neovim/build/src/nvim/auto -ID:/a/neovim/neovim/build/include
Compiled by runneradmin@fv-az152-633
```

Windows NT 19043

### checkhealth telescope

```markdown
health#telescope#check
  - OK: plenary installed.

  - OK: fd: found fd 8.2.1

## Telescope Extension: `projects`
```

### Steps to reproduce

Calling `live_grep` with following opts

```lua
  cwd = vim.fn.stdpath "data",
  search_dirs = { vim.fn.stdpath "data" }
}
```

### Expected behavior

1. Each items of result should begin with `.\` which is on befalf of `cwd` I set when calling `live_grep`
2. Open the right path of each item when press `enter`

### Actual behavior

图片
图片

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

l-kershaw:
Check if Windows drive letter (e.g. "C:") is present when parsing vimgrep

Fixes #1396

@younger-1 does this work for you?

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -115,7 +115,7 @@ do
     local start, _, filename, lnum, col, text = string.find(t.value, [[([^:]+):(%d+):(%d+):(.*)]])
 
     -- Handle Windows drive letter (e.g. "C:") at the beginning (if present)
-    if Path.path_sep == "\\" and start == 3 then
+    if Path.path.sep == "\\" and start == 3 then
       filename = string.sub(t.value, 1, 3) .. filename
     end
 
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -112,7 +112,12 @@ do
 
   -- Gets called only once to parse everything out for the vimgrep, after that looks up directly.
   local parse = function(t)
-    local _, _, filename, lnum, col, text = string.find(t.value, [[([^:]+):(%d+):(%d+):(.*)]])
+    local start, _, filename, lnum, col, text = string.find(t.value, [[([^:]+):(%d+):(%d+):(.*)]])
+
+    -- Handle Windows drive letter (e.g. "C:") at the beginning (if present)
+    if Path.path_sep == "\\" and start == 3 then
+      filename = string.sub(t.value, 1, 3) .. filename
+    end
 
     local ok
     ok, lnum = pcall(tonumber, lnum)
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -110,14 +110,29 @@ do
     ordinal = 1,
   }
 
+  local find = (function()
+    if Path.path.sep == "\\" then
+      return function(t)
+        local start, _, filename, lnum, col, text = string.find(t.value, [[([^:]+):(%d+):(%d+):(.*)]])
+
+        -- Handle Windows drive letter (e.g. "C:") at the beginning (if present)
+        if start == 3 then
+          filename = string.sub(t.value, 1, 3) .. filename
+        end
+
+        return filename, lnum, col, text
+      end
+    else
+      return function(t)
+        local _, _, filename, lnum, col, text = string.find(t.value, [[([^:]+):(%d+):(%d+):(.*)]])
+        return filename, lnum, col, text
+      end
+    end
+  end)()
+
   -- Gets called only once to parse everything out for the vimgrep, after that looks up directly.
   local parse = function(t)
-    local start, _, filename, lnum, col, text = string.find(t.value, [[([^:]+):(%d+):(%d+):(.*)]])
-
-    -- Handle Windows drive letter (e.g. "C:") at the beginning (if present)
-    if Path.path.sep == "\\" and start == 3 then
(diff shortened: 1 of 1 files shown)


lua/telescope/make_entry.lua
```diff
@@ -112,7 +112,12 @@ do
 
   -- Gets called only once to parse everything out for the vimgrep, after that looks up directly.
   local parse = function(t)
-    local _, _, filename, lnum, col, text = string.find(t.value, [[([^:]+):(%d+):(%d+):(.*)]])
+    local start, _, filename, lnum, col, text = string.find(t.value, [[([^:]+):(%d+):(%d+):(.*)]])
+
+    -- Handle Windows drive letter (e.g. "C:") at the beginning (if present)
+    if Path.path_sep == "\\" and start == 3 then
```
Conni2461, line 118: shouldnt this be `Path.path.sep` ?

Also maybe we can refactor this that we dont have to do the sep comparison for every call but rather only once.

l-kershaw: I've updated to `Path.path.sep` (sorry, typo 😅)

I'm not sure how to factor this check out apart from something like:

```lua
local parse = (function()
    if Path.path.sep == "\\" then
        return [[version of function with the extra check in]]
    else
        return [[nearly identical function without extra check]]
    end
end)()
```
which seems like its asking for trouble in terms of duplicating code and versions getting out of sync.

If you can think of another way of refactoring, then happy to do that. Or if you think its worth it to do the "two nearly identical functions" method, then I'm happy to follow your lead on that.

What do you think is best?

younger-1:
@l-kershaw It works for me. Thank you!

Conni2461:
@l-kershaw What about this diff. It obv has a little bit of duplication but i dont think its that bad
```diff
diff --git a/lua/telescope/make_entry.lua b/lua/telescope/make_entry.lua
index 7892e05..ab52741 100644
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -110,9 +110,26 @@ do
     ordinal = 1,
   }
 
+  local find = (function()
+    if Path.path.sep == "\\" then
+      return function(t)
+        local start, _, filename, lnum, col, text = string.find(t.value, [[([^:]+):(%d+):(%d+):(.*)]])
+        if start == 3 then
+          filename = string.sub(t.value, 1, 3) .. filename
+        end
+        return filename, lnum, col, text
+      end
+    else
+      return function(t)
+        local _, _, filename, lnum, col, text = string.find(t.value, [[([^:]+):(%d+):(%d+):(.*)]])
+        return filename, lnum, col, text
+      end
+    end
+  end)()
+
   -- Gets called only once to parse everything out for the vimgrep, after that looks up directly.
   local parse = function(t)
-    local _, _, filename, lnum, col, text = string.find(t.value, [[([^:]+):(%d+):(%d+):(.*)]])
+    local filename, lnum, col, text = find(t)
 
     local ok
     ok, lnum = pcall(tonumber, lnum)
```

l-kershaw:
@Conni2461 thanks, that seems like a good compromise in terms of factoring things out

If you're happy with this now, it can be merged 🙂

Conni2461 (approved):
I am happy for each contribution :) especially yours :)

Thanks for fixing our windows bugs :laughing:

PR #1505, opened 2021-11-28
add preview width for bottom pane

max397574:


lua/telescope/pickers/layout_strategies.lua
```diff
@@ -806,13 +807,22 @@ layout_strategies.bottom_pane = make_documented_layout(
     results.height = height - prompt.height - (2 * bs)
     preview.height = results.height - bs
 
+    local width_opt = layout_config.width
+    local width = resolve.resolve_width(width_opt)(self, max_columns, max_lines)
+
```
l-kershaw, line 812: Since `bottom_pane` always takes up the full width of the screen, we shouldn't have an option for it

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -806,13 +807,22 @@ layout_strategies.bottom_pane = make_documented_layout(
...
     -- Width
     prompt.width = max_columns - 2 * bs
-    -- TODO(l-kershaw): add a preview_cutoff option
-    if self.previewer then
-      -- TODO(l-kershaw): make configurable
-      results.width = math.floor(max_columns / 2) - 2 * bs
-      preview.width = max_columns - results.width - 4 * bs
+    local w_space
```
l-kershaw, line 815: We will need to add `width` as a local variable here

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -806,13 +807,22 @@ layout_strategies.bottom_pane = make_documented_layout(
...
+    if self.previewer and max_columns >= layout_config.preview_cutoff then
+      -- Cap over/undersized width (with preview)
+      width, w_space = calc_size_and_spacing(width, max_columns, bs, 2, 4, 0)
```
l-kershaw, line 818: This should be `calc_size_and_spacing(max_columns, max_columns, bs, 2, 4, 0)`

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -806,13 +807,22 @@ layout_strategies.bottom_pane = make_documented_layout(
...
+
+      preview.width = resolve.resolve_width(if_nil(layout_config.preview_width, function(_, _)
+        -- By default, previewer takes 1/2 of the layout
+        return 2 * math.floor(width / 2)
```
l-kershaw, line 822: This should be `return math.floor(width / 2)`

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -806,13 +807,22 @@ layout_strategies.bottom_pane = make_documented_layout(
...
+      end))(self, width, max_lines)
+      prompt.width = width - preview.width - w_space
+      results.width = prompt.width
```
l-kershaw, line 825: We shouldn't change the `prompt.width` here (it is always full width), but this formula is correct for `results.width`

l-kershaw (commented):
Thanks for the PR, only a couple of small changes needed.

---

Also, you have implemented the `preview_cutoff` option, but not set a default value for it. You should add a default [here](https://github.com/nvim-telescope/telescope.nvim/blob/ed4adba6d0083a14cabf5837f2511b2617a45593/lua/telescope/config.lua#L108-L111)

lua/telescope/pickers/layout_strategies.lua
max397574: just `local width`or wym?
width is added like 4 lines above

lua/telescope/pickers/layout_strategies.lua
max397574: local width = max_columns?

lua/telescope/pickers/layout_strategies.lua
max397574: oh that's why it looked so ugly

max397574:
I think I fixed everything @l-kershaw

lua/telescope/pickers/layout_strategies.lua
l-kershaw: I don't think there is any need to set `local width = max_columns`, since we already have a variable with that value, and we don't use the value of `width` again until after it is recalculated.
It's better to just introduce the local variable (probably at the same time as `w_space`).

lua/telescope/config.lua
```diff
@@ -108,6 +108,7 @@ local layout_config_defaults = {
   bottom_pane = {
     height = 25,
     prompt_position = "top",
+    preview_cutoff = 40,
```
l-kershaw, line 111: I think that 120 is a better default value here, as it matches the behaviour of the most similar other layout, `horizontal`.

max397574:
@l-kershaw

l-kershaw:
@max397574 I have responded to your question on my first review. Could you make the changes discussed there please?

lua/telescope/pickers/layout_strategies.lua
max397574: afaict I can also remove width_opt then (ls said it's unused then)

l-kershaw:
LGTM 👍
Thanks @max397574 🙂

PR #1513, opened 2021-11-29
fix(highlight): Use TS built-in is_enabled function

patrick96:
The function just replicated the logic from is_enabled and assumed the
`disable` setting is always a table. This is no longer the case (https://github.com/nvim-treesitter/nvim-treesitter/pull/2009), it can now also be a function.

Do I understand this correctly that the only reason telescope does not use `ts_configs.attach_module` and writes its own `treesitter_attach` is that `attach_module` does not return a boolean indicating success?


patrick96:
Ah, I remember seeing that, I thought they would be equivalent, but it makes sense that we don't want to set the filetype.

Conni2461:
Thanks :)

Issue #1507, opened 2021-11-28
Add highlight group for Results window


siduck:
would it be possible to add a highlight group for the results window , something like TelescopeResultstNormal?

max397574:
![telescope_highlights_wanted](https://user-images.githubusercontent.com/81827001/143845191-4d7a3d57-4691-4149-aada-86fb2f0f6bbd.jpg)

basically we want to be able to do this

this is done with photoshop

max397574:
this could be done by creating TelescopePromptNormal and TelescopeResultsNormal or just one of them and for the other normal is used (I'd prefer the first).

Another problem is that TelescopeNormal isn't used when you aren't in insert mode
![Screenshot 2021-11-29 at 10 47 47](https://user-images.githubusercontent.com/81827001/143846172-ea2b6f7e-bfe7-43af-ac3d-160e1b4a82d4.png)
![Screenshot 2021-11-29 at 10 47 54](https://user-images.githubusercontent.com/81827001/143846176-a0095287-b512-47e0-a5ed-84abdd99d6ea.png)

l-kershaw:
Happy to review a PR to add this feature if someone is interested in writing one.

I _think_ that this should only require changes in `plugin/telescope.vim` (adding links to predefined highlights) and `lua/telescope/pickers.lua` (changing the settings for the relevant popups).

max397574:
I'll try

max397574:
where would that be
> lua/telescope/pickers.lua (changing the settings for the relevant popups).

this file is 1.4k lines and I have no clue where to search

l-kershaw:
You need to search for `TelescopeNormal` which is the current highlight name

max397574:
imo it makes sense to link them to `TelescopeNormal` what do you think?

max397574:
will be closed with https://github.com/nvim-telescope/telescope.nvim/pull/1514

PR #1514, opened 2021-11-29
add Normal highlights for prompt and results

max397574:
@l-kershaw

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -320,11 +320,11 @@ function Picker:find()
 
   -- `popup.nvim` massaging so people don't have to remember minheight shenanigans
   popup_opts.results.minheight = popup_opts.results.height
-  popup_opts.results.highlight = "TelescopeNormal"
+  popup_opts.results.highlight = "TelescopeResultsNormal"
   popup_opts.results.borderhighlight = "TelescopeResultsBorder"
   popup_opts.results.titlehighlight = "TelescopeResultsTitle"
   popup_opts.prompt.minheight = popup_opts.prompt.height
-  popup_opts.prompt.highlight = "TelescopeNormal"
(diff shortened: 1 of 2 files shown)


max397574:
Screenshot 2021-11-29 at 14 30 49


this pr allows custom highlight groups for prompt and results

l-kershaw:
This works well for me; the only thing I would change right now is remove the comment in this line:
https://github.com/nvim-telescope/telescope.nvim/pull/1514/files#diff-afd2d88a2ed63b01b66015f5836a104ce81a8f6daac1b6f70692e1f4534b5f0bL19
as it (and the blank line above it) is no longer needed.

---

I also want to double check with another team member why `TelescopePreviewNormal` is linked to `Normal` rather than `TelescopeNormal`, and whether we should change it so everything matches. @Conni2461 you merged this in #460, was there a particular reason for this? Should we change it so the behaviour is consistent for all window types?

Conni2461:
No reason. It just happened because i didnt pay enough attention.

max397574:
idk if we could also try to fix this here @l-kershaw 
https://github.com/nvim-telescope/telescope.nvim/issues/1507#issuecomment-981467737

l-kershaw:
@Conni2461 thanks for the quick response 😊
I know you're trying to keep telescope work to the weekend, so I do appreciate it.
I don't think it was an attention thing, it just didn't make any difference at the time, since "preview" was the only part that was separated out 🙂

---

@max397574 I can't reproduce the issue from the comment, either on master, or with this PR checked out.
Therefore I think it would be best to open its own issue, and we can handle it separately.

For this PR, I'm happy to merge now, but will wait until you let me know that you are happy with this plan of handling things separately 🙂

max397574:
yes I think it's good as it is now @l-kershaw

PR #1480, opened 2021-11-22
fix: cleanup actions layout

Conni2461:
fix #1474
fix #1482

CC @l-kershaw @johnybx 

Why was the closing stuff changed?

Also found this funny bug

https://user-images.githubusercontent.com/15233006/142928898-04c41afa-583d-4668-ba5d-d0f7867fa64a.mp4

This happens on current master and this branch. I leave the fix to one of you guys :)

PR #1522, opened 2021-11-30
fix: keep previewer `winid` state updated on layout changes

l-kershaw:
Fixes issue pointed out by Conni [here](https://github.com/nvim-telescope/telescope.nvim/pull/1480#issue-1060554567)

Before:
preview_winid_pre

After:
preview_winid_post


Conni2461 (approved):
Big Thanks :)

Issue #1518, opened 2021-11-30
Indentation is preserved when searching


zoomido:
### Description

The indentation is preserved when searching which creates unnecessary long lines. It would be nice if this could be removed or maybe toggled somehow.
![bild](https://user-images.githubusercontent.com/35059247/144037294-49f633fc-7c0c-40a7-98b2-846cff18c5f0.png)


### Neovim version

```markdown
NVIM v0.5.1
Build type: Release
LuaJIT 2.1.0-beta3
```


### Operating system and version

macos 11.6.1

### checkhealth telescope

```markdown
health#telescope#check                                                                                                                                                                                  
========================================================================                                                                                                                                
## Checking for required plugins                                                                                                                                                                        
  - OK: plenary installed.                                                                                                                                                                              
  - OK: nvim-treesitter installed.                                                                                                                                                                      
                                                                                                                                                                                                        
## Checking external dependencies                                                                                                                                                                       
  - OK: rg: found ripgrep 13.0.0                                                                                                                                                                        
  - OK: fd: found fd 8.2.1                                                                                                                                                                              
                                                                                                                                                                                                        
## ===== Installed extensions =====                                                                                                                                                                     
                                                                                                                                                                                                        
## Telescope Extension: `fzf`                                                                                                                                                                           
  - INFO: No healthcheck provided                                                                                                                                                                       
                                                                                                                                                                                                        
## Telescope Extension: `neoclip`                                                                                                                                                                       
  - INFO: No healthcheck provided
```


### Steps to reproduce

Search for a line that has indentation

### Expected behavior

_No response_

### Actual behavior

See description screenshot

### Minimal config

```Lua
Any config
```

fdschmidt93:
Do I understand it correctly, that indentation is preserved from the original file..?

If so, two solutions come to mind:

a) maybe `rg` can preprocess the lines somehow
b) write your own [entry maker](https://github.com/nvim-telescope/telescope.nvim/blob/c1951b37229f6628a07e1290430eeb83e532dc90/lua/telescope/make_entry.lua#L139-L230) with more info in [developers.md](https://github.com/nvim-telescope/telescope.nvim/blob/master/developers.md). What you'd have to modify is the `entry.display` (for the minimum solution)

And furthermore, assuming I understand it correctly, I'd close this as `wontfix` as I cannot imagine any member steps up to do this.

zoomido:
Yes correct. Indentation is preserved from the original file. 
[...]

PR #1523, opened 2021-11-30
chore: add trim to rg

Conni2461:
fix https://github.com/nvim-telescope/telescope.nvim/issues/1518

dont know if we want this. It looks unfamiliar

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -432,7 +432,8 @@ telescope.setup({opts})                                    *telescope.setup()*
           "--with-filename",
           "--line-number",
           "--column",
-          "--smart-case"
+          "--smart-case",
+          "--trim"
         }
 
                                               *telescope.defaults.use_less*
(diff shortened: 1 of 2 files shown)


l-kershaw:
I don't have any particularly strong opinions on whether or not this default gives better behaviour on not, but whichever way we decide to go, we should add a config recipe to the wiki saying how to have it the other way.

Maybe its better to leave defaults as is and simply add the config recipe, as it seems more reasonable for a user to add an option than remove one? Not really sure, happy either way 🙂

Conni2461:
Yeah config recipe it is

Conni2461:
In the we had made it an option. I dont like this approach anymore. (e.g. we have a million options for fd).

But we should still ping the telescope-rg guy. CC @weeman1337

For completeness: https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes#ripgrep-remove-indentation

PR #1532, opened 2021-12-02
chore: add `--strip-cwd-prefix` to `fd`

numToStr:
I don't know when exactly this prefix (`./`) started to appear (maybe after `fd v0.8.3`) but I think this only adds noise to the results. So this PR adds `--strip-cwd-prefix` flag to remove the prefix.

Before:
image

After:
image

---

PS: I might need guidance on updating the test :)

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -166,7 +166,7 @@ files.find_files = function(opts)
 
   if not find_command then
     if 1 == vim.fn.executable "fd" then
-      find_command = { "fd", "--type", "f" }
+      find_command = { "fd", "--type", "f", "--strip-cwd-prefix" }
       if hidden then
         table.insert(find_command, "--hidden")
       end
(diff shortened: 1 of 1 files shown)


Conni2461:
This would break some `fd`s for example i just did `nix-shell -p fd` and the stable version doesnt have a `--strip-cwd-prefix`.

So yeah

numToStr:
@Conni2461 Ahh, You are right. And indeed this behavior was introduced in `v0.8.3` https://github.com/sharkdp/fd/pull/861 and previous versions won't be affected. We can close this If you want.

gadkadosh:
I just spent an hour trying to figure out why this `./` started showing up :)
Along the way I realized `rg` doesn't return the prefix while `find` does. So I guess there's already some variance between `find_command`s.

Conni2461:
Okay. I am closing this because we cant really require a minimum fd version. For the future: `rg --files` doesnt add a `./` so either use that or define your own `find_command` like this:

```lua
telescope.setup {
  defaults = {
    -- ....
  },
  pickers = {
    find_files = {
      find_command = { "fd", "--type", "f", "--strip-cwd-prefix" }
    },
  }
}
```

I will put this in the wiki. Still thanks for the PR and sharing this option @numToStr :)

PR #1535, opened 2021-12-02
fix: lsp_workspace_symbols

Conni2461:


clason:
Yep, looks good, thank you!

PR #1537, opened 2021-12-02
wip: virtual result lines

l-kershaw:
WIP
commit_virt_lines


l-kershaw:
I've had a bit more of a play around with this and I think this should be put on hold until there is a scrolling API (as also discussed [here](https://github.com/neovim/neovim/issues/16166#issuecomment-954661371)).
This is because otherwise I can't see a good way of ensuring that all of the virtual lines for an entry are visible when that entry is selected.

Conni2461:
Looks cool but yeah agree. For now i would schedule this for after the release

PR #1481, opened 2021-11-23
docs: fix layout_config width description

jamestrew:
Doc for `layout_config` currently:
```
...
For example, the default values below set the default width to 80% of
the screen width for all strategies except 'center', which has width
of 50% of the screen width.

Default: {
  bottom_pane = {
    height = 25,
    prompt_position = "top"
  },
  center = {
    height = 0.9,
    preview_cutoff = 40,
    prompt_position = "top",
    width = 0.8
  },
  cursor = {
    height = 0.9,
    preview_cutoff = 40,
    width = 0.8
  },
  horizontal = {
    height = 0.9,
    preview_cutoff = 120,
    prompt_position = "bottom",
    width = 0.8
  },
  vertical = {
    height = 0.9,
    preview_cutoff = 40,
    prompt_position = "bottom",
    width = 0.8
  }
}
```
`width` for 'center' is set as 80% just like the others unless I'm wildly confused.

PR #1546, opened 2021-12-03
defaults: change `center` layout default size

l-kershaw:
Now matches the description and has space for a preview above the central block

Fixes #1481


Conni2461 (approved):
Thanks for confirming :)

PR #1549, opened 2021-12-03
break: bump the minimum required neovim version to 0.6 or higher

Conni2461:


doc/telescope_changelog.txt
```diff
@@ -131,4 +131,12 @@ requires users to be on Neovim 0.5.1 (the most recent stable version) or on
 the LATEST version of Neovim nightly.
 
 
+                                                    *telescope.changelog-1549*
+
+Date: December 10, 2021
+PR: https://github.com/nvim-telescope/telescope.nvim/pull/1549
+
+Telescope requires now Neovim release 0.6.0 or a more recent nightly.
```
l-kershaw, line 139: As with the last changelog entry about this (see #1444), it is best to emphasise that we will only be supporting the latest stable and latest nightly. That way when new things are added to nightly we can safely use them behind a check for `vim.fn.has "nvim-0.7.0"`.

Conni2461: yeah i know. I'll make it more explicit again. But I already put that in the readme some time ago and i'll just close every issue that uses an unsupported neovim version.

In my opinion we dont need to include this in every changelog. People dont read that carefully anyways :laughing:

Conni2461: @l-kershaw I am not sure how often you check matrix | gitter. I posted that i am working on a first release and invited all core maintainer to a private chat (that does not exist yet) or a conference call, to discuss the future of telescope.

wookayin:
For neovim < 0.6.0, one can pin the plugin version to `nvim-0.5.1` (tag).

PR #1552, opened 2021-12-04
chore: remove dot at the beginning

Conni2461:
We added this for rg 13.0 release because it was needed but it doesnt
seem to be needed anymore. Weird

This might be due to some changes in neovim or due to changes to jobs.

I'll build 0.5.1 and test it with that right now. If it doesnt work on 0.5.1 we move this PR back after the 0.6 bump.

Edit: also seems to work for 0.5.1 What am i missing.


Conni2461:
If it breaks we can revert, I cant reproduce the behavior that we had prior to adding this `.`

PR #1553, opened 2021-12-04
feat: vim.diagnostics

### LINKED ISSUE #1378: Icon and color not properly set for diagnostic Warn
### Description

Following changes of https://github.com/neovim/neovim/issues/14090#issuecomment-921312955

MR: https://github.com/nvim-telescope/telescope.nvim/pull/1359

### Neovim version

NVIM v0.6.0-dev+475-g389a89858
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

Ubuntu 18

### checkhealth telescope

```markdown
  - OK: plenary installed.

  - OK: rg: found ripgrep 11.0.2 (rev 3de31f7527)

```

### Steps to reproduce

1. `nvim -u minimal.lua test.lua`
2. Write `test` in the file and open diagnostics with `:lua require'telescope.builtin'.lsp_document_diagnostics({previewer = false})`.

### Expected behavior

Proper icon and color of diagnostic WARN.

### Actual behavior

image

### Minimal config

```Lua
local cmd = vim.cmd

vim.fn.sign_define("DiagnosticSignError", { text = "",  texthl = "DiagnosticError" })
vim.fn.sign_define("DiagnosticSignWarn",  { text = " ", texthl = "DiagnosticWarn"  })

vim.api.nvim_command("hi DiagnosticError guifg=yellow")
vim.api.nvim_command("hi DiagnosticWarn  guifg=green")

cmd [[packadd! packer.nvim]]
cmd [[packadd! nvim-lspconfig]]
cmd [[packadd! telescope.nvim]]
cmd [[packadd! plenary.nvim]]

local use = require('packer').use
require('packer').startup(function()
  use 'wbthomason/packer.nvim' -- Package manager
  use 'neovim/nvim-lspconfig' -- Collection of configurations for built-in LSP client
  use { 'nvim-telescope/telescope.nvim', opt = true, requires = { {'nvim-lua/popup.nvim', opt = true}, {'nvim-lua/plenary.nvim', opt = true} } }
end)

local system_name = "Linux"
local sumneko_root_path = vim.fn.stdpath('cache') .. "/lspconfig/sumneko_lua/lua-language-server"
local sumneko_binary = sumneko_root_path .. "/bin/" .. system_name .. "/lua-language-server"
local lspconfig = require 'lspconfig'

local capabilities = vim.lsp.protocol.make_client_capabilities()

lspconfig.sumneko_lua.setup {
	cmd = {sumneko_binary, "-E", sumneko_root_path .. "/main.lua"};
	capabilities = capabilities,
	settings = {
		Lua = {
			runtime = {
				-- Get the language server to recognize LuaJIT globals like `jit` and `bit`
				version = 'LuaJIT',
				-- Setup your lua path
				path = vim.split(package.path, ';'),
			},
			diagnostics = {
				-- Get the language server to recognize the `vim` global
				globals = {'vim', 'use'},
			},
			workspace = {
				-- Make the server aware of Neovim runtime files
				library = {
					[vim.fn.expand('$VIMRUNTIME/lua')] = true,
					[vim.fn.expand('$VIMRUNTIME/lua/vim/lsp')] = true,
				},
			},
		},
	},
}
```

### LINKED ISSUE #1581: Stop using lsp deprecated functions
### Description

There are some lsp functions that were deprecated. [this commit](https://github.com/neovim/neovim/pull/16536/files) added warnings to those which spams deprecation warnings everytime those are used.

### Neovim version

```markdown
NVIM v0.7.0-dev+698-g2fe60905f
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by bellini@behemoth

Features: +acl +iconv +tui
```

Debian Unstable

### checkhealth telescope

```markdown
  - OK: plenary installed.

  - OK: fd: found fd 8.3.0

## Telescope Extension: `fzf`
```

### Steps to reproduce

`:lua require('telescope.builtin').lsp_document_diagnostics()` will display the warning that even needs to be dismissed

(there are probably other functions. that one is an example where the problem happens)

### Expected behavior

No warning

### Actual behavior

vim.lsp.diagnostic.get is deprecated. See :h deprecated

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

Conni2461:
fix #1378
close #1581

After 0.6 is a requirement.

Still WIP, requires more cleanup and maybe we can do something with namespace. So if we have multiple namespaces it might be useful to actually display that, maybe. or optional


lua/telescope/builtin/init.lua
```diff
@@ -435,7 +435,22 @@ builtin.lsp_workspace_symbols = require_on_exported_call("telescope.builtin.lsp"
 ---@field symbol_highlights table: string -> string. Matches symbol with hl_group
 builtin.lsp_dynamic_workspace_symbols = require_on_exported_call("telescope.builtin.lsp").dynamic_workspace_symbols
 
---- Lists LSP diagnostics for the current buffer
+builtin.lsp_document_diagnostics = function(...)
+  error("Why you still using lsp_diagnostics. 0.6 is here")
+  require("telescope.builtin.diagnostics").document(...)
+end
+builtin.lsp_workspace_diagnostics = function(...)
+  error("Why you still using lsp_diagnostics. 0.6 is here")
```
Conni2461, line 443: Make a real deprecation message!

lua/telescope/builtin/diagnostics.lua
```diff
@@ -0,0 +1,142 @@
...
+  -- convert from string to int
+  if type(severity) == "string" then
+    -- make sure that e.g. error is uppercased to ERROR
+    return severities[severity:upper()]
+  end
+  -- otherwise keep original value, incl. nil
+  return severity
+end
+
+local filter_diag_severity = function(opts, severity)
```
fdschmidt93, line 18: `vim.diagnostic` now supports this sort of functionality OOTB via `:h vim.diagnostic.severity`, which we could (but ofc don't have to) follow.

Conni2461: I thought it can only do exact. That is great. For now i wouldnt change our interface just forward it into `get`

Conni2461:
the whole severity limit | bound thing seems to work 
image

Conni2461:
`Telescope diagnostic_document` or `Telescope diagnostics_document`. I think the first one is better because of `vim.diagnostic`.

@fdschmidt93 @clason

clason:
I think the module is called `diagnostic`, but you are still looking at "diagnostics", plural.

I'd say consistency across telescope is more important here than consistency with the underlying API.

Conni2461:
Yeah makes sense thanks :)

fdschmidt93:
Why not `document_diagnostics`? Seems like the most natural name to me. `diagnostic(s)_{document, workspace}` seems a bit wordy to me.

That said, `document` and `workspace` stem from LSP lingo I suppose, but `vim.diagnostic.get` is no more akin to (all) buffers and namespaces. But not sure whether that warrants changing the naming.

But I don't care that much.

Conni2461:
I think its also not natural to say. but i dont like `document_diagnostics` I think the module should be in front.

I also just noticed that its now just the buffers so we can also just have a `Telescope diagnostics` with a option for all, default is false. If true diagnostics from all open buffers ?

clason:
Yes, diagnostics are global (by default). For a `document_diagnostics`, the telescope picker would need to explicitly restrict to the current buffer. In this case, I'd just opt for `diagnostics`.

(Related: `get` takes a namespace; it would be neat if `Telescope diagnostics <namespace>` worked. Need not be this PR, of course.)

Conni2461:
I've put in a namespace option: https://github.com/nvim-telescope/telescope.nvim/pull/1553/files#diff-79c55de55b65fbaa2b214c1f3d28277a87a5b6ac9e5e0c87168a3341a471e5d4R1487-R1488 so Telescope diagnostics namespace=7 should work.

Conni2461:
maybe it would even be better with a bufnr option. Like bufnr = nil|0 means current buffer bufnr = 7 means from buffer 7 and bufnr == all means all ? Thats more flexible than all and almost as easy

clason:
Should have actually looked at the PR before opinionating :D

> maybe it would even be better with a bufnr option. Like bufnr = nil|0 means current buffer bufnr = 7 means from buffer 7 and bufnr == all means all ? Thats more flexible than all and almost as easy

It's probably best to stick with the diagnostic convention here: 
>  Use 0 for current buffer or nil for all buffers

Conni2461:
The thing is people will call `Telescope diagnostics` which will result in `bufnr` being nil. Depending on what we want we could flip this behavior

clason:
Yes, I'm saying this should be default (`Telescope diagnostics` gets _all_ diagnostics; add modifiers to filter it down to buffer or namespace).

If you want to be _really_ fancy, you can add `:<foo>:` parsing to filter ;)

Conni2461:
Okay changelog is here: https://github.com/nvim-telescope/telescope.nvim/pull/1553/files#diff-16e85f7f2f133a7dfecf66783549b68319150bcd7b077190d22a3adb16629819

I think `diagnostics` for all open buffers and `diagnostics bufnr=0` for current document works good. I dont think we need any shortcuts. For now `lsp_document_diagnostics` and `lsp_workspace_diagnostics` still exist and work as shortcuts as intended but they will log errors because i wanna remove them. Because like tj said here: https://github.com/nvim-telescope/telescope.nvim/pull/1250#issuecomment-922364829 it has nothing to do with lsp_* anymore :)

As always Thanks for your help :)

clason:
Makes sense! And you're right, not much need for convenience commands in Telescope itself since users can just create their own bespoke mappings.

fdschmidt93:
As a side note, I'm afraid README needs to be updated for this PR :sweat_smile:

Conni2461:
`rm -rf all pickers` from the readme :sob:

PR #1554, opened 2021-12-04
docs: functions in `pickers.lua`

l-kershaw:
This is almost entirely copied from #927, but that PR was so out of date it was easier to just make a new one than to deal with the merge conflicts 😅 

I decided it wasn't worth doing docs for `entry_manager.lua` and `linked_list.lua` at the moment, since there have been discussions about reimplementing things in that area.

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -161,8 +161,8 @@ end
 
 --- Take an index and get a row.
 ---@note: Rows are 0-indexed, and `index` is 1 indexed (table index)
----@param index number: The index in line_manager
----@return number: The row for the picker to display in
+---@param index number: the index in line_manager
+---@return number: the row for the picker to display in
 function Picker:get_row(index)
   if self.sorting_strategy == "ascending" then
     return index - 1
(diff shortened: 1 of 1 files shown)


lua/telescope/pickers.lua
```diff
@@ -191,12 +193,17 @@ function Picker:get_reset_row()
   end
 end
 
+--- Check if the picker is no longer in use
+---@return boolean/nil: `true` if picker is closed, `nil` otherwise
```
Conni2461, line 197: its `|` not `/`

lua/telescope/pickers.lua
```diff
@@ -1210,6 +1293,11 @@ function Picker:_do_selection(prompt)
   end
 end
 
+--- Wrapper function for `Picker:new` that incorporates user provided `opts`
+--- with the telescope `defaults`
+---@param opts table
+---@param defaults table
+---@return any
```
Conni2461, line 1300: There are technically class annotation that you can do and then you can say `---@return Picker` and we can document fields for that class and so much more but i think that works to :laughing:

Conni2461:
If i were you i would merge this before tj comes back and rewrites pickers again: https://github.com/nvim-telescope/telescope.nvim/pull/1491

and let him update documentation :rofl:

l-kershaw:
Agreed, best to get this merged before tj rewrites it all again 🤣

I'll merge now, and if there's anything we want to add/change later, we can just do another PR 🙂

PR #1562, opened 2021-12-06
feat: Config Option to Use Unix Path Separators 

aenichols:
I work on Windows and wanted a way to make telescope treat paths with a Unix friendly separator.

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -12,7 +12,7 @@ local log = require "telescope.log"
 
 local scan = require "plenary.scandir"
 local Path = require "plenary.path"
-local os_sep = Path.path.sep
+local os_sep = utils.get_separator(conf.use_unix_friendly_paths)
 
 local flatten = vim.tbl_flatten
 local filter = vim.tbl_filter
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -676,6 +676,15 @@ append(
     Default: require("telescope.previewers").buffer_previewer_maker]]
 )
 
+append(
+  "use_unix_friendly_paths",
+  false,
+  [[
+  Boolean if should use unix friendly os seperator. "\\" -> "/"
+
+  Default: false]]
+)
+
 -- @param user_defaults table: a table where keys are the names of options,
 --    and values are the ones the user wants
 -- @param tele_defaults table: (optional) a table containing all of the defaults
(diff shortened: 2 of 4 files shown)


Conni2461:
I dont like this solution. We do path operations in plenary using `\` (on windows) just so we sub it with `/` here in telescope.
It feels like plenary.path should handle this based on `:help shellslash`. Maybe. But we use `path_separator` to determine between operating systems.
Its hard like always with windows.

aenichols:
> I dont like this solution. We do path operations in plenary using `\` (on windows) just so we sub it with `/` here in telescope. It feels like plenary.path should handle this based on `:help shellslash`. Maybe. But we use `path_separator` to determine between operating systems. Its hard like always with windows.

Thanks Conni, working with Windows is never easy! =/

I finished testing with setting plenary path to '/' using `set shellslash`. I feel like this change would have way too many consequences.

image
<sub>Note: I know <code>exists("+shellshlash")</code> is wrong- I am pretty new to vimscript. However this still returns 1 enough for testing. </sub>

Some of the bugs I ran into after making this change.
- plenary curl was trying to user /tmp instead of the correct windows temp dir.
- root path is now treated as / instead of C:/ (I seen errors when opening a term through lsp saga or harpoon)

image
image
image

Even after the change with Path.path.sep = '/' telescope still did not transform the paths as expected.

I know this may not be the most ideal solution, but I do like that it's isolated to telescope as plenary may affect a lot.
I am not sure where else to look as this may be over my head, I would appreciate some help in finding a good solution.

Conni2461:
Yes all of these bugs emerge because of what i said here:

> But we use path_separator to determine between operating systems.

aenichols:
WIP but is this more inline to what your thinking?

https://github.com/aenichols/telescope.nvim/commit/b5342ad38a893180da1139916fce09ded4e71bb8

aenichols:
@Conni2461 would I be able to get some help finding an appropriate solution to this?

Conni2461:
I can't help you. I dont like the solution. And it most likely only benefits a super small percentage of the community. But regardless we call the `path = utils.transform_path_os_sep(path)` in the `transform_path` for everyone. Which does a sub even if you are already on a unix like system. And this function is called a lot. For every rendered item.

If you only want the change in the display then please use `:help telescope.defaults.path_display` like this

```lua
require("telescope").setup {
  defaults = {
      return path:gsub("\\", "/")
    end,
  },
}
```

aenichols:
Thanks @Conni2461, even though its not the most elegant solution I was able to fix all my issues with path_display and substituting the selection value.

I was also wondering if I wanted to start contributing to telescope- is there a good place to start for a beginner?

PR #1567, opened 2021-12-07
[ADD] add windows_position option in layout_config

fedoranvar:
Good day to you!

I've added `window_position` to `layout_config`
 currently implemented it for vertical strategy.
 
image

image

image

 My main goal was to be able create `scratchpad`-like theme, but i've suggested that it could be handy for other themes (layouts)

Please make a review and give advice: is it  worth to be in telescope's core or not?

Thank you!

### DIFF
--- a/lua/telescope/pickers/layout_strategies.lua
+++ b/lua/telescope/pickers/layout_strategies.lua
@@ -94,6 +94,7 @@ local get_valid_configuration_keys = function(strategy_config)
     -- TEMP: There are a few keys we should say are valid to start with.
     preview_cutoff = true,
     prompt_position = true,
+    window_position = true,
   }
 
   for key in pairs(strategy_config) do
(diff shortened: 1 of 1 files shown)


l-kershaw:
I'm not sure if this should be added. This would add a lot of complexity to the layouts for (what I think would be) relatively little benefit. Like how do you handle if the position would put the layout off the edge of the screen?

Could you explain what your intended use case is a bit more?

fedoranvar:
My use case is to having theme, like:

* `get_dropdown` - but at top
* `get_cursor` -  not at cursor, but  custom position, so i always know where to look for my picker

image

 > Like how do you handle if the position would put the layout off the edge of the screen?
 
it can be limited like in `cursor` theme: calculate corners and check if they beyond window

Another suggestion, can it be accepted as new layout_config and theme with name, like, `get_pinned`?

Thank you!

l-kershaw:
My concern is how to implement this in a way that keeps this extra positioning stuff as separate as possible from the rest of the layout strategy. Would being able to pass an "anchor" option e.g. "NW", "E", "S" to position on the corresponding edge/corner be sufficient for you?
This seems like it would be easier to implement without completely rewriting all of the `layout_strategies` that we want this positioning to be an option for. It would also make it easier for users to configure, as if they want it in the bottom right corner, they can set the "anchor" to "SE", which can then adapt if the terminal changes size.

fedoranvar:
yes, it will be sufficient. 

i'm gonna close this PR, if you're gonna implement it

PR #1568, opened 2021-12-07
fix: [file_browser] hide "../" when in rootdir

skebanga:
Currently `file_browser` will always insert an "up dir" entry (`../`), regardless of where it is in the filesystem

This change suppresses this entry if we are in the root directory (`/`)

Note I do not have access to Windows. I don't think `\` is ever a valid path in Windows, so don't believe this will have any adverse affects on that OS, but I haven't been able to check specifically.

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -309,7 +309,9 @@ files.file_browser = function(opts)
           table.insert(data, typ == "directory" and (entry .. os_sep) or entry)
         end,
       })
-      table.insert(data, 1, ".." .. os_sep)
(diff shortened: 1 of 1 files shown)


Conni2461:
file_browser will be moved: https://github.com/nvim-telescope/telescope-file-browser.nvim please check if its already fixed there and if not open a PR.

Thanks :)

skebanga:
@Conni2461 When is telescope-file-browser due to be merged? Is there a discussion anywhere I can peruse on what the future direction is?

I looked and this change doesn't exist in that project - will make a PR there

PR #1571, opened 2021-12-07
Add layout_position option to 'vertical' layout strategy

dawsers:
Added a new option to the "vertical" layout strategy to be able to position the layout at the 'center' (current behavior and default), 'top' or 'bottom' of the current neovim window.

The idea is to have the option to see the current buffer better instead of setting the telescope popup right at the center. coc.nvim, LeaderF and other vim plugins permit this kind of layout, and some people find it tidier.


dawsers:
No need for this any more with the new `anchor` option, closing.

PR #1572, opened 2021-12-07
feat: multiselect icon

### LINKED ISSUE #1551: Multi Selection Prefix
Good day to you!

I'm always frustrated when i can't set prefix for selected entries, like in fzf
image

I would like to have parameter `multi_selection_prefix` or `selected_entry_prefix` and dedicated highlight group for it
image

For alternative, i'm using background color for multiselected entries, but it gets messy in `highlights` picker

Thank you!

l-kershaw:
Adds an icon next to multi-selected entries in a picker.
Icon defaults to "+". Functionality disabled by setting to "".
The corresponding highlight group is called `TelescopeMultiIcon`.

Demo:
image

Resolves #1551

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -184,6 +184,15 @@ telescope.setup({opts})                                    *telescope.setup()*
 
         Default: '  '
 
+                                            *telescope.defaults.multi_icon*
+    multi_icon: ~
+        Symbol to add in front of a multi-selected result entry.
+        Replaces final character of |telescope.defaults.selection_caret| and
+        |telescope.defaults.entry_prefix| as appropriate.
+        To have no icon, set to the empty string.
+
+        Default: '+'
+
                                           *telescope.defaults.initial_mode*
     initial_mode: ~
         Determines in which mode telescope starts. Valid Keys:
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -246,6 +246,18 @@ append(
 )
 
 append(
+  "multi_icon",
+  "+",
+  [[
+  Symbol to add in front of a multi-selected result entry.
+  Replaces final character of |telescope.defaults.selection_caret| and
+  |telescope.defaults.entry_prefix| as appropriate.
+  To have no icon, set to the empty string.
+
+  Default: '+']]
+)
+
+append(
   "initial_mode",
   "insert",
   [[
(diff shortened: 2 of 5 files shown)


fedoranvar:
with this PR it is possible to set some space:
*  hetween `multi_icon` and `selection_caret`
*  hetween `multi_icon` and result-line icon/text?

l-kershaw:
@fedoranvar I have refactored things a bit, so it should now be possible to add spacing wherever you want by adding spacing to `selection_caret` or `multi_icon`. Note you may also have to adjust `entry_prefix` to keep things aligned.

l-kershaw:
@Conni2461 or @fdschmidt93, could one of you have a look at this PR please?
I don't think there is anything too weird going on, but wanted to double check I haven't done anything stupid before merging 😅

Conni2461:
I am not a fan of the `+` but i see why someone wants them (because they have limited eye sight or something similar) but i think its too cluttered, like nothing can "breath" there. Thats why i only added the `under consideration` tag because i havent made a decision.

Code wise idk if this is a good solution because i have not thought about an implementation at all

l-kershaw:
Yeh, agree that it would definitely be helpful for those with limited eye sight.

I can see why you think it can appear too cluttered (although to me this only becomes a problem with a large number of entries selected, e.g. from `actions.select_all()`), and as you noted, this is configurable to allow things like `>+ entry`, `> + entry` or just disabling the multi-select icon entirely. I'm happy to adjust some defaults if this makes things more agreeable for you.

Regarding refactoring prefix options:
- did you have any thoughts about how you want it?
- was there a particular reason for waiting until post release for the refactor?

Conni2461:
Because i dont wanna introduce breaking changes that close to a release and yes we are actually close. I wanted to be done last year but i had a good vacation so maybe end of the month or mid Feb is a somewhat reasonable. We can do all this refactoring later. Its not time pressuring. 

If you wanna merge this feel free. Dont feel discouraged by me

l-kershaw:
Fair enough, having never "released" something before, I'm not familiar with what the standard procedure is. My guess was better to do it before since then you don't have to deal with issues with both versions, but perfectly happy to have it after 🙂

Sounds like a reasonable timeframe to me 👍

I think I will merge this PR as it is easy to disable if you don't want it, but could be very helpful for some users. We can have another look at default values if needed when we do the options refactor.

PR #1574, opened 2021-12-08
chore: deprecation notice for builtin.file_browser

fdschmidt93:
I'm sure it'll get users to switch. They have to press `<CR>` for the notice to go away :sweat_smile:

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -288,6 +288,9 @@ local function prepare_match(entry, kind)
 end
 
 files.file_browser = function(opts)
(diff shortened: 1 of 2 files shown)


fdschmidt93:
Sorry, too drained from work lately. I'll try to do my best to get the outstanding work for the file browser extension in today & tomorrow. Then we can deprecate here. I'd just move everything back by a week.

fdschmidt93:
`telescope-file-browser.nvim` now is in a solid enough shape to merge the notice. No time like the present :sweat_smile:

I'll try to further improve `telescope-file-browser` documentation and add the `fd` finder up until removal, but neither is critical.

jackpinto:
Thanks for all the work. It is an amazing extension. Really nice and straightforward to use.

PR #1576, opened 2021-12-08
fix: transform_path nil input

Conni2461:
Sometimes i get with live_grep:
```
Finder failed with msg:  ...nfig/nvim/plugged/telescope.nvim/lua/telescope/utils.lua:324: bad argument #1 to 'match' (string expected, got nil)
```

This happens when rg runs on a binary file, like a pdf. Then the parsing fails. We should maybe also make sure that we dont do anything if the line parsing fails. Regardless `transform_path` should not error if filepath is nil


l-kershaw (approved):
LGTM 👍

PR #1577, opened 2021-12-08
feat: `center` layout `mirror` option

l-kershaw:
Adds `mirror` option to `center` layout
This switches the position of the preview window to below the central block, rather than above.


Conni2461:
Thanks :)

PR #1582, opened 2021-12-09
feat: layout `anchor`

l-kershaw:
Adds an `anchor` option to `layout_config` for the `horizontal`, `vertical` and `center` layouts strategies.

This can be used to position these layouts at the edges/corners of the terminal window using "compass directions".
e.g. "SE" corresponds to the bottom right corner, "W" the left edge and "N" the top edge.

This PR tries to implement the aims of both #1567 and #1571 in a coherent way that doesn't require large scale refactoring of the `layout_strategies` code.
@fedoranvar and @dawsers, does this implementation work for both of you?

**Demos:**
SE
image
W
image
N
image

**Note:**
Since the center layout always extends to the top of the screen (or the bottom of the screen once #1577 is merged), the anchoring only affects this layout horizontally, i.e. it can only be pinned to the middle of the left or right edges.


dawsers:
@l-kershaw: This is definitely a more general solution than what I needed, and of course it covers my special case, so thank you!

fedoranvar:
@l-kershaw 
yes, that's good enough for me!

Thank you!

Conni2461:
I saw the +220 and thought thats a maybe for 1.0 but then i saw its mostly docs and tests (:tada:). So I'll schedule it for 1.0 If you are done you can merge. Thanks :)

PR #1583, opened 2021-12-10
Fixed module name for builtin picker, correctly jump to description

smhc:
The module name was being calculated as 'init' but the pickers are all prefixed with 'builtin'.. the directory name for 'builtin', not the 'init.lua' filename.


Conni2461 (approved):
Thanks :)

PR #1587, opened 2021-12-10
fix: `builtin` only have entries for extension functions

l-kershaw:
`:Telescope builtin include_extensions=true` currently errors when the `telescope-file-browser.nvim` extension is being used, as it exports some stuff as tables (e.g. actions) which breaks `debug.getinfo`. I don't think we want to stop extensions from exporting actions and things, so I've added a check for the `func_obj` actually being a function.

@fdschmidt93 just pinging as an FYI, in case this comes up as an issue over on the `file-browser` repo.

Also, in the case of `file-browser`, I'm not sure if we want both `file_browser` and `picker` to be included, so we could maybe add a check for beginning with a "_" to exclude from the builtin `picker`. This way extension creators can choose which exported functions to include here?

### DIFF
--- a/lua/telescope/_extensions/init.lua
+++ b/lua/telescope/_extensions/init.lua
@@ -43,6 +43,9 @@ extensions.manager = setmetatable({}, {
 ---
 ---         Only the items in `exports` will be exposed on the  resulting
 ---         module that users can access via require('telescope').extensions.foo
+---         Also, any top-level key-value pairs in exports where the value is a function and the
+---         key doesn't start with an underscore will be included when calling the `builtin` picker
+---         with the `include_extensions` option enabled.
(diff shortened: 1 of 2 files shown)


fdschmidt93 (approved):
LGTM :)

> I'm not sure if we want both file_browser and picker to be included, so we could maybe add a check for beginning with a "_" to exclude from the builtin picker.

Agreed. We'd need documentation then for that in the extensions interface on the README.

l-kershaw:
I've added the check for the underscore and a bit of explanation in the `_extensions` file (I couldn't find anywhere else that mentioned the `exports` key at all).

Unless anyone wants any more changes I will probably merge in the next day or two.

PR #1544, opened 2021-12-03
feat: chain buitlin commands

kylo252:
### Description

Allows chaining/invoking builtin commands by extending `action.run_builtin`

Fixes: #564, #930

Partially fixes: #618, #620, #1357

#### Example

- Select directory and then invoke `live_grep` or `find_files` in that directory, see #1385

#### Other useful examples

* `fuzzy_filter_results`: works by chaining `grep_string` and assuming you have `fzf` as a sorter

https://user-images.githubusercontent.com/59826753/144626240-2b3a8894-ba54-4ee3-a44a-ee9873e61dab.mp4

* `dynamic_filetype`: works by modifying the vimgrep_args passed to `ripgrep` using the filetype under the cursor to either include or exclude it

https://user-images.githubusercontent.com/59826753/144626499-3bdfb102-0b86-4ad5-a1b5-ce60eee2a3a8.mp4

PR #1589, opened 2021-12-11
chore: remove redundant run_builtin action

kylo252:
Remove `run_builtin` as it's really not that useful, as per the discussion in https://github.com/nvim-telescope/telescope.nvim/pull/1544#issuecomment-991623542

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -31,12 +31,7 @@ end
 
 local internal = {}
 
--- TODO: What the heck should we do for accepting this.
---  vim.fn.setreg("+", "nnoremap $TODO :lua require('telescope.builtin').<whatever>()<CR>")
--- TODO: Can we just do the names instead?
 internal.builtin = function(opts)
-  opts.path_display = utils.get_default(opts.path_display, "hidden")
-  opts.ignore_filename = utils.get_default(opts.ignore_filename, true)
   opts.include_extensions = utils.get_default(opts.include_extensions, false)
 
   local objs = {}
@@ -81,7 +76,27 @@ internal.builtin = function(opts)
     previewer = previewers.builtin.new(opts),
     sorter = conf.generic_sorter(opts),
     attach_mappings = function(_)
-      actions.select_default:replace(actions.run_builtin)
+      actions.select_default:replace(function(_)
+        local selection = action_state.get_selected_entry()
+        if not selection then
+          print "[telescope] Nothing currently selected"
+          return
+        end
+
+        -- we do this to avoid any surprises
+        opts.include_extensions = nil
+
+        if string.match(selection.text, " : ") then
+          -- Call appropriate function from extensions
+          local split_string = vim.split(selection.text, " : ")
+          local ext = split_string[1]
+          local func = split_string[2]
+          require("telescope").extensions[ext][func](opts)
+        else
+          -- Call appropriate telescope builtin
+          require("telescope.builtin")[selection.text](opts)
+        end
+      end)
       return true
     end,
   }):find()
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1001,7 +1001,6 @@ builtin.builtin({opts})                                    *builtin.builtin()*
         {opts} (table)  options to pass to the picker
 
     Options: ~
-        {ignore_filename}    (boolean)  dont show filenames (default: true)
         {include_extensions} (boolean)  if true will show the pickers of the
                                         installed extensions (default: false)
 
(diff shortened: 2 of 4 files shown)


lua/telescope/builtin/internal.lua
```diff
@@ -81,7 +81,15 @@ internal.builtin = function(opts)
...
     attach_mappings = function(_)
-      actions.select_default:replace(actions.run_builtin)
+      actions.select_default:replace(function(prompt_bufnr)
+        local selection = action_state.get_selected_entry()
+        if not selection then
+          print "[telescope] Nothing currently selected"
+          return
+        end
+        -- Call appropriate telescope builtin
+        require("telescope.builtin")[selection.text](opts)
```
Conni2461, line 91: You still need this part:

```lua
  if string.match(selection.text, " : ") then
    -- Call appropriate function from extensions
    local split_string = vim.split(selection.text, " : ")
    local ext = split_string[1]
    local func = split_string[2]
    require("telescope").extensions[ext][func]()
  else
    -- Call appropriate telescope builtin
    require("telescope.builtin")[selection.text]()
  end
```
so extensions also work

kylo252: Yeah, but I figured that they don't really fit in an internal function called `builtin`. 

Maybe I'm missing something obvious here, but I can see `fzf` showing up, but then it doesn't do anything. It also doesn't work on `zoxide` until I actually run it manually first, so that makes it a bit useless for extensions discoverability since it requires that you know and execute the command first.

Do you still want me to keep the ability to run extensions? or remove their related opts from this function instead?

Conni2461: `fzf` is shown because of a bad design decision in the extension interface. Pickers and internal functions will be exported the same way. But this shouldnt be addressed here. It works for me:
https://user-images.githubusercontent.com/15233006/145681549-3c991917-5e77-4708-a0d8-b6e4eba3196e.mp4

> Do you still want me to keep the ability to run extensions?

Where do i say i wanna drop this ability. Especially without adjusting documentation / removal of this option

kylo252: I'll leave it in then.

lua/telescope/builtin/internal.lua
```diff
@@ -81,7 +81,23 @@ internal.builtin = function(opts)
...
+        end
+        if string.match(selection.text, " : ") then
+          -- Call appropriate function from extensions
+          local split_string = vim.split(selection.text, " : ")
+          local ext = split_string[1]
+          local func = split_string[2]
+          require("telescope").extensions[ext][func]()
+        else
+          -- Call appropriate telescope builtin
+          require("telescope.builtin")[selection.text]()
```
Conni2461, line 98: We still need to pass in the opts so the theme works.
And it probably makes sense to nil `ignore_filename` and `include_extensions` the two opts that builtin supports

lua/telescope/builtin/internal.lua
```diff
@@ -31,12 +31,9 @@ end
 
 local internal = {}
 
--- TODO: What the heck should we do for accepting this.
---  vim.fn.setreg("+", "nnoremap $TODO :lua require('telescope.builtin').<whatever>()<CR>")
--- TODO: Can we just do the names instead?
 internal.builtin = function(opts)
-  opts.path_display = utils.get_default(opts.path_display, "hidden")
```
kylo252, line 38: This one was causing files to appear with only an icon, and I didn't see it in the docs anyway, so I removed it.

lua/telescope/builtin/internal.lua
```diff
@@ -31,12 +31,9 @@ end
...
-  opts.ignore_filename = utils.get_default(opts.ignore_filename, true)
+  opts = opts or {}
+  opts.ignore_filename = utils.get_default(opts.ignore_filename, false)
```
kylo252, line 36: @Conni2461, Is this what you meant?

Conni2461: No . Nil them before passing them into the pickers you open.

So they dont mess up things ...

Conni2461: Also you dont need to do `opts = opts or {}` in telescope builtins. We do this automatically so it doesnt error if we forget it.

Here is how we do it

https://github.com/nvim-telescope/telescope.nvim/blob/5f37fbfa837dfee7ecd30f388b271f4a71c0a9e0/lua/telescope/builtin/internal.lua#L1219-L1231

lua/telescope/builtin/internal.lua
Conni2461: We dont do anything with `path_display` and `ignore_filename` so you can remove both.

Conni2461 (approved):
Thanks :)

PR #1600, opened 2021-12-16
fix: make keepinsert work with nvim-0.7

### LINKED ISSUE #1599: actions._close(bufnr, true) exits insert mode
### Description

I have noticed that plugins that change from one picker to another or even builtin pickers (`:Telescope builtin`) quit insert mode when not intended.

### Neovim version

```markdown
NVIM v0.7.0-dev+22d7dd2ae
Build type: Release
```

macOs 12.1

### checkhealth telescope

```markdown
  - OK: plenary installed.

  - OK: rg: found ripgrep 12.1.1
  - OK: fd: found fd 8.3.0

## Telescope Extension: `fzf`
```

### Steps to reproduce

1. nvim -nu minimal.lua
2. :Telescope builtin
3. Select a builtin

### Expected behavior

I was expecting to stay in insert mode

### Actual behavior

Vim mode is changed to Normal

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

kylo252:
Ensure that `keepinsert` works correctly in `actions._close()` in nvim-0.7 after https://github.com/neovim/neovim/commit/28134f4e78819c2bbf0344326b9d44f21eb0d736

Fixes #1599

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -357,6 +357,7 @@ end
 
 --- Close the Telescope window, usually used within an action<br>
 --- Deprecated and no longer needed, does the same as |telescope.actions.close|. Might be removed in the future
+---@deprecated
 ---@param prompt_bufnr number: The prompt bufnr
 actions._close = function(prompt_bufnr)
   actions.close(prompt_bufnr)
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -348,17 +348,13 @@ end
 
 --- Close the Telescope window and specify if you want to keep insert mode or not
 ---@param prompt_bufnr number: The prompt bufnr
----@param keepinsert boolean: Remain in INSERT mode if true
-actions._close = function(prompt_bufnr, keepinsert)
+actions._close = function(prompt_bufnr)
   action_state.get_current_history():reset()
   local picker = action_state.get_current_picker(prompt_bufnr)
   local prompt_win = state.get_status(prompt_bufnr).prompt_win
   local original_win_id = picker.original_win_id
 
   actions.close_pum(prompt_bufnr)
-  if not keepinsert then
-    vim.cmd [[stopinsert]]
-  end
 
   vim.api.nvim_win_close(prompt_win, true)
   pcall(vim.cmd, string.format([[silent bdelete! %s]], prompt_bufnr))
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -343,12 +343,6 @@ end
 --- Close the Telescope window, usually used within an action
 ---@param prompt_bufnr number: The prompt bufnr
 actions.close = function(prompt_bufnr)
-  actions._close(prompt_bufnr, false)
-end
-
---- Close the Telescope window and specify if you want to keep insert mode or not
----@param prompt_bufnr number: The prompt bufnr
-actions._close = function(prompt_bufnr)
   action_state.get_current_history():reset()
   local picker = action_state.get_current_picker(prompt_bufnr)
   local prompt_win = state.get_status(prompt_bufnr).prompt_win
@@ -359,6 +353,13 @@ actions._close = function(prompt_bufnr)
   vim.api.nvim_win_close(prompt_win, true)
   pcall(vim.cmd, string.format([[silent bdelete! %s]], prompt_bufnr))
   pcall(a.nvim_set_current_win, original_win_id)
+end
+
+--- Close the Telescope window, usually used within an action<br>
+--- Deprecated and no longer needed, does the same as |telescope.actions.close|. Might be removed in the future
+---@param prompt_bufnr number: The prompt bufnr
+actions._close = function(prompt_bufnr)
+  actions.close(prompt_bufnr)
 end
 
 local set_edit_line = function(prompt_bufnr, fname, prefix, postfix)
@@ -456,7 +457,7 @@ end
 ---@param prompt_bufnr number: The prompt bufnr
 actions.insert_symbol_i = function(prompt_bufnr)
   local symbol = action_state.get_selected_entry().value[1]
-  actions._close(prompt_bufnr, true)
+  actions.close(prompt_bufnr)
   vim.schedule(function()
     vim.cmd [[startinsert]]
     vim.api.nvim_put({ symbol }, "", true, true)
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -356,14 +356,8 @@ actions._close = function(prompt_bufnr, keepinsert)
   local original_win_id = picker.original_win_id
 
   actions.close_pum(prompt_bufnr)
-
   if not keepinsert then
     vim.cmd [[stopinsert]]
-  elseif vim.fn.has "nvim-0.7" == 1 then
-    -- https://github.com/neovim/neovim/commit/28134f4e78819c2bbf0344326b9d44f21eb0d736
-    vim.schedule(function()
-      vim.cmd [[startinsert]]
-    end)
   end
 
   vim.api.nvim_win_close(prompt_win, true)
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -356,8 +356,14 @@ actions._close = function(prompt_bufnr, keepinsert)
   local original_win_id = picker.original_win_id
 
   actions.close_pum(prompt_bufnr)
+
   if not keepinsert then
     vim.cmd [[stopinsert]]
+  elseif vim.fn.has "nvim-0.7" == 1 then
+    -- https://github.com/neovim/neovim/commit/28134f4e78819c2bbf0344326b9d44f21eb0d736
+    vim.schedule(function()
+      vim.cmd [[startinsert]]
+    end)
   end
 
   vim.api.nvim_win_close(prompt_win, true)
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -178,9 +178,10 @@ internal.pickers = function(opts)
       actions.select_default:replace(function(prompt_bufnr)
         local current_picker = action_state.get_current_picker(prompt_bufnr)
         local selection_index = current_picker:get_index(current_picker:get_selection_row())
-        actions._close(prompt_bufnr, cached_pickers[selection_index].initial_mode == "insert")
+        actions.close(prompt_bufnr)
         opts.cache_picker = opts._cache_picker
         opts["cache_index"] = selection_index
+        opts["initial_mode"] = cached_pickers[selection_index].initial_mode
         internal.resume(opts)
       end)
       map("i", "<C-x>", actions.remove_selected_picker)
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -438,9 +438,7 @@ function Picker:find()
         vim.cmd [[startinsert!]]
       end)
       if self.default_text then
-        vim.schedule(function()
-          self:set_prompt(self.default_text)
-        end)
+        self:set_prompt(self.default_text)
       end
       if self.initial_mode == "normal" then
         -- otherwise (i) insert mode exitted faster than `picker:set_prompt`; (ii) cursor on wrong pos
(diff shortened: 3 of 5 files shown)


fdschmidt93:
Yeah, I was trying something similar, but that doesn't seem to work for instance with `:Telescope pickers` when you select a picker that already had text. Maybe that requires a separate fix?

kylo252:
> Yeah, I was trying something similar, but that doesn't seem to work for instance with `:Telescope pickers` when you select a picker that already had text. Maybe that requires a separate fix?

I have no idea how to use this 😅

fdschmidt93:
Sorry, I should have expressed myself more clearly. This solution has the following issue:

1. Do e.g. `builtin.find_files` and search for something; keep the search string in the prompt
2. Close the picker
3. Open the `:Telescope pickers`
4. Re-open the previously closed picker with the corresponding query
5. The picker is opened correctly in insert mode, but the query is not adequately kept in the prompt

> but the query is not adequately kept in the prompt

This part I was referring to may require a separate fix, because I don't have a good sense of how to fix the remaining in insert mode issue other than how you're doing it as part of this PR.

kylo252:
Thanks for the explanation! It led me to what I believe is the actual correct fix for this.

- works for me on both v0.6.0 and v0.7.0-dev+737-g818ae74ea
 - also works for `:Telescope pickers`

fdschmidt93:
This seems to work generally well for me for the scenarios I've tested but I'm wary that in particular scheduling `set_prompt` may have unintended interactions in for the time being untested scenarios :sweat_smile: Though that may just as well not be the case.

Since it's something rather lower level, I'd prefer a separate set of eyes on this as well prior to merging.

Conni2461:
The prompt_buffer fixes are merge so it should now be possible to just set the prompt without using feedkeys

```lua
  local prompt_text = self.prompt_prefix .. (str or "")
  vim.api.nvim_buf_set_lines(self.prompt_bufnr, 0, 1, false, { prompt_text })
  vim.api.nvim_win_set_cursor(self.prompt_win, { 1, #prompt_text })
  self:_reset_prefix_color()
```

lua/telescope/pickers.lua
```diff
@@ -414,13 +414,15 @@ function Picker:find()
     pcall(a.nvim_buf_set_option, prompt_bufnr, "filetype", "TelescopePrompt")
     pcall(a.nvim_buf_set_option, results_bufnr, "filetype", "TelescopeResults")
 
-    -- TODO(async): I wonder if this should actually happen _before_ we nvim_buf_attach.
-    -- This way the buffer would always start with what we think it should when we start the loop.
```
Conni2461, line 418: why are you removing todos ?

kylo252: I figured that this PR adds the async part, just in a different way, while the suggested way wouldn't work anymore.

Conni2461: not really but its no longer possible due to prompt buffer changes anyway. So its leave it removed

Conni2461:
Are you also fixing `set_prompt`? So it works for 0.6 and current nightly?

kylo252:
> Are you also fixing `set_prompt`? So it works for 0.6 and current nightly?

Done! 

I think I've tried most scenarios and all of them seem to work correctly for me using this configuration on both 0.6 and nightly. Please give it a try and let me know if I've missed something.

fdschmidt93:
Slight issue seems to apply here:

1. Do e.g. builtin.find_files and search for something; keep the search string in the prompt
2. Close the picker
3. Open the :Telescope pickers
4. Re-open the previously closed picker with the corresponding query
5. Correctly in insert mode but cursor is incorrectly on the second-to-last character

kylo252:
> Slight issue seems to apply here

I don't have that problem on either nightly or 0.6, at least not with find_files.

fdschmidt93:
Retried with `minimal_init.lua` with video attached. Not sure why I do and you don't have this problem :/

```lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        -- 'nvim-telescope/telescope.nvim',
        'kylo252/telescope.nvim',
        branch = "nvim-nightly-compat",
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

https://user-images.githubusercontent.com/39233597/147658023-6a3d33f3-ed70-43f6-9301-2d64066daa95.mp4

kylo252:
@fdschmidt93, I can reproduce on this machine at least, although I'm not really sure how to solve this. I tried a few different things and none of them worked. 

Do we still need to call this directly?
https://github.com/nvim-telescope/telescope.nvim/blob/1d1da664eb6505c318d405eea3d633c451edc2d8/lua/telescope/pickers.lua#L389

Why is this one adding a `+1` while the function that @Conni2461 suggested doesn't?
https://github.com/nvim-telescope/telescope.nvim/blob/1d1da664eb6505c318d405eea3d633c451edc2d8/lua/telescope/pickers.lua#L1404-L1407

Aren't these two functions the same?
https://github.com/nvim-telescope/telescope.nvim/blob/1d1da664eb6505c318d405eea3d633c451edc2d8/lua/telescope/pickers.lua#L1453-L1456

fdschmidt93:
Apologies, I'm quite tied up at the moment and hopefully will get to look more into this at some point this week but the little time I have is probably spent on `telescope-file-browser.nvim` docs and stuff so it can be finally removed from `telescope-core` prior to me (possibly) going on a hiatus till February. That said, I've also tried a few things when I reported it to no avail.

To the changes/inconsistencies you've mentioned. At very first initial glance, just looks like some debt acquired over time? I presume that can be cleaned up?

przepompownia:
Additionally, the original selection is then lost when I try to reuse it by `gv`.

lua/telescope/pickers.lua
```diff
@@ -419,11 +419,11 @@ function Picker:find()
...
     if self.initial_mode == "insert" or self.initial_mode == "normal" then
       -- required for set_prompt to work adequately
-      vim.cmd [[startinsert!]]
+      vim.schedule(function()
```
przepompownia, line 424:
```diff
diff --git a/lua/telescope/pickers.lua b/lua/telescope/pickers.lua
index d91ae5c..3889ebf 100644
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -422,6 +422,10 @@ function Picker:find()
     if self.initial_mode == "insert" or self.initial_mode == "normal" then
       -- required for set_prompt to work adequately
       vim.schedule(function()
+        if vim.fn.mode() == 'v' then
+          local key = vim.api.nvim_replace_termcodes("<ESC>", true, false, true)
+          vim.api.nvim_feedkeys(key, "i", true)
+        end
         vim.cmd [[startinsert!]]
       end)
       if self.default_text then
```
Maybe it's not the best solution but it lets me avoid manual exiting from _insert (visual)_ mode.

Conni2461: I think we should deal with opening telescope from visual mode some other time. Ill open a new issue

Conni2461: Better idea, we make your issue about problems opening telescope from visual mode

Conni2461:
@fdschmidt93 I am currently trying to fix the bug you reported here: https://github.com/nvim-telescope/telescope.nvim/pull/1600#issuecomment-1002048543

After that this one is ready

Conni2461:
Seems to be fixed for me. Do you have a couple of minutes to validate? Thanks :)

fdschmidt93:
Works flawlessly, I've spend 20 minutes to test all the weird combinations I could think of around `builtin.builtin`, launching a picker from within another picker and `Telescope pickers` them afterwards. The previous combination with outer product (hopefully complete :laughing:) of with `default_text` and varying `initial_mode`. (eg launch picker no

Truth be told, sometimes the cursor minimally glitches into insert mode, though from having worked on this a also quite bit back then and seeing your solution, I'm fairly confident that this unfortunately cannot be avoided. I'd say let's merge, great job :)

Conni2461:
Ill have to merge after #1851. My merge window today is 3 - 4pm german time.

> E: Ah, can we unify actions._close and actions.close now that keepinsert is no longer a parameter? Am I missing something?

 I added a comment that its deprecated but we can't just remove it because a lot of extensions still call it. So right now its just an "alias".

lua/telescope/builtin/internal.lua
```diff
@@ -178,9 +178,10 @@ internal.pickers = function(opts)
       actions.select_default:replace(function(prompt_bufnr)
         local current_picker = action_state.get_current_picker(prompt_bufnr)
         local selection_index = current_picker:get_index(current_picker:get_selection_row())
-        actions._close(prompt_bufnr, cached_pickers[selection_index].initial_mode == "insert")
+        actions.close(prompt_bufnr)
         opts.cache_picker = opts._cache_picker
         opts["cache_index"] = selection_index
+        opts["initial_mode"] = cached_pickers[selection_index].initial_mode
```
Conni2461, line 184: Does this capture the correct intend?

lua/telescope/pickers.lua
Conni2461: @przepompownia Its fixed because i already had to fix opening telescope from insert mode, because previously that was working. So yeah why not

Conni2461:
@clason I just validated this: https://github.com/nvim-telescope/telescope.nvim/issues/1072#issue-958930038 

Is it the correct behavior that if i do `|foobar` i end up with `fX|oobar` ? Happens on master as well as this branch. Everything else seems to work for me

clason:
No, that is not intended (and happens for me, too). It's virtually impossible to make the insert mode symbol insertion work correctly in all situations -- it's a veritable whack-a-mole game...

Conni2461:
okay thanks :) i just had to make sure i dont make it worse, after these "mode" changes

clason:
If that's the only situation that doesn't work properly (as in, inserting behaves the same as typing the character in insert mode), then you're fine ;)

(Otherwise we have a problem :])

przepompownia:
Quick manual tests after fetching new changes gives expected results both on the testing and my working environment :tada:

lua/telescope/actions/init.lua
```diff
@@ -343,28 +343,25 @@ end
...
-    vim.cmd [[stopinsert]]
-  end
 
   vim.api.nvim_win_close(prompt_win, true)
   pcall(vim.cmd, string.format([[silent bdelete! %s]], prompt_bufnr))
   pcall(a.nvim_set_current_win, original_win_id)
 end
 
+--- Close the Telescope window, usually used within an action<br>
+--- Deprecated and no longer needed, does the same as |telescope.actions.close|. Might be removed in the future
```
kylo252, line 359: maybe use the `---@deprecated` tag instead? 
ref: https://github.com/sumneko/lua-language-server/wiki/EmmyLua-Annotations#deprecated

lua/telescope/actions/init.lua
```diff
@@ -343,28 +343,25 @@ end
...
-  actions._close(prompt_bufnr, false)
-end
-
---- Close the Telescope window and specify if you want to keep insert mode or not
----@param prompt_bufnr number: The prompt bufnr
----@param keepinsert boolean: Remain in INSERT mode if true
-actions._close = function(prompt_bufnr, keepinsert)
   action_state.get_current_history():reset()
   local picker = action_state.get_current_picker(prompt_bufnr)
   local prompt_win = state.get_status(prompt_bufnr).prompt_win
```
kylo252, line 348: shouldn't we validate this value? or bail out early otherwise?

lua/telescope/builtin/internal.lua
fdschmidt93: Yes, works as it should around cached pickers with `:Telescope pickers`.

lua/telescope/pickers.lua
```diff
@@ -432,21 +432,22 @@ function Picker:find()
...
+    if self.initial_mode == "insert" then
+      vim.schedule(function()
+        -- startinsert! did not reliable do `A` no idea why, i even looked at the source code
+        -- Example: live_grep -> type something -> quit -> Telescope pickers -> resume -> cursor of by one
+        local mode = vim.fn.mode()
+        if mode == "n" then
+          vim.api.nvim_feedkeys("A", "n", true)
+        elseif mode == "v" then
+          vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("<ESC>A", true, false, true), "n", true)
+        end
```
kylo252, line 448:
```suggestion
        if mode ~= "i" then
          vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("<ESC>A", true, false, true), "n", true)
        end
```

this seems to work for me

lua/telescope/pickers.lua
przepompownia: Commented outside this thread: https://github.com/nvim-telescope/telescope.nvim/pull/1600#issuecomment-1106402968

lua/telescope/actions/init.lua
Conni2461: ?

kylo252:
```lua
if not vim.api.nvim_win_is_valid(prompt_win) then
  -- nothing to do | raise error
  return
end
```

Conni2461: worked for the past year. If you call this function outside of a telescope action, its undefined behavior

PR #1601, opened 2021-12-16
fix: too early setup on initial extension access

fdschmidt93:
This is a mundane fix for a stupid issue with extensions. For instance, when trying to set mappings from telescope-file-browser or telescope-hop with extension actions, one `require`s the extensions module prior to setup and then user configuration is not applied.

This is a minimal fix within the existing API. I'm curious, what substantive value does `require "telescope".load_extension $EXTENSION` actually add? Alternatively, we could just invoke set up when the extension config is forwarded.. otherwise any extension within telescope interface just has it's default config.

Generally speaking, this is not the most prettiest fix, but maybe more intended to get the ball rolling on a discussion as I didn't want to break the `load_extension` interface just yet.

Once something along these lines is merged, I'll move forward with deprecating the file browser. Basic remappings should be straightforwardly possible :sweat_smile:

### DIFF
--- a/lua/telescope/_extensions/init.lua
+++ b/lua/telescope/_extensions/init.lua
@@ -14,7 +14,6 @@ end
 
 extensions.manager = setmetatable({}, {
   __index = function(t, k)
-    -- See if this extension exists.
 
     local ext = load_extension(k)
     t[k] = ext.exports or {}
--- a/lua/telescope/_extensions/init.lua
+++ b/lua/telescope/_extensions/init.lua
@@ -14,7 +14,6 @@ end
 
 extensions.manager = setmetatable({}, {
   __index = function(t, k)
-
     local ext = load_extension(k)
     t[k] = ext.exports or {}
     extensions._health[k] = ext.health
--- a/lua/telescope/_extensions/init.lua
+++ b/lua/telescope/_extensions/init.lua
@@ -4,18 +4,19 @@ extensions._loaded = {}
 extensions._config = {}
 extensions._health = {}
 
+local load_extension = function(name)
+  local ok, ext = pcall(require, "telescope._extensions." .. name)
+  if not ok then
+    error("This extension doesn't exist or is not installed: " .. name .. "\n" .. ext)
+  end
+  return ext
+end
+
 extensions.manager = setmetatable({}, {
   __index = function(t, k)
     -- See if this extension exists.
-    local ok, ext = pcall(require, "telescope._extensions." .. k)
-    if not ok then
-      error("This extension doesn't exist or is not installed: " .. k .. "\n" .. ext)
-    end
 
-    if ext.setup then
-      ext.setup(extensions._config[k] or {}, require("telescope.config").values)
-    end
-
+    local ext = load_extension(k)
     t[k] = ext.exports or {}
     extensions._health[k] = ext.health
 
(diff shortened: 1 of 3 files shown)


lua/telescope/_extensions/init.lua
```diff
@@ -54,7 +55,10 @@ extensions.register = function(mod)
 end
 
 extensions.load = function(name)
-  return extensions.manager[name]
+  local ext = load_extension(name)
```
Conni2461, line 58: Isnt the extension not inside `extensions.manager` do we need to do load twice? Am i missing something

fdschmidt93: This is a compromise to maintain the current interface. We now load the extension upon access and with `telescope.load_extension $EXT` here, such that.

```lua

-- first access triggers setup without config passed, also status quo
local fb_actions = require "telescope".extensions.file_browser.actions

local telescope = require "telescope"
telescope.setup {
  extensions = {
    -- we only now set up the extension
    file_browser = { ... }
  }
}
-- now we apply the new settings and have to load extension again
telescope.load_extension "file_browser"
```

I would probably only "load" (ie applying settings) the extension in

```lua
telescope.setup {
  extensions = {
    -- we only now set up the extension
    file_browser = { ... }
  }
}
```
and maybe deprecate `load_extension`? But might not work easily with extensions like `telescope-fzf-native` (opting out then annoying I suppose?).

Compromise: just load setup on `load_extension` then?

Conni2461: Oh i was just saying. Cant you do 
```lua
local ext = extensions.manager[name]
if ext.setup then
  ext.setup(extensions._config[name] or {}, require("telescope.config").values)
end
```

rather than having another local function ?

fdschmidt93: No, `extensions.manager[name]` gives you the `exports` in

```lua
extension = {
  setup = { ... },
  exports = { ... },
}
```

Thus, `setup` is not part of `ext` in your example code.

Conni2461: Oh right. Yeah i was missing something :laughing: You can merge this PR if its done.

ikerurda:
I've noticed that now commands simmilar to `:Telescope packer` no longer work, but their lua interface (e.g. `:lua require ("telescope"). extensions.packer.packer ()`) do.
I think this issue is related to this PR...

GreatestOfAllTimee:
> I've noticed that now commands simmilar to `:Telescope packer` no longer work, but their lua interface (e.g. `:lua require ("telescope"). extensions.packer.packer ()`) do. I think this issue is related to this PR...

Yah, I have same issue. Did you find how to fix this ?

GreatestOfAllTimee:
Also `Telescope media_files` seems like no longer work. [Telescope-media-files](https://github.com/nvim-telescope/telescope-media-files.nvim)

clason:
Yes, all user commands for extensions no longer work out of the box (no need to add each single one ;)).

They become available once the extension is triggered (e.g., through the Lua API) the first time.

fdschmidt93:
Sorry guys, I can only look into it tomorrow. Not perfectly clear to me what would cause that (also only able to look from phone now) as with very latest commit it should only add redoing setup with `load_extension`.

clason:
As always, a workaround until this is fixed is to pin telescope to the commit right before this PR got merged (3dafcb419b9b15855731f52965143ad6006726d4) in your plugin manager.

GreatestOfAllTimee:
> Sorry guys, I can only look into it tomorrow. Not perfectly clear to me what would cause that (also only able to look from phone now) as with very latest commit it should only add redoing setup with `load_extension`.

I just noticed that 😂 I hope this will fix or just reinstall telescope will a difference commit while waiting for this issue to resolve.

PR #1609, opened 2021-12-18
fix: handle filenames with colons in `live_grep`

### LINKED ISSUE #1604: Colon in filenames when searching.
### Description

Trying to search emails, the filename gets cut off because of it containing a colon.
1639392127_3.3457503.speedo,U=35,FMD5=7e33429f656f1e6e9d79b29c3f82c57e:2,S becomes 2,S in telescope searches.
Emails gets created with these kinds of weird filenames.

### Neovim version

```markdown
NVIM v0.6.0
Build type: Release
LuaJIT 2.0.5
Compiled by builduser

Features: +acl +iconv +tui
```

Archlinux

### checkhealth telescope

```markdown
│ - OK: plenary installed.
│ - OK: nvim-treesitter installed.

│ - OK: rg: found ripgrep 13.0.0
│ - OK: fd: found fd 8.3.0

## Telescope Extension: `cheat`
│ - INFO: No healthcheck provided

## Telescope Extension: `fzf`
│ - INFO: No healthcheck provided

## Telescope Extension: `zk`
│ - INFO: No healthcheck provided

## Telescope Extension: `zoxide`
│ - INFO: No healthcheck provided
```

### Steps to reproduce

Telescope grep_string in a directory with files having a colon like these:
1639392127_3.3457503.speedo,U=35,FMD5=7e33429f656f1e6e9d79b29c3f82c57e:2,S

find_files works correctly (though it seem to think the files are binary)

### Expected behavior

_No response_

### Actual behavior

pic

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

l-kershaw:
Fixes #1604
@dagle does this work for you?

---

The only time that I can think of this new regex breaking would be if there are sequences of exclusively digits between colons in the filename or the text. However, I'm not the best with regexes, so it would be good if someone else could have a look at it. 

Do we want to add `^` and `$` to the regex too? Can't find a circumstance where doing this breaks anything, but, again, I'm not the best with regexes 😅

This change also means that we don't need to have separate functions for windows/non-windows users, as I introduced in #1494. So its good that we can simplify the code a bit too 🙂

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -110,29 +110,9 @@ do
     ordinal = 1,
   }
 
-  local find = (function()
-    if Path.path.sep == "\\" then
-      return function(t)
-        local start, _, filename, lnum, col, text = string.find(t, [[([^:]+):(%d+):(%d+):(.*)]])
-
-        -- Handle Windows drive letter (e.g. "C:") at the beginning (if present)
-        if start == 3 then
-          filename = string.sub(t.value, 1, 3) .. filename
-        end
-
-        return filename, lnum, col, text
-      end
-    else
-      return function(t)
-        local _, _, filename, lnum, col, text = string.find(t, [[([^:]+):(%d+):(%d+):(.*)]])
-        return filename, lnum, col, text
-      end
-    end
-  end)()
-
   -- Gets called only once to parse everything out for the vimgrep, after that looks up directly.
   local parse = function(t)
-    local filename, lnum, col, text = find(t.value)
+    local _, _, filename, lnum, col, text = string.find(t.value, [[(.+):(%d+):(%d+):(.*)]])
 
     local ok
     ok, lnum = pcall(tonumber, lnum)


mtoohey31:
Hi, I'm testing this right now, it seems to work for me, but I'll post an update if I run into anything.

dagle:
This is working for me. Haven't hit any problems so far. Thanks.

l-kershaw:
Great, thanks guys 👍
Having another look at this, I don't think that the regex will be any different with or without the `^` and `$`, as the groups on the end are "greedy" and accept all characters. So I'm happy with this for Windows now

@Conni2461 could you test this to make sure that I didn't massively break things for Linux 😅

Conni2461:
Seems to work thanks :)

PR #1610, opened 2021-12-18
add filter options for diagnostics builtin

smackesey:
Adds support for filtering diagnostics by:

1. A root directory. Useful to browse only project-scoped diagnostics.
2. `buflisted` status. Useful to exclude diagnostics from unlisted buffers, which are created by language servers that analyze the entire project/workspace (e.g. `sumneko-lua`).

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1458,7 +1458,7 @@ builtin.lsp_dynamic_workspace_symbols({opts}) *builtin.lsp_dynamic_workspace_sym
 
 
 builtin.diagnostics({opts})                            *builtin.diagnostics()*
-    Lists diagnostics for current or all open buffers
+    Lists diagnostics
     - Fields:
       - `All severity flags can be passed as `string` or `number` as per
         `:vim.diagnostic.severity:`
@@ -1481,6 +1481,14 @@ builtin.diagnostics({opts})                            *builtin.diagnostics()*
         {severity_bound} (string|number)  keep diagnostics equal or less
                                           severe wrt severity name (string) or
                                           id (number)
+        {root_dir}       (string|boolean) if set to string, get diagnostics only
+                                          for buffers under this directory. If
+                                          set to true, uses current working
+                                          directory.
+        {no_unlisted}    (boolean)        if true, get diagnostics only for
+                                          listed buffers. Useful to exclude
+                                          diagnostics from language servers that
+                                          report diagnostics for whole projects.
         {no_sign}        (boolean)        hide DiagnosticSigns from Results
                                           (default: false)
         {line_width}     (number)         set length of diagnostic entry text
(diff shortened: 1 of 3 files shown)


fdschmidt93:
Thanks for the PR!

While this is a use case telescope should be able to inherently be able to address, I'm afraid though long-term we have more generic solutions in mind

See #1082 #1168

smackesey:
Generic filtering sounds fantastic-- I hope it merges soon.

In the meantime, I hope you consider merging this or adding similar to diagnostic search (even if the functionality is later removed in favor of the generic mechanism), because:

1. There are already picker-specific filter options (for severity) on the diagnostics picker.
2. It will likely make diagnostic search much more useful for a large number of users until generic filters are available.

doc/telescope.txt
```diff
@@ -1459,7 +1459,7 @@ builtin.lsp_dynamic_workspace_symbols({opts}) *builtin.lsp_dynamic_workspace_sym
 
 
 builtin.diagnostics({opts})                            *builtin.diagnostics()*
-    Lists diagnostics for current or all open buffers
```
fdschmidt93, line 1462: Documentation is auto-generated and therefore any changes should stem from the corresponding comments. In this case from `/telescope/builtin/init.lua` at `builtin.diagnostics`.

lua/telescope/builtin/diagnostics.lua
```diff
@@ -41,25 +41,42 @@ local diagnostics_to_tbl = function(opts)
     end
   end
 
+  local root_dir
```
fdschmidt93, line 44: `opts.root_dir = opts.root_dir == true and vim.loop.cwd or opts.root_dir` I think already suffices and is more readable.

`vim.loop.cwd` is not needed here but generally favored (slightly faster and can be used in "fast"-context, IIRC).

lua/telescope/builtin/diagnostics.lua
```diff
@@ -41,25 +41,42 @@ local diagnostics_to_tbl = function(opts)
...
-  local preprocess_diag = function(diagnostic)
+  local filter_diag = function(diagnostic)
     if bufnr_name_map[diagnostic.bufnr] == nil then
       bufnr_name_map[diagnostic.bufnr] = vim.api.nvim_buf_get_name(diagnostic.bufnr)
     end
 
-    local buffer_diag = {
+    local root_dir_test = not root_dir or
+      string.sub(bufnr_name_map[diagnostic.bufnr], 1, #root_dir) == root_dir
+    local listed_test = not opts.no_unlisted  or
```
fdschmidt93, line 59: `vim.api.nvim_buf_get_option($BUFNR, "buflisted")` already returns `boolean`. `vim.api` also favored (speed, etc.)

fdschmidt93 (commented):
I very much prefer having a stackable sorters or something similar rather than picker specific options.

I guess here this can be simplified largely to the degree that I'd be ok with having this. Happy to hear from others.

smackesey:
Thanks for the feedback, I made the requested adjustments.

Conni2461 (approved):
The symbols pickers have a couple of filter options as well so we can also add those.

Thanks for the PR :) I rebased and generated docs

PR #1614, opened 2021-12-19
feat: adds multiselect counter in prompt status_text

l-kershaw:
Adds a counter for the number of entries in the multi-selection.

Example:
image

Note: new counter isn't displayed when nothing in the multi-selection

### DIFF
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -332,6 +332,7 @@ append(
 append(
   "get_status_text",
   function(self)
+    local ww = #(self:get_multi_selection())
     local xx = (self.stats.processed or 0) - (self.stats.filtered or 0)
     local yy = self.stats.processed or 0
     if xx == 0 and yy == 0 then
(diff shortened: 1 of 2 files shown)


Conni2461:
Mixed feelings about this one. I personally think its too cluttered. 

But we can merge the `get_status_updater` stuff so people could just set it themselves?

l-kershaw:
I'm happy to revert the default and keep other changes if that's what you think is best.

My reasoning for adding it was that if you add an entry to the multiselection, then filter so that entry is no longer visible, then there is no visible indicator for whether you have any entries in your multiselection at all. I thought that this could cause confusion, particularly as we are trying to improve the multiselection experience at the moment.

fdschmidt93:
Related: https://github.com/nvim-telescope/telescope-file-browser.nvim/pull/48

Conni2461:
Okay i actually used it and i dont mind it :grimacing: 

Thanks :)

PR #1619, opened 2021-12-22
feat: workspace & document diag in title

fdschmidt93:
This PR is the minimum viable solution to disambiguate between "Workspace" and "Document" diagnostics (ie once `bufnr`) is passed.

Of course we could also add some more info in the prompt title.

One other way I like, which would be a departure from how we've been doing it would be:

* Prompt title: "Diagnostics"
* Results title: "Workspace" or filename (or buf number)

Any thoughts/preferences here?


Conni2461 (approved):
Thanks :)

PR #1624, opened 2021-12-25
[RFC]: Ensure 'cb_filter' defined when creating new index-bias sorter

levouh:
Re-creating #781 from the correct branch

---

I use this sorter to do things like bias terminal history towards those that are newer when using _Telescope_ to query it, etc. as you'd expect. I noticed that this finder isn't broken, but throws a handful of warnings due to the `sorter:sort` not being called with the new `cb_add`, etc. functions.

If we use the locally-defined sorter, we get double the results back so I just ended up calling `:sorting_function` directly.

Not sure if this is the best way to do this, but looking for feedback.

### DIFF
--- a/lua/telescope/sorters.lua
+++ b/lua/telescope/sorters.lua
@@ -138,14 +138,14 @@ function Sorter:score(prompt, entry, cb_add, cb_filter)
     if self.tags then
       self.tags:insert(entry)
     end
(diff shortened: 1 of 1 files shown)


levouh:
@Conni2461 fixed this one (based on the correct branch, rather than `master` of my fork).

Conni2461 (approved):
LGTM thanks

PR #1628, opened 2021-12-27
fix: setup solely on access as well

fdschmidt93:
@clason does that fix your issue? 

I'll quickly check that this doesn't interfere with previous commit


clason:
Indeed, that restores the old behavior for me!

fdschmidt93:
Since it doesn't interfere with what the previous commit solved, I'll merge this as it is a strictly positive change since setup through accessing the extension is only invoked anyways and facilitates lazy loading.

PR #1630, opened 2021-12-28
Prevent an exception when err.message is nil

kdarkhan:
When trying to use go-to-definition using telescope's
LSP helper while Sumneko lua-language-server has not completed loading yet,
it returns an error response which does not have `message`
field.

This error handler tries to log an error by accessing
`message` field which is a `nil`. Concatenating `nil` to a string
throws an exception in Lua. Using `tostring()` prevents that.


kdarkhan:
The root cause of my issue was fixed in lua-language-server https://github.com/sumneko/lua-language-server/issues/879

So this change is not necessary for my use case anymore, but you can still merge it if you find it useful.

Conni2461:
Yeah. I think we can skip it if the standard requires it to be not null

PR #1642, opened 2021-12-30
fix: allow find_command along with options

### LINKED ISSUE #1635: setup.pickers options do not work
### Description

This does not set the `hidden` option
```lua
require('telescope').setup{
  pickers = {
    find_files = {
      find_command = { "fd", "--type", "f", "--strip-cwd-prefix" },
      hidden = true,
    },
  },
}
```
This works
```lua
require('telescope').setup{
  pickers = {
    find_files = {
      find_command = { "fd", "--type", "f", "--strip-cwd-prefix", "--hidden" },
    },
  },
}
```
But the docs says that there's a `hidden` option we can set

https://github.com/nvim-telescope/telescope.nvim/blob/9d0034df79506af0ee0f102f1f5c6eec59e20e90/doc/telescope.txt#L745-L767

`:lua require"telescope.builtin".find_files()`

### Neovim version

```markdown
NVIM v0.7.0-dev+761-g0d7a97224
```

Arch linux

### checkhealth telescope

```markdown
  - OK: plenary installed.

  - OK: fd: found fd 8.3.0

## Telescope Extension: `git_worktree`

## Telescope Extension: `projects`
```

### Steps to reproduce

.

### Expected behavior

I expect with the `hidden` option shows the hidden files if true

### Actual behavior

The `hidden` option does not work.
The `find_command` option works

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

pedro757:
fix #1635

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -145,13 +145,13 @@ files.grep_string = function(opts)
   }):find()
 end
 
-local function which_find_command(opts)
   local command = opts.find_command
   if command then
     return command[1]
   end
   if 1 == vim.fn.executable "fd" then
     return "fd"
+local function get_find_command(opts)
   elseif 1 == vim.fn.executable "fdfind" then
     return "fdfind"
   elseif 1 == vim.fn.executable "rg" then
@@ -163,9 +163,9 @@ local function which_find_command(opts)
   end
 end
 
-local function find_command_options(opts)
   local command = which_find_command(opts)
   local find_command = opts.find_command
+local function get_find_command_options(opts)
   local hidden = opts.hidden
   local no_ignore = opts.no_ignore
   local follow = opts.follow
@@ -261,7 +261,7 @@ end
 -- TODO: Maybe just change this to `find`.
 -- TODO: Support `find` and maybe let people do other stuff with it as well.
 files.find_files = function(opts)
-  local find_command = find_command_options(opts)
+  local find_command = get_find_command_options(opts)
 
   if not find_command then
     print(
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -186,22 +186,9 @@ files.find_files = function(opts)
       table.insert(find_command, "-L")
     end
     if search_dirs then
-      table.insert(find_command, ".")
-      for _, v in pairs(search_dirs) do
-        table.insert(find_command, v)
+      if command ~= "rg" then
+        table.insert(find_command, ".")
       end
-    end
-  elseif command == "rg" then
-    if hidden then
-      table.insert(find_command, "--hidden")
-    end
-    if no_ignore then
-      table.insert(find_command, "--no-ignore")
-    end
-    if follow then
-      table.insert(find_command, "-L")
-    end
-    if search_dirs then
       for _, v in pairs(search_dirs) do
         table.insert(find_command, v)
       end
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -255,7 +255,6 @@ end
 -- TODO: Maybe just change this to `find`.
 -- TODO: Support `find` and maybe let people do other stuff with it as well.
 files.find_files = function(opts)
-  local find_command = opts.find_command
   local search_dirs = opts.search_dirs
 
   if search_dirs then
@@ -264,7 +263,7 @@ files.find_files = function(opts)
     end
   end
 
-  find_command = find_command_options(opts)
+  local find_command = find_command_options(opts)
 
   if not find_command then
     print(
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -175,23 +175,7 @@ local function get_find_command_options(opts)
     end
   end
 
-  if command == "fd" then
-    if hidden then
-      table.insert(find_command, "--hidden")
-    end
-    if no_ignore then
-      table.insert(find_command, "--no-ignore")
-    end
-    if follow then
-      table.insert(find_command, "-L")
-    end
-    if search_dirs then
-      table.insert(find_command, ".")
-      for _, v in pairs(search_dirs) do
-        table.insert(find_command, v)
-      end
-    end
-  elseif command == "fdfind" then
+  if command == "fd" or command == "fdfind" then
     if hidden then
       table.insert(find_command, "--hidden")
     end
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -163,6 +163,15 @@ files.find_files = function(opts)
       return { "where", "/r", ".", "*" }
     end
   end)()
+
+  if not find_command then
+    print(
+      "You need to install either find, fd, or rg. "
+        .. "You can also submit a PR to add support for another file finder :)"
+    )
+    return
+  end
+
   local command = find_command[1]
   local hidden = opts.hidden
   local no_ignore = opts.no_ignore
@@ -223,14 +232,6 @@ files.find_files = function(opts)
     if search_dirs ~= nil then
       log.warn "The `search_dirs` key is not available for the Windows `where` command in `find_files`."
     end
-  end
-
-  if not find_command then
-    print(
-      "You need to install either find, fd, or rg. "
-        .. "You can also submit a PR to add support for another file finder :)"
-    )
-    return
   end
 
   if opts.cwd then
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -175,7 +175,7 @@ files.find_files = function(opts)
     end
   end
 
-  if command == "fd" or command == "fdfind" then
+  if command == "fd" or command == "fdfind" or command == "rg" then
     if hidden then
       table.insert(find_command, "--hidden")
     end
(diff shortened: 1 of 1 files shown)


lua/telescope/builtin/files.lua
```diff
@@ -145,106 +145,126 @@ files.grep_string = function(opts)
...
 end
 
--- TODO: Maybe just change this to `find`.
--- TODO: Support `find` and maybe let people do other stuff with it as well.
-files.find_files = function(opts)
+local function which_find_command(opts)
+  local command = opts.find_command
+  if command then
+    return command[1]
+  end
```
pedro757, line 152: Is this a good idea?

Are there cases when the first element of the table is not the command?

Conni2461: What if someone implemented his own find implementation and uses this as `find_command`. Then we will return nil.

Conni2461: Then its not handled by the `find_command_options` if ?

lua/telescope/builtin/files.lua
```diff
@@ -145,106 +145,126 @@ files.grep_string = function(opts)
...
   local find_command = opts.find_command
   local hidden = opts.hidden
   local no_ignore = opts.no_ignore
   local follow = opts.follow
   local search_dirs = opts.search_dirs
 
-  if search_dirs then
-    for k, v in pairs(search_dirs) do
-      search_dirs[k] = vim.fn.expand(v)
+  if command == "fd" then
```
Conni2461, line 174: Cant we refactor this now ? if `command == "fd" or command == "fd_find" or command == 'rg'` all do the same if i am not wrong

lua/telescope/builtin/files.lua
```diff
@@ -145,106 +145,126 @@ files.grep_string = function(opts)
...
+    end
+    if search_dirs ~= nil then
+      log.warn "The `search_dirs` key is not available for the Windows `where` command in `find_files`."
+    end
   end
+  return find_command
+end
+
+-- TODO: Maybe just change this to `find`.
+-- TODO: Support `find` and maybe let people do other stuff with it as well.
```
Conni2461, line 256: Yeah no, why rename it? Scope of PR also screens no

lua/telescope/builtin/files.lua
```diff
@@ -145,106 +145,126 @@ files.grep_string = function(opts)
...
+  elseif 1 == vim.fn.executable "find" and vim.fn.has "win32" == 0 then
+    return "find"
+  elseif 1 == vim.fn.executable "where" then
+    return "where"
+  end
+end
+
+local function find_command_options(opts)
+  local command = which_find_command(opts)
   local find_command = opts.find_command
```
Conni2461, line 168: You need to make adjustments here. If `find_command` is not passed in its nil and shortly after you add doing `table.insert` thats bad

Conni2461: Or for me it doesnt even work. No config

lua/telescope/builtin/files.lua
```diff
@@ -145,106 +145,126 @@ files.grep_string = function(opts)
...
--- TODO: Maybe just change this to `find`.
--- TODO: Support `find` and maybe let people do other stuff with it as well.
-files.find_files = function(opts)
+local function which_find_command(opts)
+  local command = opts.find_command
+  if command then
+    return command[1]
+  end
+  if 1 == vim.fn.executable "fd" then
+    return "fd"
```
Conni2461, line 154: With only that we will go from `{ "fd", "--type", "f" }` to only `"fd"` which is something different.

lua/telescope/builtin/files.lua
pedro757: The `which_find_command` function is just to know the options to apply depending on which of these `fd, fdfind, rg, find or where` are using, if someone use a custom `find_command` generally `(idk if it's always the case)` the first element on that list will be the command

pedro757: if that custom command is not one of the mentioned, then it will bypass all the options `(all the if statements in find_command_options)` and return just the `find_command` the user set

lua/telescope/builtin/files.lua
pedro757: `fd` and `fdfind` apparently are the same, `rg` is diferent.

pedro757: But I could not say they will be the same in every option they have. I only know `fd`, I wont put my hands in the fire for the other commads, that's why I didn't want to change anything there.

lua/telescope/builtin/files.lua
pedro757: You're right

lua/telescope/builtin/files.lua
pedro757: I'm going to fix this

lua/telescope/builtin/files.lua
Conni2461: `fd` and `fdfind` are 100% the same tool. the command is just called `fdfind` on ubuntu vs `fd` on all other systems (arch, nix, ...)

Conni2461: Yeah i see that rg is slightly different. We could still pull it in like this

```lua
if command == "fd" or command == "fd_find" or command == "rg" then
  if hidden then
    table.insert(find_command, "--hidden")
  end
  if no_ignore then
    table.insert(find_command, "--no-ignore")
  end
  if follow then
    table.insert(find_command, "-L")
  end
  if search_dirs then
    if command ~= "rg" then
      table.insert(find_command, ".")
    end
    for _, v in pairs(search_dirs) do
      table.insert(find_command, v)
    end
  end
else if ...
```
cant we?

lua/telescope/builtin/files.lua
```diff
@@ -145,10 +145,25 @@ files.grep_string = function(opts)
...
+  elseif 1 == vim.fn.executable "rg" then
+    return { "rg", "--files" }
+  elseif 1 == vim.fn.executable "find" and vim.fn.has "win32" == 0 then
+    return { "find", ".", "-type", "f" }
+  elseif 1 == vim.fn.executable "where" then
+    return { "where", "/r", ".", "*" }
+  end
+end
+
+local function get_find_command_options(opts)
```
Conni2461, line 164: can we not make a new function for this ?

lua/telescope/builtin/files.lua
pedro757: Haha, I just saw that [here](https://github.com/sharkdp/fd#on-ubuntu), let me fix that.

lua/telescope/builtin/files.lua
```diff
@@ -145,10 +145,25 @@ files.grep_string = function(opts)
   }):find()
 end
 
--- TODO: Maybe just change this to `find`.
--- TODO: Support `find` and maybe let people do other stuff with it as well.
-files.find_files = function(opts)
-  local find_command = opts.find_command
+local function get_find_command(opts)
```
Conni2461, line 148: either move this local function inside `builtin.find_files` or do something like this

```lua
local find_command = (function()
  if opts.find_command then
    return opts.find_command
  elseif 1 == vim.fn.executable "fd" then
    return { "fd", "--type", "f" }
  elseif 1 == vim.fn.executable "fdfind" then
    return { "fdfind", "--type", "f" }
  elseif 1 == vim.fn.executable "rg" then
    return { "rg", "--files" }
  elseif 1 == vim.fn.executable "find" and vim.fn.has "win32" == 0 then
    return { "find", ".", "-type", "f" }
  elseif 1 == vim.fn.executable "where" then
    return { "where", "/r", ".", "*" }
  end
end)()
```

There is no need to have a local function that is only used once

lua/telescope/builtin/files.lua
Conni2461: yep

pedro757:
Should I writte docs? I don't know how to use the treesitter docgen showed in the [CONRTIBUTING](https://github.com/nvim-telescope/telescope.nvim/blob/master/CONTRIBUTING.md) guide

The [doc/telescope.txt](https://github.com/nvim-telescope/telescope.nvim/blob/948f5adff28928d21207035035c0827c9ad92cbc/doc/telescope.txt#L754) looks good to me

lua/telescope/builtin/files.lua
```diff
@@ -148,7 +148,22 @@ end
...
+      return { "fdfind", "--type", "f" }
+    elseif 1 == vim.fn.executable "rg" then
+      return { "rg", "--files" }
+    elseif 1 == vim.fn.executable "find" and vim.fn.has "win32" == 0 then
+      return { "find", ".", "-type", "f" }
+    elseif 1 == vim.fn.executable "where" then
+      return { "where", "/r", ".", "*" }
+    end
+  end)()
+  local command = find_command[1]
```
Conni2461, line 166: One last thing i just noticed while doing the final review. If we never find a command `find_command` is nil which would cause a error here. 

In line 228 is the check. We need to do that check now earlier before we actually do anything with `find_command`

Conni2461:
No i think you dont need to write docs

lua/telescope/builtin/files.lua
pedro757: done

Conni2461 (approved):
LGTM dont think it has any regression. I tested a couple of configurations.

Thanks :)

PR #1644, opened 2021-12-31
Fix documentation for lsp_range_code_actions

OleStrohm:
The current documentation accidentally used start_line for both fields, making the documentation both incomplete and confusing.
Renaming the second field will bring this in line with the intended documentation, as well as the behaviour of the function.

Hopefully this is the correct file to change, as the docs/telescope.txt seems to be generated from this.


Conni2461:
Thanks :)

PR #1646, opened 2022-01-02
fix: picker for keymaps error with new api for keymaps

seblj:
With the new API for keymaps, the picker errors on trying to concatinating the `rhs` if a Lua function is used. This fixes the error and also uses the `desc` opt if it is specified.


Conni2461:
Thanks :)

PR #1649, opened 2022-01-03
reload buffers and syntax-hi after git_reset

xnorly:
After calling the git_reset function from telescope,
- the buffers do not refresh and do not show the changes that are invoked by the git_reset command. Therefore, ```vim.cmd("bufdo e")``` was added such that buffers are reloading
- ```bufdo e``` will reload all open buffers, but may end up viewing another buffer (and not, as expected, the one that was previously viewed). Therefore, ```vim.cmd("b" ..currentBuff)``` was added to restore the view to the currentBuffer.
- ```bufdo e``` may also remove syntax highlighting (for reasons of speed). Therefore ```vim.cmd("syntax on")``` was added.


xnorly:
This may also be used in other telescope git_functions. For example: ```git_switch_branch```

Issue #1579, opened 2021-12-09
C-W stopped working in the prompt


Gelio:
### Description

`<C-W>` in insert mode is often used to remove the current word. It used to work in the Telescope picker. I recently noticed it stopped working and instead it seems to work as-if it was not in Insert mode (invoking the regular `<C-W>` from normal mode).

### Neovim version

```markdown
NVIM v0.7.0-dev+697-g6063e07a0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc-11 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/runner/work/neovim/neovim/build/config -I/home/runner/work/neovim/neovim/src -I/home/runner/work/neovim/neovim/.deps/usr/include -I/usr/include -I/home/runner/work/neovim/neovim/build/src/nvim/auto -I/home/runner/work/neovim/neovim/build/include
Compiled by runner@fv-az32-430
```

### Operating system and version

Ubuntu 20.04 LTS

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - WARNING: nvim-treesitter not found.

## Checking external dependencies
  - OK: rg: found ripgrep 13.0.0
  - OK: fd: found fd 7.4.0
```

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. `:Telescope live_grep`
3. Type in any text in the prompt (e.g. `pack`)
4. Press C-W in insert mode

### Expected behavior

The word is removed and the picker is empty. Neovim remains in insert mode.

### Actual behavior

The cursor changes to a block. The word is not removed. It looks as if `<C-W>` from normal mode was executed (the window manipulation prefix).

[...]

PR #1650, opened 2022-01-03
feat: Add <C-w> to default mappings

davegallant:
As described in https://github.com/nvim-telescope/telescope.nvim/issues/1579, this adds `<C-w>` to the default mappings.

### DIFF
--- a/lua/telescope/mappings.lua
+++ b/lua/telescope/mappings.lua
@@ -34,6 +34,7 @@ mappings.default_mappings = config.values.default_mappings
       ["<M-q>"] = actions.send_selected_to_qflist + actions.open_qflist,
       ["<C-l>"] = actions.complete_tag,
       ["<C-_>"] = actions.which_key, -- keys from pressing <C-/>
+      ["<C-w>"] = { "<c-s-w>", type = "command" },
     },
 
     n = {


davegallant:
@tjdevries I've added this to the default mapping as described in https://github.com/nvim-telescope/telescope.nvim/issues/1579#issuecomment-1000969628.

Looks like this needs workflow approval.

lua/telescope/actions/init.lua
```diff
@@ -846,6 +846,12 @@ actions.remove_selected_picker = function(prompt_bufnr)
   end
 end
 
+--- Delete the word before the cursor.
+---@param prompt_bufnr number: The prompt bufnr
+actions.delete_word = function(prompt_bufnr)
+  vim.api.nvim_input "<c-s-w>"
```
tjdevries, line 852: The only thing might be that we should make sure we're in insert mode here (otherwise I don't think it would do anything, right?)

Conni2461: Can't we just add this mapping ?

```lua
        ["<c-w>"] = { "<c-s-w>", type = "command" },
```

The only question i have is what happens with 0.6.1

davegallant: Removed the action. You're right. It doesn't seem necessary.

As for 0.6.1, I'm not sure I have enough context to understand.

Conni2461: I meant neovim release 0.6.1. dont worry i just tested it

Conni2461 (approved):
Thanks :)

PR #1653, opened 2022-01-04
Xnorly buffer reload

xnorly:
After undoing commits with telescope's ```git_reset_branch``` function, the buffers do not reload and still show the old state. Therefore I added a ```restore_buffers``` function in actions/utils.lua, that does exactly that:

- reloading all the buffers with ```vim.cmd("bufdo e")```
- turn syntax on again with ```vim.cmd("syntax on")``` (because ```bufdo``` may turn off syntax to increase speed)
- show previously shown buffer with ```vim.cmd("b" .. currentBuff)``` (because with multiple buffers opened, ```bufdo``` will change the view and not return back to the previously shown buffer)

The ```restore_buffers``` function could also be used in other git functions. Which is why I added it to the utils module.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -550,6 +550,8 @@ local git_reset_branch = function(prompt_bufnr, mode)
   else
     print(string.format('Error when resetting to: %s. Git returned: "%s"', selection.value, table.concat(stderr, "  ")))
   end
+
+  action_utils.restore_buffers()
 end
 
 --- Reset to selected git commit using mixed mode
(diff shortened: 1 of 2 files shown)


tami5:
Interesting fix. I have auto-read enabled so when the file on disk changes the buffer gets  updated but this seems to be better way to avoid some errors, in particular, I'm interested to know what will happen in the following cases:

- buffers open but not saved yet.
- buffers open but no longer exists after a git command

Additionally, what other function do u think we should have this in.

Thanks for the detailed explanation

xnorly:
Hi tami5
Thanks for your answer.

I have no clue yet about how to treat the situations you mentioned.

Here are the other functions I think could make usage of the **restore_buffer** function:

JoseEscobarBendezu:
Any update ?

PR #1684, opened 2022-01-13
feat: make results pane of builtin `keymaps` more readable

daangoossens22:
Updated the result pane of the builtin picker `keymaps` to be more coherent with some other pickers by using the `|` to seperate information for each entry (e.g. like `commands`, `quickfix`). The width of the middle column is adjusted dynamically so the lhs of the keymap doesn't get cut off.

Before:
before

After:
after

Some feedback if everything is ok, would be nice, as I'm still a bit new to lua.

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -955,30 +955,27 @@ internal.keymaps = function(opts)
   local modes = { "n", "i", "c" }
   local keymaps_table = {}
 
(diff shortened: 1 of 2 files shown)


Conni2461 (approved):
Thanks :)

Conni2461:
Can you fix the formatting issue?

daangoossens22:
Done

PR #1688, opened 2022-01-14
add offset encoding to vim.lsp.util.jump_to_location() call

p00f:


p00f:
earlier the function itself had one argument so this should not break anything

Conni2461:
Thanks :)

PR #1690, opened 2022-01-15
feat: add optional plugin picker

kuntau:
Make it easier to load optional plugin (`packadd`)


kuntau:
Uff.. first timer here.

> Auto-updates from CI

I need help with this

Conni2461:
We are currently not looking for new builtins

kuntau:
> We are currently not looking for new builtins

May I know why? This is just a simple vim command that I use all the time. It's much more convenient to have picker for them.

fdschmidt93:
As to why, much of is discussed here: https://github.com/nvim-telescope/telescope.nvim/issues/1228

kuntau:
> You could just create an extension for this I think

Sure, but it's just too small in functionality and code to be worth as extension IMO

max397574:
Then it would be a small extension ig

Conni2461:
I am sorry but i have to deny this PR. You should put this picker in your local config if it benefits you or create a small extension.

Still thanks for the PR :) and i am very sorry.

PR #1691, opened 2022-01-15
Update doc to remove file browser

adalessa:
On the PR #1453 the `file_browser` has been remove from the builtint, it was remove from almost all places but the documentation, so it still part of the help when checked from neovim.


Conni2461:
Thanks :)

PR #1695, opened 2022-01-17
feat(rg): add glob and type filter opts by default

kylo252:
- `glob_pattern` string: argument to be used with `--glob`,
e.g. "*.toml", can use the opposite "!*.toml"
- `type_filter` string: argument to be used with `--type`,
e.g. "rust", see `rg --type-list`

```vim
:Telescope live_grep glob_pattern=!*.txt
:Telescope live_grep type_filter=lua
:Telescope live_grep type_filter=make
```

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -764,6 +764,11 @@ builtin.live_grep({opts})                      *telescope.builtin.live_grep()*
         {search_dirs}         (table)     directory/directories to search in,
                                           mutually exclusive with
                                           `grep_open_files`
+        {glob_pattern}        (string)    argument to be used with `--glob`,
+                                          e.g. "*.toml", can use the opposite
+                                          "!*.toml"
+        {type_filter}         (string)    argument to be used with `--type`,
+                                          e.g. "rust", see `rg --type-list`
         {additional_args}     (function)  function(opts) which returns a table
                                           of additional arguments to be passed
                                           on
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -72,6 +72,14 @@ files.live_grep = function(opts)
     additional_args = opts.additional_args(opts)
   end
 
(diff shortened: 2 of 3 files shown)


kylo252:
Since this is a commonly asked for feature, I figured I could add these opts to make it a bit more accessible.

I know that it can be achieved by either modifying `vimgrep_arguments` directly or using the unfortunately-named `additional_args`  function, but I felt that pre-filtering opts deserved special treatment because of how frequently it is needed.

antosha417:
Do you think it is possible to be able to specify glob_pattern or type_filter after I entered search query and saw that there are too many results?

Right now what I have to do is:
1. Call live grep 
2. Enter some search pattern.
3. Get upset because I see too many results and exit live grep.
4. Call live grep with the right type_filter
5. Enter same search pattern as before and be happy with the results.

I want to be able to do something like this:
1. Call live grep 
2. Enter some search pattern.
3. See too many results and enter type_filter and be happy with the results.

antosha417:
I figured how to do it. 
```lua

actions.glob_filter = function(prompt_bufnr)
  local glob = vim.fn.input("Enter glob: ")
  vim.api.nvim_command("normal :esc<CR>")
  if not glob or #glob == 0 then
    return
  end
  local promt = picker:_get_prompt()
  actions.close(prompt_bufnr)
  vim.schedule(function ()
      default_text = promt,
      additional_args = function () return { '-g', glob} end
    }
  end)
end
```
And than map this to some shortcut
```lua 
require('telescope').setup {
  pickers = {
    live_grep = {
      mappings = { i = { ["<c-r>"] = "glob_filter" } }
    }
  }
}

```

Conni2461:
We can do these shorthands. They can benefit a lot of people.

Thanks for the PR :)

weeman1337:
To unleash the power of live grep there is also https://github.com/nvim-telescope/telescope-rg.nvim :smile:

PR #1698, opened 2022-01-18
feat: show a message for unknown command

zeertzjq:
For a week I've been typing `:Telescope lsp_reference` instead of `:Telescope lsp_references` and wondering why it stopped working, then I found out that typo.


Conni2461:
Thanks :)

Issue #1700, opened 2022-01-18
Quickfix escaped spaces in paths on Windows


jonasw234:
### Description

On Windows when using the `:Telescope find_files` command and using the `<C-q>` mapping to send items to the quickfix list, spaces in filenames are escaped with `\` (e.g. `filename\ with\ spaces.txt`). However, this makes the quickfix list entries unusable, because on Windows this escaping is not only unnecessary, the entries get interpreted as going into a different folder (i.e. the actual filename would be ` spaces.txt` inside the `filename\ with` directory).

### Neovim version

```markdown
NVIM v0.6.1
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

### Operating system and version

Windows 10

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - OK: nvim-treesitter installed.

## Checking external dependencies
  - OK: rg: found ripgrep 13.0.0 (rev af6b6c543b)
  - OK: fd: found fd 8.3.1

## ===== Installed extensions =====

## Telescope Extension: `coc`
  - INFO: No healthcheck provided

## Telescope Extension: `fzf`
  - INFO: No healthcheck provided
```

### Steps to reproduce

1. Switch to directory where files with spaces in the filename exist
2. `nvim -nu minimal.lua`
3. :Telescope find_files<CR>
4. <C-q>

### Expected behavior

Quickfix list should open and contain entries like:
```
.\filename with spaces.txt|1 col 1| .\filename with spaces.txt
.\another filename with spaces.txt|1 col 1| .\another filename with spaces.txt
```

### Actual behavior

Quickfix list opens and contains entries like:
```
.\filename\ with\ spaces.txt|1 col 1| .\filename with spaces.txt
.\another\ filename\ with\ spaces.txt|1 col 1| .\another filename with spaces.txt
```

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
[...]

PR #1712, opened 2022-01-24
Skip path escaping when adding items to quickfix list

dimfeld:
# Problem

Items added to the quickfix list should not have their paths escaped because then Neovim can't open them.

My specific case is paths such as `src/routes/api/trackables/[trackable_id]/index.ts`, where the brackets would be escaped when adding to the quickfix list. 

I think #1700 is the same problem but I don't have a Windows box to test on so can't be 100% sure. I did also test this fix on MacOS with a filename containing spaces and it works fine.

# Reproduction

1. Create a file with some escapable character such as a space or `[`.
2. Open neovim in that directory, use a telescope picker to show the file in the list, and hit `<c-q>`.
3. Notice the extra backslashes in the escaped filename in the quickfix list, and that trying to open the file from the quickfix list fails. Neovim tries to open the path containing the backslash instead of the actual path.

# Solution

This PR adds a new parameter `escape` to `from_entry.path`. It defaults to `true` if omitted, but if `false` then the call to `vim.fn.fnameescape` is skipped.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -623,7 +623,7 @@ local entry_to_qf = function(entry)
 
   return {
     bufnr = entry.bufnr,
-    filename = from_entry.path(entry, false),
+    filename = from_entry.path(entry, false, false),
     lnum = vim.F.if_nil(entry.lnum, 1),
     col = vim.F.if_nil(entry.col, 1),
     text = text,
--- a/lua/telescope/from_entry.lua
+++ b/lua/telescope/from_entry.lua
@@ -10,8 +10,15 @@ This will provide standard mechanism for accessing information from an entry.
 
 local from_entry = {}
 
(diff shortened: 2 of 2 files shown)


dimfeld:
Ah I see #1306 is already open. Feel free to close this if you prefer that one.

Conni2461:
Yeah i acutally have 0 idea whats happening in the other PR. Was kinda abounded and I didnt understand what was missing

See: https://github.com/nvim-telescope/telescope.nvim/pull/1306#discussion_r735074527

NICHTJ3:
Just started running into this in a next project. Would definitely appreciate an update 😂

Conni2461:
I didn't wanna do a type string comparison every time the function gets called so we should only set a default if its nil. I updated this.

I haven't gotten an answer from the other PR in over a month so i merge your now. Thanks :)

PR #1717, opened 2022-01-26
fix: picker `autocommands` doesn't show all autocommands

daangoossens22:
When using this picker I noticed that some autocommands where not showing. From my own setup, calling `:verbose au *` retrieves 4 kinds of metadata (from most common to least common):
1. `Last set from {filepath} line {linenr}`
2. `Last set from anonymous :source`
3. `Last set from anonymous :source line {linenr}`
4. `Last set from {filepath}`

Case 2 and 4 are now also picked up and defaulted to the first line in the document (even tho the `anonymous :source` one doesn't go anywhere)

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -1096,7 +1096,7 @@ internal.autocommands = function(opts)
     end
 
     if current_ft and cmd then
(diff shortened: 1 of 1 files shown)


Conni2461 (approved):
LGTM thanks :)

PR #1718, opened 2022-01-26
add offset encoding

p00f:


p00f:
I think this is the last one

Conni2461:
:rofl: 

We'll see. Thanks :)

PR #1719, opened 2022-01-26
Change center layout strategy to allow vertical anchoring

alefpereira:
Affects `telescope.pickers.layout_strategies.center` that is used by dropdown (`telescope.themes.get_dropdown`) theme, allowing more customization for dropdown theme, i. e. setting it in the top with mirrored preview.

Some may argue that it is already possible to do that with vertical layout, although vertical layout uses fixed height, so toggling preview squeezes the result prompt, while center layout has dynamic height and shows the preview without squeezing the result prompt.

[Example of the initial proposal](https://i.imgur.com/9LHjKT3.gif) (obsolete)

EDIT: 

As discussed (see comments), beyond changing the center layout to allow vertical anchoring, I updated the change to make preview window fill the rest of the screen (which is the original purpose of center layout) and also changed ´anchor´ to override `mirror` behavior as follows:

Vertical anchoring, i.e. `anchor` containing `"N"` or `"S"` will ignore `mirror` config. For `"N"` anchoring preview will be placed below prompt result block. For `"S"` anchoring preview will be placed above prompt result block. For horizontal only anchoring preview will be placed according to `mirror` config, default is above the prompt result block.

horizontalanchoring

### DIFF
--- a/lua/telescope/pickers/layout_strategies.lua
+++ b/lua/telescope/pickers/layout_strategies.lua
@@ -395,11 +395,18 @@ layout_strategies.horizontal = make_documented_layout(
 
 --- Centered layout with a combined block of the prompt
 --- and results aligned to the middle of the screen.
---- The preview window is then placed in the remaining space above.
+--- The preview window is then placed in the remaining
+--- space above or below, according to `anchor` or `mirror`.
 --- Particularly useful for creating dropdown menus
---- (see |telescope.themes| and |themes.get_dropdown()|`).
+--- (see |telescope.themes| and |themes.get_dropdown()|).
 ---
---- Note that the `anchor` option can only pin this layout to the left or right edges.
+--- Note that vertical anchoring, i.e. `anchor` containing
+--- `"N"` or `"S"`, will override `mirror` config. For `"N"`
+--- anchoring preview will be placed below prompt/result
+--- block. For `"S"` anchoring preview will be placed above
+--- prompt/result block. For horizontal only anchoring preview
+--- will be placed according to `mirror` config, default is
+--- above the prompt/result block.
 ---
 --- <pre>
 --- ┌──────────────────────────────────────────────────┐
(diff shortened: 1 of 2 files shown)


l-kershaw:
Hi @alefpereira, thanks for the PR.

Unfortunately this change means that the pickers are not laid out in a reasonable way for the layout config like `{anchor="S",mirror=true}` or `{anchor="N",mirror=false}`.
This is the reason that I decided not to allow vertical anchoring for this layout when I implemented it.

This problem is caused by the fact that for the `center` layout strategy, the `height` option only defines the height of the "central block" (the prompt and results) and the remaining vertical space (either above or below) is taken up by the preview (if present).

I couldn't think of a way around this issue that wouldn't involve making configuration of the center layout considerably more confusing to users, so decided it would be best to just disable this sort of anchoring.
If you have any ideas on how to do this, then feel free to update this PR and ping me if you want feedback 🙂

alefpereira:
Hello @l-kershaw, thank you for your attention.

I saw those problems when I was testing. Thinking about it, an option to "padding" preview height seems to be the simplest solution, and a good option to have, but I personally don't think it would be ideal (not the prettier one in this case). Also, I don't know if its possible/viable to resize/redraw the telescope window on toggling preview, but this would be an alternative, for example  with `{anchor="N",mirror=false}`, toggling preview off would shift "central block" up, and toggling preview on again,  shift "center block" down to give space for preview. I'll go more deep in the code to know what I can do, my knowledge is still very superficial, so let me know if those options are viable or already exist.

Although, what do you think about allowing center layout to accept vertical anchoring (or adding an option to it like ´layout_config.allow_vertical_anchoring´ with `false` as default) and changing the dropdown theme to choose the `mirror` according to the `anchor`? For example, for `anchor="S"`, set `mirror=true` like below, and for `anchor="N"` keep set `mirror=false` like in the PR description example (users could still override `mirror` of course). I think this would give more flexibility to users to create more customized dropdown layouts. If you think it is a good idea I can work on that.

dropdown-bottom

alefpereira:
Sure, it allows vertical anchoring for `center` strategy while still keeps its configs simple. I'll work on that, as soon I have something I'll push here. Thanks @l-kershaw.

alefpereira:
Dear @l-kershaw , after some studying on center layout, I wrote with this solution. Core of the change was to adjust preview line and height only after anchor adjustment on the prompt and result block, in this way I could set preview line and height values relative to the final position of the "central block" and to the "border" of the screen.

I experimented on a lot of cases and as far as I could validate, It is behaving like we discussed for horizontal anchoring, and it keeps the same behavior as before for vertical/central anchoring. If you could take a look to check I'm not missing anything I'll appreciate, I'm always open to suggestions.

Some notes:
- Cap over/undersized height/width are working the same.
- Tabline shift is working the same.
- Preview cutoff working normally.
- Also changed the the layout docs.

horizontalanchoring

alefpereira:
Fixed some typos on my change. Now I think this is it.

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -395,11 +395,18 @@ layout_strategies.horizontal = make_documented_layout(
...
+--- space above or below, according to `anchor` or `mirror`.
 --- Particularly useful for creating dropdown menus
 --- (see |telescope.themes| and |themes.get_dropdown()|`).
 ---
---- Note that the `anchor` option can only pin this layout to the left or right edges.
+--- Note that vertical anchoring, i.e. `anchor` containing
+--- `"N"` or `"S"` will ignore `mirror` config. For `"N"`
+--- anchoring preview will be placed below prompt result
+--- block. For `"S"` anchoring preview will be placed above
+--- prompt result block. For horizontal only anchoring preview
```
l-kershaw, line 407: I would say "prompt/result block" rather than "prompt result block".

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -476,25 +483,38 @@ layout_strategies.center = make_documented_layout(
...
+      mirror = layout_config.mirror
+    end
+
+    -- Set preview position
+    local block_line = math.min(results.line, prompt.line)
+    if not mirror then -- Preview at top
+      preview.line = 1 + bs
+      preview.height = block_line - (2 + 2*bs)
+    else -- Preview at bottom
+      preview.line = block_line + results.height + 2 + 2*bs
```
l-kershaw, line 510: Stylua is complaining about the `2*bs` here, so should add spaces around the `*`

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -395,11 +395,18 @@ layout_strategies.horizontal = make_documented_layout(
...
 --- and results aligned to the middle of the screen.
---- The preview window is then placed in the remaining space above.
+--- The preview window is then placed in the remaining
+--- space above or below, according to `anchor` or `mirror`.
 --- Particularly useful for creating dropdown menus
 --- (see |telescope.themes| and |themes.get_dropdown()|`).
 ---
---- Note that the `anchor` option can only pin this layout to the left or right edges.
+--- Note that vertical anchoring, i.e. `anchor` containing
+--- `"N"` or `"S"` will ignore `mirror` config. For `"N"`
```
l-kershaw, line 404: "override" seems better than "ignore" here

l-kershaw (commented):
This seems like a good implementation overall, thanks 🙂 

---

Just a couple of small changes needed, as mentioned.

Also, we need to make sure that docs are automatically generated by the changed comments, as they don't seem to have been so far. This might require you changing some settings for actions on your fork.

alefpereira:
@l-kershaw you're welcome. Thank your for your feedback. I just fixed your comments, and activated docgen action, it is generating the docs now.

Additionally, if its ok, I took the opportunity to removed the alone backtick (`) at the end of this line from the center layout docs:
```
--- (see |telescope.themes| and |themes.get_dropdown()|`).
                                                       ^
```

l-kershaw (approved):
Apologies for the delayed review.

This looks good to me, merging now.
Thanks for the hard work 😄 (and patience 😅)

PR #1618, opened 2021-12-21
fix(actions): force insertmode after inserting symbol

clason:
required after upstream changes to prompt buffer behavior (see https://github.com/neovim/neovim/commit/28134f4e78819c2bbf0344326b9d44f21eb0d736)

Fixes #1608

(I doubt anyone besides me has ever used this action... @Conni2461)

PR #1722, opened 2022-01-27
fix(action): insert symbol in insert mode

clason:
followup to #1618, ensure insert after cursor

(turns out bang means `A`, not `a`. classic reading fail 🤦🏻 )


Conni2461:
makes sense. thanks

PR #1723, opened 2022-01-27
add filemask action for live_grep picker

antosha417:
When I use live_grep in a huge repo sometimes there are too many results to scroll through. This is when filemask filters like `some/path/*/*.{lua,vim}` come in handy. 
The only problem is that I can see that there are too many results only after I typed a search query. So I need to close live_grep, open it again with a right filemask and then retype my query. 
I want to be able to call `filemask_filter` action to filter out results I don't need in live_grep

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -9,6 +9,7 @@
 
 local a = vim.api
 
+local builtin = require "telescope.builin"
 local config = require "telescope.config"
 local state = require "telescope.state"
 local utils = require "telescope.utils"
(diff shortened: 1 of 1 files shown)


antosha417:
I also have an idea to be able to filter out results by file type. `rg` have builtin file types that can be passed to telescope. But I couldn't figure out a proper way to save prompt content for live_grep when another telescope picker is opened

Conni2461:
Not sure if we want this action in core. Could just be a wiki entry somewhere here: https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes

aidos:
Thanks @antosha417 for the idea on how to do this. I've been looking myself for a way to better control `rg`. I was wondering if it made sense to allow live_grep to parse the command as arguments instead of as the search text (say if the prompt begins with `--`).

So then we could just use `--type python -- my search here` (or any other `rg` flags we wanted)

aidos:
Ah ha, just found this [PR](https://github.com/nvim-telescope/telescope.nvim/pull/670) and [this offshoot extension](https://github.com/nvim-telescope/telescope-rg.nvim) which is exactly the idea I was looking for above. Will point my attention that way instead.

Conni2461:
https://github.com/nvim-telescope/telescope-rg.nvim is a good solution that does this in the prompt.

Closing this as this can be done from user side

PR #1706, opened 2022-01-21
fix: add opts.bufnr and opts.winnr to builtin picker

jamestrew:
When buffer/window specific pickers are selected from the `:Telescope builtin` picker, the current buffer & window becomes Telescope itself making many of the pickers practically useless.

This PR sets `opts.bufnr` and `opts.winnr` when `:Telescope builtin` is ran. Buffer and window sensitive pickers can then reference those values when called via the `builtin` picker (if not set, use the current bufnr/winnr via `vim.api.nvim_get_current_buf/win()`).

closes #1678

Considering the builtin picker is probably where many new users try out the various builtin pickers, I think getting all the pickers functional from within the builtin picker is quite important.

Fixes or tweaks to allow for fix:
 - `files.treesitter`
 - `files.current_buffer_fuzzy_find`
 - `files.current_buffer_tags`
 - `git.bcommits`
 - `lsp.references`
 - `lsp.definitions`
 - `lsp.type_definitions`
 - `lsp.implementations`
 - `lsp.document_symbols`
 - `lsp.code_actions`
 - `lsp.range_code_actions`
 - `lsp.workspace_symbols`
 - `lsp.dynamic_workspace_symbols`
 - `diagnostics.get`

I think I got them all.

TODO: 
- [x] `current_buffer_ctags` still needs work in its `make_entry`
 - [x] double check and test all changes
 - [x] clean up

PR #1731, opened 2022-02-01
refactor: switch `vim.lsp.diagnostic` with `vim.diagnostic`

jamestrew:
As discussed in #1706, replaced the deprecated `vim.lsp.diagnostic.get_line_diagnostics` with `vim.diagnostic.get`.

Not sure why it required some rearranging of the params but this does what the LSP specs for code action needs.

Pending merging of #1706

### DIFF
--- a/lua/telescope/builtin/lsp.lua
+++ b/lua/telescope/builtin/lsp.lua
@@ -146,11 +146,36 @@ lsp.document_symbols = function(opts)
   end)
 end
 
+local function get_code_action_diagnostics(bufnr, lnum)
+  local diag = vim.diagnostic.get(bufnr, { lnum = lnum - 1 })[1]
+  -- reshape according to
+  -- https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#diagnostic
+  diag = {
+    range = {
+      ["end"] = {
+        character = diag.end_col,
+        line = diag.lnum,
+      },
+      start = {
+        character = diag.col,
+        line = diag.lnum,
+      },
+    },
+    severity = diag.severity,
+    code = diag.user_data.lsp.code,
+    source = diag.source,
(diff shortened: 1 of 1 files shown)


Conni2461:
Hmm thinking about this again. neovim also still uses the same function for code_actions

https://github.com/neovim/neovim/blob/1bd6c0a05c18e10e9439b4090e6f57f0d1e1fc21/runtime/lua/vim/lsp/buf.lua#L597-L599

Question is how long until the deprecated code will be removed. Maybe neovim could expose this function 

https://github.com/neovim/neovim/blob/1bd6c0a05c18e10e9439b4090e6f57f0d1e1fc21/runtime/lua/vim/lsp/diagnostic.lua#L120-L139

Or we should really remove `lsp_code_action` from telescope and suggest that people should use `vim.lsp.buf.code_action()` with `ui.select`

jamestrew:
I wasn't aware neovim core was still using the deprecated function. I can look into creating a PR/RFC to expose that private function.

I don't know how popular `lsp_code_action` is so I wonder how disruptive removing that picker will be to the average user.
Personally, I use it fairly frequently with the `get_cursor` theme so for me it's a better solution than if I were to use `ui.select`. Me knowing a little bit about telescope, I could always create a custom picker to preserve `lsp_code_action` if it were to be removed. But again, I don't know how most other telescope users would feel.

Conni2461:
So `vim.lsp.buf.code_action()` uses `ui.select` which can be overriden to use telescope: https://github.com/nvim-telescope/telescope-ui-select.nvim

So in the end it will still be the telescope ui

jamestrew:
Oh yeah I almost forgot that extension.
I can close this PR for now. If the private neovim function gets exposed, we can use that or maybe this picker gets removed.

Issue #1727, opened 2022-01-29
Picker autocommands throws error `cursor position outside buffer`


daangoossens22:
### Description

When changing previewers for the autocommands picker, it sometimes will throw the error `Error executing vim.schedule lua callback: ...scope.nvim/lua/telescope/previewers/buffer_previewer.lua:933: Cursor position outside buffer`. Looking at the source code this error is thrown when setting the cursor position in the previewer window.

Observations:
- It usually happens when changing the contents of the previewer
- Mostly happens when changing to a previewer with a lot of lines in it and immediately selecting the lines further at the bottom of the previewer. (e.g. augroup filetypedetect)
- When a previewer of an augroup has already been shown before, it doesn't happen as often (e.g. selecting the first line of augroup filetypedetect and then selecting the last line of filetypedetect doesn't throw an error while selecting the last line immediately does throw it)

Maybe related to: #957

### Neovim version

```markdown
NVIM v0.7.0-dev+963-g082ff2190
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc-11 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -DNVIM_TS_HAS_SET_ALLOCATOR -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/runner/work/neovim/neovim/build/config -I/home/runner/work/neovim/neovim/src -I/home/runner/work/neovim/neovim/.deps/usr/include -I/usr/include -I/home/runner/work/neovim/neovim/build/src/nvim/auto -I/home/runner/work/neovim/neovim/build/include
Compiled by runner@fv-az32-759

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/share/nvim"

Run :checkhealth for more info
```

### Operating system and version

[...]

PR #1732, opened 2022-02-01
fix: builtin `autocommands` throws error

daangoossens22:
Resolves: #1727

The issue seems to be that this snippet in `new_buffer_previewer` is sometimes scheduled after setting the cursor in the `define_preview` function.

https://github.com/nvim-telescope/telescope.nvim/blob/f262e7d56d37625613c5de0df5a933cccacf13c5/lua/telescope/previewers/buffer_previewer.lua#L366-L370

This will result in an error when going from a small preview buffer to a large preview buffer and trying to set the cursor at the end of the larger buffer, because it will try to set the cursor in the smaller buffer.

I tried to quickly replicate the issue with other previewers using `nvim_win_set_cursor`, but wasn't able to do so, so I didn't change anything to them.

But since this is kinda an easy mistake to make and hard to debug, maybe the `define_preview` function call in `new_buffer_previewer` needs to be scheduled instead or maybe add an util function around `nvim_win_set_cursor` or add something in the docs to be carefull/schedule `nvim_win_set_cursor`. @Conni2461 What do you think?

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -930,7 +930,9 @@ previewers.autocommands = defaulter(function(_)
       end
 
       vim.api.nvim_buf_add_highlight(self.state.bufnr, ns_previewer, "TelescopePreviewLine", selected_row + 1, 0, -1)
-      vim.api.nvim_win_set_cursor(status.preview_win, { selected_row, 0 })
+      vim.schedule(function()
+        vim.api.nvim_win_set_cursor(status.preview_win, { selected_row, 0 })
+      end)
 
       self.state.last_set_bufnr = self.state.bufnr
     end,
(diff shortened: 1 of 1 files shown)


tami5:
Does this issue still happen? On my system I can't reproduce it

daangoossens22:
@tami5 Hmm that is weird. The issue still persists for me on both the stable and nightly builds of neovim, using the minimal config and steps to reproduce in #1727. Are you sure when trying the reproduce the issue that the terminal is large enough so that the preview window is showing when cycling through entries.

Conni2461:
We should document it but i would just merge anyway right now. Changes look good to me.

It doesn't happen with the other previewers because they call the highlight function on return of results from e.g. `uv.read_file` which needs to be scheduled anyway.

Thanks for debugging :)

Issue #1296, opened 2021-09-29
option for wrap file path 


grimgort:
Hello,
In general, it is not possible to display the absolute path of the files because they overflow the window. Is it possible to add an option to wrap the paths? 
![image](https://user-images.githubusercontent.com/44238325/135284984-418075aa-7cff-4f96-99ae-653498775f6f.png)

tami5:
you can do that by `telescope.defaults {    path_display = { truncate = 2 } | "truncate" } `

grimgort:
Ho yes, sorry. 
However I had tried to modify this variable to "truncate" but without success. I must have made a mistake... sorry for the inconvenience...

Conni2461:
wrapping was just implemented in #1735

Issue #701, opened 2021-03-29
Wrap lines in the results window


divagant-martian:
**Is your feature request related to a problem? Please describe.**
Using the builtin picker for lsp diagnostics gives me cut version of the error, ,when I really need to be able to read it completely.

**Describe the solution you'd like**
Ideally, something like what's described in #582, but something as simple as `autocmd User TelescopePreviewerLoaded setlocal wrap` with `TelescopePreviewerLoaded` changed for the appropriate name would be enough

**Describe alternatives you've considered**
At least being able to change the layout for the default picker so that I can see the errors in a longer line. This wouldn't really be enough, but I haven't found how to do this either.

**Additional context**
The amount of info that can be read in this width is very little
![image](https://user-images.githubusercontent.com/26765164/112911129-681d4e80-90ba-11eb-95e0-0588b38639f6.png)

Conni2461:
Could be possible. You could change this from `false` to `true`. But this results in a buggy selection. Also it wouldn't fix the truncate problem because the width is column width is hardcoded for diagnostics here. Maybe we should move the table around and do `file:lnum:col | message`.

CC @fdschmidt93

fdschmidt93:
I personally would not move the table around since file names are optional and not shown in document diagnostics.

The following options can be passed to `lsp_{document,workspace}_diagnostics` to reformat the entries.

- `no_sign`: remove E in case of your screenshot to save one char; highlighting indicates error type and is linked to `LspDiagnosticsDefault{Error, Warning, Hint, Information}`
- `line_width`: increase line width of diagnostic message, possibly at the expense of truncating the filename
- `hide_filename`: remove file name entirely by default (`workspace only`)

As a reference, the relevant code can also be found [here](https://github.com/nvim-telescope/telescope.nvim/blob/aefc8317354426864a361d7b7d84848145d218c1/lua/telescope/make_entry.lua#L857).

More generally, I'd prefer to  add an action that calls `vim.lsp.diagnostic.show_line_diagnostic` to "print" the full diagnostic message for the line as that appears clean and rather straightforward though may include other messages. Possible alternative `actions`: (a) put entry in a floating window, (b) horizontal scrolling for entry with `h`, `l` in normal mode? might be of more general use

[...]

PR #1735, opened 2022-02-05
feat: allow user to wrap picker results (#701, #1296)

sultanahamer:
Added a config option `wrap_results` to wrap the results in all pickers. Requested in #701  and #1296.

Default option is false. Same as current behavior. If user chooses, they can set this value true in setup call and see the results wrapped.

Unwrapped behavior:
unwrapped

Wrapped behavior:
wrapped

On a git_commits picker
image

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -179,6 +179,12 @@ telescope.setup({opts})                                    *telescope.setup()*
 
         Default: 0
 
+                                          *telescope.defaults.wrap_results*
+    wrap_results: ~
+        Word wrap the search results
+
+        Default: false
+
                                          *telescope.defaults.prompt_prefix*
     prompt_prefix: ~
         The character(s) that will be shown in front of Telescope's prompt.
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -237,6 +237,15 @@ append(
 )
 
 append(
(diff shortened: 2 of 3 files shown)


sultanahamer:
@Conni2461 didn't get your question. This is inbuilt wrap(set nowrap) functionality of vim and it has to work good. Unless the picker results are very lengthy lines, in that case UX might be a bit bad but I am using this every day and is working fine for me.

Conni2461:
I know that its builtin. I just wanted to know if there are any problems with selecting or similar issues. Because we have a lot of custom stuff on top of that.

sultanahamer:
I didn't encounter any issues as of now.

sultanahamer:
@artfulrobot here is what I use
```
require("telescope").setup {
  defaults = {
    wrap_results = true,
    sorting_strategy = "ascending"
  },
  pickers = {
    oldfiles = {
      initial_mode = "normal"
    }
  }
}
```

atercygnus:
Is it possible to have file preview wrapped as well?

jamestrew:
@atercygnus 

Something like this should work.
```lua
vim.api.nvim_create_autocmd("User", {
  pattern = "TelescopePreviewerLoaded",
  callback = function()
    vim.wo.wrap = true
  end,
})
```

Issue #1739, opened 2022-02-09
Telescope should make use of the quickfix list stack


cgsheeh:
**Is your feature request related to a problem? Please describe.**
Sometimes I make a search for A and send it to the quickfix list. Then I work a little more, and make another search for B, which I also send to the quickfix list. After that, I realize I need to see the results of my search for A again. Since I searched for B already, the quickfix list is overwritten and I need to re-open Telescope, search for A and re-send to the quickfix list.

I looked online for some extension that might solve this issue for me, but as it turns out this is already a feature of the quickfix list in Vim/Neovim! See [this page](https://neovim.io/doc/user/quickfix.html) and search for "Using more than one list of errors" (sorry, I can't figure out how to permalink the exact section). Neovim supports a stack of up to 10 quickfix lists, which can be moved between using `:colder`, `:cnewer` in a similar fashion to `:cnext` and `:cprev`.

**Describe the solution you'd like**
So the feature request (bug?) is that Telescope should be sending results to new quickfix lists instead of replacing all entries in the current quickfix list.

**Describe alternatives you've considered**
Alternatively we could create a new mapping that specifically sends results to a new quickfix list, keeping "overwrite all existing entries in the current list" as the default.

**Additional context**
I'm a relative newbie to Neovim and its rapidly evolving ecosystem, so it's possible there is a better solution that I haven't considered. I've never actually used the quickfix stack before so I might also be wrong about it's usefulness here. If I'm way off the mark here or there's a much better workflow I haven't been made aware of, please point me in the right direction!

I took a brief look at the code and it seems when Telescope calls [`setqflist`](https://neovim.io/doc/user/builtin.html#setqflist()) it always passes `'r'` as the mode on "send to qf list", when it should probably pass `''` to create a new list. I'd be interested in taking this on as a contribution if it would be accepted into the codebase.

cgsheeh:
The first PR which enables Telescope using the quickfix list stack has landed. Nice!

I'm going to keep this open as I plan to hack on this a little more. As I mentioned previously I think having a picker for the quickfix list stack would be a nice feature. We may also be able to leverage the `what` dict passed to `setqflist` to add the Telescope query as the title of the quickfix list in the stack, which would make search much easier.

Conni2461:
we are currently not looking for new pickers. See https://github.com/nvim-telescope/telescope.nvim/issues/1228

We can expose the local function `send_selected_to_qf ` in the `actions/init.lua`?

[...]

PR #1742, opened 2022-02-10
fix(actions.qflist): make it work with vim quickfix commands

cgsheeh:
Changes functions that use `setqflist` in `r` mode to instead use
` ` mode. This makes the Telescope quickfix list action send results
to a new list in the quickfix stack instead of overwriting the
entries in the current list. Doing so enables `:chistory`, `:colder`
and `:cnewer` functions to review results of previous Telescope
queries.

The location list uses a similar API and is updated in this PR also
(ie this also enables `:lhistory`, `:lolder` and `:lnewer`).

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -667,7 +667,7 @@ end
 
 --- Sends the selected entries to the quickfix list, replacing the previous entries.
 actions.send_selected_to_qflist = function(prompt_bufnr)
-  send_selected_to_qf(prompt_bufnr, "r")
+  send_selected_to_qf(prompt_bufnr, " ")
 end
 
 --- Adds the selected entries to the quickfix list, keeping the previous entries.
@@ -677,7 +677,7 @@ end
 
 --- Sends all entries to the quickfix list, replacing the previous entries.
 actions.send_to_qflist = function(prompt_bufnr)
-  send_all_to_qf(prompt_bufnr, "r")
+  send_all_to_qf(prompt_bufnr, " ")
 end
 
 --- Adds all entries to the quickfix list, keeping the previous entries.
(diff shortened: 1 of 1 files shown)


tami5:
🤔 hmmm I get how practical it is to have new items added to an existing list. But I have a feeling that maybe it should rather be a global option or new function. The thing is many users of telescope expect the behavior of having new quickfix list on send_to_quickfix, like me. 

Since it's vim default behavior, then it should be the default here, however, with an way for current users to opt-out this change.

Otherwise, LGTM. btw great argument #1739

cgsheeh:
> The thing is many users of telescope expect the behavior of having new quickfix list on send_to_quickfix, like me.

@tami5 Just to clarify, after this change you will still see a new quickfix list on `send_to_quickfix`. The difference is the previous quickfix list will now be in the quickfix list stack, recoverable with `:colder` after you send another Telescope query to the quickfix list (and the same change is true of the location list). The current behaviour of `send_to_quickfix` is to remove all the existing entries in the current quickfix list and replace them with the new entries. You can already use `add_selected_to_qflist` to append to the current quickfix list (with `<M-Q>` IIRC). See `:h setqflist`, the section on `actions`. Confusingly the `' '` action is in a paragraph at the end instead of formatted similarly to the `r` and `a` actions (which is perhaps a bug in and of itself).

Apologies in advance if you already understood all this, just want to make sure we're on the same page. If you could pull down the PR and give it a try I'd greatly appreciate it!

tami5 (approved):
Oh I see, I totally misunderstood it as feature rather than a bug. I did some testing and indeed no big behavior changes to what the user would expected. but rather an access to a new suite of quickfix driven workflows.

Thanks for the clarification @cgsheeh

PR #1744, opened 2022-02-12
feat(spell_suggest): ignore spell setting

ModProg:
`z=` works even when `spell` is not set. I think it would be nice if Telescope would behave the same.

If you think this should be a setting, I'll look into how they work in telescope an add it.


tami5:
indeed it works without spell being set. I had refactored some internal code in my dot files because of this information so thanks :).

LGTM

PR #1745, opened 2022-02-12
feat(lsp): ignore_symbols

tami5:
#### Purpose 

Adds new option to ignore symbols. Essentially same as `opts.symbols` but only ignores given symbols and doesn't do any checking whether the symbol is valid or not.

extracted from #1449.

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1410,7 +1410,6 @@ builtin.lsp_document_symbols({opts})          *builtin.lsp_document_symbols()*
                                             false)
         {symbols}           (string|table)  filter results by symbol kind(s)
         {ignore_symbols}    (string|table)  list of symbols to ignore
-                                            (default: nil)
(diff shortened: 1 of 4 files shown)


fdschmidt93 (commented):
Generally LGTM :)

Two things come to mind:

- `ignore_symbols` unlike `symbols` only allows to pass a table; I'd say let's probably harmonize that
- I guess the filtering with catching all the cases originally written by me was a bit needlessly complicated, looking at it now (lines 110-141). Should we refactor that here to harmonize the mechanics `{ignore_}symbols` (largely simplifying the latter)? Up to you. It works one way or another I suppose.

Conni2461:
I'll finish it.
`ignore_symbols` now takes both `string` and `table` and i'll also try to refactor `symbols`. And would merge then. I'll try to finish some of these long standing PRs in the next couple of days. So don't be sad that i finished it for you @tami5 :)

PR #1748, opened 2022-02-14
feat: teach git_files to use the current repository

davvid:
git_files only looks at the current repository, which is limiting
when editing files across several repositories.

A simple solution is to scope the "git ls-files" queries
to the current buffer's directory. This causes Telescope to
discover the git repository in the context of the current
file and return results relevant to that file.

Enable this behavior behind a new "use_per_file_cwd" so that
existing behavior is unchanged.

This is arguably better behavior in general and could be made the
default behavior, but is left as an opt-in feature.

### DIFF
--- a/lua/telescope/builtin/git.lua
+++ b/lua/telescope/builtin/git.lua
@@ -20,6 +20,7 @@ git.files = function(opts)
 
   local show_untracked = utils.get_default(opts.show_untracked, true)
   local recurse_submodules = utils.get_default(opts.recurse_submodules, false)
+  local use_per_file_cwd = utils.get_default(opts.use_per_file_cwd, false)
   if show_untracked and recurse_submodules then
     error "Git does not support both --others and --recurse-submodules"
   end
@@ -36,6 +37,7 @@ git.files = function(opts)
         git_command,
         show_untracked and "--others" or nil,
         recurse_submodules and "--recurse-submodules" or nil,
+        use_per_file_cwd and "--full-name" or nil,
       },
       opts
     ),
@@ -349,9 +351,19 @@ git.status = function(opts)
 end
 
 local set_opts_cwd = function(opts)
+  local use_per_file_cwd = utils.get_default(opts.use_per_file_cwd, false)
   if opts.cwd then
     opts.cwd = vim.fn.expand(opts.cwd)
+  elseif use_per_file_cwd then
+    -- Run git commands in the context of the current file's repository.
+    -- A per-buffer git repository is used when finding files when used in conjunction
+    -- with the default use_git_root option.
+    opts.cwd = vim.fn.expand('%:h:p')
   else
+    opts.cwd = vim.loop.cwd()
+  end
+  -- Fallback to the current directory when vim.fn.expand('...') returns an empty result.
+  if opts.cwd == nil or opts.cwd == '' then
     opts.cwd = vim.loop.cwd()
   end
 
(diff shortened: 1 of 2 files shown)


davvid:
The problem:

```
project/.git  # Not using submodules
project/lib-repo/.git
project/client-repo/.git

nvim lib-repo/lib.h client-repo/client.c


# When client-repo/client.c is the current buffer
: Telescope git_files

# results reflect project/.git but we're editing files form project/lib-repo/ and the file
```

The solution
```
: Telescope git_files
# When client.c is the current buffer then files from project/client-repo/ are displayed.
# When lib.h is the current buffer then the files from project/lib-repo/ are displayed.
# When an anonymous buffer or a file from project/ is used then the files from project/ are displayed.
```

This seems more intuitive and correct all around to me, but I've left this as an opt-in feature since it probably warrants discussion etc.

lua/telescope/builtin/git.lua
```diff
@@ -349,11 +351,21 @@ git.status = function(opts)
...
 
 local set_opts_cwd = function(opts)
+  local use_per_file_cwd = utils.get_default(opts.use_per_file_cwd, false)
   if opts.cwd then
     opts.cwd = vim.fn.expand(opts.cwd)
+  elseif use_per_file_cwd then
+    -- Run git commands in the context of the current file's repository.
+    -- A per-buffer git repository is used when finding files when used in conjunction
+    -- with the default use_git_root option.
+    opts.cwd = vim.fn.expand('%:h:p')
```
davvid, line 361: I could not find a lua native way to do get the directory of the current buffer which is why it's doing `%:h:p`.

`vim.api.nvim_buf_get_dir(0)` would have been useful if it existed. I can get the filename but we would need to `os.path.dirname(_)` it. This works for me, but if there's a better way to get that I'm all ears.

davvid:
I didn't realize it, but this option actually works for all of the `git_*` helper functions. I should maybe make `--full-name` unconditional -- it doesn't hurt for it to be there always.

davvid:
Sweet, thanks, yes, this does work:

```
vim.keymap.set('n', '<leader>f', function()
    telescope_builtin.git_files({ cwd = vim.fn.expand('%:h:p') })
end)
```

I'm going to close this MR for now.

----

As a follow-up, do we think `%:h:p` should be the default behavior for `Telescope git_files` when provided no arguments?

It's arguably not very useful in its default form in a lot of situations where the user might expect `git_files` to work.

For example, if my cwd is a directory without `.git` and I edit a subdirectory `some-repo/filename.rs` then `Telescope git_files` currently throws an error condition. Defaulting to `%:h:p` would make it a non-error and smooth.

If you think this would be worth changing then I can look at changing the default behavior.

Conni2461:
> telescope_builtin.git_files({ cwd = vim.fn.expand('%:h:p') })


You dont even need `vim.fn.expand` you can just pass `cwd = "%:h:p"` because we expand already

> As a follow-up, do we think %:h:p should be the default behavior for Telescope git_files when provided no arguments?

Nope it should not. We are currently moving towards not having additional arguments set for any picker. If you run `Telescope git_files` it should just run `git ls-files` everything else is just not intuitive

davvid:
The current behavior is currently not intuitive, though.

fugitive.vim follows the basic idea that I'm proposing -- it uses the git repo that corresponds with your current buffer.

I really think of it as a refinement of this existing feature:

```
        {use_git_root}       (boolean)  if we should use git root as cwd or
                                        the cwd (important for submodule)
                                        (default: true)
```

What I'm suggesting is that the above should be "git root of the current buffer" so that we match fugitive's intuitive behavior. The current behavior is less useful in more situations.

davvid:
That makes sense, thanks for explaining.

If others complain about this behavior in the future then I'd be happy to submit a documentation patch showing the `%:h:p` trick.

Issue #1754, opened 2022-02-20
builtin.autocommands previewer throws: Cursor position outside buffer


Vedant36:
### Description

Context: I've only started converting my init.vim to init.lua since day before yesterday and got this problem while checking if a specific autocommand was set(it was)

The bug seems to be in the following lines(lua/telescope/previewers/buffer_previewer.lua:935-937):
```
vim.schedule(function()
  vim.api.nvim_win_set_cursor(status.preview_win, { selected_row, 0 })
end)
```
Error disappears when the lines are removed but I don't know what their purpose is
No error when letters typed slower than 0.5s apart so the error showing that `Cursor position outside buffer` seems confusing
I don't understand `vim.schedule` well but this seems to suggest that the error is caused when another callback starts before the previous one finishes but I'm probably wrong

Note: the error margin is actually 0.45s(always error) - 0.46s(never error) (multiple tests) but I doubt the time accuracy of sleep at that scale

Telescope commit: df0b35c

Related but fixed Issues:
#957 fixed by #1157
#1727 fixed by #1732

### Neovim version

```markdown
NVIM v0.7.0-dev+1109-gfaeff49cb
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

### Operating system and version

Linux 5.16.10-zen1-1-zen

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - OK: nvim-treesitter installed.

## Checking external dependencies
  - OK: rg: found ripgrep 13.0.0
  - OK: fd: found fd 8.3.2

## ===== Installed extensions =====
```

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. `:Telescope autocommands`
3. type something fast(less than 0.4s between letters, tested with xdotool). i could get a consistent error by typing `asd` fast

### Expected behavior

No error is thrown

### Actual behavior

Error thrown:
```
Error executing vim.schedule lua callback: ...scope.nvim/lua/telescope/previewers/buffer_previewer.lua:936: Cursor position outside buffer
stack traceback:
        [C]: in function 'nvim_win_set_cursor'
        ...scope.nvim/lua/telescope/previewers/buffer_previewer.lua:936: in function <...scope.nvim/lua/telescope/previewers/buffer_previewer.lua:935>
```

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
[...]

PR #1756, opened 2022-02-21
fix builtin.autocommands giving error

Vedant36:
im `pcall` ing it as per the suggestion of @Conni2461
i also removed the vim.schedule as the other previewers didn't need it too
Fixes: #1754


Vedant36:
sorry made a syntax error will close this and pull request again

PR #1758, opened 2022-02-22
fix builtin.autocommands throwing error

Vedant36:
im `pcall` ing it as per the suggestion of @Conni2461
i also removed the vim.schedule as the other previewers didn't need it either
Fixes: #1754

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -930,11 +930,7 @@ previewers.autocommands = defaulter(function(_)
       end
 
       vim.api.nvim_buf_add_highlight(self.state.bufnr, ns_previewer, "TelescopePreviewLine", selected_row + 1, 0, -1)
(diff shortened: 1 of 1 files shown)


Vedant36:
sure i'll add the commit.
i am able to scroll the preview window without that line there, but you probably understand it better

daangoossens22:
I didn't intend to sound nosy, I added the vim.schedule part in a previous pull request so I wanted to chime in.
It is not about being able to scroll the preview window manually, but that it does so automatically while fuzzy finding. Here to show it visually (without manually scrolling the preview window):
- pcall, no vim.schedule:
no_schedule
- pcall, with vim.schedule:
with_schedule

Vedant36:
10/10 explanation. i understand the reason now

Conni2461 (approved):
yep i only wanted an additional `pcall` to suppress the error message :laughing: 

LGTM thanks :)

PR #1770, opened 2022-03-03
Added system dependency in readme.md

mmkumr:


Conni2461:
We kinda already have this a couple of lines higher. https://github.com/nvim-telescope/telescope.nvim/blob/76120285f88c1becb5728695f6df77c545437c53/README.md?plain=1#L47-L50

PR #1773, opened 2022-03-06
feat: allow change the prompt title when refreshing the picker 

przepompownia:


Conni2461:
You can just call `picker.prompt_border:change_title("asdf")` after your `refresh` call right?

przepompownia:
Indeed, `prompt_border` is accessible directly from `picker` so no need to modify telescope part. :bulb:

Issue #1775, opened 2022-03-06
Telescope sends results to quickfix incorrectly when there are spaces in the path.


lucypero:
### Description

When there are spaces in the file's path, Telescope sends the results to the quickfix incorrectly. It seems to escape the spaces with `\`. I think that's the reason why the files don't open when you try to open them from the quickfix list. It just opens an empty buffer.

### Neovim version

```markdown
NVIM v0.7.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

### Operating system and version

Ubuntu 20.04.3 LTS on Windows 10 x86_64

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - WARNING: nvim-treesitter not found. 

## Checking external dependencies
  - OK: rg: found ripgrep 11.0.2
  - WARNING: fd: not found. Install [sharkdp/fd](https://github.com/sharkdp/fd) for extended capabilities

## ===== Installed extensions =====
```

### Steps to reproduce

1. Go to a directory with spaces
2. Select some files on telescope
3. Send them to quickfix :   `require("telescope.actions").smart_send_to_qflist(prompt_bufnr)`
4. Try to open one of the files from quickfix

### Expected behavior

It should open the correct file on vim.

### Actual behavior

It opens a new empty buffer.

### Minimal config

```Lua
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-telescope/telescope.nvim'
```

eruizc-dev:
I managed to reproduce this in native linux as well:

![image](https://user-images.githubusercontent.com/58179604/156932607-f5079d82-325f-4cba-bca0-e72b58c7a31a.png)

The file `test dir/file.json` is sent to the quickfixlist as `test\ dir/file.json` with unnecessary escaping thus opening the wrong file afterwards.

There is no special configuration required for this to happen, it's the default behavior

Conni2461:
Dup of #1700

Conni2461:
fixed by https://github.com/nvim-telescope/telescope.nvim/commit/b501d9ecb7d8181e2238620c919740025e8b2096

PR #1776, opened 2022-03-06
Not escape spaces when sending files to qf

lucypero:
This fixes https://github.com/nvim-telescope/telescope.nvim/issues/1775 but it might cause other issues.


Conni2461:
We sometimes need an escape but not always. There are already 2 PRs that do that:
- https://github.com/nvim-telescope/telescope.nvim/pull/1306 I poked twice and havent received an answer so i am not gonna merge that one.
- https://github.com/nvim-telescope/telescope.nvim/pull/1712 will merge that now because i want it fixed and i think #1712 is good enough and i didnt get the issue they had in #1306

Still thanks for the PR

PR #1787, opened 2022-03-12
fix: follow with find command

i3Cheese:
This doesn't work
`$ find . -type f -L`
find: -L: unknown primary or operator
But this works well
`$ find -L.  -type f`


Conni2461 (approved):
Thanks :)

PR #1788, opened 2022-03-12
chore: delete unused functions

Conni2461:
Functions are not used in telescope.
First 3 functions are not used by anyone according to sourcegraph

Last function `quickfix_items_to_entries` is https://sourcegraph.com/search?q=context:global+quickfix_items_to_entries+lang:Lua+&patternType=literal

both usages are kinda copy pasta functions that we have builtin


Conni2461:
I'll merge this without a deprecation notice. I am also thinking we should split up the util file. First move so much as possible as local functions in the modules. Like function that are only used in one place.

Then we do an `internal` or similar file where we could put function that are only for internal use. And if someone uses that for their extension or worse dotfiles its not our fault if we change something because we said dont anything here.

The second file is a util file that extension can use with proper interfaces that simplify things. Like what we currently have with `transform_path`, `get_devicons` ...

PR #1792, opened 2022-03-15
Update picker docs. there -> their

dylanbyars:
A tiny improvement to the documentation.


Conni2461:
Thanks documentation is autogenerated. Needs to be fixed here as well: https://github.com/nvim-telescope/telescope.nvim/blob/a36a813d5d031e6f5d52b74986915e68130febd9/lua/telescope/builtin/init.lua#L213

PR #1795, opened 2022-03-19
fix: prevent double calling of `setup`

jamestrew:
Minor fix but was previously calling the `setup` function twice upon `require'telescope'.load_extension(...)`


Conni2461:
CC @fdschmidt93

#1601 #1628

jamestrew:
Got it :+1: Nevermind this then.
Sorry to bother you all.

PR #1797, opened 2022-03-19
docs: fix typo 'grown' in documentation

SuvanshKumar:
In the history section of the documentation description (docs/telescope.txt), the warning line sounds like an incorrect grammar:

    Warning: If limit is set to nil it will grown unbound.


Conni2461:
docs is autogenerated needed to be fixed at another place as well, ive quickly done that. Thanks for the PR :)

PR #1807, opened 2022-03-25
fix: detect binary files before detecting filetype

### LINKED ISSUE #1637: Telescope find_files freezes nvim when select big binary file(use previewer)
### Description

when I use `Telescope find_files` select a big binary file(test file: unzip [ccls](https://github.com/fcying/tools/releases/download/tools/ccls_linux_amd64.zip), about 38M), nvim be freezes, nvim cpu 90+%.
wait several seconds, display `Binary cannot be previewed`

if disable previers`Telescope find_files previewer=false`, work fine, select  binary file not freeze nvim.

telescope version: master latest: 9d0034d

### Neovim version

```markdown
NVIM v0.6.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by runner@fv-az32-74
```

debian 11

### checkhealth telescope

```markdown
  - OK: plenary installed.

  - OK: rg: found ripgrep 12.1.1 (rev 7cb211378a)

```

### Steps to reproduce

1. nvim -u init.lua
2. Telescope find_files
3. select ccls(freezes nvim, cpu 90+%)  
4. wait and quit nvim
5. nvim -u init.lua
6. Telescope find_files previewer=false
7. select ccls(work fine)

### Expected behavior

if select binary file, only display `Binary cannot be previewed`, not freeze nvim

### Actual behavior

freeze nvim

### Minimal config
init.lua
```Lua
local fn, cmd = vim.fn, vim.cmd
local root_dir = fn.fnamemodify(fn.resolve(fn.expand('<sfile>:p')), ':h')

cmd('set packpath=' .. root_dir .. '/plugged')
cmd('set runtimepath^=' .. root_dir .. '/plugged')
local install_path = root_dir .. '/plugged/pack/packer/opt/packer.nvim'
if fn.empty(fn.glob(install_path)) > 0 then
    Bootstrap = fn.system({'git', 'clone', '--depth', '1', 'https://github.com/wbthomason/packer.nvim', install_path})
end

cmd([[
    set mouse=n
    packadd packer.nvim
    nnoremap <space>pu :PackerSync<CR>
    nnoremap <space>pc :PackerCompile<CR>

    nnoremap <silent> ff :Telescope find_files previewer=false<CR>
    nnoremap <silent> fp :Telescope find_files<CR>
    set cmdheight=3
]])

local packer = require('packer')
local use = packer.use
packer.init({
    package_root = root_dir .. '/plugged/pack',
    compile_path  = root_dir .. '/plugged/plugin/packer_compiled.lua',
})

use {'wbthomason/packer.nvim', opt = true}
use 'nvim-lua/plenary.nvim'
use 'nvim-telescope/telescope.nvim'

if Bootstrap then
    packer.sync()
end


```
```

dancek:
Fixes #1637 for files recognized as binary by `file --mime-type`.

The filetype detection from plenary was run before checking if the file
is binary and if it is over the size limit. Moving the detection after
them fixes most cases of freezing while previewing a large binary file.

---

Note that https://github.com/nvim-lua/plenary.nvim/issues/307 is still needed for large files detected as `text/plain` with no newlines. Still, I think this commit is also useful: why detect the filetype if we later decide not to show a preview after all?

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -160,7 +160,6 @@ previewers.file_maker = function(filepath, bufnr, opts)
   if opts.use_ft_detect == nil then
     opts.use_ft_detect = true
   end
-  opts.ft = opts.use_ft_detect and pfiletype.detect(filepath)
   if opts.bufname ~= filepath then
     if not vim.in_fast_event() then
       filepath = vim.fn.expand(filepath)
@@ -187,7 +186,7 @@ previewers.file_maker = function(filepath, bufnr, opts)
           end),
         })
       else
-        if opts.preview.check_mime_type == true and has_file and opts.ft == "" then
+        if opts.preview.check_mime_type == true and has_file then
           -- avoid SIGABRT in buffer previewer happening with utils.get_os_command_output
           local output = capture(string.format([[file --mime-type -b "%s"]], filepath))
           local mime_type = vim.split(output, "/")[1]
@@ -222,6 +221,8 @@ previewers.file_maker = function(filepath, bufnr, opts)
             return
           end
         end
+
+        opts.ft = opts.use_ft_detect and pfiletype.detect(filepath)
 
         opts.start_time = vim.loop.hrtime()
         Path:new(filepath):_read_async(vim.schedule_wrap(function(data)


dancek:
Oh, that's a good point. I didn't realize spawning `file` is that slow.

Still, I think filesize should be checked before detecting filetype. I'll write such a commit.

dancek:
And now I realize there are quite a bit of dependencies in the order things are done, so if I were to check filesize first the logic needed to still have `filetype_hook` work is a bit unreadable.

I'll try a plenary.nvim patch first. I guess it's the best approach.

PR #1817, opened 2022-04-03
Allow results_title and prompt_title to have global default values

mbenford:
Currently both `results_title` and `prompt_title` can only be changed when a picker is executed. This PR makes those fields global so that they can be set in Telescope setup's `defaults` property.

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -307,6 +307,21 @@ telescope.setup({opts})                                    *telescope.setup()*
 
         Default: false
 
+                                         *telescope.defaults.results_title*
+    results_title: ~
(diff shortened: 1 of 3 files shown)


mbenford:
I'm not sure if I'm supposed to update `docs/telescope.txt` myself or some [build action](https://github.com/nvim-telescope/telescope.nvim/blob/master/CONTRIBUTING.md#auto-updates-from-ci) will do it for me. Any guidance would be much appreciated.

Conni2461:
This action doesn't like new contributors. I dont know why. If it doesnt run i'll do it from hand :)

Thanks for the PR :)

PR #1819, opened 2022-04-05
Fix calc_result_length can throw

pgherveou:
git-worktree extension is throwing an error when this function get executed. 
I think this is because I have the truncate set, and this plugin does not operate on buffer
adding a type check here seems to fix it

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -202,7 +202,12 @@ end
 
 local calc_result_length = function(truncate_len)
   local status = get_status(vim.api.nvim_get_current_buf())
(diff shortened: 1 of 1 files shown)


Conni2461:
https://github.com/ThePrimeagen/git-worktree.nvim/blob/d7f4e2584e81670154f07ca9fa5dd791d9c1b458/lua/telescope/_extensions/git_worktree.lua#L166 This is not allowed. `transform_path` is meant to be used inside `make_entry`.

So i prefer not to fix it here. And rather add a documentation bit where transform_path is suppose to be used. In `make_entry` and more specific in a display function because then we can optimize it by only transforming the paths that will be displayed. Because its only a visual thing

Conni2461:
Closing. Will not be addressed in telescope core! 

Issue for documentation #1844

Issue #1827, opened 2022-04-08
Make documentation more friendly


danielo515:
First of all, let me say thank you for this amazing tool. I have tried other fuzzy finders before, and none was as easy to use, configure and extend as this one is. It really made me go back to nvim, and that is something that makes me super happy, so thank you very much.

**Is your feature request related to a problem? Please describe.**
My request is about developer/advanced user experience and my relation with the docs, which has not been very nice and a bit frustrating.

**Describe the solution you'd like**
I want the docs to be a bit more structured and easier to navigate. Probably adding more advanced examples will help.

**Additional context**
Here are some of the pain points I found. I think it will be easier to act on things that are clearly defined.

**The structure of the first part of the doc is confusing**. 
It starts very good, but then I had to verify several times that I was not skipping parts, because there are several points tha I consider key parts missing. So, this is the beginning:

```
telescope.setup({opts})                                    *telescope.setup()*
    Setup function to be run by user. Configures the defaults, pickers and
    extensions of telescope.

    Usage:
    >
    require('telescope').setup{
      defaults = {
        -- Default configuration for telescope goes here:
        -- config_key = value,
        -- ..
      },
      pickers = {
        -- Default configuration for builtin pickers goes here:
        -- picker_name = {
        --   picker_config_key = value,
        --   ...
        -- }
        -- Now the picker_config_key will be applied every time you call this
        -- builtin picker
BLABLA
```

So far so good, makes sense. Then you start to document the valid keys for the first par of the config, which is great:

```
   Valid keys for {opts.defaults}

                                      *telescope.defaults.sorting_strategy*
    sorting_strategy: ~
```

And then, after a lot of text it jumps straight to registering extensions `telescope.register_extension`. What? Where are all the other options documented? What are the concrete config bits that a picker accepts? From there it then goes to builtin pickers. I think I am supposed to understand that all the options documented inside each builtin.picker_name are also possible to be provided to the initial setup? If that is the case, a single line saying `see builtin.picker` (or whatever) for configuration options. But even then, no single picker documents an option to define custom mappings, and seems that it is possible to define mappings per picker. I don't even remember where I read that to be honest, but it is true.
I think documenting all the options the setup function expects explicitly and before anything else is the best thing for most readers, but I may be mistaken.

**Certain options are documented outside the main project**

By reading the documentation of telescope-file-browser I discovered that mappings can be functions:

```
      mappings = {
        ["i"] = {
          -- remap to going to home directory
[...]

PR #1829, opened 2022-04-08
docs: document more actions and add introduction to actions

Conni2461:
Might merge tomorrow. I'll need to think about it over night. If i wanna add more

Close #587

Ref #1827

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1887,6 +1887,54 @@ resolver.resolve_anchor_pos()                  *resolver.resolve_anchor_pos()*
 
 Actions functions that are useful for people creating their own mappings.
 
+Actions can be either normal functions that expect the prompt_bufnr as first
+argument (1) or they can be a custom telescope type called "action" (2).
+
+(1) The `prompt_bufnr` of a normal function denotes the identifier of your
+picker which can be used to access the picker state. In practice, users most
+commonly access from both picker and global state via the following:
+
+>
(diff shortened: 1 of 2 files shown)


doc/telescope.txt
```diff
@@ -1868,6 +1868,34 @@ resolver.resolve_anchor_pos()                  *resolver.resolve_anchor_pos()*
 
 Actions functions that are useful for people creating their own mappings.
 
+Actions can be either normal functions that expect the prompt_bufnr as first
```
fdschmidt93, line 1871: I'd suggest to extend this paragraph with something like

> The `prompt_bufnr` denotes the unique identifier of your picker to access the picker state. Beyond the picker state, telescope.nvim also retains a global state. In practice, users most commonly access from both picker and global state via 
local current_picker = action_state.get_current_picker(prompt_bufnr) -- picker state
local entry = action_state.get_selected_entry() -- global state

I guess part on global state is debatable , but I'd suggest to add how to most commonly access relevant info of telescope(-picker) state here (= the code segment). Maybe it's worthwhile adding a small example function here as well, even if it's just printing the `entry.value`. I'm 100% certain that we surely mention that already somewhere, but telescope is too big/scattered and users don't get across all the points of information :sweat_smile:

Conni2461: that seems like a good idea. I'll try to integrate it. Thanks :)

PR #1849, opened 2022-04-18
feat: add default hl group for devicons

Diaoul:
Adds the ability to choose the color that is used for rendering the icon when `color_devicons` is set to `false`.

```lua
local colors = require 'gruvbox.colors'

require('telescope').setup {
    defaults = {
        color_devicons = false,
    },
}
vim.api.nvim_set_hl(0, 'TelescopeResultsFileIcon', {fg = colors.neutral_red})
```

image

_This is ugly but just to illustrate_


Conni2461:
Thanks :)

PR #1856, opened 2022-04-19
fix: closing process handle after termination

### LINKED ISSUE #1159: rg don't exit
### Description

When using `live_grep ` the query, if you enter and switch entries quickly, it will get stuck and rg will have multiple processes

### Neovim version

NVIM v0.5.0

macos 11.4

### Steps to reproduce

1. [LunarVim](https://github.com/LunarVim/LunarVim)
2. open big project
3. open live_grep()
4. enter and switch entries quickly

### Expected behavior

_No response_

### Actual behavior

image

https://user-images.githubusercontent.com/2859981/130627873-d4a287dd-0cb2-4582-b30f-8453a23e7849.mp4

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

quetz:
fix #1159

### DIFF
--- a/lua/telescope/_.lua
+++ b/lua/telescope/_.lua
@@ -54,6 +54,8 @@ function AsyncJob:close(force)
     p:close(force)
   end)
 
+  uv.process_kill(self.handle, "SIGTERM")
+
   log.debug "[async_job] closed"
 end
 
--- a/lua/telescope/_.lua
+++ b/lua/telescope/_.lua
@@ -53,9 +53,6 @@ function AsyncJob:close(force)
   self:_for_each_pipe(function(p)
     p:close(force)
   end)
-  if not self.handle:is_closing() then
-    self.handle:close()
-  end
 
   log.debug "[async_job] closed"
 end
@@ -67,6 +64,9 @@ M.spawn = function(opts)
     self.uv_opts,
     async.void(function()
       self:close(false)
+      if not self.handle:is_closing() then
+        self.handle:close()
+      end
     end)
   )
 


lua/telescope/_.lua
```diff
@@ -67,6 +64,9 @@ M.spawn = function(opts)
     self.uv_opts,
     async.void(function()
       self:close(false)
```
Conni2461, line 66: Isn't this line calling the above function `:close` anyway? So you just moved a block of code? Did you test and this actually fixes the issue?

quetz: I moved calling  `close` on process handle to close callback (that is called when process is terminated). Calling `close` on handle of non-terminated process results in libuv's SIGCHLD handler removed and process remains zombie after eventual termination. I'm not a lua programmer, but after this fix I could not trigger issue and have zombie `rg` processes pile up.

Conni2461: oh i know what the actual problem is. If you use `live_grep` and start typing `a` we spawn `rg --options a` and if you type `b` we stop the process (with `:close()` from the finder) and start a new one `rg --options ab`.

With your change we no longer do that. We only say we are no longer interested in the pipes. `rg` still runs till all things are outputed and then we close it with

```lua
if not self.handle:is_closing() then
  self.handle:close()
end
```

But that is also not what we want, we dont need to keep the `rg` processes around if we are done with them. Maybe we can additionally do `uv.process_kill(self.handle, "SIGINT")`before `self.handle:close()` in the close method (reverting your changes and making it basically a kill method).

Does this make sense? Could you test it?

Conni2461: What we can also do is just introduce a `kill` method doing only `uv.process_kill` (maybe an additional field to check we don't do this twice). The callback will then be called and calls `self:close(false)` which then closes the process.

And then we replace all the `:close` function calls with the new kill function from example in this file https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/finders/async_job_finder.lua#L31 and L71

quetz: Here is what I think is happening (take this with a grain of salt, as I have zero experience with lua):
There can be at most one `live_grep` job running (local job in async_job_finder.lua). When we type `a` there's `rg` process spawned, if we happen to type `b` while this `rg` is still running `:close` is called  on that job and new job is started.

How that `job:close` works? Well, it closes process fds (pipes): input, output, err and then calls `:close` on process handle. Looks legit, but there's a problem -- if `:close` is called on process handle while it is still running libuv removes SIGCHLD handler for that process and when process eventually terminates it becomes a zombie (because without SIGCHLD handler there's noone to `waitpid` and reap it). Original `job:close` implementation closes process handle too fast (yes, fds are closed and `rg` will detect it and terminate, but process handle is closed before that) resulting in zombie.

Proposed fix is to call `:close` on process handle only in `spawn` callback (that is called only when process is terminated) and have `job:close` only close fds. Closing fds will lead to process termination and callback will `:close` process handle. 

Sorry for somewhat long explanation, I hope it is more clear what I was trying to do with that PR. After that fix I saw no `rg` zombies anymore (does not mean I have not failed elsewhere, but still...)

Conni2461: > Proposed fix is to call :close on process handle only in spawn callback (that is called only when process is terminated) and have job:close only close fds. Closing fds will lead to process termination and callback will :close process handle.

Yes i got that and i understand why that works but then `rg` will always run till its done, even thought we no longer need it. For big searches this could waste seconds of computation. Thats why i proposed maybe we want to send `SIGINT` to the process if we no longer need it. I need to test this after work today

Conni2461: This diff fixes the issue for me and stops rg processes we no longer need. (only implemented for live_grep right now, one_shot_job should also be done)

```diff
diff --git a/lua/telescope/_.lua b/lua/telescope/_.lua
index eaea824..9be8c2b 100644
--- a/lua/telescope/_.lua
+++ b/lua/telescope/_.lua
@@ -45,7 +45,18 @@ function AsyncJob:_for_each_pipe(f, ...)
   end
 end
 
+function AsyncJob:kill()
+  if self.running then
+    -- We need to get rid of the pipes otherwise kill wont work
+    self:_for_each_pipe(function(p)
+      p:close(true)
+    end)
+    uv.process_kill(self.handle, "SIGINT")
+  end
+end
+
 function AsyncJob:close(force)
+  self.running = false
   if force == nil then
     force = true
   end
@@ -69,6 +80,7 @@ M.spawn = function(opts)
       self:close(false)
     end)
   )
+  self.running = true
 
   if not self.handle then
     error(debug.traceback("Failed to spawn process: " .. vim.inspect(self)))
diff --git a/lua/telescope/finders/async_job_finder.lua b/lua/telescope/finders/async_job_finder.lua
index f2c4f39..a476366 100644
--- a/lua/telescope/finders/async_job_finder.lua
+++ b/lua/telescope/finders/async_job_finder.lua
@@ -28,7 +28,7 @@ return function(opts)
 
   local callable = function(_, prompt, process_result, process_complete)
     if job then
-      job:close(true)
+      job:kill()
     end
 
     local job_opts = fn_command(prompt)
@@ -68,7 +68,7 @@ return function(opts)
   return setmetatable({
     close = function()
       if job then
-        job:close(true)
+        job:kill()
       end
     end,
   }, {
```

CC @tjdevries

quetz: > Yes i got that and i understand why that works but then `rg` will always run till its done, even thought we no longer need it. For big searches this could waste seconds of computation. Thats why i proposed maybe we want to send `SIGINT` to the process if we no longer need it. I need to test this after work today

Why would it run till its done? It will run until it tries to write next line to closed stdout, detect it and terminate.

Conni2461: Okay i did not know about `SIGPIPE`. I had to read up on it. I think my solution also relied on it. Some thoughts
- if rg still has to run for a couple of seconds, searching for matches it will never receive a `SIGPIPE`, this is bad
- I think i read that `SIGPIPE` does not exist on windows, somewhat bad
- `_.lua` with its AsyncJob, was originally planned as library, that will replace plenary.job in the future. Have a function `:close()` that will only close jobs that actually write something to a pipe is kinda bad, for a library. 

I think `:close` should instantly close and stop/kill the process not when the next `write()` happens. This is not good for processes that write irregularly.

quetz: Sounds reasonable. What are downsides of just adding `uv.process_kill` call to `job:close` in my PR?

nesqi: Thank you for the solution!
I'll just comment out the self.handle:close() locally for now until your final solution is available upstream.

tjdevries: don't worry about making it work perfectly as a library -- just make it work for us for now (I can investigate more at a different time). If it closes the processes, then that's good for me.

quetz: @tjdevries I have no zombie `rg` processes with this PR, but extra testing is always welcome :smile:

Conni2461: I for a kill in addition. I clean it up sometime else.

Thanks :)

PR #1867, opened 2022-04-21
feat: none strategy & control attachment

fdschmidt93:
I've recently played a bit around with Emacs magit (blasphemous I know) to see what the fuzz is about (I also started playing with mlua to essentially expose the sync library of [gitui](https://github.com/extrawurst/gitui) which can blow magit diff parsing parallelized out of the water).

Anyways, neither here nor there, but Vertico as a really nicely integrated framework got me thinking whether we can have similar things with telescope (i.e. telescope as `vim.ui.input` with some niceties I suppose). This PR and example code were 20 minutes playing around with the idea (we can polish `attach/detach` later).

Showcase

https://user-images.githubusercontent.com/39233597/164539336-1df5953d-f246-4d91-a52a-9316df4347ee.mp4

Example code

```lua

local toggle_attach = function(prompt_bufnr)
  current_picker.attach = not current_picker.attach
end

local set_prompt = function(prompt_bufnr)
  local value = action_state.get_selected_entry().value
  current_picker:reset_prompt(value)
end

builtin.find_files({
  selection_strategy = "none",
  attach_mappings = function()
    actions.move_selection_next:enhance({
      pre = function(prompt_bufnr)
        toggle_attach(prompt_bufnr)
      end,
      post = function(prompt_bufnr)
        set_prompt(prompt_bufnr)
        toggle_attach(prompt_bufnr)
      end,
    })
    actions.move_selection_previous:enhance({
      pre = function(prompt_bufnr)
        toggle_attach(prompt_bufnr)
      end,
      post = function(prompt_bufnr)
        set_prompt(prompt_bufnr)
        toggle_attach(prompt_bufnr)
      end,
    })
    return true
  end,
})
```

This PR of course is not per se needed, but enables a `nvim-cmp`-like completion experience from within the prompt. I guess `"none"` is not really controversial and some users may like it anyways, but intermediate detaching like this probably is I suppose a bit more controversial (proper `nvim_buf_detach` apparently is only available via RPC). Primarily putting this up for discussion. I'll probably do some more work around `telescope` for `vim.ui.input` in the next couple of days.

Furthermore, attaching/detaching also could nicely enable uses cases for `telescope-file-browser.nvim` to leverage the prompt for renaming/creating files while being able to glance current results.

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -97,6 +97,7 @@ telescope.setup({opts})                                    *telescope.setup()*
         - "follow"
         - "row"
         - "closest"
+        - "none"
 
                                        *telescope.defaults.scroll_strategy*
     scroll_strategy: ~
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -182,7 +182,8 @@ append(
   - "reset" (default)
   - "follow"
   - "row"
-  - "closest"]]
+  - "closest"
+  - "none"]]
 )
 
 append(
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -82,6 +82,7 @@ function Picker:new(opts)
     initial_mode = get_default(opts.initial_mode, config.values.initial_mode),
     debounce = get_default(tonumber(opts.debounce), nil),
 
+    _finder_attached = true,
     default_text = opts.default_text,
     get_status_text = get_default(opts.get_status_text, config.values.get_status_text),
     _on_input_filter_cb = opts.on_input_filter_cb or function() end,
@@ -499,10 +500,12 @@ function Picker:find()
   -- Register attach
   vim.api.nvim_buf_attach(prompt_bufnr, false, {
     on_lines = function(...)
-      find_id = self:_next_find_id()
+      if self._finder_attached then
+        find_id = self:_next_find_id()
 
-      status_updater { completed = false }
-      self._on_lines(...)
+        status_updater { completed = false }
+        self._on_lines(...)
+      end
     end,
 
     on_detach = function()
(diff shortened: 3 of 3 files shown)


Conni2461:
This looks cool, i dont see a reason why we could not add it. Do you want it in the 0.6 tag

fdschmidt93:
Great :)

>  Do you want it in the 0.6 tag?

I don't have a strong preference, since users should be using 0.7 anyways :sweat_smile: 

I'll add a private interface for toggling attach ala `picker:_toggle_attach` and I think this is GTG.

fdschmidt93:
I've renamed `attach` to `_finder_attached` and introduced a `Picker:_toggle_finder_attach`.

It's not the greatest of naming, but there already is `Picker:_detach` for a slightly different purpose and this way it's unambiguously clear I suppose (at least to people familiar with how telescope works) :sweat_smile: . If you mind the naming, I'm also happy to reconsider and hopefully think of something more elegant.

max397574:
Perhaps you could explain rq what we can see in the video?
(I think I know what but not sure)

fdschmidt93:
A list of what's going on hopefully roughly in order:
- `selection_strategy = "none"` makes s.t. there is no initial selection when something is typed in prompt
- You can now detach the prompt from the finder
- Selecting an entry among the results detaches prompt from finder, sets prompt to `entry.value`, and reattaches the finder

That way, it looks/feels very similar to completion with `nvim-cmp` which is a cool-use for when I bridge a telescope picker to `vim.ui.input` (for which you can pass lists etc.).

Furthermore, I probably plan on using this approach for creating/renaming/etc files for `telescope-file-browser.nvim` because it makes for a nicely unified interface to input stuff.

fdschmidt93:
Maybe an eventually-to-come `vim.ui.input` telescope preview leveraging the features from this PR :sweat_smile: 

https://user-images.githubusercontent.com/39233597/165178718-c1bde432-5c3a-43bc-9837-72142a2b0530.mp4

Conni2461:
do we want to merge this into dev? Or whats missing?

fdschmidt93:
Yeah, let's merge this to `dev` :) Can you please take care of that? :sweat_smile:

PR #1871, opened 2022-04-22
feat: show assigned function in actions.which_key

### LINKED ISSUE #1869: Mappings do not appear on which key
I am using the telescope extension of harpoon, and despite the extension registers some custom mappings for its picker, those mappings do not appear when I trigger which key inside the popup of the picker. Take a look at the additional context section for links and screenshots

I heavily rely on the inline help that which-key provides, so I would love to this to actually work for all the mappings

None other than having a cheatsheet.

**Additional context**
Here is the code of the extension that registers the mappings:
https://github.com/ThePrimeagen/harpoon/blob/master/lua/telescope/_extensions/marks.lua#L102-L115

Here you can see an screenshot with which key open, where the mappings created do not exist
image

fdschmidt93:
Out of laziness, I now left `action_mt` docs out, because their main purpose otherwise (apart from originally for `which_key`) I suppose is sufficiently described in `:h telescope.actions`

Works as intended in my tests. If you don't like it @Conni2461 , it's totally fine :sweat_smile:  felt like a challenge and my experiments are running on GPUs now anyways.

Closes #1869 

My test code

```lua

local M = {}

M.print_prompt = function(prompt_bufnr)
  current_picker:_get_prompt(value)
  print(prompt)
end

function print_value(prompt_bufnr)
  local value = action_state.get_current_history().value
  print(value)
end

local say_hi = function(prompt_bufnr)
  print("hi")
end

builtin.find_files({
  attach_mappings = function(prompt_buf, map)
    map("i", "X", function()
      print("test")
    end)
    map("i", "p", M.set_prompt)
    map("i", "P", print_value)
    map("i", "H", say_hi)
    map("i", "O", require "fds.utils".resize)
    return true
  end,
})
```

image

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -1128,6 +1128,16 @@ actions.which_key = function(prompt_bufnr, opts)
           table.insert(mappings, { mode = v.mode, keybind = v.keybind, name = name })
         end
       end
+    elseif type(v.func) == "function" then
+      if not opts.only_show_current_mode or mode == v.mode then
+        local fname = action_utils._get_anon_function_name(v.func)
+        table.insert(mappings, { mode = v.mode, keybind = v.keybind, name = fname })
+        utils.notify("actions.which_key", {
+          msg = "No name available for anonymous functions.",
+          level = "INFO",
+          once = true,
+        })
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


Conni2461:
Lets make this the last commit part of the 0.6 tag

Thanks :) LGTM

Conni2461:
Oh yeah. I write docs/spec for `actions.mt`.

PR #1873, opened 2022-04-22
feat: switch to telescope.lua

Conni2461:
less vim files. This is nice

especially `vim.api.nvim_create_user_command`


plugin/telescope.lua
```diff
@@ -0,0 +1,144 @@
...
+        \ default_text = [=[" . escape(getcmdline(), '"') . "]=]
+        \ }"<CR><CR>
+]==]
+
+vim.api.nvim_create_user_command("Telescope", function(opts)
+  require("telescope.command").load_command(opts.line1, opts.line2, opts.count, unpack(opts.fargs))
+end, {
+  nargs = "*",
+  range = true,
+  complete = function(_, line)
```
Conni2461, line 113: I hope the complete function still works as expected

plugin/telescope.lua
```diff
@@ -0,0 +1,144 @@
...
+}
+
+for k, v in pairs(highlights) do
+  vim.api.nvim_set_hl(0, k, v)
+end
+
+-- This is like "<C-R>" in your terminal.
+--     To use it, do `cmap <C-R> <Plug>(TelescopeFuzzyCommandSearch)
+--TODO(conni2461): Make it lua make it awesome
+vim.cmd [==[
```
Conni2461, line 101: Still todo, Didn't even know that is a thing

plugin/telescope.lua
```diff
@@ -0,0 +1,144 @@
+if 1 ~= vim.fn.has "nvim-0.7.0" then
+  vim.api.nvim_err_writeln "Telescope.nvim requires at least nvim-0.7.0. See `:h telescope.changelog-1851`"
+  return
+end
+
+if vim.g.loaded_telescope then
+  return
+end
+vim.g.loaded_telescope = true
```
clason, line 9:
```suggestion
vim.g.loaded_telescope = 1
```
for vimscript compatibility

plugin/telescope.lua
clason: Yeah, that was (just) for me :)

I've been trying to convert it myself, unsuccessfully...

Probably easiest to use `vim.keymap.set`, but keep the rhs as a vimscript command string.

plugin/telescope.lua
Conni2461: makes sense

Conni2461: oh also have to switch to == 1 then

clason: Ah, I thought you just wanted to test for existence, basically. But yes, to be on the safe side, you should absolutely test for equality with `1`!

clason:
superseded by #18219

Conni2461:
I havent had any issues so far. I'll merge and see what issues will be raised :grimacing:

PR #1879, opened 2022-04-25
docs: reflect behavior of diagnostics bufnr nil

ratmice:


Conni2461:
Needs to be updated here: https://github.com/nvim-telescope/telescope.nvim/blob/2b8c7b1ec7a3b27f1d24478c404ce4b2a82cc055/lua/telescope/builtin/init.lua#L431 because docs are autogenerated

parser does not support `@field` multiline, yet and we have a 120 line limit, so ... hard to document

ratmice:
Did the best I could must given the space constraints, couldn't figure out the magic sequence to make docgen build it however

PR #1881, opened 2022-04-26
feat: add support for builtin.commands to show buffer commands.

milanglacier:
`builtin.keymaps` show the buffer local keymap and global keymaps both.
`builtin.commands` should follow the same as the keymap as the buffer local commands should also be shown in the picker

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1101,6 +1101,9 @@ builtin.commands({opts})                        *telescope.builtin.commands()*
     Parameters: ~
         {opts} (table)  options to pass to the picker
 
+    Options: ~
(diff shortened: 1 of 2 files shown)


Conni2461:
Thanks :) Documentation is autogenerated and needs to be changed here

https://github.com/nvim-telescope/telescope.nvim/blob/6a54433038ce6d37e506ff9102ad7fcca121d58a/lua/telescope/builtin/init.lua#L235-L237

milanglacier:
Ok I will create a new PR.

Conni2461:
why open a new PR? Okay nevermind, replaced with #1889 closing

PR #1883, opened 2022-04-26
Fix the reference to vim.lsp.buf.code_action

schmidma:
This PR fixes a misspelled reference to `vim.lsp.buf.code_action`.
#1866 removed code actions from telescope and referred to `vim.lsp.buf.code_actions`. There only exists `vim.lsp.buf.code_action`


Conni2461:
My bad.

Thanks :)

PR #1887, opened 2022-04-27
Set tagstack when we jump with lsp in telescope

dagle:
Atm telescope doesn't populate the tagstack making it confusing when you jump with a telescope function, where it populates the tagstack if there is only 1 item.

### DIFF
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -133,12 +133,19 @@ action_set.edit = function(prompt_bufnr, command)
   local picker = action_state.get_current_picker(prompt_bufnr)
   require("telescope.actions").close(prompt_bufnr)
   local win_id = picker.get_selection_window(picker, entry)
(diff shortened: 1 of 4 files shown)


Conni2461:
I think we can do a better solution, similar to the `jumplist` solution https://github.com/nvim-telescope/telescope.nvim/commit/8af0d384d21104be3cb055494aab9a6f86d94067

just with this snippet

```lua
  local from = {vim.fn.bufnr('%'), vim.fn.line('.'), vim.fn.col('.'), 0}
  local items = {{tagname=vim.fn.expand('<cword>'), from=from}}
  vim.fn.settagstack(vim.fn.win_getid(), {items=items}, 't')
```

Conni2461:
I pushed my proposed version to your PR. In my opinion its more flexible. Can you test if this works for you?

dagle:
I been running it for a while and it works fine and I haven't seen any bugs so far. Sorry for the late response.

PR #1889, opened 2022-04-27
Feat: add support for builtin.commands to show buffer commands (update).

milanglacier:
`builtin.commands` should keep the same with `builtin.keymap`: show buffer local commands.


Conni2461 (approved):
LGTM thanks :)

PR #1895, opened 2022-04-29
fix(pickers): misc inital_mode setting fixes

### LINKED ISSUE #1894: Insert mode is not set at startup when A is mapped
### Description

As was mentioned in the comments on this PR(#1886), when A is mapped, it doesn't start in insert mode by default.
@fdschmidt93

### Neovim version

```markdown
NVIM v0.8.0-dev+71-g188537efb
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

ArchLinux

### checkhealth telescope

```markdown
  - OK: plenary installed.

  - OK: fd: found fd 8.3.2

```

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. `:Telescope`
3. Not in insert mode.

### Expected behavior

When telescope is started, it will be set to insert mode.

### Actual behavior

When telescope is started, it does not set to insert mode.

### Minimal config

```Lua
vim.cmd [[syntax enable]]
vim.cmd [[filetype plugin indent on]]

local install_path = vim.fn.stdpath('data')..'/site/pack/packer/opt/packer.nvim'
if vim.fn.empty(vim.fn.glob(install_path)) > 0 then
  vim.api.nvim_command('!git clone https://github.com/wbthomason/packer.nvim '..install_path)
end
vim.cmd [[packadd packer.nvim]]

vim.keymap.set("n", "A", function()
        return vim.fn.len(vim.fn.getline(".")) ~= 0 and "A" or '"_cc'
end, { noremap = true, expr = true, silent = true })

return require('packer').startup(function()
  use {'wbthomason/packer.nvim', opt = true}
  use {'nvim-telescope/telescope.nvim',requires = { {'nvim-lua/plenary.nvim'} }, event = "VimEnter", config = function() require("telescope").setup({}) end}
end)
```

fdschmidt93:
Um, yeah, not gonna lie, this is satirically stupid but seems to work.

Importantly, this does not work appropriately if startinsert is invoked otherwise (`vim.cmd` or `vim.fn.execute`)

I think we can rely on `<ESC>` though :laughing:

Fixes:
- Code cleanup between initial modes `"normal"` and `"insert"`
- `initial_mode="normal"` wouldn't function properly if original mode was not `"normal"`
- `initial_mode="insert"`: raw input `startinsert!` to avoid `A` which might be mapped otherwise and cannot be fed with `nvim_feedkeys` (see code comments)

Closes #1894

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -436,11 +436,11 @@ function Picker:find()
 
     if self.initial_mode == "insert" then
       vim.schedule(function()
-        -- startinsert! did not reliable do `A` no idea why, i even looked at the source code
         -- Example: live_grep -> type something -> quit -> Telescope pickers -> resume -> cursor of by one
+        -- vim.cmd doesn't schedule appropriately, bypass with `nvim_input` (echon to clear "startinsert" msg)
         local mode = vim.fn.mode()
         if mode ~= "i" then
-          a.nvim_input(mode ~= "n" and "<ESC>A" or "A")
+          a.nvim_input(mode ~= "n" and "<ESC>:startinsert!|echon ''<CR>" or ":startinsert!|echon ''<CR>")
         end
       end)
     elseif self.initial_mode ~= "normal" then
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -434,16 +434,22 @@ function Picker:find()
       self:set_prompt(self.default_text)
     end
 
+    local mode = vim.fn.mode()
     if self.initial_mode == "insert" then
       vim.schedule(function()
         -- Example: live_grep -> type something -> quit -> Telescope pickers -> resume -> cursor of by one
-        -- vim.cmd doesn't schedule appropriately, bypass with `nvim_input` (echon to clear "startinsert" msg)
-        local mode = vim.fn.mode()
+        -- vim.cmd doesn't schedule appropriately, bypass with `nvim_input`
+        -- echon to clear "startinsert" msg & histdel('cmd', -1) to not pollute command history with startinsert
         if mode ~= "i" then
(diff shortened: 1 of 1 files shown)


Conni2461:
image

fdschmidt93:
No worries, I feel the same about this diff. It's a cursed fix and absurd that this works haha.

I'm not on my laptop today, will triple check tomorrow and then most likely merge :)

fdschmidt93:
Fixes:
- Code cleanup between initial modes `"normal"` and `"insert"`
- `initial_mode="normal"` wouldn't function properly if original mode was not `"normal"`
- `initial_mode="insert"`: raw input `startinsert!` to avoid `A` which might be mapped otherwise and cannot be fed with `nvim_feedkeys` (see code comments)

fdschmidt93:
I'll quadruple check this throughout the day because there's quite some scenarios now to cover. Maybe I'll look into how to write appropriate tests for this, though I think if this now works we should be good apart from a person remapping `<ESC>` :laughing:

Conni2461:
Thats a lot. I think it looks good. The only thing that is still bothering me is the fact that we `vim.schedule` every time. I think we only need to do that if we go from picker to picker. Of the top of my head, i dont have a good idea how to determine that. But no vim.schedule should fix #1898 and ( #1911 ) (even thought that doesnt matter, would still be a good test case).

We could do a new PR, thought to figure out that `vim.schedule` stuff

fdschmidt93:
Yeah, let's merge this.

I've attempted writing tests for this but I think the tests always pass before the `assert` gets actually triggered :sweat_smile: 

Let's do these things in a follow-up PR.

Conni2461:
> I've attempted writing tests for this but I think the tests always pass before the assert gets actually triggered

Tj also rewrote "test framework" in fps :laughing:

PR #1902, opened 2022-04-30
chore: adapt to resolved_capabilities deprecation

clason:
Neovim deprecated `resolved_capabilities` in https://github.com/neovim/neovim/pull/17814

Replace with `server_capabilities` and adapted feature table, which is backwards compatible with 0.7.


Conni2461 (approved):
I saw the breaking change notice this morning. I wasn't aware of the backwards compatibility. 

LGTM thanks :)

clason:
Would have been much more tedious without it ;)

PR #1906, opened 2022-05-02
fix: termpreviewer scrolling

Conni2461:
CC @bluz71

This solves this issue for me


bluz71:
Yes, scrolling issue is fixed with this PR.

Great work @Conni2461

Issue #1898, opened 2022-04-30
Keys get lost when typing during Telescope's initialization


AlxHnr:
**Is your feature request related to a problem? Please describe.**

There is a small time window during Telescope's initialization, at which the keys I press are not captured. In contrast, commands from [fzf.vim](https://github.com/junegunn/fzf.vim) allow me to start mashing keys immediately, even when the UI is not visible yet and still loads. When fzf finally pops up later, all the keys I've pressed are there.

**Describe the solution you'd like**

Telescope should start capturing keys as early as possible when running `:Telescope`. The goal is to get rid of percieved latency without disrupting my flow. Otherwise I have to retry often because my search got messed up.

**Describe alternatives you've considered**

Staying with fzf.vim.

**Additional context**

This happens a lot with heavier commands like `:Telescope help_tags`. We are talking about up to 200ms in some cases, which is _definitely_ noticeable. I run Fedora 35 on an i7-3632qm.

Here is how I reproduce it 70% of the time:
1. Run `:Telescope help_tags`
2. Start typing _immediately_, e.g. a string like `telescope`
3. The search box only contains `escope`

Conni2461:
Might be fixed by https://github.com/nvim-telescope/telescope.nvim/pull/1917

Unrelated: next time open bug reports as bug reports not as feature requests

caifara:
I'm still having this problem, but after examination, it seems to be more subtle. Maybe it rings a bell …

* I have `:Telescope git_files<CR>` mapped to `<leader>t`. I also have whichkey installed. This is slow and it got worse (swallows more characters) after this (but maybe also other unrelated updates).
* I have `:Telescope live_grep<CR>` mapped to `<leader>p`. This is fast and doesn't swallow any characters. This is also setup with whichkey.
* If I remap the first command to just `T`, Telescope initializes much, much faster and doesn't swallow any characters.

Any suggestions that can explain why the first map performs so bad?

Conni2461:
You probably have another mapping under `<leader>t` (e.g. `<leader>tx`) and vim waits for you to hit the `x` key.

You can do `nnoremap <nowait><leader>t <cmd>Telescope find_files<CR>` or find a different mapping.

If that is not the issue. Please open a new bug report

caifara:
@Conni2461 Thanks, that solved the last problem!

PR #1917, opened 2022-05-04
fix: initial mode setting issues

fdschmidt93:
After some analysis, I've realized that we can avoid `vim.schedule` when switching pickers by just force-feeding going into insert mode.

Avoiding `vim.schedule` also avoids the race condition.

This way, we should also be able to use `nvim_feedkeys`.

@clason please tell me this solution doesn't have a regression for your symbol insertion? :) The reason there maybe one is because we always intermittently leave insert mode, to back into it. I'm afraid you'll now insert the symbol one to the left.. let's see.


clason:
I (very) quickly tested it and didn't see any regressions!

(Again, inserting before an existing character in column 1 doesn't work, but that's no regression.)

fdschmidt93:
Yeah, I've also got `telescope-symbols.nvim` and inserted some symbols. Seems to work fine apart from the exception you've mentioned, thanks :)

Ok, then I'll populate a list of test cases in which this worked for me and then we can hopefully merge this later today.

fdschmidt93:
@Conni2461 once I have your OK from diff I'll rerun as many tests as I can briefly and then I'd merge. Fingers crossed, but it seems like a longer lasting fix :)

Conni2461:
Yep yep yep. Nice good job. Seems to work for me and i think this also fixes #1898

fdschmidt93:
Yes, also seems to fix that issue for me.

Ok, this should hopefully stick a bit longer :laughing:

levouh:
This has worked flawlessly after using it at work all day @fdschmidt93, thanks for the quick turnaround on the fixes here.

clason:
This has introduced a regression for `actions.insert_symbol_i`, where inserting a symbol at the end of the line inserts it _before_ the last character, i.e., `foo|` -> `fo<symbol>|o`.

PR #1929, opened 2022-05-08
feat: show lua fn name if available

### LINKED ISSUE #1927: Make builtin.keymaps() work with Lua functions
In the current version of neovim there is now a possibility to map keys to a lua function. However, in this case the `telescope.builtin.keymaps()` shows simply "Lua function", which is not very informative.

 It would be great if the name of the function could be printed there instead, i.e., for a ` vim.keymap.set('n', 'gca', vim.lsp.buf.code_action)` a message "vim.lsp.buf.code_action" could be printed.

Using the old syntax with ":lua command" works, but it is not very friendly.

fdschmidt93:
Closes #1927 

Does this work for you @drybalka ? Very minimal diff and I'm not really a `builtin.keymaps` user.

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -684,8 +684,12 @@ end
 
 function make_entry.gen_from_keymaps(opts)
   local function get_desc(entry)
-    return entry.desc or entry.rhs or "Lua function"
+    if entry.callback and not entry.desc then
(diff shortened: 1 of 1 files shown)


Conni2461:
imo this is good enough. Thanks :)

You can only see the function name and not the module where it is located but you that person needs more it should just set the `desc` key in the opts tbl

PR #1935, opened 2022-05-13
get mark list from opts.bufnr instead of using :marks

pjmtdw:
`:marks` command shows local marks for current buffer and global marks.
However, current buffer is not identical to original buffer when called from `builtin.builtin`, so it won't show buffer local marks.
This P-R uses `vim.fn.getmarklist()` to mimic the behavior of `:marks` command.

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -993,11 +993,10 @@ internal.colorscheme = function(opts)
 end
 
 internal.marks = function(opts)
-  local bufnr = opts.bufnr or vim.api.nvim_get_current_buf()
   local local_marks = {
-    items = vim.fn.getmarklist(bufnr),
+    items = vim.fn.getmarklist(opts.bufnr),
     name_func = function(_, line)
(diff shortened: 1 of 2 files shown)


lua/telescope/builtin/internal.lua
```diff
@@ -993,11 +993,46 @@ internal.colorscheme = function(opts)
 end
 
 internal.marks = function(opts)
-  local marks = vim.api.nvim_exec("marks", true)
-  local marks_table = vim.fn.split(marks, "\n")
-
-  -- Pop off the header.
-  table.remove(marks_table, 1)
+  local bufnr = opts.bufnr or vim.api.nvim_get_current_buf()
```
Conni2461, line 996: this can be omitted: https://github.com/nvim-telescope/telescope.nvim/blob/39b12d84e86f5054e2ed98829b367598ae53ab41/lua/telescope/builtin/init.lua#L458

pjmtdw: Thanks, I fixed it.

Conni2461:
Thanks :)

PR #1937, opened 2022-05-15
fix: correctly use nvim-web-devicons

xeluxee:
Before:
wrong_icons

After:
correct_icons

Telescope integration with `nvim-web-devicons` used to work well with most files but not with files like `.bashrc`, `.zshrc`, `CMakeLists.txt` because their extension is different from their filetypes.
Other plugins like barbar, NvimTree and feline display these icons correctly

Also when calling `get_icon` the `extension` parameter is set to `nil` because extension is already computed by `nvim-web-devicons`: https://github.com/kyazdani42/nvim-web-devicons/blob/cde67b5d5427daeecfd7c77cf02ded23a26980bb/lua/nvim-web-devicons.lua#L1531-L1532

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -179,10 +179,14 @@ end)()
 
 utils.path_tail = (function()
   local os_sep = utils.get_separator()
-  local match_string = "[^" .. os_sep .. "]*$"
 
   return function(path)
-    return string.match(path, match_string)
+    for i = #path, 1, -1 do
+      if path:sub(i, i) == os_sep then
+        return path:sub(i + 1, -1)
+      end
+    end
+    return path
   end
 end)()
 
@@ -431,7 +435,7 @@ utils.transform_devicons = load_once(function()
         return display
       end
 
-      local icon, icon_highlight = devicons.get_icon(filename, string.match(filename, "%a+$"), { default = true })
+      local icon, icon_highlight = devicons.get_icon(utils.path_tail(filename), nil, { default = true })
       local icon_display = (icon or " ") .. " " .. (display or "")
 
       if conf.color_devicons then
@@ -461,7 +465,7 @@ utils.get_devicons = load_once(function()
         return ""
       end
 
-      local icon, icon_highlight = devicons.get_icon(filename, string.match(filename, "%a+$"), { default = true })
+      local icon, icon_highlight = devicons.get_icon(utils.path_tail(filename), nil, { default = true })
       if conf.color_devicons then
         return icon, icon_highlight
       else


Conni2461:
yep i get what you are saying. But lets not use `fnamemodify`, its not `api-fast`. we can get the tail by doing 
```lua
local parts = vim.split(filename, os_sep, true)
local name = parts[#parts]
```

xeluxee:
> yep i get what you are saying. But lets not use `fnamemodify`, its not `api-fast`. we can get the tail by doing
> 
> ```lua
> local parts = vim.split(filename, os_sep, true)
> local name = parts[#parts]
> ```

What do you mean with `api-fast`?
Also please note that `/` can be included in file names in many filesystems, and `fnamemodify` seems to handle it well

fdschmidt93:
> Most API functions are "deferred": they are queued on the main loop and
processed sequentially with normal input.  So if the editor is waiting for
user input in a "modal" fashion (e.g. the |hit-enter-prompt|), the request
will block.  Non-deferred (fast) functions such as |nvim_get_mode()| and
|nvim_input()| are served immediately (i.e. without waiting in the input
queue).  Lua code can use |vim.in_fast_event()| to detect a fast context.

`:h api-fast`

Conni2461:
> Also please note that `/` can be included in file names in many filesystems, and fnamemodify seems to handle it well

Does this really matter. We only need the filename for special files like `CMakeLists.txt` and as long as none of these files contains a `/` it doesn't matter that a file like `as/df.lua` isn't recognized correctly, because only the `.lua` part is important for that.

Using `fnamemodify` would introduce a breaking change because it would make `transform_devicons` no longer viable inside a `entry_maker` without a `make_display` when using `one_shotjob finder` e.g. No idea if someone does that but its just unnecessary.

xeluxee:
Got it

> One thing to note is that all lua seems to be 3 times slower than `fnamemodify`

Maybe that's caused by the allocation of a new table containing names between slashes.

Instead we could use lua patterns or implement custom funcion to get path tail

Conni2461:
> lua patterns

That was even slower for me

xeluxee:
@Conni2461 what about the latest commit?
Custom lua function should be faster than patterns

Conni2461:
Nice. Thanks :) It actually faster than everything else

```
|| Benchmark Group: 'path tail' -----------------------
|| Benchmark #1: 'vim.fn'
||   Time(mean ± σ):    400.1 ms ±  23.8 ms
||   Range(min … max):  378.5 ms … 463.3 ms  10 runs
|| Benchmark #2: 'vim.split'
||   Time(mean ± σ):    1.7 s ± 101.3 ms
||   Range(min … max):  1.6 s … 1.9 s  10 runs
|| Benchmark #3: 'ffi'
||   Time(mean ± σ):    199.1 ms ±   6.8 ms
||   Range(min … max):  190.3 ms … 209.3 ms  10 runs
|| Benchmark #4: 'current path_tail'
||   Time(mean ± σ):    2.4 s ±  89.2 ms
||   Range(min … max):  2.3 s … 2.6 s  10 runs
|| Benchmark #5: 'new path_tail'
||   Time(mean ± σ):    132.0 ms ±  34.8 ms
||   Range(min … max):  114.8 ms … 230.7 ms  10 runs
|| Summary
||   'new path_tail' ran
||   3.0 ± 0.8 times faster than 'vim.fn'
||   13.1 ± 3.5 times faster than 'vim.split'
||   1.5 ± 0.4 times faster than 'ffi'
||   18.5 ± 4.9 times faster than 'current path_tail'
```

Can you fix stylua CI after that i merge. Thank you very much :)

xeluxee:
> Nice. Thanks :) It actually faster than everything else

Nice to hear that!

> Can you fix stylua CI after that i merge. Thank you very much :)

Well, stylua is going to format 6646 lines in the entire project. Should I format only `utils.lua`?

Conni2461:
> Well, stylua is going to format 6646 lines in the entire project. Should I format only utils.lua?

Excuse me? :laughing:  Did you run `stylua lua/` or `stylua .` because only the `lua/` directory is covered by stylua (currently)

xeluxee:
> Excuse me? Did you run `stylua lua/` or `stylua .` because only the `lua/` directory is covered by stylua (currently)

My bad, I've run `stylua .` instead of `stylua lua/` 😅

Conni2461:
Thanks :)

xeluxee:
Thank you for co-maintaining this awesome plugin!

Issue #1940, opened 2022-05-17
Make delete_buffer force deletion for terminal windows


TheMeaningfulEngineer:
**Is your feature request related to a problem? Please describe.**

The delete_buffer isn't handling the edge cases for the moment. 

**Describe the solution you'd like**

## Acceptance criteria:

- [x] Force buffer deletion for terminal
- [x] If the non-terminal buffer can't be deleted (because of unsaved changes)
    -  Print a human friendly error message
    -  Don't remove the item from the selection
- [ ] Add debug logs which would allow for deeper diagnostic if needed (if that exists as a mechanism)

fdschmidt93:
If I understand it correctly, a PR for this seems very straightforward and would be very welcome :)

TheMeaningfulEngineer:
Yup, planning to make a PR from this, looks like a nice first issue :)

Conni2461:
#1943 was merged in the dev branch.

Issue #1938, opened 2022-05-16
telescope 0.1 release announcement


Conni2461:
Hey,

we(tj, @fdschmidt93 and me) had our first telescope maintainer meeting a couple of days ago, and we discussed the first release, which is only a couple of weeks away. We plan to release telescope 0.1 on July 12, the second birthday of telescope.
 
Foremost, I have to thank a tone of people. Thanks to all 220 people who contributed to telescope, thanks to everyone that opened issues and discussed problems and improvements. We almost have 1000 commits on the main branch, most of them squashed. It's insane what the community achieved in the past two years. We are really grateful for your support.

Coming back to the release. We discussed a couple of things, including features we want to have included in the first release and some breaking changes/cleanup we want to do prior to the release. To make these breaking changes easier for end users, we decided we merge them into a separate [dev](https://github.com/nvim-telescope/telescope.nvim/tree/dev) branch and write a changelog for every breaking change before everything will then be merged in master.
The general timeline will be a feature freeze in around 4 weeks, 1/2 weeks later we merge the [dev](https://github.com/nvim-telescope/telescope.nvim/tree/dev) branch into master and release a first release candidate. Following by the 0.1 release 2/3 weeks later. We suggest not running the [dev](https://github.com/nvim-telescope/telescope.nvim/tree/dev) branch, though we appreciate everyone who is willing to testing it :) The master branch will for the time being only receiving necessary fixes.

You can follow the current development state here: https://github.com/nvim-telescope/telescope.nvim/projects/2 and I will go over all open issues and PRs in the next couple of days and check what could be done.

I'll leave this issue open for discussion, but I kindly ask you to open new issues for features you want to see in this release (or comment under the issue, if one already exists).

Edit: The dev PR can be tracked here: https://github.com/nvim-telescope/telescope.nvim/pull/1945

Edit 2:

> **Note**
> dev branch will receive regular rebases and force pushes. Only run it if you know what you do :)

Edit 3: Updated time schedule, see comment below (after a bunch of mentions)

Conni2461:
took us a long time, a lot of work but here it is https://github.com/nvim-telescope/telescope.nvim/releases/tag/0.1.0

mehalter:
Hey @Conni2461 regarding these tagged releases, is there a reason you are avoiding including features in the 0.1.x releases? I believe you are doing it because of the semantic versioning specification summary saying:

> PATCH version when you make backward compatible bug fixes

[...]

PR #1943, opened 2022-05-18
Force buffer delete for terminal

TheMeaningfulEngineer:
Closes https://github.com/nvim-telescope/telescope.nvim/issues/1940

Acceptance criteria listed in the issue.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -1021,8 +1021,8 @@ actions.delete_buffer = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   current_picker:delete_selection(function(selection)
     local force = vim.api.nvim_buf_get_option(selection.bufnr, "buftype") == "terminal"
-    local success, function_return = pcall(vim.api.nvim_buf_delete, selection.bufnr, { force = force })
-    return success
+    local ok = pcall(vim.api.nvim_buf_delete, selection.bufnr, { force = force })
+    return ok
   end)
 end
 
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -1020,7 +1020,9 @@ end
 actions.delete_buffer = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   current_picker:delete_selection(function(selection)
-    vim.api.nvim_buf_delete(selection.bufnr, { force = false })
+    local force = vim.api.nvim_buf_get_option(selection.bufnr, "buftype") == "terminal"
+    local success, function_return = pcall(vim.api.nvim_buf_delete, selection.bufnr, { force = force })
+    return success
   end)
 end
 
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -692,7 +692,7 @@ end
 ---
 --- Example usage in telescope:
 ---   - `actions.delete_buffer()`
----@param delete_cb function: called with each deleted selection
+---@param delete_cb function: called for each selection fn(s) -> bool|nil (true|nil removes the entry from the results)
 function Picker:delete_selection(delete_cb)
   vim.validate { delete_cb = { delete_cb, "f" } }
   local original_selection_strategy = self.selection_strategy
@@ -718,10 +718,9 @@ function Picker:delete_selection(delete_cb)
     return x > y
   end)
   for _, index in ipairs(selection_index) do
-    local selection = self.finder.results[index]
-    local delete_cb_return = delete_cb(selection)
-    if delete_cb_return then
-        local selection = table.remove(self.finder.results, index)
+    local delete_cb_return = delete_cb(self.finder.results[index])
+    if delete_cb_return == nil or delete_cb_return == true then
+      table.remove(self.finder.results, index)
     end
   end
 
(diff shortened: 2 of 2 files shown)


lua/telescope/actions/init.lua
```diff
@@ -1017,7 +1017,11 @@ end
 actions.delete_buffer = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   current_picker:delete_selection(function(selection)
-    vim.api.nvim_buf_delete(selection.bufnr, { force = false })
+    if vim.api.nvim_buf_get_option(selection.bufnr, "buftype") == "terminal" then
+      vim.api.nvim_buf_delete(selection.bufnr, { force = true })
```
Conni2461, line 1021: how about

```lua
local force = vim.api.nvim_buf_get_option(selection.bufnr, "buftype") == "terminal"
vim.api.nvim_buf_delete(selection.bufnr, { force = force })
```

TheMeaningfulEngineer: Sure, don't have a strong opinion on this. Will change it.

TheMeaningfulEngineer:
@Conni2461 
I was hoping to add some prettier error handling for cases where one tries to close an unsaved buffer.
If you try doing it now there is a lot of ugly red errors printed.

Do you know how can the errors from `vim.api.nvim_buf_delete(selection.bufnr, { force = force })` be handled to show a human friendlier error?

I've tried wrapping it with `pcall` but that didn't prevent the errors from printing.

Conni2461:
Doing 

`pcall(vim.api.nvim_buf_delete, selection.bufnr, { force = force })` goes from this error:
```
E89: No write since last change for buffer 18 (add ! to override)
E5108: Error executing lua [string ":lua"]:1: Failed to unload buffer.
stack traceback:
        [C]: in function 'nvim_buf_delete'
```
to this
```
E89: No write since last change for buffer 18 (add ! to override)
```

I think the latter one in an improvement, and you could add the `pcall`. You can't catch that error because `pcall` only catches lua `error`s and this seems to come from somewhere in neovim (old viml land).

TheMeaningfulEngineer:
Thanks @Conni2461.
However enclosing the call with `pcall` doesn't change the verbosity level for me.
I've pushed the [second commit](https://github.com/nvim-telescope/telescope.nvim/pull/1943/commits/b9bd96e4957160cad5687041fbb5626f084656ba) with this.
This code produces the exact same error as when `pcall` isn't present.

Does running this on your machine gives you the reduced error log?

lua/telescope/pickers.lua
```diff
@@ -716,8 +716,14 @@ function Picker:delete_selection(delete_cb)
     return x > y
   end)
   for _, index in ipairs(selection_index) do
-    local selection = table.remove(self.finder.results, index)
-    delete_cb(selection)
+    -- WIP: Trying to implement logic which will not remove the item from selection if buffer deletion failed
+    --
+    local selection = self.finder.results[index]
+    local no_error, function_return = pcall(delete_cb, selection)
```
Conni2461, line 722: hmm we can do this but we should not do a interface around pcall.

We can say that `delete_cb` returns either `nil` or a `boolean`. If its `true` or `nil` (so we dont break backwards compat) we remove it otherwise we dont

lua/telescope/pickers.lua
```diff
@@ -716,8 +716,14 @@ function Picker:delete_selection(delete_cb)
...
+    if not no_error then
+        local selection = table.remove(self.finder.results, index)
+      return
```
Conni2461, line 725: We should not return here! We are in a loop of maybe 10 multiselected items. And i expect that all are removed on delete, even if the first one can't be closed (or something like this)

lua/telescope/pickers.lua
TheMeaningfulEngineer: Will gladly implement it this way, but would like to understand the logic behind the backwards compatibility argument.
I didn't find that the `delete_cb` is run within `pcall` somewhere else.

What would break if the exception gets cough at that point?
Would it not propagate correctly to the upper layers and that's the compatibility your're worried about?

(full disclosure, I come mostly from Python in terms for how I think exceptions work, from what I read, lua is similar in the sense that exceptions propagate up the stack if uncouth. If something I say doesn't make sense, it's likely coming from that incorrect mental model)

Conni2461: lua works the same, there is just no reason here to build an interface around passing exceptions up the stack. You already caught the exception in the callback.

`delete_selection` can be used outside of this project as well for other pickers. We dont have to keep backwards compatibility and have the callback be `function(selection) -> boolean` only, its still better than only building this interface on exceptions

TheMeaningfulEngineer: Thanks

> We dont have to keep backwards compatibility and have the callback be function(selection) -> boolean only, its still better than only building this interface on exceptions

I guess the docs also need a change (defining the return values [here](https://github.com/nvim-telescope/telescope.nvim/blob/ee6111952a8c76b65e59b26a7efe62efb1998590/lua/telescope/pickers.lua#L693)) as that's the way to communicate the interface of the callback function?

TheMeaningfulEngineer:
@Conni2461 
Hey, can this be merged?

Conni2461:
Thanks for the PR :) i cleaned it up, wrote docs and changed the target branch to the `dev` branch following https://github.com/nvim-telescope/telescope.nvim/issues/1938 because i dont consider this a fix and only fixes are currently merged into master (you've also opened this as feature request). It will hit master in about 2 weeks.

Thanks again for improving `actions.delete_buffer` as well as `Picker:delete_selection`

PR #1945, opened 2022-05-18
dev branch (including breaking changes)

Conni2461:
> **Note**
> This branch will receive regular rebases and force pushes. Only run it if you know what you do :)

Prior to merge, write changelog of breaking changes

- break(git_files): change show_untracked default to false
- break: deprecate `utils.get_default utils.if_nil`
- break: drops `ignore_filename`, use `path_display= { "hidden" }`
- break: prefix internal interfaces with `__` so `require("telescope.builtin.files").find_files` will show a notify error but still works for now. The error will be removed prior to 0.1! You should use `require("telescope.builtin").find_files` because we wrap all the functions that are exposed in this module. You should only access that one this applies to all git, lsp, ... builtins
- break: `defaults.preview.treesitter` rework that allows you to either enable a list of languages or enable all and disable some. Please read `:help telescope.defaults.preview` for more information


lua/telescope/actions/init.lua
```diff
@@ -1020,7 +1020,9 @@ end
 actions.delete_buffer = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   current_picker:delete_selection(function(selection)
-    vim.api.nvim_buf_delete(selection.bufnr, { force = false })
+    local force = vim.api.nvim_buf_get_option(selection.bufnr, "buftype") == "terminal"
+    local ok = pcall(vim.api.nvim_buf_delete, selection.bufnr, { force = force })
```
jamestrew, line 1024: In my testing, wrapping `nvim_buf_delete` with a `pcall` will not catch the vim error caused by trying to delete an unsaved/changed buffer.

Maybe we can add something like below above this line so users don't see a big error
```lua
if vim.fn.getbufinfo(selection.bufnr)[1].changed then
  -- maybe add in a less intrusive msg
  return
end
```

Conni2461:
According to schedule this is being merged in time. :laughing: 

Here goes nothing. `:help telescope.changelog-1945`

doc/telescope_changelog.txt
```diff
@@ -195,5 +195,64 @@ https://github.com/stevearc/dressing.nvim which has support for multiple
...
+- feat: Add entry_index for entry_makers (#1850)
+- feat: refine with new_table (#1115)
+
+The last one is one of the most existing new features, because it  allows  you
+to go from live_grep into a fuzzy environment with the following mapping
+`<C-Space>`. It's a general interface we now implemented for `live_grep` and
+`lsp_dynamic_workspace_symbols` but it could also be easily implemented for
+other builtins, by us or the user. It's now available for extension developers.
+We will add documentation in the next couple of days and improve it by adding
+more options to configure it after the initial 0.1 release.
```
clason, line 224: I hope you will also add a name for it since the current anonymous function
1. looks weird in `which-key` 
2. cannot be remapped (which is a problem on macOS where `c-space` is a default system shortcut)

:D

Conni2461: yeah makes sense. I'll make the proposed changes

PR #1947, opened 2022-05-18
chore: cleanup autocmd builtin

Conni2461:
This is so much cleaner, but we no longer have the source file and linenumber where the autocmd was defined. Currently we get the information with `verbose` and open the appropriate file on `<CR>`

I dont see an open other than parsing `verbose autocmd` again on `<CR>` or removing it or patching neovim. `help nvim_get_option_info()` also returns

```
                • last_set_sid: Last set script id (if any)
                • last_set_linenr: line number where option was set
```

So i dont think it is that wrong to include it (maybe hide behind a verbose option)

### DIFF
--- a/lua/telescope/builtin/internal.lua
+++ b/lua/telescope/builtin/internal.lua
@@ -1170,95 +1170,44 @@ internal.highlights = function(opts)
 end
 
 internal.autocommands = function(opts)
-  local autocmd_table = {}
-
-  local pattern = {}
-  pattern.BUFFER = "<buffer=%d+>"
-  pattern.EVENT = "[%a]+"
-  pattern.GROUP = "[%a%d_:]+"
-  pattern.INDENT = "^%s%s%s%s" -- match indentation of 4 spaces
-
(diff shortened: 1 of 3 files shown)


milanglacier:
After this rework. the autocommand picker does not show `desc` when an autocmd  is defined via `nvim_create_autocmd` and use a lua callback function rather than a cmd string (vimscript),
as an example, I have the following autocommand:

```lua
local augroup = vim.api.nvim_create_augroup
local autocmd = vim.api.nvim_create_autocmd
local command = vim.api.nvim_create_user_command

M.my_augroup = augroup('MyAugroup', {})

autocmd('FileType', {
    group = M.my_augroup,
    pattern = { 'r', 'rmd' },
    desc = 'set r, rmd keyword pattern to include .',
    callback = function()
        vim.bo.iskeyword = vim.bo.iskeyword .. ',.'
    end,
})

```

In older version, this autocommand will show the `desc` option in the picker, but in the new version, it only shows nothing

PR #1954, opened 2022-05-21
chore: get started with picker docs

fdschmidt93:
I hope to have it finished by 0.1 :laughing: we should have this, then finally users can find some important stuff.

Lots to do

- [ ] Intro segment with examples
- [ ] Annotate functions
- [ ] Complete `Picker.new`

Surely more things I cannot think of yet.

If anyone wants to help, please just make PRs / push straight away to the branch :)


Conni2461:
Note that fps is still on going and is changing pickers.lua significant

PR #1963, opened 2022-05-24
Add support for no-ignore-parent option in find_files.

radglob:
`ripgrep`, `fd` and `find` all support the `no-ignore` option, which doesn't obey ignore files present in the local directory. This is currently exposed in the API for `find_files`. A similar option, `no-ignore-parent`, allows the underlying program to not obey ignore files in parent directories.

This adds support for the latter, and covers the cases of other programs that do not support this option.

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -878,19 +878,22 @@ builtin.find_files({opts})                    *telescope.builtin.find_files()*
         {opts} (table)  options to pass to the picker
 
     Options: ~
-        {cwd}          (string)   root dir to search from (default: cwd, use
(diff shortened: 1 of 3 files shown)


Conni2461:
Thanks :) I changed the target to the current dev branch, because its not a fix that needs to hit right now but rather a new feature. For more information on the dev branch, please read this announcement #1938 :)

Can you write docs for it here https://github.com/nvim-telescope/telescope.nvim/blob/cd9e6aaee01624c1a5423a783e7fd5bc8fc955c5/lua/telescope/builtin/init.lua#L82-L83

Issue #1617, opened 2021-12-21
builtin current_buffer_tags show nothing when using vim-gutentags to generate tags file


phongnh:
### Description

Maybe dupplicate of https://github.com/nvim-telescope/telescope.nvim/issues/1309.

I use https://github.com/ludovicchabant/vim-gutentags to generate tags files, because tag files are stored in other folder, so the tag path is absolute path.

Another way to reproduce, it uses `ctags -R --tag-relative=never` to generate tags file with tag path is absolute.

### Neovim version

```markdown
$ nvim --version
NVIM v0.7.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

### Operating system and version

Pop!OS 20.04

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - WARNING: nvim-treesitter not found.

## Checking external dependencies
  - OK: rg: found ripgrep 13.0.0 (rev af6b6c543b)
  - OK: fd: found fd 8.3.0

## ===== Installed extensions =====

## Telescope Extension: `fzf`
  - INFO: No healthcheck provided
```

### Steps to reproduce

1. `mkdir telescope-test`
2. `cd telescope-test`
3. `echo "let g:telescope = 1" > test.vim`
4. `ctags -R --tag-relative=never`
5. nvim -u minimal.lua
6. `:edit test.vim`
7. `:Telescope current_buffer_tags`

### Expected behavior

Show tags for the buffer `test.vim`

### Actual behavior

Show nothing.

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          { 'nvim-telescope/telescope-fzf-native.nvim', run = 'make' },
        },
      },
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
      display = { non_interactive = true },
    },
  }
end
[...]

PR #1314, opened 2021-10-07
fix(tags): respect tagrelative option

Conni2461:
fix #1309

We could refactor some of that tags code. So i might not merge soon

PR #1970, opened 2022-05-26
Force ctags entries to use relative paths

dsummersl:
This fixes https://github.com/nvim-telescope/telescope.nvim/issues/1617 by ensuring any ctags file entries are relative to the cwd.

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -954,12 +954,11 @@ function make_entry.gen_from_ctags(opts)
     end
 
     if opts.only_current_file then
-      local cache_key = current_file .. file
-      if current_file_cache[cache_key] == nil then
-        current_file_cache[cache_key] = Path:new(file):normalize(cwd) == current_file
+      if current_file_cache[file] == nil then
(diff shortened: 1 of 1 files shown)


Conni2461:
Thanks :)

Also related https://github.com/nvim-telescope/telescope.nvim/pull/1314 (we probably need both) I have an idea on how to improve this a little bit, i will add a commit to it in the next couple of days :)

Conni2461:
The current merge target is the `dev` branch. I rebased and improved it a little bit. Can you retest if it still works (it does for me)

dsummersl:
@Conni2461 didn't work for me. I think the cache needs to contain the `file` and `current_file`, otherwise whatever the first cached value for the first `file` is used for all subsequent checks. Updated to reflect that logic.

Conni2461:
Oh yeah i am stupid, I should have used `file` as key. We don't even need to do `current_file .. file` because for `current_buffer_fuzzy_find` `current_file` is always the same filename.

Can you update that, check again, after that i would merge into `dev` branch thanks :)

dsummersl:
worked, updated!

Conni2461:
Thanks :)

PR #1974, opened 2022-05-27
feat: add git_live_grep

antoinemadec:
git_live_grep is to live_grep what git_files is to find_files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,7 +263,7 @@ Built-in functions. Ready to be bound to any key you like.
 | `builtin.git_files`                 | Fuzzy search through the output of `git ls-files` command, respects .gitignore, optionally ignores untracked files                |
 | `builtin.grep_string`               | Searches for the string under your cursor in your current working directory                                                       |
 | `builtin.live_grep`                 | Search for a string in your current working directory and get results live as you type (respecting .gitignore)                    |
-| `builtin.git_grep`                  | Search for a string in the files tracked by Git, uses the output of `git grep` command                                            |
+| `builtin.git_live_grep`             | Search for a string in the files tracked by Git, uses the output of `git grep` command                                            |
 
 ### Vim Pickers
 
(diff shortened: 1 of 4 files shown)


antoinemadec:
@Conni2461 maybe that could be its own separate plugin ?
I was thinking of creating:

- git_browse live_grep
- git_browse grep_string
- git_browse commit_msg
- git_browse commit_diff
- git_browse commit_author

What do you think ?

Conni2461:
I wanted to argue that this doesn't really need to be in telescope core. Its already somewhat possible with this, there are just no supported shorthands and we dont have git cwd detection.

`lua require("telescope.builtin").live_grep({ vimgrep_arguments = {"git", "grep", "--line-number", "--column", "-I", "--ignore-case"} })`

If you want to make an extension feel free, we already argued for this some time ago anyway :laughing: see #1228

antoinemadec:
@Conni2461 got it.
To me, the git grep feature is a must have, but after searching for similar requests on the internets it seems that people don't really care about it :laughing: 

Another issue is the fact that ```git_commits``` does not respect the chronological order of commits once you start filtering them.
I looked for a builtin sorter which only filters and don't reorder, but I didn't find one. So I created one myself.
However, there is no generic ```get_score()``` that could be overriden by extensions like ```fzf-native``` etc. So I had to point to fzy's ```get_score()``` specifically.

Long story short: I will create an extension.
When it's done, we can discuss including code in the builtin such as the **sorter which respects the order**.

Thanks for the excellent plugin :heart:

Conni2461:
I think this can be done with any sorter. See `:help telescope.defaults.tiebreak`

You then can pass a opt `tiebreak = function() return false end` to `git_commits` / `git_bcommits`

We can even argue that we should make that default for those. If that works, PR welcome :)

antoinemadec:
It does not work for me, here is what I tried:
```lua
  local tiebreak = function() return false end
  opts.defaults.tiebreak = tiebreak;
  opts.tiebreak = tiebreak;
```

After reading the help, it seems not to do what I want.
It's supposed to only break a tie. What I want is to keep the order whatever the score is (unless it is -1 i.e. filtered).
This is what I wrote, and what works for me:
```lua
local sorter_preserve_order = function(opts)
  opts = opts or {}
  local fzy = opts.fzy_mod or require "telescope.algos.fzy"

  return sorters.Sorter:new {
    scoring_function = function(_, prompt, line, _)
      if not fzy.has_match(prompt, line) then
        return -1
      end
      return 1
    end,

    highlighter = function(_, prompt, display)
      return fzy.positions(prompt, display)
    end,
  }
end
```

Issue #1953, opened 2022-05-21
The results of lsp_definitions are different.


ryoppippi:
### Description

The results are different

![スクリーンショット 2022-05-21 14 34 01](https://user-images.githubusercontent.com/1560508/169654113-38390b98-8ee7-461d-ba21-f8747eb099a1.png)

![スクリーンショット 2022-05-21 14 34 26](https://user-images.githubusercontent.com/1560508/169654115-14961c0b-2538-4732-b718-12c83a32211e.png)

### Neovim version

```markdown
NVIM v0.7.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@HMBRW-A-001-M1-004.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.7.0/share/nvim"

Run :checkhealth for more info
```

### Operating system and version

macOS 12.3.1

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - OK: nvim-treesitter installed.

## Checking external dependencies
  - OK: rg: found ripgrep 13.0.0
  - OK: fd: found fd 8.3.2

## ===== Installed extensions =====

## Telescope Extension: `changes`
  - INFO: No healthcheck provided

## Telescope Extension: `file_browser`
  - INFO: No healthcheck provided

## Telescope Extension: `frecency`
  - INFO: No healthcheck provided

## Telescope Extension: `fzf`
  - INFO: No healthcheck provided

## Telescope Extension: `heading`
  - INFO: No healthcheck provided

## Telescope Extension: `media_files`
  - INFO: No healthcheck provided

## Telescope Extension: `notify`
  - INFO: No healthcheck provided

## Telescope Extension: `ui-select`
  - INFO: No healthcheck provided
```

### Steps to reproduce

1. clone this repository https://github.com/ryoppippi/telescope_type_definitions_bug
2. execute `yarn`
3. open src/main.ts
4. move cursor to the word `message` at 6:32
5. execute `:lua vim.lsp.buf.lsp_type_definition()` and `:Telescope lsp_type_definition` on  `message`

### Expected behavior

Should be the same results

### Actual behavior

The results are different

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
[...]

PR #1976, opened 2022-05-27
feat: add `no_wildignore` option to `find_files`

loqusion:
This PR makes `telescope.builtin.find_files` respect the `'wildignore'` vim option by default, unless the user overrides it by setting `no_wildignore` to `true`.

## Motivation

`telescope.builtin.find_files` hides hidden files (`*/.*`) by default, but sometimes I want to show hidden files, for example configuration files like `.eslintrc`, `.prettierrc`, `.travis.yml`, etc. But setting `hidden` to `true` shows files in `.git` first and foremost, which shoves aside the files I actually want to work with. (`.gitignore` typically does not include `.git` since it is ignored by git anyway.)

Screen Shot 2022-05-27 at 11 07 04 AM

## Solution

The solution I propose is to respect the user's `'wildignore'` by default and allow them to override it by passing `no_wildignore = true` to `telescope.builtin.find_files`.

One possible solution would be instead to hide files in `.git` by default regardless of what `hidden` is set to, and possibly supply a `git` option to disable this behavior. Here are a few reasons why I think respecting `'wildignore'` by default is better:

- Hiding `.git` when `hidden` is `true` might be confusing to users
- `'wildignore'` is empty by default, so no weird behavior unless the user overrides it
- `'wildignore'` offers more flexibility
- `'wildignore'` exists in vanilla vim and is already respected by vim functions such as `expand()`, `glob()`, and `globpath()`; this orthogonality multiplies the utility of the concept the user already knows

Screen Shot 2022-05-27 at 11 37 07 AM

I'm new to writing lua and vim/neovim plugins in general, so feedback is appreciated!

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -200,7 +200,12 @@ files.find_files = function(opts)
     end
     if not no_wildignore then
       for _, ignore_pattern in pairs(vim.opt.wildignore:get()) do
-        table.insert(find_command, "--exclude")
+        if command == "rg" then
+          table.insert(find_command, "--glob")
+          ignore_pattern = "!" .. ignore_pattern
+        else
+          table.insert(find_command, "--exclude")
+        end
         table.insert(find_command, ignore_pattern)
       end
     end
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -181,6 +181,7 @@ files.find_files = function(opts)
   local command = find_command[1]
   local hidden = opts.hidden
   local no_ignore = opts.no_ignore
+  local no_wildignore = opts.no_wildignore
   local follow = opts.follow
   local search_dirs = opts.search_dirs
 
@@ -196,6 +197,12 @@ files.find_files = function(opts)
     end
     if no_ignore then
       table.insert(find_command, "--no-ignore")
+    end
+    if not no_wildignore then
+      for _, ignore_pattern in pairs(vim.opt.wildignore:get()) do
+        table.insert(find_command, { "--exclude", ignore_pattern })
+      end
+      find_command = flatten(find_command)
     end
     if follow then
       table.insert(find_command, "-L")
@@ -216,6 +223,19 @@ files.find_files = function(opts)
     if no_ignore ~= nil then
       log.warn "The `no_ignore` key is not available for the `find` command in `find_files`."
     end
+    if not no_wildignore then
+      local wildignore = vim.opt.wildignore:get()
+      local wildignore_end = #wildignore
+      table.insert(find_command, { "-not", "(" })
+      for idx, ignore_pattern in pairs(wildignore) do
+        table.insert(find_command, { "-path", ignore_pattern })
+        if idx ~= wildignore_end then
+          table.insert(find_command, "-or")
+        end
+      end
+      find_command = flatten(find_command)
+      table.insert(find_command, ")")
+    end
     if follow then
       table.insert(find_command, 2, "-L")
     end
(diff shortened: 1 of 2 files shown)


lua/telescope/builtin/files.lua
```diff
@@ -197,6 +198,12 @@ files.find_files = function(opts)
     if no_ignore then
       table.insert(find_command, "--no-ignore")
     end
+    if not no_wildignore then
+      for _, ignore_pattern in pairs(vim.opt.wildignore:get()) do
+        table.insert(find_command, { "--exclude", ignore_pattern })
+      end
+      find_command = flatten(find_command)
```
Conni2461, line 205: you dont need to flatten if you do 

```lua
table.insert(find_command, "--exclude")
table.insert(find_command, ignore_pattern)

File: lua/telescope/builtin/files.lua [hunk 2]
```diff
@@ -216,6 +223,19 @@ files.find_files = function(opts)
     if no_ignore ~= nil then
       log.warn "The `no_ignore` key is not available for the `find` command in `find_files`."
     end
+    if not no_wildignore then
+      local wildignore = vim.opt.wildignore:get()
+      local wildignore_end = #wildignore
+      table.insert(find_command, { "-not", "(" })
```
Conni2461, line 229: same here

Conni2461:
Thanks :) I changed the target to the current dev branch, because its not a fix that needs to hit right now but rather a new feature. For more information on the dev branch, please read this announcement https://github.com/nvim-telescope/telescope.nvim/issues/1938 :)

fdschmidt93:
> So can we rather have a respect_wildignore option that is false on default and people need to opt in?

Conni2461:
It makes sense to do this through the command and not thought `file_ignore_patterns`. It could lead to better performance and `file_ignore_patterns` could cause these problems described here: https://github.com/nvim-telescope/telescope.nvim/issues/1953#issuecomment-1133655114 (I know you can only set it as opts where you need it but who gonna do that?)

Initially i only had a problem with the default behaivor but now i am thinking that `opts.find_command` could maybe be either a string (like it is right now) or a function returning a `find_command` table. So you can setup whatever you want, like support for `wildignore`. And then we can do another configuration receipt

loqusion:
I just realized that the `file_ignore_patterns` you guys were discussing is the one passed to `telescope.setup`, not `telescope.builtins.find_files`, my bad 😅. That's why I was talking about passing it to `find`'s `-regex`.

`file_ignore_patterns` being applied to all `telescope` commands could definitely cause unexpected behavior, but what about passing it as an option to `find_files`? The only issue I have with `opts.find_command` is that it implicitly makes the command it uses a dependency of the user's vim config. `opts.file_ignore_patterns` is a more portable solution.

fdschmidt93:
> I just realized that the file_ignore_patterns you guys were discussing is the one passed to telescope.setup, not telescope.builtins.find_files, my bad

Both is possible. I'm pretty sure Conni's point was that people sometimes just forget about the 

```lua
require "telescope".setup {
  pickers = {
    find_files = {
      file_ignore_pattern = ...
    }
  }
}
```

> The only issue I have with opts.find_command is that it implicitly makes the command it uses a dependency of the user's vim config.

That's what we want though. A user can tie the `opts.find_command` to their user config, if `opts.find_command` could be a lua function. As in, you would like to use `wildignore` as per this PR, then you are happy to incorporate that as part of the function passed `opts.find_command`. Then, we'd add this functionality as a configuration recipe to the wiki.

loqusion:
> Both is possible. I'm pretty sure Conni's point was that people sometimes just forget about the
> 
> ```lua
> require "telescope".setup {
>   pickers = {
>     find_files = {
>       file_ignore_pattern = ...
>     }
>   }
> }
> ```
> 
> way to config stuff which I think was the problem of the linked issue (to lazy to look at it yet again).

Ah, I didn't realize this existed. In that case, this PR is probably unnecessary. Thanks anyway for your consideration!

Conni2461:
I just merged that `find_command` can be a function into dev branch

So you can now do the following, to have support for this feature
```lua
require "telescope".setup {
  pickers = {
    find_files = {
      find_command = function()
        local command = { "fd", "--type", "f" }
        for _, ignore_pattern in pairs(vim.opt.wildignore:get()) do
          table.insert(find_command, "--exclude")
          table.insert(find_command, ignore_pattern)
        end
        return command
      end
    }
  }
}
```

So this PR is no longer needed. Still thanks for your work here :)

PR #1979, opened 2022-05-30
fix: cursor move in actions.insert_symbol_i

fdschmidt93:
@clason I hope this fixes your issue.

The issue sort of persists if you insert at beginning of a line right before a word, but IIRC you've mentioned before that this is problem already existed.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -459,7 +459,8 @@ end
 ---@param prompt_bufnr number: The prompt bufnr
 actions.insert_symbol_i = function(prompt_bufnr)
   local symbol = action_state.get_selected_entry().value[1]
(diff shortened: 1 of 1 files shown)


clason:
Yes, that fixes the issue, thank you very much!

I really need to find a better way of doing that; the current way is just too brittle and breaks too easily when making reasonable changes elsewhere...

fdschmidt93:
Ok great! Your comment got me thinking that we maybe want to solve this another way; @Conni2461 any preference between the two commits to solve the problem? I suppose maybe the second one is more generic.

clason:
(still working 😁 )

Conni2461:
second one seems good :)

Conni2461 (approved):

PR #1982, opened 2022-06-02
fix: error in action.close when picker.original_win_id is not valid

jedrzejboczar:
A concrete scenario that leads to an error can be found in [this issue](https://github.com/jedrzejboczar/toggletasks.nvim/issues/3#issuecomment-1144574061), but in short: if `picker.original_win_id` has already been closed and is no longer valid, an error occurs inside `actions.close` when calling `nvim_win_get_cursor`. This PR just wraps the call using `pcall`.


Conni2461:
Makes sense. Thanks :)

PR #1985, opened 2022-06-02
fix(health): update to upstream changes

clason:
The `health` module was moved to `vim.health` in https://github.com/neovim/neovim/pull/18720


clason:
looks like there's an echo here 😅

Issue #1981, opened 2022-06-01
need to wait until finder command is done


antoinemadec:
### Description

When a finder takes multiple seconds to generate all the entries, there are 2 issues:

1. the screen refresh is frozen unless we move the cursor (up/down arrow)
2. the sorter does not take effect until the command is (almost) done

### Neovim version

```markdown
NVIM v0.8.0-dev
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

ManjaroLinux 21.2.6

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - WARNING: nvim-treesitter not found.

## Checking external dependencies
  - OK: rg: found ripgrep 13.0.0
  - OK: fd: found fd 8.3.2
```

### Steps to reproduce

1. cd <big_project> (e.g.: neovim git repo)
2. `nvim --clean -u minimal.vim`
3. `:lua require("telescope.builtin").grep_string({search = ""})`
4. type pattern in prompt, use up and down arrow

**I tried with and without [telescope-fzf-native.nvim](https://github.com/nvim-telescope/telescope-fzf-native.nvim) and observed the same behavior**

### Expected behavior

I expect the screen to be refreshed without user input and the sort to take effect immediately.
This is how ```fzf.vim``` deals with commands that take time.

It should behave like (even if slower than) this command:
`rg --color=never --no-heading --with-filename --line-number --column --smart-case -- "" | fzf`
 

### Actual behavior

1. the screen refresh is frozen unless we move the cursor (up/down arrow)
2. the sorter does not take effect until the command is (almost) done

### Minimal config

```Lua
set rtp+=/home/antoine/.local/share/nvim/site/pack/packer/start/plenary.nvim
set rtp+=/home/antoine/.local/share/nvim/site/pack/packer/start/telescope.nvim
lua << EOF
require('telescope').setup()
EOF
```

fdschmidt93:
For `grep_string` `telescope` does essentially _everything_ (processing `rg` output, sorting, drawing, ...) on the neovim main thread in lua. There was an essentially similar note on reddit recently about telescope performance in large `find_files` (50K+) that essentially reports the same thing, see my [reply](https://www.reddit.com/r/neovim/comments/udx0fi/comment/i91uwmj/?utm_source=share&utm_medium=web2x&context=3) to this here which goes into bit more context.

[...]

PR #1989, opened 2022-06-06
feat(tags): process tagfiles on the fly

antoinemadec:
This allow to start grepping before having to preprocess all the tagfiles entries.

Can be interested if the tag we are looking for is at the beginning of the tagfile.
Still issues of refresh (https://github.com/nvim-telescope/telescope.nvim/issues/1981).

The `Path:new(vim.fn.expand(ctags_file, true)):iter()` performance hit was not as big as when I first tried to implement it.
This could be because I was on a local machine with a SSD, or that the repo I was using previously used a bunch of hardlink.

### DIFF
--- a/lua/telescope/builtin/files.lua
+++ b/lua/telescope/builtin/files.lua
@@ -417,20 +417,11 @@ files.tags = function(opts)
     })
     return
   end
-
-  local results = {}
-  for _, ctags_file in ipairs(tagfiles) do
-    for line in Path:new(vim.fn.expand(ctags_file, true)):iter() do
-      results[#results + 1] = line
-    end
-  end
+  opts.entry_maker = utils.get_default(opts.entry_maker, make_entry.gen_from_ctags(opts))
 
   pickers.new(opts, {
     prompt_title = "Tags",
-    finder = finders.new_table {
-      results = results,
-      entry_maker = opts.entry_maker or make_entry.gen_from_ctags(opts),
-    },
+    finder = finders.new_oneshot_job(flatten {"cat", tagfiles}, opts),
     previewer = previewers.ctags.new(opts),
     sorter = conf.generic_sorter(opts),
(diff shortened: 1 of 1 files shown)


lua/telescope/builtin/files.lua
```diff
@@ -417,20 +417,11 @@ files.tags = function(opts)
...
-  end
+  opts.entry_maker = utils.get_default(opts.entry_maker, make_entry.gen_from_ctags(opts))
 
   pickers.new(opts, {
     prompt_title = "Tags",
-    finder = finders.new_table {
-      results = results,
-      entry_maker = opts.entry_maker or make_entry.gen_from_ctags(opts),
-    },
+    finder = finders.new_oneshot_job(flatten {"cat", tagfiles}, opts),
```
Conni2461, line 424: i think we still should vim.fn.expand every path in the tbl.

Everything else LGTM. Thanks :)

lua/telescope/builtin/files.lua
```diff
@@ -410,27 +410,21 @@ end
...
     return
   end
-
-  local results = {}
-  for _, ctags_file in ipairs(tagfiles) do
-    for line in Path:new(vim.fn.expand(ctags_file, true)):iter() do
-      results[#results + 1] = line
-    end
-  end
+  opts.entry_maker = utils.get_default(opts.entry_maker, make_entry.gen_from_ctags(opts))
```
Conni2461, line 423: please dont use `utils.get_default` its already deprecated in current `dev` branch and will be removed prior 0.1 (see also #1938 )

use `vim.F.if_nil`

antoinemadec: Fixed in https://github.com/nvim-telescope/telescope.nvim/pull/1989/commits/9ff4fe1c5495ca32dec6243f8c15bf46c23fa971

Issue #1714, opened 2022-01-24
Pickers do not use their own settings when invoked through `builtin.builtin` picker.


sonata-chen:
### Description

https://github.com/nvim-telescope/telescope.nvim/blob/0011b1148d3975600f5a9f0be8058cdaac4e30d9/lua/telescope/builtin/internal.lua#L92-L100

Because `opts` is passed to the function, the selected picker will use the same setting as `builtin.builtin` picker's. 
Maybe we can change these two lines to 

``` lua
 require("telescope").extensions[ext][func]() 
 -- skip
 require("telescope.builtin")[selection.text]() 
``` 

so the selected picker will behave as if it is invoked manually. ( through `:Telescope help_tags`)
Or maybe it is not a bug. Is there any reasons you decide to pass `opts` to the selected picker?

### Neovim version

```markdown
NVIM v0.6.1
Build type: Release
LuaJIT 2.0.5
```

### Operating system and version

Arch linux 5.16.2-zen1-1-zen

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - WARNING: nvim-treesitter not found. 

## Checking external dependencies
  - OK: rg: found ripgrep 13.0.0
  - OK: fd: found fd 8.3.1

## ===== Installed extensions =====

## Telescope Extension: `fzf`
  - INFO: No healthcheck provided
```

### Steps to reproduce

1. `nvim -u minimal.lua`
2. `:Telescope builtin`
3. select a picker (help_tags)
4. `:Telescope help_tags`

### Actual behavior

skip

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          { 'nvim-telescope/telescope-fzf-native.nvim', run = 'make' },
        },
      },
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
[...]

PR #1996, opened 2022-06-09
feat: add option `use_default_opts` to `builtin.builtin`

mrbjarksen:
This is a simple PR that adds the option `use_default_opts` to the builtin picker `builtin`. When true, this option opens the selected picker with its default options, rather than with those given to the `builtin` picker. This option is disabled by default, so as not to interfere with existing workflows.
The behavior this PR aims to address is discussed in #1714.


Conni2461:
Thanks :)

PR #1997, opened 2022-06-09
Add `--follow` to bcommits

briandipalma:
The diff to parent preview is blank but at least a user can follow a file around various moves and get access to the SHA-ID for the move.


Conni2461:
Thanks :) i fixed the formatting so i can merge.

Current merge target for anything other than bugfixes is dev branch, so i change that. For more information see #1938

PR #2003, opened 2022-06-12
feat(rg): pass multiple globs to rg by default

antosha417:
ripgrep supports multiple `--glob` flags. 
This gives us the ability to do interesting queries. A simple example will be: search in all `lua` files that are not in test directory.  
This query in telescope will look like this
```:lua require"telescope.builtin".live_grep{ glob_pattern = {"*.lua", "!**/test/**/*"} }```


Conni2461:
Sometimes the action doesnt trigger (for some reason). Rebased to current `dev` branch. see #1938 for more information

Thanks for the PR :)

PR #2004, opened 2022-06-13
chore: clear previewer if no item is selected

### LINKED ISSUE #1702: Clear preview window when no results are available
In `telescope.nvim` repo:

image

Wouldn't it be nice if the preview window got cleared?

Conni2461:
close #1702

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -965,6 +965,9 @@ function Picker:set_selection(row)
   state.set_global_key("selected_entry", entry)
 
   if not entry then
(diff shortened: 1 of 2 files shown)


fdschmidt93:
LGTM & works for me :)

One thing that maybe comes to mind from quick test drive is that dynamic titles persist. I guess we could let those try to fall back to the static title, which should typically be accessible via `picker._previewer._title_fn` I suppose. (`previewers.pickers` from a quick glance I think sole offender -- my bad -- that wouldn't have a static title.)

PR #2010, opened 2022-06-15
fix: block autocommands when setting preview buffer

zeertzjq:
Telescope creates most floating windows with `noautocmd = true`, so
these windows do not trigger autocommands, but preview buffer is set in
window using `nvim_win_set_buf()`, which triggers buffer autocommands.
This may be unwanted, so block them using `'eventignore'`.


Conni2461 (approved):
Thanks :)

PR #2012, opened 2022-06-15
chore: PR template

### LINKED ISSUE #1495: PR template
We should add a PR template so we dont end up with less PRs without any context

Conni2461:
close #1495

CC @fdschmidt93 thoughts

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,10 +9,10 @@ Fixes # (issue)
 
 Please delete options that are not relevant.
 
-- [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
-- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-- [ ] This change requires a documentation update
+- Bug fix (non-breaking change which fixes an issue)
+- New feature (non-breaking change which adds functionality)
+- Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- This change requires a documentation update
 
 # How Has This Been Tested?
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+Please include a summary of the change and which issue is fixed. Please also
+include relevant motivation and context
+
+Fixes # (issue)
+
+## Type of change
+
+Please delete options that are not relevant.
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
```
fdschmidt93, line 12: I'd just make this a 

- list

rather than

- [ ] checklist

since users delete unrequired options.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,37 @@
...
+- [ ] Test A
+- [ ] Test B
+
+**Configuration**:
+* Neovim version (nvim --version):
+* Operating system and version:
+
+# Checklist:
+
+- [ ] My code follows the style guidelines of this project (stylua)
```
fdschmidt93, line 30: I'd keep the checklist shorter, I think this suffices:

- [ ] My code is formatted with `telescope`'s [stylua configuration](https://github.com/nvim-telescope/telescope.nvim/blob/master/.stylua.toml)

To omit imo -

> [ ] `vimdoc` was regenerated (CI should do it, otherwise maintainer will step in)

CI will or won't generate docs, the above "I have made changes to [...] lua annotations" I feel already suffices. I'd keep this if we now fully understand why it exactly sometimes doesn't generate vimdocs (both first time and repeated contributors.. but I suppose problem roots in former, dunno) and can provide guidance.

> ~~[ ] My changes generate no new linter warnings~~ 

Users typically see that post-hoc, so added value of bullet here in practice slim.
 
> I have added tests that prove my fix is effective or that my feature works (if possible)
New and existing unit tests pass locally with my changes

I don't think users would locally test their PRs as too few PRs for the time being actually fail tests. Maybe that changes once TJ revamps our testing framework. I'd omit it for the time being.

I don't have strong feelings, just commenting on what I'd maybe do differently. Feel free to just leave unchanged.

Conni2461: makes sense. thanks :)

Conni2461:
Merging into master so it will be picked up. Lets see how it goes, we can always improve (or remove) it in the future depending on how useful it will be

PR #2018, opened 2022-06-20
wip: added on_picker_find_pre hook

### LINKED ISSUE #1939: Allow opening of a picker with default text selected in **select** mode
I imagine this request is weird/unconventional because select mode (and not just visual mode), so apologies for that. I blame other editors for conditioning me this way :)

I am used to behavior where a search dialog is pre-populated with the last input provided. This input is then highlighted to either search again or allow you to type over it immediately without any special hotkeys. This allows me the option to open the picker with the same keymap to either a) reuse the previous input string for a new search, optionally editing it if i need to; or b) start a new search seamlessly by typing as if the buffer was empty (that is, no extra keypresses to contextually remember to do to clear the prompt)

Telescope provides the picker option [`default_text`](https://github.com/nvim-telescope/telescope.nvim/blob/0b5f1d69d793660900552ad4606c3adcc9934963/lua/telescope/pickers.lua#L85) that I've been able to use to solve the first part of the problem. However, it doesn't look like Telescope provides the means to enable solving the second part of the problem since it seems the plugin is limited to [just insert and normal mode](https://github.com/nvim-telescope/telescope.nvim/blob/0b5f1d69d793660900552ad4606c3adcc9934963/lua/telescope/pickers.lua#L432-L448).

Allowing for passing `initial_mode = 'select'` in the picker configuration. I am not sure if it makes sense to pass in a separate, explicit config option to highlight all of the input, or if just specifying `initial_mode = 'select'` would be enough to implicitly select the whole buffer (because it seems like select mode would otherwise be worthless if there was nothing highlighted)

Setting `initial_mode = 'normal'`, and then in the [`attach_mappings`](https://github.com/nvim-telescope/telescope.nvim/blob/0b5f1d69d793660900552ad4606c3adcc9934963/lua/telescope/pickers.lua#L560) hook, programmatically entering visual mode, selecting the whole line, and then switching to select mode somehow? This feels like a huge hack though, and I'm not actually sure if this is even possible given that I am very new to Lua and NeoVim's API 😅 

**Additional context**
This plugin is amazing!!!!!!! Thank you so much for your hard work!! 🙏 Please let me know if what I'm describing is not making sense

rkoval:
# Description

WIP - moving discussion here from #1939 so that we have a live branch we can talk through

Partially Fixes #1939

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Configuration**:
```
NVIM v0.7.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@BigSur

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "/usr/local/Cellar/neovim/0.7.0/share/nvim"

```

# Checklist:


rkoval:
@fdschmidt93 - I'm moving the convo here from #1939 so that we can talk through some code here.

So far, I have added a simple hook per [this recommendation](https://github.com/nvim-telescope/telescope.nvim/issues/1939#issuecomment-1156135646) to allow for my entering of select mode use case generically. However, when I consume this new hook by passing the following function to it for the `live_grep` finder: 

```lua
local function enter_select_mode(picker)
  local mode = vim.fn.mode()
  local keys = mode ~= 'n' and '<ESC>' or ''
  vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes(keys .. [[^v$<C-g>]], true, false, true), 'n', true)
end
```

... my neovim will hang indefinitely with 100% CPU utilization as telescope opens. I then must force close my neovim and restart it, as it doesn't appear to exit this state.  I am unsure what could be the cause of this though... can you advise here please? My guess is that this isn't the right spot for it because of some event loop conflict leading to an infinite loop somewhere, but that's a stab in the dark.

fdschmidt93:
Just as heads up, I'll probably attend to this towards mid-July as I am very tightly pre-occupied by then.

rkoval:
No problem at all. Thanks for the attentiveness + patience so far :). Quite rare for an open source project haha. Just ping me whenever you can take a long at this

rkoval:
closing this in favor of https://github.com/nvim-telescope/telescope.nvim/pull/2092

PR #2040, opened 2022-07-04
feat: invert_match for search=""

Conni2461:
# Description

I am thinking about "lying" to our users and changing a `search=` query to an inverted query that removes empty lines (which are imo useless). This would make `Telescope grep_string search=` at least more useful for me. It needs a little bit of changing to the parser matching because we no longer have a column.

Thoughts?

CC @tjdevries @fdschmidt93 

## Type of change

# Checklist:


fdschmidt93:
Agreed, makes perfect sense and may improve performance a bit (although very small ofc).

Conni2461:
I hope this doesn't break things :grimacing: I tipple checked

PR #2061, opened 2022-07-11
feat(make_entry) allow replacement symbol_type string

muemmel:
# Description
Minor change to allow replacing symbol type names (variable, function, etc.) via function. No posted issue addressed, I just like having symbols there for easier recognition.

Wasn't too sure on where you would document this, so I didn't for now.

## Type of change
New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Not much to test since it's a simple string replacement.

- [x] Relevant LSP Pickers used with and without config option type_transform

**Configuration**:

* Neovim version (nvim --version): 0.72
* Operating system and version: Fedora 36

# Checklist:
- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~


muemmel:
Updated to latest upstream state

rameshsanth:
Useful feature. Can be used to translate long symbol type names into single character icons and reclaim valuable space for messages.
I ended up doing similar except i passed an explicit lookup table to the Telescope and have the function do a lookup into it if its present. I used a new variable, symbol_icons for this purpose.
However  approach in this PR is much more flexible.

Screenshot 2023-07-09 at 9 39 28 PM

with the transformation, it appears as below
Screenshot 2023-07-09 at 9 43 07 PM

without the tranformation, the symbol types would appear as below
Screenshot 2023-07-09 at 9 50 52 PM

PR #2117, opened 2022-08-09
fix(buffer_previewer): terminal path validation

fdschmidt93:
# Description

Recent regression of path validation affects terminal buffer preview as terminal buffers always return invalid paths.

Fixes # (issue)

Reddit issue @siduck

## Type of change

# How Has This Been Tested?

Create terminal buffer and open `:Telescope buffers`. It now shows up again.

**Configuration**:
Irrelevant

# Checklist:

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -462,13 +462,18 @@ previewers.vimgrep = defaulter(function(opts)
     end,
 
     define_preview = function(self, entry, status)
-      local p = from_entry.path(entry, true)
-      if p == nil or p == "" then
-        return
+      -- bypass path validation for terminal buffers that don't have appropriate path
+      local has_buftype = vim.api.nvim_buf_get_option(entry.bufnr, "buftype") ~= ""
+      local p
+      if not has_buftype then
+        p = from_entry.path(entry, true)
+        if p == nil or p == "" then
+          return
+        end
       end
 
       -- Workaround for unnamed buffer when using builtin.buffer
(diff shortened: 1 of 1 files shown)


siduck:
@fdschmidt93 thx a lot it works now! 

btw is it normal that the preview is colorless?

image


I remember it using colors before

fdschmidt93:
Check my reply on [reddit](https://www.reddit.com/r/neovim/comments/wjpq7o/comment/ijk4izr/?utm_source=share&utm_medium=web2x&context=3).

One of the reasons to avoid raising issues on reddit ;) I originally implemented the buffers previewer since you raised my attention towards that issue in Matrix at the time and it turned out to be a huge sunk cost of 20+ hours of work for a marginally helpful feature that led to a lot of unforeseen side effects.

siduck:
i wanted to make an issue here but it seemed like the recent issues werent getting attention 
image

so thought my issue being more recent wouldnt get any attention either :fearful:

fdschmidt93:
> so thought my issue being more recent wouldnt get any attention either fearful

bug fix != feature request. Only one issue of the screenshot is what I'd consider an actual bug and not feature requests, and even that bug is not _that_ important because there is a straightforward workaround that off of the top of my head is more commonly used anyways. You are very welcome to contribute and make PRs for the other feature requests ;)

> So the colored preview of terminal buffers wont be added right?

Have you read the linked issue and PR?

> If anyone comes up with a way to extract the lines from a terminal buffer including the ANSI color codes, it would be relatively straightforward. I've spent some time on this and maybe I'm missing something, but I unfortunately haven't yet found a way to do so.

As per my final comment of the issue, this should unambiguously answer your question from my point of view. Maybe there's another solution, I won't work on it though.

siduck:
oh my bad i missed that comment on the reddit, good job with the plugin tho!

Issue #2077, opened 2022-07-15
Previewer not displaying files that are not a child of the present working directory tree


jessevernon:
### Description

If I open a file `~/foo/bar.txt`, the previewer for `:Telescope buffers` works as expected. If I cd to a directory that is not a parent of `bar.txt` such as `:cd ~/baz/`, the previewer for `:Telescope buffers` no longer displays anything.

Works when pwd is `~`:
![image](https://user-images.githubusercontent.com/13456400/179294159-44210826-0c7e-425e-ab32-87cd6c98645f.png)

Does not work when pwd is not a parent of file (in this case `~/Downloads`):
![image](https://user-images.githubusercontent.com/13456400/179294247-b8ca4f0f-a817-4e32-949d-732a04677255.png)
![image](https://user-images.githubusercontent.com/13456400/179294415-11d2813d-912c-4a96-920a-aa0fbb16aa64.png)

### Neovim version

```markdown
NVIM v0.8.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/cc -g -O2 -ffile-prefix-map=/build/neovim-KoQPuQ/neovim-0.7.0~ubuntu1+git202205061518-d9ec57e16-dd05b3569=. -flto=auto -ffat-lto-objects -fstack-protector-strong -Wf
ormat -Werror=format-security -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -DNVIM_TS_HAS_SET_ALLOCATOR -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -
Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -Wimplicit-fallthrough -Wsuggest-attribute=pure -
Wsuggest-attribute=const -Wsuggest-attribute=malloc -Wsuggest-attribute=cold -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SO
[...]

PR #2122, opened 2022-08-10
Expand tilde in path because path in filereadable and isdirectory is used as is

dagle:
# Description
For pickers that can generate entries with absolute paths containing a tilde, the previewer will fail.
This is because ~ isn't expanded in filereadable and isdirectory and take the path at face value.

## Type of change

# How Has This Been Tested?

I have tested this on some previewers that uses from_entry
- [x] vimgrep
- [x] git_file_diff

And now the pickers give the correct preview.

### DIFF
--- a/lua/telescope/from_entry.lua
+++ b/lua/telescope/from_entry.lua
@@ -24,6 +24,7 @@ function from_entry.path(entry, validate, escape)
     return
   end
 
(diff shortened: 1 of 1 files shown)


Conni2461:
there is already a issue https://github.com/nvim-telescope/telescope.nvim/issues/2077 and a fix #2084 i am going to merge that one now.

Still thanks for opening

PR #2132, opened 2022-08-13
feat(picker): command history filter

kkharji:
# Description

I've recently start using command history. For sometime I was a bit annoyed of unrelevant commands
like edit/write and others (most likely only used once)

I've considered using lua patterns, however, logical `or` isn't a thing. Additionally, passing a list of lua patterns and checking each pattern for each command history entry felt tedious.

This PR introduce a new optional function to filter command history items.

For example, in my configurations

~~~lua
local command_history_ignore = vim.regex "edit\\|Move\\|write\\|Write\\|e\\s\\|lua\\sI("
overrides.command_history = minimal {
  prompt_prefix = "CMDHistory> ",
  filter_fn = function(item)
    if #item < 3 then
      return false
    else
      return not command_history_ignore:match_str(item)
    end
  end,
}
~~~
## Type of change

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1213,6 +1213,10 @@ builtin.command_history({opts})          *telescope.builtin.command_history()*
     Parameters: ~
         {opts} (table)  options to pass to the picker
 
+    Options: ~
+        {filter_fn} (function)  filter fn(cmd:string). true if the history
+                                command should be presented.
+
 
 builtin.search_history({opts})            *telescope.builtin.search_history()*
     Lists searches that were executed recently, and reruns them on `<cr>`
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -544,10 +544,20 @@ internal.command_history = function(opts)
   local history_list = vim.split(history_string, "\n")
 
   local results = {}
+  local filter_fn = opts.filter_fn
+
   for i = #history_list, 3, -1 do
     local item = history_list[i]
     local _, finish = string.find(item, "%d+ +")
-    table.insert(results, string.sub(item, finish + 1))
+    local cmd = string.sub(item, finish + 1)
+
+    if filter_fn then
+      if filter_fn(cmd) then
+        table.insert(results, cmd)
+      end
+    else
+      table.insert(results, cmd)
+    end
   end
 
   pickers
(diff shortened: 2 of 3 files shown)


Conni2461 (approved):
Thanks :) we can do this. But for the future, i am currently thinking, we could use plenary.iterator to implement this on finder level for all pickers at the same time.

So when we iterate we can do either use default iterator, which does not do any filtering, or use the iterator returned by a `fn(iter) -> iter`. No idea if this can work, for both finders, one_shot and new_table and i am currently not sure what obers plenary.iterator implementation even offers.

Just something to think about, if you are interested in implementing you can give it a shot :laughing: So you can either merge this (you should have merge rights) or investigate this idea. I am leaving this evening for a 10 day vacation and i am generally not reachable so i leave it up to you. 

If you end up implementing this iterator idea, i am interested in reading it prior to merging :)

Thanks for contributing kkharji (tami), good to see you. Also lets talk after my vacation, i am terrible at keeping in touch lately

kkharji:
> Thanks :) we can do this. But for the future, i am currently thinking, we could use plenary.iterator to implement this on finder level for all pickers at the same time.
> 
> So when we iterate we can do either use default iterator, which does not do any filtering, or use the iterator returned by a `fn(iter) -> iter`. No idea if this can work, for both finders, one_shot and new_table and i am currently not sure what obers plenary.iterator implementation even offers.

This quiet interesting. i quickly skim through plenary.iterator.

> Just something to think about, if you are interested in implementing you can give it a shot 😆 So you can either merge this (you should have merge rights) or investigate this idea. I am leaving this evening for a 10 day vacation and i am generally not reachable so i leave it up to you.

Yah maybe, it has bigger scope and far much complexity that I'd rather this merged then see if we can somehow discuss and further play with the idea later.

> If you end up implementing this iterator idea, i am interested in reading it prior to merging :)
> 
> Again thanks for contributing kkharji (tami), good to see you. Also lets talk after my vacation, i am terrible at keeping in touch lately


🌝 I'm still here, not as much as before but telescope is still a joy to work with. Have fun

kkharji:
@Conni2461 could you please merge this? or is there something need to be done?

Conni2461:
I kinda said you can merge this ;) in this comment https://github.com/nvim-telescope/telescope.nvim/pull/2132#pullrequestreview-1071935517

I just think we can find a even better solution and you said that you wanted to look into it. 
Did you have time to do that? otherwise i would give it a shot sometime :)

kkharji:
> I kinda said you can merge this ;) in this comment https://github.com/nvim-telescope/telescope.nvim/pull/2132#pullrequestreview-1071935517
> 
> 
> 
> I just think we can find a even better solution and you said that you wanted to look into it. 
> 
> Did you have time to do that? otherwise 


I would've but I couldn't merge. I had the permissions for a day or so then it was taken away. Do I have the right permissions? 

No, I got carried with other projects but hopefully next week I will re-read the suggestion and see what I can do

Conni2461:
Every member should have merge rights?! I'll look into this?

PR #2143, opened 2022-08-23
feat: added support for tabdrop

rkoval:
# Description

This PR adds support for the `tab` modifier of `drop` to open a file in a new tab if it is not currently open in an existing window. Existing functionality did not support the `tab` modifier, so this would always replace the existing window, which was not what was desired for me.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] I referenced `actions.select_tab_drop` in a keymap, and then executed this keymap while Telescope was open for `git_files`

**Configuration**:

# Checklist:

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -299,6 +299,23 @@ actions.select_tab = {
   end,
 }
 
+--- Perform 'drop' action on selection, usually something like<br>
+---`:drop <selection>`
+---
+--- i.e. open the selection in a new tab
+---@param prompt_bufnr number: The prompt bufnr
+actions.select_drop = {
+  pre = function(prompt_bufnr)
+    action_state.get_current_history():append(
+      action_state.get_current_line(),
+      action_state.get_current_picker(prompt_bufnr)
+    )
+  end,
+  action = function(prompt_bufnr)
+    return action_set.select(prompt_bufnr, "drop")
+  end,
+}
+
 --- Perform 'tab drop' action on selection, usually something like<br>
 ---`:tab drop <selection>`
 ---
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -329,7 +329,7 @@ actions.select_tab_drop = {
     )
   end,
   action = function(prompt_bufnr)
-    return action_set.select(prompt_bufnr, "tab drop")
+    return action_set.select(prompt_bufnr, "tab_drop")
   end,
 }
 
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -329,7 +329,7 @@ actions.select_tab_drop = {
     )
   end,
   action = function(prompt_bufnr)
-    return action_set.select(prompt_bufnr, "tab_drop")
+    return action_set.select(prompt_bufnr, "tab drop")
   end,
 }
 
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -302,7 +302,7 @@ actions.select_tab = {
 --- Perform 'drop' action on selection, usually something like<br>
 ---`:drop <selection>`
 ---
---- i.e. open the selection in a new tab
+--- i.e. open the selection in a window
 ---@param prompt_bufnr number: The prompt bufnr
 actions.select_drop = {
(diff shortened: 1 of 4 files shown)


lua/telescope/actions/state.lua
```diff
@@ -33,6 +33,7 @@ local select_to_edit_map = {
   horizontal = "new",
   vertical = "vnew",
   tab = "tabedit",
+  ["tab drop"] = "tab drop",
```
Conni2461, line 36: uff did we forget to add drop in this table, in the previous PR?

lua/telescope/actions/init.lua
```diff
@@ -299,6 +299,23 @@ actions.select_tab = {
   end,
 }
 
+--- Perform 'tab drop' action on selection, usually something like<br>
+---`:tab drop <selection>`
+---
+--- i.e. open the selection in a new tab
+---@param prompt_bufnr number: The prompt bufnr
+actions.select_tab_drop = {
```
Conni2461, line 307: do we also want a `select_drop`? and we forgot to add it in the previous PR

(same comment as above)

lua/telescope/actions/init.lua
```diff
@@ -299,6 +299,23 @@ actions.select_tab = {
...
+  pre = function(prompt_bufnr)
+    action_state.get_current_history():append(
+      action_state.get_current_line(),
+      action_state.get_current_picker(prompt_bufnr)
+    )
+  end,
+  action = function(prompt_bufnr)
+    return action_set.select(prompt_bufnr, "tab drop")
```
Conni2461, line 315: not sure if i like this key with a ` `,  maybe `_` is a little bit better

lua/telescope/actions/state.lua
rkoval: i wasn't involved in the previous PR, so i am not sure what this is referencing. i can definitely add it here though

Conni2461: sorry thought you were the same person who added: https://github.com/nvim-telescope/telescope.nvim/commit/d793de0f12d874c463e81edabee741b802c1a37a

just because its also related to drop :laughing:

rkoval: no problem!

lua/telescope/actions/init.lua
rkoval: https://github.com/rkoval/telescope.nvim/commit/5cc4bf49fe72e583e2529bdf7b6e0f108c07dcbf

lua/telescope/actions/init.lua
rkoval: https://github.com/rkoval/telescope.nvim/commit/87557efcd1a1475dcb436302fbeac5bd9150b14c

rkoval:
@Conni2461 - this should be ready for re-review!

lua/telescope/actions/set.lua
```diff
@@ -68,6 +68,7 @@ local edit_buffer
 do
   local map = {
     drop = "drop",
+    ["tab drop"] = "tab drop",
```
Conni2461, line 71: We should use the same key everywhere: https://github.com/nvim-telescope/telescope.nvim/pull/2143/files#diff-536f772c8bf9390acea7443c9dc1d570cde4901c85d77ba1334d47e3239e242bR37 so `tab_drop` also i dont think you need to `["` `"]` the key anymore

rkoval: this doesn't seem to work when i change this to `tab_drop` unfortunately, though i am unsure why. i was under the impression these keys should be valid nvim commands though that just map to the local buffer cloning commands. for example, all of the existing keys are all runnable commands that just map to some local buffer cloning commands. perhaps there's some detail there that makes this not work as `tab_drop` (which is not a valid nvim command) as a result

i did update [this line](https://github.com/nvim-telescope/telescope.nvim/pull/2143/files#diff-536f772c8bf9390acea7443c9dc1d570cde4901c85d77ba1334d47e3239e242bR37) though to remove the surrounding `["`/`"]` though for that `tab_drop` command

Conni2461:
You were right. Appearantly i didn't 100% remember my code :laughing: I reverted all the key stuff back to `tab drop`, tested it, generated docs and ran stylua.

Thanks for the PR :) will now merge

PR #2174, opened 2022-09-21
docs: Add Lua-only key mappings examples

sestrella:
# Description

Adding some Lua-only key mapping examples taking advantage of NeoVim 0.7 features

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ to get an understanding of how to use Telescope and how to configure it.
 Try the command `:Telescope find_files<cr>`
   to see if `telescope.nvim` is installed correctly.
 
+Using VimL:
+
 ```viml
 " Find files using Telescope command-line sugar.
 nnoremap <leader>ff <cmd>Telescope find_files<cr>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -130,6 +132,16 @@ nnoremap <leader>fb <cmd>lua require('telescope.builtin').buffers()<cr>
 nnoremap <leader>fh <cmd>lua require('telescope.builtin').help_tags()<cr>
 ```
 
+Using Lua:
+
+```lua
+local builtin = require('telescope.builtin')
+vim.keymap.set('n', 'ff', builtin.find_files, {})
```
Conni2461, line 139: mappings dont have parity with the viml because of missing `noremap = true`

sestrella: @Conni2461 thank you for the comment, however, I don't think `noremap` is a valid option in the context of `vim.keymap.set`. There is an option called `remap` with `false` as the default value which I think is equivalent to `noremap = true` in the context of `nvim_set_keymap`


image

Conni2461: image

It clearly links to `map-arguments`, noremap should be a map argument.  But remap clearly meantions that the default is already false. maybe the neovim team changed the default to the better.

Conni2461:
Thanks :)

CI failure is unrelated and already fixed on master

Issue #996, opened 2021-07-15
sorter in :Telescope registers is not working as expected


freezboltz:
### Description

Not able to sort through registers as expected

**Expected Behavior**

- After typing first letter the sorter should sort using in capital Alphabet present in between square bracket
- After pressing second letter the sorter should filter through register such as fuzzy finder

**Actual Behavior**

- In first case it works as expected
- In second case it is not fuzzy finding content but freeze in last selected item
- While typing letters all capital the last letter get sorted

#### step to reproduce issue
1. open nvim nightly
2. in command mode type Telescope registers
3. type all capital letter and see the behavior and compare with expected behavior
4. try to search for any register content the sorter will freeze in last selected item while searching
![reg_bug_sorter](https://user-images.githubusercontent.com/58996975/125788883-61d6f8c7-fdb0-4dfb-9553-9ca163b59ed4.gif)

<details><summary>Environment</summary>

- nvim --version output: NVIM v0.5.0-dev+5aaa1a1 
- Operating system: Manjaro latest
- Telescope commit: 37a3a68

</details>

<details><summary>Configuration</summary>
<p>

```viml
set nocompatible hidden laststatus=2

if !filereadable('/tmp/plug.vim')
  silent !curl --insecure -fLo /tmp/plug.vim
    \ https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
endif

source /tmp/plug.vim
call plug#begin('/tmp/plugged')
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-telescope/telescope.nvim'
call plug#end()

autocmd VimEnter * PlugClean! | PlugUpdate --sync | close
lua << EOF

local actions = require('telescope.actions')
require('telescope').setup {
    defaults = {
        file_sorter = require('telescope.sorters').get_fzy_sorter,
        prompt_prefix = ' $ ',
        color_devicons = true,

        file_previewer   = require('telescope.previewers').vim_buffer_cat.new,
        grep_previewer   = require('telescope.previewers').vim_buffer_vimgrep.new,
        qflist_previewer = require('telescope.previewers').vim_buffer_qflist.new,

        mappings = {
            i = {
                ["<C-x>"] = false,
                ["<C-q>"] = actions.send_to_qflist,
            },
        }
    },
    extensions = {
        fzy_native = {
            override_generic_sorter = false,
            override_file_sorter = true,
        }
    }
}

[...]

PR #2175, opened 2022-09-21
fix: builtin register picker better sorting

tusqasi:
Now using `:Telescope registers` will sort with register content.

# Description

Default `:Telescope registers` does weird stuff like sorting with register names. Change the ordinal to sort on the basis of content of the registers.

This will be much better than the default behaviour I think.

Fixes #(996)
https://github.com/nvim-telescope/telescope.nvim/issues/996

## Type of change

# How Has This Been Tested?

😅. I ran it a few times.

**Configuration**:

- Neovim version (v0.8.0-dev-1132-g37a71d1f2):
- Operating system and version: manjaro

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -7,7 +7,6 @@ local Path = require "plenary.path"
 local pickers = require "telescope.pickers"
 local previewers = require "telescope.previewers"
 local p_window = require "telescope.pickers.window"
-local sorters = require "telescope.sorters"
 local state = require "telescope.state"
 local utils = require "telescope.utils"
 
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -1089,9 +1089,7 @@ internal.registers = function(opts)
         results = registers_table,
         entry_maker = opts.entry_maker or make_entry.gen_from_registers(opts),
       },
-			-- levenshtein doesn't seem to work with ordinal = register content
-			-- so changed to get_generic_fuzzy_sorter
-      sorter = sorters.get_generic_fuzzy_sorter(),
+      sorter = conf.generic_sorter(opts),
       attach_mappings = function(_, map)
         actions.select_default:replace(actions.paste_register)
         map("i", "<C-e>", actions.edit_register)
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -1089,8 +1089,9 @@ internal.registers = function(opts)
         results = registers_table,
         entry_maker = opts.entry_maker or make_entry.gen_from_registers(opts),
       },
-      -- use levenshtein as n-gram doesn't support <2 char matches
-      sorter = sorters.get_levenshtein_sorter(),
+			-- levenshtein doesn't seem to work with ordinal = register content
+			-- so changed to get_generic_fuzzy_sorter
+      sorter = sorters.get_generic_fuzzy_sorter(),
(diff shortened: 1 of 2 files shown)


lua/telescope/builtin/__internal.lua
```diff
@@ -1089,8 +1089,9 @@ internal.registers = function(opts)
         results = registers_table,
         entry_maker = opts.entry_maker or make_entry.gen_from_registers(opts),
       },
-      -- use levenshtein as n-gram doesn't support <2 char matches
-      sorter = sorters.get_levenshtein_sorter(),
+			-- levenshtein doesn't seem to work with ordinal = register content
+			-- so changed to get_generic_fuzzy_sorter
+      sorter = sorters.get_generic_fuzzy_sorter(),
```
Conni2461, line 1094: would be better to do `conf.generic_sorter(opts)` we also dont really need that comment

lua/telescope/make_entry.lua
```diff
@@ -794,7 +794,7 @@ function make_entry.gen_from_registers(opts)
   return function(entry)
     return make_entry.set_default_entry_mt({
       value = entry,
-      ordinal = entry,
+      ordinal = vim.fn.getreg(entry),
```
Conni2461, line 797: lets only do `vim.fn.getreg` once. can you move the result in a `local v` and use that for ordinal as well as content? thanks

lua/telescope/builtin/__internal.lua
tusqasi: What does that do though? Seems to work find in both ways

lua/telescope/make_entry.lua
tusqasi: Yeah that makes sense.

lua/telescope/builtin/__internal.lua
Conni2461: using the sorter configured by the user rather than an hardcoded one.

So people use fzf-native if they have that one installed, or fzy-native, or fzy-lua,, ....

Conni2461 (approved):
Thanks :)

CI failure is unrelated

lua/telescope/make_entry.lua
```diff
@@ -792,10 +792,11 @@ function make_entry.gen_from_registers(opts)
   end
 
   return function(entry)
+    local contents = vim.fn.getreg(entry)
     return make_entry.set_default_entry_mt({
       value = entry,
-      ordinal = entry,
-      content = vim.fn.getreg(entry),
+      ordinal = contents,
```
Conni2461, line 798: Re: https://github.com/nvim-telescope/telescope.nvim/issues/996#issuecomment-1258966444

what we could do here is `string.format("%s %s", entry, contents)`

Try it and if it works as expected, feel free to open a new PR :)

tusqasi: A while back I tried to solve this by changing the ordinal conditionally.

This is much better

tusqasi: I tried to add that, and did a little git oopsy, and that branch doesn't exist anymore. I will add that later on. Also the pull request closed. Is that supposed to happen

Conni2461: i already merged the PR

Conni2461: you would need to open a new PR, like i said above.

So create a new branch from current master with the new changes.

I just saw the comment on the issue after merging

tusqasi: My bad, didn't read the whole thing. I will just create a new PR.

Issue #410, opened 2021-01-10
git_files shows submodules


timbedard:
### Description
The `git_files` picker shows git submodules. This is both annoying and can break previewers that are expecting to act only on files, such as the new `vim_buffer_cat` previewer.

I haven't found a suitable option for `git ls-files` or an alternative single command that gives us what we want, so we might just need to filter out the submodules from the list. From cursory exploration, it looks like we can use the output of `git submodule status` or just parse `.gitmodules` to determine what the submodules for the repo are.

Alternatively, `git grep --exclude-standard --cached --untracked -l ""` comes _really_ close, but excludes empty files and symlinks as well.

**Expected Behavior**
The `git_files` picker does not show git submodules.

**Actual Behavior**
The `git_files` picker shows git submodules.

### Details

<details><summary>Reproduce</summary>

1. Open nvim in a git repo with one or more submodules.
2. `:Telescope git_files`

</details>

<details><summary>Environment</summary>

- nvim version: v0.5.0-dev+b535575ac
- Operating system: macOS
- Telescope commit: de80a98

</details>

<details><summary>Configuration</summary>
<p>

(default)

```viml
set nocompatible hidden laststatus=2

if !filereadable('/tmp/plug.vim')
  silent !curl --insecure -fLo /tmp/plug.vim
    \ https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
endif

source /tmp/plug.vim
call plug#begin('/tmp/plugged')
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-telescope/telescope.nvim'
call plug#end()

autocmd VimEnter * PlugClean! | PlugUpdate --sync | close
lua << EOF

require('telescope').setup{
  defaults = {
    vimgrep_arguments = {
      'rg',
      '--color=never',
      '--no-heading',
      '--with-filename',
      '--line-number',
      '--column',
      '--smart-case'
    },
    prompt_position = "bottom", 
    prompt_prefix = ">",
    selection_strategy = "reset",
    sorting_strategy = "descending",
    layout_strategy = "horizontal",
    layout_defaults = {},
    file_ignore_patterns = {},
    shorten_path = true,
    winblend = 0,
    width = 0.75,
    preview_cutoff = 120,
    results_height = 1,
    results_width = 0.8,
    border = {},
    borderchars = { '─', '│', '─', '│', '╭', '╮', '╯', '╰'},
[...]

PR #2181, opened 2022-09-29
feat(git): notify when not in a git repo instead of error

### LINKED ISSUE #2180: use utils.notify() instead of error() when not in a git repo
When running git builtin pickers outside of a git repo, telescope reports an error.
This spits out 10 read lines and forces to press Enter.

Use utils.notify() with ERROR level instead and stop telescope prompt from showing up.

Be perfect and never invoke git pickers outside of git repo. Not realistic.

**Additional context**
If you agree, I can create the PR.

antoinemadec:
# Description

Fixes #2180

## Type of change

# How Has This Been Tested?

- [x] git_files in git repo: works as before
- [x] git_files outside git repo: notification pops-up, Telescope prompt does not show up, no more error

**Configuration**:
* Neovim version (nvim --version): NVIM v0.8.0-dev-1200-gf9228577e
* Operating system and version: Linux Manjaro 22

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__git.lua
+++ b/lua/telescope/builtin/__git.lua
@@ -392,7 +392,11 @@ local set_opts_cwd = function(opts)
     local in_bare = utils.get_os_command_output({ "git", "rev-parse", "--is-bare-repository" }, opts.cwd)
 
     if in_worktree[1] ~= "true" and in_bare[1] ~= "true" then
-      error(opts.cwd .. " is not a git directory")
+      utils.notify("builtin.git", {
+        msg = opts.cwd .. " is not a git directory",
+        level = "ERROR",
+      })
+      return false
     elseif in_worktree[1] ~= "true" and in_bare[1] == "true" then
       opts.is_bare = true
     end
@@ -401,6 +405,8 @@ local set_opts_cwd = function(opts)
       opts.cwd = git_root[1]
     end
   end
+
+  return true
 end
 
 local function apply_checks(mod)
@@ -408,8 +414,10 @@ local function apply_checks(mod)
     mod[k] = function(opts)
       opts = vim.F.if_nil(opts, {})
 
-      set_opts_cwd(opts)
-      v(opts)
+      local ok = set_opts_cwd(opts)
+      if ok then
+        v(opts)
+      end
     end
   end
 


Conni2461 (approved):
Thanks :)

Conni2461:
Hmm this broke a use case i forgot about: https://github.com/nvim-telescope/telescope.nvim/issues/410#issuecomment-1264426662

antoinemadec:
@Conni2461 , this is how to do it know: https://github.com/nvim-telescope/telescope.nvim/issues/2183#issuecomment-1264492824

It seems reasonable and rather intuitive to me, what do you think ?

Some users were recommanding updating the Wiki with this.

sQVe:
@antoinemadec @Conni2461 This change complicates my use-case by quite a bit. I run the following function:
```lua
M.find_files = function(use_buffer_cwd)
  local builtin = require('telescope.builtin')
  local utils = require('telescope.utils')
  local opts = {
    follow = true,
    hidden = true,
    show_untracked = true,
    use_git_root = false,
  }

  if use_buffer_cwd then
    opts.cwd = utils.buffer_dir()
  end

  local ok = pcall(builtin.git_files, opts)
  if not ok then
    builtin.find_files(opts)
  end
end
```

This allows me to run `find_files` for the `cwd` of the current buffer. Now I would have to feed the `cwd` into a `systemlist` call, instead of being able to rely on the built-in functionality.

I think this should be reverted. There's a lot of power in it erroring when failing. If you want to be notified instead you can always wrap the call yourself.

Conni2461:
I agree, i dont consider calling `vim.fn.systemlist"git rev-parse --is-inside-work-tree"[1] == 'true'` a acceptable "workaround". its doing the same work twice, which is less elegant compared to before.

I was thinking about returning a bool if we started the picker, but this would still result in the vim.notify error message, which isn't ideal either. In fact you can no longer suppress the vim.notify error message easily (at all, without suppressing all notify messages with level error, or providing a no op for vim.notify). This isn't great either.

So i was thinking about reverting because other people could always just do

```lua
local wrap = function(builtin, opts)
  local ok, msg = pcall(builtin, opts)
  if not ok then
    vim.notify(msg)
  end
end

wrap(require("telescope.builtin").git_files, {})
```

something like this would achieves the same as the pr (not tested, written in github right now)

So yeah, i'll need to think about it for a day or so. If you need that feature rn, please checkout a early commit or checkout branch 0.1.x

antoinemadec:
The "not doing the same check twice" argument is sound.

On my end, I like to be able to chose from find_files and git_files myself (when you are looking for untracked or ignored files in your repo for instance), and I'd rather have a default behavior that notifies me rather than errors out.

I think it just boils down to:

- how many people are negatively impacted by keeping or removing this commit
- consistency of the erroring out behavior across the pickers
- personal taste

sQVe:
@Conni2461 👋🏼 

Have you possibly landed in a decision here? I've been pushing off the required changes in my config but this issue is getting more frustrating.

Conni2461:
Yeah. I know i put the decision of because its a very opinionated issue.

We could do a config option and i probably will do it on a more global level. Like what i am currently thinking is doing `error` everywhere, which enables users to do scripting things and then providing a `no_error` option (or so) and this option then pcalls everything. you can then obviously do `no_error = true` on a global level and override it for `Telescope git_files no_error=false`  to bring `error`s back. https://github.com/nvim-telescope/telescope.nvim/issues/2216

I will revert this commit now and then work on that issue this weekend. I hope this works for everyone :)

Conni2461:
https://github.com/nvim-telescope/telescope.nvim/pull/2225 does implement config value / option `error_mode` with the following values "modes":
- "lua" (default, because current behavior)
- "notify"
- "silent"

`Telescope git_files error_mode=notify` or `:lua require("telescope.builtin").git_files { error_mode = "notify" }` or configure it in the default table in `telescope.setup`

feedback is appreciated

PR #2197, opened 2022-10-15
docs: Add <leader> to Lua-only key mappings examples

Bonen0209:
# Description

Adding \<leader\> to key mapping examples


Conni2461:
thanks :)

PR #2205, opened 2022-10-21
perf(builtin): faster list extend for find_files

kylo252:
# Description

replace `table.insert` with direct indexing for performance gainz

## Type of change

Performance

# How Has This Been Tested?

still works :)

# Checklist:


Conni2461:
did you ever benchmark `table.insert` vs `tbl[#tbl + 1]` vs `idx = idx + 1; tbl[idx]`?

Its one of these placebo things, there is no difference within a margin of error. Try it with `plenary.benchmark`.

Its also startup code, its not running in a loop. But whatever

PR #2218, opened 2022-11-01
Add jump_type option for lsp_references

### LINKED ISSUE #2189: [feat] lsp_implementations option
Add new option to lsp_implementations. When i open telescope with lsp_implementations picker, and telescope found only one reference, telescope closes and opens file, with this reference. It will be more comfortable for me.

gutyina70:
# Description

Add `jump_type` option to `lsp_references`, since it already exists for other lsp pickers, like `lsp_definitions`.
But the main reason is it was annoying when I know there is only 1 reference, use `lsp_references` and I still get the picker opened instead of immediately jumping to the only reference.

Fixes #2189 (probably)

# How Has This Been Tested?

(for all tests `lsp_references = require 'telescope.builtin'.lsp_references`)

- [x] Test A:
```py
def foo(): # cursor is here, `lsp_references {}` jumps to the only result
  foo()
```

- [x] Test B:
```py
def foo(): # cursor is here, `lsp_references { jump_type = 'never' }` opens the picker with 1 result
  foo()
```

- [x] Test C:
```py
def foo(): # cursor is here, `lsp_references { jump_type = 'split' }` splits the window and opens the result
  foo()
```

- [x] Test D:
```py
def bar(): # cursor is here, `lsp_references {}` opens the picker with 2 results
  bar()
  bar()
```

**Configuration**:
* Neovim version (nvim --version): v0.8.0-dev-1093-g982fef601
* Operating system and version: Arch Linux 6.0.5-arch1-1

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1420,6 +1420,9 @@ builtin.lsp_references({opts})            *telescope.builtin.lsp_references()*
                                           lsp references (default: true)
         {include_current_line} (boolean)  include current line (default:
                                           false)
+        {jump_type}            (string)   how to goto reference if there is
+                                          only one, values: "tab", "split",
+                                          "vsplit", "never"
         {fname_width}          (number)   defines the width of the filename
                                           section (default: 30)
         {show_line}            (boolean)  show results text (default: true)
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -39,6 +39,20 @@ lsp.references = function(opts)
       return
     end
 
+    if #locations == 1 and opts.jump_type ~= "never" then
+      if opts.jump_type == "tab" then
+        vim.cmd "tabedit"
+      elseif opts.jump_type == "split" then
+        vim.cmd "new"
+      elseif opts.jump_type == "vsplit" then
+        vim.cmd "vnew"
+      end
+      -- jump to location
+      vim.api.nvim_win_set_buf(0, opts.bufnr)
+      vim.api.nvim_win_set_cursor(0, { locations[1].lnum, locations[1].col - 1 })
+      return
+    end
+
     pickers
       .new(opts, {
         prompt_title = "LSP References",
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -372,6 +372,7 @@ builtin.jumplist = require_on_exported_call("telescope.builtin.__internal").jump
 ---@param opts table: options to pass to the picker
 ---@field include_declaration boolean: include symbol declaration in the lsp references (default: true)
 ---@field include_current_line boolean: include current line (default: false)
+---@field jump_type string: how to goto reference if there is only one, values: "tab", "split", "vsplit", "never"
 ---@field fname_width number: defines the width of the filename section (default: 30)
 ---@field show_line boolean: show results text (default: true)
 ---@field trim_text boolean: trim results text (default: false)


Conni2461 (approved):
Thanks :)

kashing1999:
Hi @gutyina70 , I think jump_type should be tested for emptiness, because if jump_type is not set, `opts.jump_type ~= "never"` will still be set to true, and` vim.api.nvim_win_set_cursor(0, { locations[1].lnum, locations[1].col - 1 })` could be setting the cursor to the wrong location because the previous if statements from line 43 to 49 would evaluate to false as well.

gutyina70:
> `vim.api.nvim_win_set_cursor(0, { locations[1].lnum, locations[1].col - 1 })` could be setting the cursor to the wrong location because the previous if statements from line 43 to 49 would evaluate to false as well.

Hi @kashing1999, when all the conditions from lines 43-49 evaluate to false, it will set the cursor position in the current window as intended (instead of creating a new tab or split and setting the cursor position in that).

Also this behaviour is present in [list_or_jump](https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/builtin/__lsp.lua#L181-L189) which is used by other lsp methods, for example [lsp_definitions](https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/builtin/__lsp.lua#L209-L211). If you still think it should be changed, please open a new issue or PR.

kashing1999:
Ah okay, because i faced a problem where when the only reference is not in the file that im editing, it'll not properly jump to the file that contains that reference, and I'm not sure if that is intended behaviour or not.

gutyina70:
@kashing1999 you're right, when the reference is in a different buffer, it still jumps in the current buffer. My bad I haven't tested it with multiple files. I will submit a PR soon, thanks for noticing it :)

kashing1999:
Thanks! @gutyina70

Conni2461:
someone opened this 2 days ago https://github.com/nvim-telescope/telescope.nvim/pull/2229

Conni2461:
why couldn't we use `vim.lsp.util.jump_to_location`?

Gordin:
Crossposting my answer to @Conni2461, because he asked in both places ^^
> why couldn't we use vim.lsp.util.jump_to_location?

>@Conni2461 I actually tried using that, but the `location` parameter that `vim.lsp.util.jump_to_location` wants is different from the locations that the lsp provides. The `jump_to_location` looks for the attributes `uri`, `targetUri`, `range` and `targetSelectionRange` on the `location` variable, but the `location` in the `telescope.lsp.references` only has the attributes `lnum`, `col`, `filename` and `text`.
I guess you could turn one location format into the other one, but I didn't know what I should put for `range` and I just wanted this to work, so I just used the `set_buf` + `set_cursor` thing that was already there.
(Also, I literally migrated my vim config to lua stuff a week ago, I have no idea if there is a meaningful difference between using `set_buf` + `set_cursor` or `vim.lsp.util.jump_to_location`, or if one would be preferable...)

@gutyina70 Maybe looking at #2229 saves you some time, I'm not sure if it's the best way to fix the jumping, but it does fix it :)

gutyina70:
> why couldn't we use `vim.lsp.util.jump_to_location`?

@Conni2461 yes, I wanted to use that in the first place, but I ran into the same issue that @Gordin mentioned.

I was trying to link together the 2 different location formats, so I can filter using the converted format, and give `jump_to_location` the lsp provided format, but it didn't go well, since it looks like `locations_to_items` sorts the output. So for example taking the first element of the the lsp provided references table, and the first element of the converted references table, may correspond to a seperate reference.

I'm thinking of doing the filter on the lsp provided locations instead of the converted locations, but I need to convert the uri to a file path which may introduce other bugs.

Let me know if you have a solution that doesnt involve reinventig something that neovim already does.

gutyina70:
When you select a location entry in the picker (that gets the converted locations format), this code runs, and at first sight it reimplements `jump_to_location`. This is probably how this issue should be solved, I'll get back to it when I have time.

PR #2225, opened 2022-11-06
feat: error_mode

### LINKED ISSUE #2216: consistent error handling for "frontend" functions
Like we should do `error` everywhere in every single frontend function and then have a `no_error` option, which defaults to false.

If this option is set, we will pcall the frontend function and `if not ok`, call `vim.notify` with the message.

Option should behave like all conf, opts options and should be overridable on a local level.

Option name is still open for debate.

Conni2461:
Close #2216

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -561,6 +561,15 @@ telescope.setup({opts})                                    *telescope.setup()*
 
         Default: true
 
+                                            *telescope.defaults.error_mode*
+    error_mode: ~
+        How frontend errors are handled. Possible options:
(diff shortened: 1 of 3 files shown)


antoinemadec:
`Telescope git_files error_mode=notify` works fine, but `Telescope gitfiles` seems not to take the follow config into account:
```lua
require('telescope').setup{
  defaults = {
    error_mode = 'notify',
    dynamic_preview_title = true,
    mappings = {
      i = {
        ["<C-k>"] = "move_selection_previous",
        ["<C-j>"] = "move_selection_next",
      }
    }
  },
}
```

Is this expected ?

PR #2229, opened 2022-11-08
Fixes not changing files when request only returns 1 result

Gordin:
# Description

When using `telescope.lsp_references` and there are only 2 references (the one where the cursor is on, and another one), telescope will try to jump to the other location without showing a picker. But if the reference is in another file, it will just put the cursor in the correct row/column, but it will stay in the same buffer/file.

## Type of change

# How Has This Been Tested?

I called `telescope.lsp_reference` on a symbol that has 2 references, the one where the cursor is, and one in a different file. After this change, it jumps to the other file, before the change, it did not.

**Configuration**:
* Neovim version (nvim --version): NVIM v0.9.0-dev-5e1687e-dirty
* Operating system and version: Archlinux

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -48,8 +48,13 @@ lsp.references = function(opts)
         vim.cmd "vnew"
       end
       -- jump to location
-      vim.api.nvim_win_set_buf(0, opts.bufnr)
-      vim.api.nvim_win_set_cursor(0, { locations[1].lnum, locations[1].col - 1 })
+      local location = locations[1]
+      local bufnr = opts.bufnr
+      if location.filename then
+        bufnr = vim.uri_to_bufnr(vim.uri_from_fname(location.filename))
+      end
+      vim.api.nvim_win_set_buf(0, bufnr)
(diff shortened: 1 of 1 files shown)


Conni2461:
This is a new feature and apparently didnt do a good review job, sorry. https://github.com/nvim-telescope/telescope.nvim/pull/2218

why cant we just use `vim.lsp.util.jump_to_location`?

Gordin:
@Conni2461 I actually tried using that, but the `location` parameter that `vim.lsp.util.jump_to_location` wants is different from the locations that the lsp provides. The `jump_to_location` looks for the attributes `uri`, `targetUri`, `range` and `targetSelectionRange` on the `location` variable, but the `location` in the `telescope.lsp.references` only has the attributes `lnum`, `col`, `filename` and `text`.
I guess you could turn one location format into the other one, but I didn't know what I should put for `range` and I just wanted this to work, so I just used the `set_buf` + `set_cursor` thing that was already there.
(Also, I literally migrated my vim config to lua stuff a week ago, I have no idea if there is a meaningful difference between using `set_buf` + `set_cursor` or `vim.lsp.util.jump_to_location`, or if one would be preferable...)

Conni2461 (approved):
Thanks :)

PR #2232, opened 2022-11-10
Add configuration to fix encoding of output of find_command in find_files

zbq:
# Description

The encoding of output of find_command (such as 'where.exe' on Windows) could be non-utf8.

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -74,6 +74,7 @@ builtin.grep_string = require_on_exported_call("telescope.builtin.__files").grep
 ---@param opts table: options to pass to the picker
 ---@field cwd string: root dir to search from (default: cwd, use utils.buffer_dir() to search relative to open buffer)
 ---@field find_command function|table: cmd to use for the search. Can be a fn(opts) -> tbl (default: autodetect)
+---@field file_entry_encoding string: encoding of output of `find_command`
 ---@field follow boolean: if true, follows symlinks (i.e. uses `-L` flag for the `find` command)
 ---@field hidden boolean: determines whether to show hidden files or not (default: false)
 ---@field no_ignore boolean: show files ignored by .gitignore, .ignore, etc. (default: false)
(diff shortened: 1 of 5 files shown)


Conni2461:
hmm we dont need a global config, just for find_files :| we can just do a opts, because we can configure opts with `telescope.setup { pickers = { find_files = { file_entry_encoding = true } } }` see `help telescope.setup`

Can you rework this?

Also lets not do this check in return function, because this function is running inside a loop.

lets do something like this

```lua
if opts.file_entry_encoding then
  return function(line)
    line = vim.iconv(line, opts.file_entry_encoding, 'utf8')
    return setmetatable({ line }, mt_file_entry)
  end
else
  return function(line)
    return setmetatable({ line }, mt_file_entry)
  end
end
```

zbq:
yes, it's better to be an option of find_files.

lua/telescope/builtin/init.lua
```diff
@@ -74,6 +74,7 @@ builtin.grep_string = require_on_exported_call("telescope.builtin.__files").grep
 ---@param opts table: options to pass to the picker
 ---@field cwd string: root dir to search from (default: cwd, use utils.buffer_dir() to search relative to open buffer)
 ---@field find_command function|table: cmd to use for the search. Can be a fn(opts) -> tbl (default: autodetect)
+---@field file_entry_encoding string: encoding of output of `find_command`
```
Conni2461, line 77: Why is the type a string, option works like a boolean. description could also be: "whether to re encode output of `find_command` to utf8" or something similar

zbq: It is encoding name for ```vim.iconv```.  See ```help iconv```:
```
iconv({string}, {from}, {to})
    The result is a String, which is the text {string} converted
    from encoding {from} to encoding {to}.
```

Conni2461: oh i misslocked, i thought it was used like a bool in the if. my bad

PR #2235, opened 2022-11-14
support for multi-part file extensions like *.test.js

mrjones2014:
# Description

Updates code used to get icons from `nvim-web-devicons` to pass the filename as well as extension.

## Type of change

# How Has This Been Tested?

Set up `nvim-web-devicons` with configuration:

```lua
require('nvim-web-devicons').setup({
  override = {
    ['js'] = {
      icon = '',
      color = '#efd81d',
      cterm_color = '220',
      name = 'Js',
    },
    ['test.js'] = {
      icon = 'ﭧ',
      color = '#ffca28',
      cterm_color = '220',
      name = 'TestJavascript',
    },
  },
})
```

Then create a `*.js` and a `*.test.js` file, and open `:Telescope find_file`, and make sure the icons appear as configured.

**Configuration**:

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -838,25 +838,26 @@ builtin.find_files({opts})                    *telescope.builtin.find_files()*
         {opts} (table)  options to pass to the picker
 
     Options: ~
-        {cwd}              (string)          root dir to search from (default:
-                                             cwd, use utils.buffer_dir() to
-                                             search relative to open buffer)
-        {find_command}     (function|table)  cmd to use for the search. Can be
-                                             a fn(opts) -> tbl (default:
-                                             autodetect)
-        {follow}           (boolean)         if true, follows symlinks (i.e.
-                                             uses `-L` flag for the `find`
-                                             command)
-        {hidden}           (boolean)         determines whether to show hidden
-                                             files or not (default: false)
-        {no_ignore}        (boolean)         show files ignored by .gitignore,
-                                             .ignore, etc. (default: false)
-        {no_ignore_parent} (boolean)         show files ignored by .gitignore,
-                                             .ignore, etc. in parent dirs.
-                                             (default: false)
-        {search_dirs}      (table)           directory/directories/files to
-                                             search
-        {search_file}      (string)          specify a filename to search for
+        {cwd}                 (string)         root dir to search from (default:
+                                               cwd, use utils.buffer_dir() to
+                                               search relative to open buffer)
+        {find_command}        (function|table) cmd to use for the search. Can be
+                                               a fn(opts) -> tbl (default:
+                                               autodetect)
+        {file_entry_encoding} (string)         encoding of output of `find_command`
+        {follow}              (boolean)        if true, follows symlinks (i.e.
+                                               uses `-L` flag for the `find`
+                                               command)
+        {hidden}              (boolean)        determines whether to show hidden
+                                               files or not (default: false)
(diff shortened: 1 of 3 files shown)


DasGandlaf:
Please merge this, I need this.

lua/telescope/utils.lua
```diff
@@ -446,7 +446,9 @@ utils.transform_devicons = load_once(function()
         return display
       end
 
-      local icon, icon_highlight = devicons.get_icon(utils.path_tail(filename), nil, { default = true })
+      local basename = utils.path_tail(filename)
+      -- the double :e:e gets multipart extensions, if one exists, like *.test.js, for example
+      local icon, icon_highlight = devicons.get_icon(basename, vim.fn.fnamemodify(basename, ":e:e"), { default = true })
```
Conni2461, line 451: can we not use `vim.fn.fnamemodify`? Using `vim.fn` here makes this function no longer api-fast and i prefer to keep it api-fast because entry-maker can run inside a tight lua loop and this function is primarily used there

DasGandlaf, line 451: If we add something like this:
```
local icon, icon_highlight = devicons.get_icon(basename, basename:match("[^a-z][^A-Z]*$"), { default = false })
if icon == nil then
  icon, icon_highlight = devicons.get_icon(basename, vim.fn.fnamemodify(basename, ":e:e"), { default = true })
end
```
(Not a good solution, but just to get the point), we could have different icons for files names like these: `ControllerTests.java` and `Controller.java`. (Notice, that `Tests.java` in `ControllerTests.java` does not have a preceding dot, which means if there is an icon mapped to the name `Tests.java`, it would not get picked up with the current changes)
I think this would be cool to add.

mrjones2014: How would we know what the "extension" is in that case? Also we'll be looking to remove `fnamemodify` as per @Conni2461 's comment

mrjones2014:
Whoops. Accidentally synced the fork and blew away my changes. I'll open a new PR.

lua/telescope/utils.lua
DasGandlaf: I don't know if I understand you correctly, but in nvim-devicons you can set the icons for a file extension like so:
```lua
require "nvim-web-devicons".set_icon {
  "spec.ts" = { icon = "A", color = "#00ffff", name = "AngularTest" }
}
```
This would set the icon for all `.spec.ts` files.
With my change however you can add another key like so:
```lua
require "nvim-web-devicons".set_icon {
  "spec.ts" = { icon = "A", color = "#00ffff", name = "AngularTest" }
  "Events.spec.ts" = { icon = "B", color = "#00ffff", name = "AngularExampleFileTest" }
}
```
And this would set the icon `B` for every file that is named `somethingEvents.spec.ts`. (notice, there is no dot preceding Events.spec.ts)

This is useful in languages like java, where test classes end with `Tests.java`, and NOT `.Tests.java`.  (Again, notice the preceding dot)
The code would look like this:

This would set the icon for all `.spec.ts` files. With my change however you can add another key like so:
```lua
require "nvim-web-devicons".set_icon {
  "java" = { icon = "J", color = "#00ffff", name = "Java" }
  "Tests.java" = { icon = "T", color = "#00ffff", name = "JavaTests" }
}
```
With my changes applied, it would have the icon T for all files ending with `Tests.java`, and the icon j for all files ending with `.java` (the dot)

PR #2241, opened 2022-11-23
Add treesitter as a required dep

fseitun:
# Description

Treesitter is a required dep as shown by checkhealth
image

## Type of change

- This change requires is documentation update


Conni2461:
treesitter is not a required dependency! we have some integration with treesitter + previews and have an Telescope treesitter picker but thats it. The plugin is perfectly usable without treesitter.

I'd rather make it visible in the health check that treesitter is optional

fseitun:
Great. I'll make that PR.

PR #2242, opened 2022-11-23
Change treesitter form required to optional

fseitun:
# Description

While performing `checkhealth telescope` treesitter gets wrongly listed as required. This PR changes it to optional.


fseitun:
Opposite solution to #2241

PR #2258, opened 2022-12-03
feat(builtin.buffers): add filter_fn option

trmckay:
# Description

Adds an option `filter_fn` to `telescope.builtins.buffers` to filter out buffers from the picker. `filter_fn` is a function that takes a buffer number and returns `true` if the buffer should be kept.

Some example use-case:

- Search for buffers with the same filetype as current buffer.
- Search for terminal or other special buffers.
- Exclude special buffers from the search.
- Search for buffers with diagnostics.

## Type of change

# How Has This Been Tested?

With a keymap to test different predicates. For example, do not show terminal buffers:

```lua
vim.api.nvim_set_keymap("<leader>fb", function()
    telescope.buffers({
            map("i", "<C-q>", require("telescope.actions").delete_buffer)
            return true
        end,
        filter_fn = function(bufnr)
            return vim.api.nvim_buf_get_option(bufnr, "buftype") ~= "terminal"
        end,
    })
end)
```

**Configuration**:

- Neovim version (nvim --version): NVIM v0.8.1
- Operating system and version: macOS Ventura 13.0.1

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -884,23 +884,25 @@ internal.buffers = function(opts)
   local buffers = {}
   local default_selection_idx = 1
   for _, bufnr in ipairs(bufnrs) do
-    local flag = bufnr == vim.fn.bufnr "" and "%" or (bufnr == vim.fn.bufnr "#" and "#" or " ")
+    if not opts.filter_fn or opts.filter_fn(bufnr) then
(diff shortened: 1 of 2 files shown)


fdschmidt93:
I'm not opposed to having this, but this can also be achieved by something like

```lua
require "telescope.builtin".buffers { file_ignore_patterns = { "term" } }
```

which levers the fact that buffer names of terminal buffers are prefixed with "term://" (though they could be changed).

Issue #2304, opened 2023-01-05
Warning when running live_grep that ripgrep is not installed


Gian-Laager:
The `live_grep` command didn't work on my machine because i hadn't installed `ripgrep` . The problem has already been discussed in the #1214 issue. 

**Warning if dependency doesn't exist**
I thought it would be nice if telescope printed a warning that `ripgrep` is not installed when running `live_grep`. This warning could be a simple echo or if its possible in the Telescope prompt.

Also it would be nice if the dependency was mentioned in the README (*).

\* I will make a pull request with the updated README.

Conni2461:
Yeah we kinda already have this and an error message is shown. Take a look at this video, i recorded. But I merged your PR and its now mentioned in another place that ripgrep is required for grep

https://user-images.githubusercontent.com/15233006/210999520-d432af99-3410-4289-83c6-478b239794ec.mp4

PR #2305, opened 2023-01-05
Note that live_grep uses ripgrep in README

Gian-Laager:
As mentioned in issue  #2304 I added a note that the live_grep picker requires ripgrep.


Conni2461:
Thanks :)

CI failure is unrelated

PR #2312, opened 2023-01-08
feat: jump to the line specific by colon

luyuhuang:
# Description

When using `find_files`, we can specify a line number to jump to when opening this file, by typing a colon plus the line number, just like ctrlp.vim and vscode. E.g. `sorters.lua:42` to open `sorters.lua` and jump to line 42 of `sorters.lua`.

## Type of change

# How Has This Been Tested?

- [x] Open `find_files` and type a file name followed by a colon and a line number, then press enter, it'll open the file and jump to the specific line.
- [x] Type a file name and a line number, then press <C-t>, it'll open the file in a new tab and jump to the specific line.
- [x] Type a file name and a line number, then press <C-v>, it'll jump to the specific line.
- [x] Type a file name and a line number, then press <C-x>, it'll jump to the specific line.
- [x] Just type a file name without colon and line number, then press enter, it'll just open the file and stay at the original position.

**Configuration**:
* Neovim version (nvim --version): NVIM v0.8.2
* Operating system and version: Ubuntu 20.04 LTS (WSL2)

# Checklist:

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -235,13 +235,6 @@ actions.center = function(prompt_bufnr)
   vim.cmd ":normal! zz"
 end
 
-local function jump_to_line()
-  local line_no = state.get_global_key "line_no"
-  if line_no then
-    vim.cmd["normal!"](line_no .. "G")
-  end
-end
-
 --- Perform default action on selection, usually something like<br>
 --- `:edit <selection>`
 ---
@@ -255,9 +248,6 @@ actions.select_default = {
   end,
   action = function(prompt_bufnr)
     return action_set.select(prompt_bufnr, "default")
-  end,
-  post = function()
-    jump_to_line()
   end,
 }
 
@@ -275,9 +265,6 @@ actions.select_horizontal = {
   action = function(prompt_bufnr)
     return action_set.select(prompt_bufnr, "horizontal")
   end,
-  post = function()
-    jump_to_line()
-  end,
 }
 
 --- Perform 'vertical' action on selection, usually something like<br>
@@ -293,9 +280,6 @@ actions.select_vertical = {
   end,
   action = function(prompt_bufnr)
     return action_set.select(prompt_bufnr, "vertical")
-  end,
-  post = function()
-    jump_to_line()
   end,
 }
 
@@ -313,9 +297,6 @@ actions.select_tab = {
   action = function(prompt_bufnr)
     return action_set.select(prompt_bufnr, "tab")
   end,
-  post = function()
-    jump_to_line()
-  end,
 }
 
 --- Perform 'drop' action on selection, usually something like<br>
@@ -332,9 +313,6 @@ actions.select_drop = {
   action = function(prompt_bufnr)
     return action_set.select(prompt_bufnr, "drop")
   end,
-  post = function()
-    jump_to_line()
-  end,
 }
 
 --- Perform 'tab drop' action on selection, usually something like<br>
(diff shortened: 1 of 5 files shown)


lua/telescope/sorters.lua
```diff
@@ -444,6 +445,14 @@ sorters.get_fzy_sorter = function(opts)
...
 
     scoring_function = function(_, prompt, line)
+      local i = prompt:find(":", 1, true) -- Specify the line number by colon
+      if i then
+        global_state.set_global_key("line_no", tonumber(prompt:sub(i + 1)))
+        prompt = prompt:sub(1, i - 1)
+      else
+        global_state.set_global_key("line_no", nil)
+      end
+
```
Conni2461, line 455: lets not do that in sorter, if we do it here we need to do that in all sorters (also external sorters like fzf-native) and that doesnt scale.

we should rather do that in the main loop, here or in the `_get_next_filtered_prompt` function

https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/pickers.lua#L473-L474

lua/telescope/actions/init.lua
```diff
@@ -235,6 +235,13 @@ actions.center = function(prompt_bufnr)
   vim.cmd ":normal! zz"
 end
 
+local function jump_to_line()
+  local line_no = state.get_global_key "line_no"
+  if line_no then
+    vim.cmd["normal!"](line_no .. "G")
```
Conni2461, line 241: we can drop that function and all post calls, by just changing something here:

https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/actions/set.lua#L172-L177

Like we can change `row` to the global key if the global key is not nil (we just have to make sure we nil it again, if we delete `:123`)

Conni2461:
Feature makes sense, but i think the implementation can be more straight forward. Please ask if you have any questions.

Thanks for the PR :)

luyuhuang:
Thanks for your suggestions. I've modified the code, and modified previewers so that it can jump to the specific line when we type the line number.

tjdevries:
Hm, i'm not sure I love where the code is currently. I'd have to think about it a bit more -- perhaps it could be passed via one of the existing callbacks. Do you have any thought @luyuhuang about somewhere to put this that isn't hard coded into the picker? I don't like putting this kind of logic there.

luyuhuang:
I think if you don't like it to be hard-coded into the picker, we might need something like a "prompt handler" to do these. Now we don't have such a place. Prompt handler, I mean like `_get_prompt` and `_get_next_filtered_prompt`, is indeed hard-coded into the picker.  When it becomes complex, you may prefer to move it into an independent module.

I think it might be not that clear to put it into a callback like `on_input_filter_cb` if we regard it as a part of the "main process". Each part of the main process should be visible in the main procedure code.

Innei:
Are there any updates?

lua/telescope/pickers.lua
```diff
@@ -1540,6 +1541,33 @@ function Picker:_get_next_filtered_prompt()
   return prompt
 end
 
+function Picker:_handle_line_number(prompt)
```
dmtrKovalenko, line 1544: Just an idea, I can help on that or iterate after. But what do you think about adding column here as well? It will be able to handle the paths output from the compiler that looks like 

```
/some/path/file.rs:121:64
```

luyuhuang: I think that'll be better. to me, however, it's enough to just jump to a specific line.

Conni2461: like i already said [here](https://github.com/nvim-telescope/telescope.nvim/issues/2650) i kinda dont want this, but it also dependens on the complexity of the resulting code. So i'll do my final judgement once i see the code.

arnevm123:
Hi, I'd love to get this feature in telescope, what is the current status of this?
And could I provide any help?

Conni2461:
experimental support for this was implemented here: https://github.com/nvim-telescope/telescope.nvim/pull/2791

PR #2314, opened 2023-01-08
feat(pickers): add opt.show_remote_tracking_branches to git_branches

64:
# Description

Add a new option `show_remote_tracking_branches` to the `git_branches` picker so that the user can hide remote tracking branches like `origin/main` if they only care about local branches. Defaults to true which is the current behaviour.

## Type of change

# How Has This Been Tested?

- [x] Use `git_branch` picker with `opts.show_remote_tracking_branches = true` and verify tracking branches are shown
- [x] Use `git_branch` picker with `opts.show_remote_tracking_branches = false` and veryify tracking branches are not shown
- [x] Use `git_branch` picker with no options and verify that remote tracking branches are still shown

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1028,10 +1028,17 @@ builtin.git_branches({opts})                *telescope.builtin.git_branches()*
         {opts} (table)  options to pass to the picker
 
     Options: ~
-        {cwd}          (string)   specify the path of the repo
-        {use_git_root} (boolean)  if we should use git root as cwd or the cwd
-                                  (important for submodule) (default: true)
-        {pattern}      (string)   specify the pattern to match all refs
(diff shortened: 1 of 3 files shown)


Conni2461:
Thanks for the PR :)

64:
> I ran docgen and looking at the output, is the option name too long? what do you think about making it shorter?

Don't mind -- I've seen 'remote tracking' branches referred to as both 'remote branches' and 'tracking branches' so figured it'd be clearest if I just put both words in 😄 but `show_remote_branches` would be my second choice, unless you have an alternative suggestion?

Conni2461:
yeah, lets just merge.

Thanks for the PR :)

PR #2317, opened 2023-01-09
feat: delete file with C-r keybind

skykosiner:
# Description
A lot of the time I find my self having a file in telescope and I wanted a quick way to delete instead of going into NETRW. This `<C-r>` keybind quickly allows you to fuzzy search for a file then delete it instead of having to find where it is in your filetree

## Type of change

# How Has This Been Tested?

Opend fuzzy file searcher, searched for file, ran <C-r> to delete file pressed y to confirm and the file was gone

**Configuration**:
* Neovim version (nvim --version): NVIM v0.9.0-dev-642+g50f03773f
* Operating system and version: Arch Linux 6.0.12

# Checklist:

- [x ] I have performed a self-review of my own code

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -267,6 +267,22 @@ actions.select_horizontal = {
   end,
 }
 
+--- Perform deletes the current file telescope has selcted
+actions.delete_file = {
+  action = function(prompt_bufnr)
+    local content = require("telescope.actions.state").get_selected_entry()
+
+    local path = content.cwd .. "/" .. content.value
+    local shortPath = vim.fn.pathshorten(vim.fn.fnamemodify(path, ":."))
+
+    vim.ui.input({ prompt = "Do you want to delete the file " .. shortPath .. " (y/n): " }, function(input)
+      if input == "y" then
+        vim.cmd("!rm -rf " .. path)
+      end
+    end)
+  end,
+}
+
 --- Perform 'vertical' action on selection, usually something like<br>
 ---`:vnew <selection>`
 ---
--- a/lua/telescope/mappings.lua
+++ b/lua/telescope/mappings.lua
@@ -148,6 +148,8 @@ mappings.default_mappings = config.values.default_mappings
       ["<C-v>"] = actions.select_vertical,
       ["<C-t>"] = actions.select_tab,
 
+      ["<C-r>"] = actions.delete_file,
+
       ["<C-u>"] = actions.preview_scrolling_up,
       ["<C-d>"] = actions.preview_scrolling_down,
 


skykosiner:
idk why it says so much code was changed, I only added about 30 lines and removed 0. I think it was my styling changed some stuff then I changed it back to how the `.stylelua` how is it

lua/telescope/mappings.lua
```diff
@@ -148,6 +148,8 @@ mappings.default_mappings = config.values.default_mappings
       ["<C-v>"] = actions.select_vertical,
       ["<C-t>"] = actions.select_tab,
 
+      ["<C-r>"] = actions.delete_file,
```
Conni2461, line 151: doesnt make sense to map that on all pickers. Why would we want this for a picker like: `help_tags` ...

Conni2461:
hmm, i dont think we need this in core. we have functions like that already in telescope-file-browser: https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/master/lua/telescope/_extensions/file_browser/actions.lua#L429

I am currently not sure if its portable and can be mapped to other pickers but yeah. We kinda made the decision to not have filesystem operations in telescope core and moved out (fdschmidt did move out everything) the filebrowser picker into a extension.
Because if we add a delete action you can also argue for other filesystem actions, copy, move, touch ..... and thats kinda out of scope, imo.

If you need that you can just have a action in your telescope config and add an entry in our wiki: https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes

lua/telescope/mappings.lua
tjdevries: agreed, should not be default mappings, but I'm OK with adding the action I think.

lua/telescope/actions/init.lua
```diff
@@ -267,6 +267,22 @@ actions.select_horizontal = {
...
+actions.delete_file = {
+  action = function(prompt_bufnr)
+    local content = require("telescope.actions.state").get_selected_entry()
+
+    local path = content.cwd .. "/" .. content.value
+    local shortPath = vim.fn.pathshorten(vim.fn.fnamemodify(path, ":."))
+
+    vim.ui.input({ prompt = "Do you want to delete the file " .. shortPath .. " (y/n): " }, function(input)
+      if input == "y" then
+        vim.cmd("!rm -rf " .. path)
```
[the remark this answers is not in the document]
tjdevries: we can use `os.remove` instead and check the return value I think.

Conni2461:
yeah i dont think we need this in core, feel free to add a configuration recipe, still thanks for opening a PR :)

PR #2324, opened 2023-01-14
feat: apply jump_type only if the definition file is different from the current file

luyuhuang:
# Description

Apply `jump_type` rule only if the definition file is different from the current file. I think most people expect open the definition on a new tab (or a split pane) only when we need to go to another file.

## Type of change

# How Has This Been Tested?

Set a key mapping `vim.keymap.set('n', 'gd', function() builtin.lsp_definitions{jump_type = 'tab'} end, bufopts)`:

- [x] Move the cursor on a symbol that has only one definition and type `gd`. If the definition is in the current file, it'll just jump to the corresponding position of the current file; otherwise it'll open the definition on a new tab.

Set a key mapping `vim.keymap.set('n', 'gr', function() builtin.lsp_references{jump_type = 'tab'} end, bufopts)`:

- [x] Move the cursor on a symbol that has only one reference and type `gr`. If the reference is in the current file, it'll just jump to the corresponding position of the current file; otherwise it'll open the reference on a new tab. 

**Configuration**:
* Neovim version (nvim --version): NVIM v0.8.2
* Operating system and version: Ubuntu 20.04 LTS (WSL2)

# Checklist:

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -24,3 +24,11 @@ globals = {
 read_globals = {
   "vim",
 }
+
+files = {
+  ["lua/telescope/builtin/init.lua"] = {
+    ignore = {
+      "631", -- allow line len > 120
+    }
+  },
+}
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1454,8 +1454,10 @@ builtin.lsp_references({opts})            *telescope.builtin.lsp_references()*
         {include_current_line} (boolean)  include current line (default:
                                           false)
         {jump_type}            (string)   how to goto reference if there is
-                                          only one, values: "tab", "split",
-                                          "vsplit", "never"
+                                          only one and the definition file is
+                                          different from the current file,
+                                          values: "tab", "split", "vsplit",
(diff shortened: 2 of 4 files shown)


Conni2461:
thanks :)

ahouseago:
Hi @luyuhuang. I'd like to add a config value to control this because I don't want to reuse the same window when I've specified a jump type.

I wanted to check before I did that: if I were to revert the change from this PR, would using the `reuse_win` option work for you? Or do you want it so it only ignores the `jump_type` when the definition exists in the same file?

luyuhuang:
> Hi @luyuhuang. I'd like to add a config value to control this because I don't want to reuse the same window when I've specified a jump type.
> 
> I wanted to check before I did that: if I were to revert the change from this PR, would using the `reuse_win` option work for you? Or do you want it so it only ignores the `jump_type` when the definition exists in the same file?

Yes it's OK for me.

leoperegrino:
@ahouseago @luyuhuang so, should this be reverted?

PR #2332, opened 2023-01-18
0.1.x

HUAHUAI23:
`Telescope current_buffer_fuzzy_find` change cursor position but it doesn't seem to add current cursor position to jumplist, so i add `vim.cmd [[ normal! m' ]]` to lua/telescope/builtin/__files.lua cols 412  <https://neovim.discourse.group/t/how-to-add-the-current-cursor-position-to-the-jump-list/1585>

**no add `vim.cmd [[ normal! m' ]]`**
https://user-images.githubusercontent.com/43649186/213096883-f72d0ef9-0715-4560-9e9b-97710a9062fe.mp4

**after add `vim.cmd [[ normal! m' ]]`**
https://user-images.githubusercontent.com/43649186/213097016-1025d74b-c27d-41ed-8bd0-08b1258555cd.mp4


Conni2461:
can you please open a PR against master, dev is done against master and fixes are backported to 0.1.x (by cherry-picking commits)

Also this shouldn't be done every time `current_buffer_fuzzy_find` is opened but rather its already done in the select action we only have to enable it. So all that needs to be done is to do `push_cursor_on_edit = true,` in the picker configuration

HUAHUAI23:
ok,thanks

Conni2461:
i resolved that myself because this pr includes like 77 unrelated already merged commits and it was to much hustle for me to figure out what went wrong with git here.

None the less, thanks :) bug is fixed on master and will be backported to 0.1.x later today

HUAHUAI23:
thanks,ha ha ,i just change little lines，and don't konw why some much diff 😅

Issue #1048, opened 2021-07-25
Open multiple files at once


kishikaisei:
**Is your feature request related to a problem? Please describe.**
When using find_files, I would like to be able to open multiple results at once.

**Describe the solution you'd like**
In telescope, in normal mode, having the ability to toggle selection using space for instance, and then open them at once.

**Describe alternatives you've considered**
The only alternative was to open each file one by one by using the same search text, and try to remember which file I opened already.

**Additional context**
I find the need for this feature in 2 occasions, having multiple files with same name but in different folders and when I wanted to open all files specific to a feature in a codebase.

caojoshua:
With default bindings, you can select entries with `<Tab>` or `<Shift>+<Tab>`. You can somewhat get what you're looking for with `<Alt> + q`, which send the selected entries to the quickfix list. I believe that there is currently not an action to open select entries in a buffer, but could be implemented.

```
# toggle entry
require('telescope.actions').toggle_selection

# send selected to quickfix
require('telescope.actions').send_selected_to_qflist
```

defaults defined [here](https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/mappings.lua)

kishikaisei:
That's an alternative way which seems promising, but how can you the said files from the qflist after sending the files there?

caojoshua:
> but how can you the said files 

I'm gonna guess you mean how to **open** the said files

I'm not sure if there is a quick one liner to do this, but a quick google search show me something like: https://gist.github.com/larsthegeek/4063239

fdschmidt93:
I'm currently working on multi-select style actions and will be opening a PR with my WIP branch soon. It's quite convoluted to support everything nicely, so it'll be ready when it's ready :)

shlomocarmeter:
Maybe until @fdschmidt93 finishes his PR something like this can help (That's what I'm currently using):
```
local actions = require('telescope.actions')
local action_state = require('telescope.actions.state')

function custom_actions.fzf_multi_select(prompt_bufnr)
  local picker = action_state.get_current_picker(prompt_bufnr)
  local num_selections = table.getn(picker:get_multi_selection())

  if num_selections > 1 then
    local picker = action_state.get_current_picker(prompt_bufnr)
    for _, entry in ipairs(picker:get_multi_selection()) do
      vim.cmd(string.format("%s %s", ":e!", entry.value))
    end
    vim.cmd('stopinsert')
  else
[...]

Issue #416, opened 2021-01-11
[feature] Multi-selection


tjdevries:
Currently half implemented.

Need to fix highlights and other items related to it.

Needs other actions as well.

baldore:
Hey, first at all, thanks for this great plugin. I'm wondering if, after this feature is added, you can make multiple selections of Rg results and add them to the quickfix. I ask because, if this is possible already, I would love to migrate to this plugin from Fzf.vim.

Conni2461:
We can send entries to qflist. You can already do this with all results and the function for sending only selected to qflist is also done. You need to manually set it up right now:
```lua
local actions = require('telescope.actions')
require('telescope').setup{
  defaults = {
    mappings = {
      i = {
        ["<C-w>"] = actions.send_selected_to_qflist,
        ["<C-q>"] = actions.send_to_qflist,
      },
      n = {
        ["<C-w>"] = actions.send_selected_to_qflist,
        ["<C-q>"] = actions.send_to_qflist,
      },
    },
  }
}
```

So this will be possible when multi selection is done and enabled.

larrybotha:
For anyone looking for a basic fzf-type multiselection, where:

- `<CR>` with selections <= 1 opens a file for editing
- `<CR>` with selections > 1 opens the quick fix window (currently doesn't open the file for editing)

I've done the following:

```lua
local actions = require("telescope.actions")
local action_state = require("telescope.actions.state")

local custom_actions = {}

function custom_actions.fzf_multi_select(prompt_bufnr)
    local picker = action_state.get_current_picker(prompt_bufnr)
    local num_selections = table.getn(picker:get_multi_selection())

    if num_selections > 1 then
        -- actions.file_edit throws - context of picker seems to change
        --actions.file_edit(prompt_bufnr)
        actions.send_selected_to_qflist(prompt_bufnr)
        actions.open_qflist()
    else
        actions.file_edit(prompt_bufnr)
    end
end

require("telescope").setup {
    defaults = {
        mappings = {
            i = {
                -- close on escape
                ["<esc>"] = actions.close,
                ["<tab>"] = actions.toggle_selection + actions.move_selection_next,
                ["<s-tab>"] = actions.toggle_selection + actions.move_selection_previous,
                ["<cr>"] = custom_actions.fzf_multi_select
            },
            n = {
                ["<tab>"] = actions.toggle_selection + actions.move_selection_next,
[...]

PR #2337, opened 2023-01-20
feat: extend git_delete_branch to delete of all selected branches

### LINKED ISSUE #2363: Feature Request: Delete multiple git branches at once
Having worked on a few branches in parallel it is always irritating to clean up my local workspace, to the point that I prefer using the command line over telescope. 

Telescope has already a very nice multi select functionality it would be great if it would have an effect when deleting branches. Simply tabbing through the branches I want gone then press the delete branch key binding.

using telescope for this task is clunky, apart from that I could go with other plugins or the command line, which would solve the problem but I like telescope and the functionality feels like it should work.
I added a script in my config to make it work like I want but I think this should be standard functionality.

**Additional context**
I made a [PR](https://github.com/nvim-telescope/telescope.nvim/pull/2337) for the change, it is a small change and it would make my neovim config smaller to have it merged ;)

mostley:
# Description

Extends the git_delete_branch action to delete all selected branches. It also keeps the picker open.
The default behavior of deleting the branch on which the cursor stands on still works.

## Type of change

# How Has This Been Tested?

- [x] tested in local vim

**Configuration**:
* Neovim version (nvim --version): 0.8.2
* Operating system and version: OSX 12.6

# Checklist:

Closes #2363

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -2719,7 +2719,7 @@ actions.git_track_branch({prompt_bufnr}) *telescope.actions.git_track_branch()*
 
 
 actions.git_delete_branch({prompt_bufnr}) *telescope.actions.git_delete_branch()*
-    Delete the currently selected branch
+    Delete all the selected branches
 
 
     Parameters: ~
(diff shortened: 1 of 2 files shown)


mostley:
Is this not something that is interesting for core? I felt like this would actually get the behaviour closer to how the general UX patterns. The problem is fixed for me since I added something similar to my config. This MR is just me wanting to contribute.

Conni2461:
sorry for the late response, i am currently the only maintainer and i am swamped with everything. 

yeah that is something we could add this but i wanna go back and resolve this on multiselection action level see #416 and #1048

thanks for the PR :) i am going to resolve stylua ci so i can merge :)

PR #2342, opened 2023-01-23
pick lsp_reference from multpile clients

hunterliao29:
# Description

When one buffer is attached to multiple lsp clients, current implementation will only allow one client's result be shown. 
This cause issues when two client return different set of references. 

## Type of change

# How Has This Been Tested?

- [x] Test A: Create an angular project using
```
ng new test
nvim test/src/app/app.component.ts
``` 
```Telescope lsp_references``` on ```title```(line 9)

**Configuration**:
* Neovim version (nvim --version): 0.8.2 and master
* Operating system and version: macOS 13.1

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -14,6 +14,24 @@ lsp.references = function(opts)
   local lnum = vim.api.nvim_win_get_cursor(opts.winnr)[1]
   local params = vim.lsp.util.make_position_params(opts.winnr)
   local include_current_line = vim.F.if_nil(opts.include_current_line, false)
+  local buf_clients_num = #vim.tbl_keys(vim.lsp.get_active_clients { bufnr = opts.bufnr })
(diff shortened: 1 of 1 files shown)


pidgeon777:
Do you think the same could be done for LSP definitions and implementations?

tjdevries:
This is a good idea -- I think maybe we should do this as well for definitions and implementations. Do you want to do that or would you prefer if I did that in a separate PR?

pidgeon777:
> Do you want to do that or would you prefer if I did that in a separate PR?

@tjdevries It would be great I think if you could proceed with it 👍.

eslam-allam:
perhaps this should be extended to other lsp functionality as well e.g. lsp_document_symbols ?

pidgeon777:
> perhaps this should be extended to other lsp functionality as well e.g. lsp_document_symbols ?

Yes, also that, preventing any duplicate.

PR #2288, opened 2022-12-27
feat(git): git_commits shows the current branch graph

CyanJoeng:
# Description

I'd like to see a graph view in git_commits picker, and notice the closed issue (#1041) when I try to find some solutions.
I'm not familiar with the Lua language, so, a little modifications are written to achieve the basic demand.

Feat # (#1041)
show the current branch graph when viewing git_commits

## Type of change

- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Invoking the git picker function **builtin.git_commits** by a vim key mapping, the the commits log with graph will displayed in the popup window.

_NB._ commit logs showing with graph will insert extra entries with no SHA value, so a error msg will shown in the vim msg bar if these extra entries selected.

PR #2357, opened 2023-01-31
revise the #2288

CyanJoeng:
# Description

the revised version of #2288 

Fixes # (issue)
-   Fix: remove graph option in bcommits
-   Feat: add config to control the graph behaviour in commits

# How Has This Been Tested?

invoking the function  `builtin.git_bcommits` and `builtin.git_commits` with configuration 

**Configuration**:
```
require('telescope').setup{
  defaults = {
    -- config_key = value,
    git_log_graph = {show = true, all_branches = true, inverted = false},
  },
}
```

### DIFF
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -831,12 +831,12 @@ append(
     Default: require("telescope.previewers").buffer_previewer_maker]]
 )
 
- append(
-   "git_log_graph",
-   {
+append(
+  "git_log_graph",
+  {
     show = false,
     all_branches = false,
-    inverted = false
+    inverted = false,
   },
   [[
   Show git log with --graph option.
@@ -851,7 +851,7 @@ append(
         false (default)
         true
     Default: {show = false, all_branches = false, inverted = false}]]
- )
+)
 
 -- @param user_defaults table: a table where keys are the names of options,
 --    and values are the ones the user wants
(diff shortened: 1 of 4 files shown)


CyanJoeng:
Sorry to cause so many complaints, hope this version will be helpful.

configurations are added with not open this feature by default 
and the descriptions are noted in the document file.

Please thoroughly inspect the code before merging.
Cheers.

lua/telescope/config.lua
```diff
@@ -831,6 +831,28 @@ append(
...
+        false (default)
+        true
+    - all_branches              showing all branches with graph
+        false (default)
+        true
+    - inverted                  if first git log showing on the top
+        false (default)
+        true
+    Default: {show = false, all_branches = false, inverted = false}]]
+)
```
Conni2461, line 854: not gonna happen, there is no need to add a global configuration that only affects one builtin.

Conni2461:
at this point i think we should just revert, this feature. I don't think its that useful and it makes our code just bloated. On top of that this feature could already be realized with opts with configuration.

None the less, its still broken:

git_commits with 0.1.x (which did not get 2288)
image

with your current branch (2357)
image

with the suggested options enabled (`git_log_graph = {show = true, all_branches = true, inverted = false},`)
image

for good measure with current master (prior to revert)
image

Its just straight up broken and i should have caught that in #2288 

on top of that its so much slower, i'm getting commits instantly with 0.1.x but with graph i end up waiting up to 20s

CyanJoeng:
No problem, you are right.

PR #2374, opened 2023-02-10
live_grep: Allow disabling of filename in result list

kstrafe:
= Description =

This patch adds a new option to the `live_grep` picker named `disable_filenames`. It will hide the filename in the result buffer.

This is useful when we are using `live_grep` for grepping inside a single file; it is not useful to see the path to the file since that is already known. This is especially true for files in deep paths.

In this case I am using `live_grep` to create an interactive log filtering mechanism. By using a custom sorter we can sort by line numbers. By removing the filename we can see a lot more of the logs.

Fixes #<issue not filed>

== Type of change ==

== How Has This Been Tested? ==

=== Test ===
Run the following in neovim with telescope available, inside a populated buffer. The results window will not show filenames: `:lua require('telescope.builtin').live_grep { search_dirs = { vim.api.nvim_buf_get_name(0) }, disable_filenames = true }`

Setting `disable_filenames = false` or removing the option makes the filenames reappear.

**Configuration**:
* Neovim version (nvim --version): 0.8.1
* Operating system and version: Linux 6.0.10 (NixOS 23.05.20221130.e76c78d (Stoat) x86_64)

= Checklist: =

- [-] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -801,6 +801,8 @@ builtin.live_grep({opts})                      *telescope.builtin.live_grep()*
         {max_results}         (number)          define a upper result value
         {disable_coordinates} (boolean)         don't show the line & row
                                                 numbers (default: false)
+        {disable_filenames}   (boolean)         don't show the file path
+                                                (default: false)
 
 
 builtin.grep_string({opts})                  *telescope.builtin.grep_string()*
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -272,6 +272,7 @@ do
 
     local disable_devicons = opts.disable_devicons
     local disable_coordinates = opts.disable_coordinates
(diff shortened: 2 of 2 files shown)


EpsilonKu:
Is it possible to make opposite? To show only filename?

tjdevries:
Would it make more sense to allow `path_display` style options here? We already have a lot of logic for formatting and managing path displays.

Conni2461:
this is already possible with `path_display`: `require("telescope.builtin").live_grep({ path_display = function() return "" end })` but then you rend up with an additional `:` in front of the coordinates, i would be open to fixing that but not introduce another option value.

If you wanna fix that please open a new PR. thanks :)

> To show only filename?

for only showing filenames we also already have a option: `disable_coordinates`

PR #2381, opened 2023-02-14
[POC] compact theme

jrnxf:
# Description

I really love telescope.nvim, but I've never been a big fan of having three separate (looking) popups for the prompt, results, and preview. Looking at the plenary popup options, I noticed there were options for z-indexing that weren't being used within telescope. By adding those values, and doing some slight modifications to the vertical and horizontal layout strategies, I was able to produce the following compact theme.

To date it works with both the horizontal / vertical themes (not mirrored - yet) and either prompt_position.

If there is interest in me completing this work I would be happy to do so!

Here are a few example screenshots of the theme in action. (Note: I prefer to not have prompt/results/preview titles, but they do work as you'd expect due to the z-indexing):

image

image

## Type of change

# How Has This Been Tested?

TBD

**Configuration**:
* Neovim version (nvim --version): nightly
* Operating system and version: MacOS Ventura

# Checklist:

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -77,6 +77,7 @@ function Picker:new(opts)
     entry_prefix = vim.F.if_nil(opts.entry_prefix, config.values.entry_prefix),
     multi_icon = vim.F.if_nil(opts.multi_icon, config.values.multi_icon),
 
+    theme = vim.F.if_nil(opts.theme, "dropdown"),
     initial_mode = vim.F.if_nil(opts.initial_mode, config.values.initial_mode),
     _original_mode = vim.api.nvim_get_mode().mode,
     debounce = vim.F.if_nil(tonumber(opts.debounce), nil),
--- a/lua/telescope/pickers/layout_strategies.lua
+++ b/lua/telescope/pickers/layout_strategies.lua
@@ -344,7 +344,7 @@ layout_strategies.horizontal = make_documented_layout(
     height, h_space = calc_size_and_spacing(height, max_lines, bs, 2, 4, 1)
 
     prompt.height = 1
-    results.height = height - prompt.height - h_space
+    results.height = height - prompt.height - h_space + (self.theme == "minimal" and 1 or 0)
 
     if self.previewer then
       preview.height = height - 2 * bs
@@ -367,10 +367,12 @@ layout_strategies.horizontal = make_documented_layout(
     preview.line = math.floor((max_lines - height) / 2) + bs + 1
     if layout_config.prompt_position == "top" then
       prompt.line = preview.line
-      results.line = prompt.line + prompt.height + 1 + bs
+      results.line = prompt.line + prompt.height + bs + (self.theme == "minimal" and 0 or 1)
+      results.zindex = prompt.zindex + (self.theme == "minimal" and 1 or 0)
     elseif layout_config.prompt_position == "bottom" then
       results.line = preview.line
-      prompt.line = results.line + results.height + 1 + bs
+      prompt.line = results.line + results.height + bs + (self.theme == "minimal" and 0 or 1)
+      prompt.zindex = results.zindex + (self.theme == "minimal" and 1 or 0)
     else
       error(string.format("Unknown prompt_position: %s\n%s", self.window.prompt_position, vim.inspect(layout_config)))
     end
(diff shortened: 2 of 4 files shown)


craigmac:
Can we call it Helix theme? It looks like helix defaults, or very close to my eye.

jrnxf:
> Can we call it Helix theme? It looks like helix defaults, or very close to my eye.
> 
> Also, if this accepted then it looks we now have to check for "minimal" hardcoded in a few places, can we not refactor these out to something more general so we are not special casing in our calculations here?

Could you elaborate on this? I have a few opinion on how the current layout strategies work as is. It seems to me that having knowledge of the theme in the layout strategy is a cleaner approach then having to supply extra parameters to the layout config that essentially hint at what theme is being used. For example, I originally passed a value called `paddingless` to the layout config, but felt like that was silly when in reality all I wanted was to let the layout config know the theme was `minimal`. Also, with the way things work currently, some layout configs flat out don't work with certain layout strategies, with no errors thrown. If the layout strategies could check for the presence of the theme being supplied at the very least a warning message could be shown to the user.

max397574:
I'm clearly against the "helix" but I also think that minimal doesn't describe what it is. There aren't like things left out. I suggest `compact`.

craigmac:
Ok I get it now, you've obviously thought about this more than I have, my mistake there. On the naming: minimal is subjective but I can't think of anything better, it's helix-y ;)

max397574:
I'd have no example what helix style would be
and I bet there are many people who perhaps know that helix got a thing like telescope builtin (ig a lot even don't know that) but most of them won't know how exactly this looks. Especially not regarding to whether the three windows are overlapped or not. 
Compact perfectly describes this and people could even imagine what it'd look like without ever having seen it

uloco:
Looks nice, thanks. I like compact a lot, too!

jrnxf:
Yeah I'm down for calling it compact!

lua/telescope/pickers.lua
```diff
@@ -77,6 +77,7 @@ function Picker:new(opts)
     entry_prefix = vim.F.if_nil(opts.entry_prefix, config.values.entry_prefix),
     multi_icon = vim.F.if_nil(opts.multi_icon, config.values.multi_icon),
 
+    theme = vim.F.if_nil(opts.theme, "dropdown"),
```
tjdevries, line 80: doesn't this change the default theme?

lua/telescope/pickers/layout_strategies.lua
```diff
@@ -344,7 +344,7 @@ layout_strategies.horizontal = make_documented_layout(
     height, h_space = calc_size_and_spacing(height, max_lines, bs, 2, 4, 1)
 
     prompt.height = 1
-    results.height = height - prompt.height - h_space
+    results.height = height - prompt.height - h_space + (self.theme == "minimal" and 1 or 0)
```
tjdevries, line 347: i think instead we could perhaps have an option that themes pass of whether they want the lines to overlap or not. So you could have an opt for `overlap`, which would apply these changes.

lua/telescope/pickers.lua
jrnxf: oh my bad, I must have though dropdown was the default when I wrote this! Is the default one just unnamed?

lua/telescope/pickers/layout_strategies.lua
jrnxf: oooh I like that idea! would be nice to have the option for all themes

ditsuke: I really like this idea too!

MordechaiHadad:
Is this PR dead?

jamestrew:
my preference is with moving forward with #2572 but i'll defer to tj/conni

NormTurtle:
when merge?

Conni2461:
> when merge?

probably never. The PR is most likely dead and we are currently more interested in #2572

HoldenLucas:
time to close this? #2572 is merged

jamestrew:
Technically we don't have a compact theme yet. Just the mechanism to facilitate one better.

jamestrew:
@MunifTanjim has kindly added a snippet in our wiki to achieve this using the new `create_layout` API from #2572 

https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes#fused-layout

I think we can close this PR with this.

PR #2398, opened 2023-02-22
feat(git): Add bcommits_range picker

aaronkollasch:
# Description

[fzf.vim](https://github.com/junegunn/fzf.vim#commands) has a feature where the `:BCommits` command shows only commits affecting the selected lines, if run from visual mode.

This functionality cannot be replicated in telescope with config options alone. Fzf.vim uses the [arguments](https://github.com/junegunn/fzf.vim/blob/dc71692255b62d1f67dc55c8e51ab1aa467b1d46/autoload/fzf/vim.vim#L1278) `git log --no-patch -L start,end:file`, which are not compatible with the filename that telescope appends to `git_commands`. (I tried some hacks like adding an argument to `git_commands` to make git ignore the filename argument, but they didn't work).

## Type of change

~~This PR only affects bcommits behavior when run in visual mode, so I think it's a minor change but could be considered breaking.~~

This PR adds `bcommits_range` to show commits affecting a range of lines. `bcommits_range` can get the range of selected lines in visual mode, and supports an operator-pending mode to get lines from the next text object or motion. To support operators, this PR also adds a new `telescope.operators` module.

# How Has This Been Tested?

- [x] Test A: Visually select lines of code, then activate the `bcommits_range` picker. Only commits affecting the selected lines should be shown. Check all previewer outputs to make sure they make sense.
- [x] Test B: Run test A from a subdirectory of the git directory. Behavior should be the same as before.
- [x] Test C: Run `bcommits_range` in normal mode, all commits for the file should be shown.
- [x] Test D: Run `<cmd>Telescope git_bcommits_range first=1 last=10<CR>` in normal mode. Commits affecting the first 10 lines should be shown.
- [x] Test E: Run `<cmd>Telescope git_bcommits_range operator=true<CR>` in normal mode followed by a motion or text object, e.g. `ip` for inner paragraph. All commits affecting the selected range should be shown.

**Configuration**:
* Neovim version (nvim --version): v0.8.3
* Operating system and version: macOS 13.2.1 Ventura

# Notes

- The line range in the file may not match the range used to filter commits, if there are uncommitted changes to the file (whether staged or not, it seems). This affects fzf.vim as well, and solving it could be tricky.
- I've supported visual block mode by including a ^V in `__git.lua`, not sure if that's ok, or if there's a better way to detect visual mode.

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__git.lua
+++ b/lua/telescope/builtin/__git.lua
@@ -198,8 +198,10 @@ git.bcommits_range = function(opts)
   opts.current_line = (opts.current_file == nil) and get_current_buf_line(opts.winnr) or nil
   opts.current_file = vim.F.if_nil(opts.current_file, vim.api.nvim_buf_get_name(opts.bufnr))
   opts.entry_maker = vim.F.if_nil(opts.entry_maker, make_entry.gen_from_git_commits(opts))
-  local git_command =
-    vim.F.if_nil(opts.git_command, { "git", "log", "--pretty=oneline", "--abbrev-commit", "--no-patch", "-L" })
+  opts.git_command = vim.F.if_nil(
+    opts.git_command,
+    git_command({ "log", "--pretty=oneline", "--abbrev-commit", "--no-patch", "-L" }, opts)
+  )
   local visual = string.find(vim.fn.mode(), "[vV]") ~= nil
 
   local line_number_first = opts.first
@@ -227,7 +229,7 @@ git.bcommits_range = function(opts)
       prompt_title = "Git BCommits in range",
       finder = finders.new_oneshot_job(
         vim.tbl_flatten {
-          git_command,
+          opts.git_command,
           line_range,
         },
         opts
--- a/lua/telescope/builtin/__git.lua
+++ b/lua/telescope/builtin/__git.lua
@@ -215,6 +215,12 @@ git.bcommits_range = function(opts)
   elseif opts.operator_callback then
     line_number_first = vim.fn.line "'["
     line_number_last = vim.fn.line "']"
+  elseif line_number_first == nil then
+    table.remove(git_command)
+    table.insert(git_command, "--follow")
+    opts.git_command = git_command
+    git.bcommits(opts)
+    return
   end
   local line_range =
     string.format("%d,%d:%s", line_number_first, line_number_last, Path:new(opts.current_file):make_relative(opts.cwd))
--- a/lua/telescope/builtin/__git.lua
+++ b/lua/telescope/builtin/__git.lua
@@ -216,11 +216,8 @@ git.bcommits_range = function(opts)
     line_number_first = vim.fn.line "'["
     line_number_last = vim.fn.line "']"
   elseif line_number_first == nil then
-    table.remove(git_command)
-    table.insert(git_command, "--follow")
-    opts.git_command = git_command
-    git.bcommits(opts)
-    return
+    line_number_first = vim.F.if_nil(line_number_first, vim.fn.line ".")
+    line_number_last = vim.F.if_nil(line_number_last, vim.fn.line ".")
   end
   local line_range =
     string.format("%d,%d:%s", line_number_first, line_number_last, Path:new(opts.current_file):make_relative(opts.cwd))
--- a/lua/telescope/builtin/__git.lua
+++ b/lua/telescope/builtin/__git.lua
@@ -193,6 +193,14 @@ git.bcommits = function(opts)
     :find()
 end
 
+local last_bcommits_range_opts = {}
+
+local bcommits_range_callback = function()
+  last_bcommits_range_opts.operator = false
+  last_bcommits_range_opts.operator_callback = true
+  git.bcommits_range(last_bcommits_range_opts)
+end
+
 git.bcommits_range = function(opts)
   opts.current_line = (opts.current_file == nil) and get_current_buf_line(opts.winnr) or nil
   opts.current_file = vim.F.if_nil(opts.current_file, vim.api.nvim_buf_get_name(opts.bufnr))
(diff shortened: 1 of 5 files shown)


tjdevries:
What do you think about just having a separate picker for this rather than overriding the original?

For example, we could have `bcommits_range` that accepts a start and end range, and we could generate those by default if it's in visual mode (but you could generate those other ways if you weren't in visual mode. For example, you could get the start and end range of a function with tree-sitter and then call this to get the commits over this function).

aaronkollasch:
I pushed a commit with a separate `bcommits_range` picker. My one hesitation with this is the amount of repeated code/documentation, as any updates to `bcommits` would be expected to propagate to `bcommits_range`.

aaronkollasch:
I pushed a commit which adds operator mode as an option, and it works as expected. For example, `<cmd>Telescope git_bcommits_range operator=true<CR>` followed by `if` shows all commits affecting the function under the cursor.

tjdevries:
Love the operator pending stuff -- my only thought is that do you think it's possible to maybe move the operator pending logic outside of this function?

Perhaps we can make a new module `telescope.operators`. I think this idea is pretty cool and we may want to use the concept for some other pickers.

What are your thoughts on that?

aaronkollasch:
That sounds good. I've updated the PR to move the operators functionality to `telescope.operators`.

lua/telescope/operators.lua
```diff
@@ -0,0 +1,23 @@
...
+--- See `:h map-operator`
+---
+---@param callback function: the function to call after exiting operator-pending
+---@param opts table: options to pass to the callback
+operators.run_operator = function(callback, opts)
+  last_operator = { callback = callback, opts = opts }
+  vim.o.operatorfunc = "v:lua.require'telescope.operators'.operator_callback"
+  -- feed g@ to enter operator-pending mode
+  -- 'i' required for which-key compatibility, etc.
+  vim.api.nvim_feedkeys("g@", "mi", false)
```
tjdevries, line 20: do you think after running this line we should clear the operator? Or perhaps on line 7? This way we drop the reference to the picker and it can be garbage collected later, which seems like a good idea.

Maybe it should be:

```lua
operators.operator_callback = function()
  last_operator.callback(last_operator.opts)
  last_operator = { callback = function(_) end, opts = {} }
end

Comment (aaronkollasch) [reply lua/telescope/operators.lua:20]: Right, it's only a reference to the most recent picker/callback function used with `operators.run_operator`. At least for this picker it would just keep a reference to `git.bcommits_range` and its opts.

**Comment by aaronkollasch** [comment]
By the way, what do you think about the names of the arguments to specify the range? I'd originally used `start` and `end` but switched to `first` and `last` because the `end` keyword is a bit inconvenient in lua.
Would `start` and `end` make more sense?

**Comment by aaronkollasch** [comment]
I've rebased this PR on master to include the changes in #2597.

**Comment by jamestrew** [comment]
Thanks for sticking with this^^ 
I'll try to review this in the coming days.

**Comment by jamestrew** [comment]
> By the way, what do you think about the names of the arguments to specify the range? I'd originally used `start` and `end` but switched to `first` and `last` because the `end` keyword is a bit inconvenient in lua. Would `start` and `end` make more sense?
> 
> edit: I will keep `first` and `last` due to the inconvenience of creating tables with an element named "end".

Maybe `from` and `to`? Open to keeping `first` and `last` -- I think they work just as well

File: lua/telescope/builtin/__git.lua [hunk 2]
```diff
@@ -193,6 +194,112 @@ git.bcommits = function(opts)
...
+        opts
+      ),
+      previewer = {
+        previewers.git_commit_diff_to_parent.new(opts),
+        previewers.git_commit_diff_to_head.new(opts),
+        previewers.git_commit_diff_as_was.new(opts),
+        previewers.git_commit_message.new(opts),
+      },
+      sorter = conf.file_sorter(opts),
+      attach_mappings = function()
```
jamestrew, line 244: I think the `attach_mappings` here is identical to that for `bcommits` right? Is it possible to extract this as it's own function closing over `opts` and share this new function between the two pickers?

aaronkollasch:
> > By the way, what do you think about the names of the arguments to specify the range? I'd originally used `start` and `end` but switched to `first` and `last` because the `end` keyword is a bit inconvenient in lua. Would `start` and `end` make more sense?
> > edit: I will keep `first` and `last` due to the inconvenience of creating tables with an element named "end".
> 
> Maybe `from` and `to`? Open to keeping `first` and `last` -- I think they work just as well

`from` and `to` sound good, I think it's a bit more natural than `first` and `last` for specifying a range in a command.

lua/telescope/builtin/__git.lua
aaronkollasch: Yes, they are identical. I suppose a unified `attach_mappings` function makes sense as `bcommits` and `bcommits_range` should be identical aside from the way each generates its list of commits. The only difference between the pickers should be the `prompt_title` and `finder`.

aaronkollasch: Given that almost all of the picker is shared, I wonder if it makes sense to extract the `pickers.new()` call into a shared function that accepts `opts`, `prompt_title`, and `finder`? Those are the only aspects where the pickers differ, and it would ensure that changes to one picker also show up in the other.

aaronkollasch: I added a commit that creates the bcommits picker in a shared function between `bcommits` and `bcommits_range`. What do you think?

jamestrew: Sounds and looks good :+1:

jamestrew:
Beautiful. Thanks again :+1:

aaronkollasch:
Thanks for reviewing!

PR #2434, opened 2023-03-23
feat: Improve Git error output when calling Telescope git command in non-git directory

RoRep1ay:
# Description

Currently, when you're calling `:Telescope git_command` on a non-git directory, it's outputting an error exception detailing that current directory is not a directory. I don't think this error trace is necessary, just outputting message that current directory is not a git directory is enough .

Before: 
image

After:
image

## Type of change

# How Has This Been Tested?

1. On a non-git directory, run one of `Telescope git` command

**Configuration**:
* Neovim version (nvim --version): NVIM v0.8.3
* Operating system and version: Mac OS Ventura

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__git.lua
+++ b/lua/telescope/builtin/__git.lua
@@ -395,7 +395,11 @@ local set_opts_cwd = function(opts)
     local in_bare = utils.get_os_command_output({ "git", "rev-parse", "--is-bare-repository" }, opts.cwd)
 
     if in_worktree[1] ~= "true" and in_bare[1] ~= "true" then
-      error(opts.cwd .. " is not a git directory")
+      utils.notify("builtin.git", {
+        msg = opts.cwd .. " is not a git directory",
+        level = "ERROR",
+      })
+      return false
     elseif in_worktree[1] ~= "true" and in_bare[1] == "true" then
       opts.is_bare = true
     end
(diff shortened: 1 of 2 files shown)


lua/telescope/builtin/__git.lua
```diff
@@ -395,7 +395,11 @@ local set_opts_cwd = function(opts)
     local in_bare = utils.get_os_command_output({ "git", "rev-parse", "--is-bare-repository" }, opts.cwd)
 
     if in_worktree[1] ~= "true" and in_bare[1] ~= "true" then
-      error(opts.cwd .. " is not a git directory")
+      utils.notify("builtin.git", {
+        msg = opts.cwd .. " is not a git directory",
+        level = "ERROR",
+      })
+      return false
```
Conni2461, line 402: i prefer not to use vim.notify for some things that can be used for scripting. using vim.notify completly removes this ability to allow users to script around errors.

https://github.com/nvim-telescope/telescope.nvim/pull/2225 tries to resolve this and i plan to finish it.

we already merged this and had a whole discussion here on why we probably dont want to do it like that. https://github.com/nvim-telescope/telescope.nvim/pull/2181

Conni2461:
so yeah i am going to close this in favor of 2225 or another solution, i am going to figure this out today

PR #2440, opened 2023-03-25
Simplify vimgrep_arguments args

snelling-a:
# Description

ripgrep has an argument that returns results in vim format

```man
       --vimgrep
           Show results with every match on its own line, including line numbers and column numbers. With this option, a line with more than one match will be printed more than once.
```

which would replace `"--color=never", "--no-heading", "--with-filename", "--line-number", "--column"` and still produce the same result
## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Configuration**:

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -532,16 +532,13 @@ telescope.setup({opts})                                    *telescope.setup()*
         Hint: Make sure that color is currently set to `never` because we do
         not yet interpret color codes
         Hint 2: Make sure that these options are in your changes arguments:
-          "--no-heading", "--with-filename", "--line-number", "--column"
+          ("--no-heading", "--with-filename", "--line-number", "--column") or
+          "--vimgrep"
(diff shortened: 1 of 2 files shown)


Conni2461:
thanks :) i've run docgen, so the documentation is also up to date :)

nmrtv:
This change causes crashing in some repositories because Neovim runs out of memory. With old arguments there are no problems. Probably something is wrong with parsing of results, because running rg from a command line works fine with both argument styles.

Using ripgrep 13.0.0.
Neovim 0.9.0 and nightly (crashes on both).

To cause crashing in live grep, I only need to enter a single character, then cpu and memory raises instantly.

nmrtv:
I found that these problems are caused by large minimized js files from here - https://github.com/swagger-api/swagger-ui/tree/master/dist

After removing them everything works fine.

PR #2442, opened 2023-03-27
Always add index to entry

stefanmh:
# Description

https://github.com/nvim-telescope/telescope.nvim/commit/59497d664086c4c7aa24a15a1c3b9ccf718ce9c7 introduced `sorters.fuzzy_with_index_bias`, which gives a scoring boost to earlier entries.

However, this sorter relies on an `index` key existing for the entry, which is only populated by the static finder currently. We should set it from the other finders, too.

This will allow us to use said sorter everywhere. It will also let us replicate the behaviour of `fzf --tiebreak=index`:

```
  return pickers.new(opts, {
    finder = finders.new_oneshot_job(...)
    sorter = telescope.extensions.fzf.native_fzf_sorter(),
    tiebreak = function(current_entry, existing_entry, _)
      return current_entry.index < existing_entry.index
    end
  }):find()
```

This gives me better results for my "recently opened files" picker. Other builtin pickers might benefit from this, too.

## Type of change

# How Has This Been Tested?

Tested against my own picker using `tiebreak`, similarly to the snippet above.

**Configuration**:
* Neovim version (nvim --version): 0.8.0
* Operating system and version: N/A

# Checklist:

### DIFF
--- a/lua/telescope/finders.lua
+++ b/lua/telescope/finders.lua
@@ -65,16 +65,22 @@ function JobFinder:_find(prompt, process_result, process_complete)
     self.job:shutdown()
   end
 
+  local line_num = 0
   local on_output = function(_, line, _)
+    line_num = line_num + 1
     if not line or line == "" then
       return
     end
 
+    local entry
     if self.entry_maker then
-      line = self.entry_maker(line)
+      entry = self.entry_maker(line)
+      entry.index = line_num
+    else
+      entry = line
     end
 
-    process_result(line)
+    process_result(entry)
   end
 
   local opts = self:fn_command(prompt)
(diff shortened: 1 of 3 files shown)


Conni2461:
sorry for the really late response. tiebreak index sounds awesome, i'm not up to date with a lot so i am not sure if there is an issue already for something like this. i am gonna check later and if not i probably gonna open one, maybe we can provide some helpers for tiebreaker.

Thanks for the addition :)

Conni2461:
i think this just broke something, i am getting weird errors in `Telescope man_pages` ... i am going to look into it and hotfix (prefer not to revert)

PR #2451, opened 2023-04-03
fix: `vim.treesitter.get_query` deprecated

TheSafdarAwan:
# Description

vim.treesitter.get_query is deprecated use vim.treesitter.query.get

# Checklist:

- [x ] My code follows the style guidelines of this project (stylua)
- [ x] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [ x] I have made corresponding changes to the documentation (lua annotations)


jamestrew:
I think this will break thing for people not using nightly neovim.
Can instead do something like `local get_query = vim.treesitter.query.get or vim.treesitter.get_query` and then `pcall` using the `get_query` function.

Also that checklist formatting is hilarious :rofl:

TheSafdarAwan:
> 

thanks for the suggestion. will do this.

Conni2461:
thanks :)

PR #2452, opened 2023-04-06
add 'full' for line_width in diagnostics options

### LINKED ISSUE #2020: Full messages visible in diagnostics picker
I am always frustrated when I cannot see the full diagnostics message in the diagnostics picker results, even with wrap_results enabled

A nicely looking full diagnostics messages and their file of origin. I guess this may be possible with enabled wrap_results and by having the line_width parameter set to the message length, different for each message

The only relevant option seems to be line_width, but it is a fixed number. Setting it to a large number produces quite horribly-looking results

blue-pitaya:
# Description

Added new value "full" for line_width option in builtin.diagnostics. When "full" is set there are no restrictions on diagnostics message line width and so whole message is shown.

image

Until this change, line_width could only be number. Maximum line width was "one width" of picker window and longer messages were cut.

image

This was annoying since you generally want to read whole error message.

Fixes #2020

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list 
relevant details about your configuration

- [x] Verify change
- in config defaults set `wrap_line = true`
- open project where there is error with long message
- open diagnostics with `require('telescope.builtin').diagnostics({line_width='full'})`

**Configuration**:
* Neovim version (nvim --version): `NVIM v0.8.3`
* Operating system and version: `Linux 6.2.6-artix1-1`

# Checklist:

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -483,7 +483,7 @@ builtin.lsp_dynamic_workspace_symbols = require_on_exported_call("telescope.buil
 ---@field root_dir string|boolean: if set to string, get diagnostics only for buffers under this dir otherwise cwd
 ---@field no_unlisted boolean: if true, get diagnostics only for listed buffers
 ---@field no_sign boolean: hide DiagnosticSigns from Results (default: false)
----@field line_width string|number: Set length of diagnostic entry text in Results. Use 'full' for full untruncated text
+---@field line_width string|number: set length of diagnostic entry text in Results. Use 'full' for full untruncated text
 ---@field namespace number: limit your diagnostics to a specific namespace
 builtin.diagnostics = require_on_exported_call("telescope.builtin.__diagnostics").get
 
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -483,7 +483,7 @@ builtin.lsp_dynamic_workspace_symbols = require_on_exported_call("telescope.buil
 ---@field root_dir string|boolean: if set to string, get diagnostics only for buffers under this dir otherwise cwd
 ---@field no_unlisted boolean: if true, get diagnostics only for listed buffers
 ---@field no_sign boolean: hide DiagnosticSigns from Results (default: false)
----@field line_width number: set length of diagnostic entry text in Results
+---@field line_width string|number: Set length of diagnostic entry text in Results. Use 'full' for full untruncated text
 ---@field namespace number: limit your diagnostics to a specific namespace
 builtin.diagnostics = require_on_exported_call("telescope.builtin.__diagnostics").get
 
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1165,15 +1165,8 @@ function make_entry.gen_from_diagnostics(opts)
   }
   local line_width = vim.F.if_nil(opts.line_width, 0.5)
   local line_width_opts = { width = line_width }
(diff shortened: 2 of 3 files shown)


blue-pitaya:
Is this PR good to merge, or should I make some adjustments?

lua/telescope/builtin/init.lua
```diff
@@ -483,7 +483,7 @@ builtin.lsp_dynamic_workspace_symbols = require_on_exported_call("telescope.buil
 ---@field root_dir string|boolean: if set to string, get diagnostics only for buffers under this dir otherwise cwd
 ---@field no_unlisted boolean: if true, get diagnostics only for listed buffers
 ---@field no_sign boolean: hide DiagnosticSigns from Results (default: false)
----@field line_width number: set length of diagnostic entry text in Results
+---@field line_width string|number: Set length of diagnostic entry text in Results. Use 'full' for full untruncated text
```
jamestrew, line 486: super nitpicking but for the sake on consistency, can we lowercase the "Set length..."?

lua/telescope/make_entry.lua
```diff
@@ -1157,9 +1157,17 @@ function make_entry.gen_from_diagnostics(opts)
     { remaining = true },
   }
   local line_width = vim.F.if_nil(opts.line_width, 0.5)
+  local line_width_opts = { width = line_width }
+  if type(line_width) == "string" then
+    if line_width == "full" then
+      line_width_opts = {}
+    else
+      error(string.format("'%s' is not a valid value for line_width", line_width))
```
jamestrew, line 1165: I think we should move the validation of `line_width` to declaration of the picker and use `utils.notify` instead of `error`.

jamestrew (changes requested):
Hey sorry for the late response. Just a couple of suggested changes and I think we're good.

jamestrew:
Also just as a side note, noticed you're on `NVIM v0.8.3` but telescope is only supporting v0.9.0+. You might want to update your neovim if you haven't already lest you might run into various issues.

lua/telescope/make_entry.lua
blue-pitaya: Hi, thanks for checking my PR :)

I changed error to utils.notify, but left it in the same place. I couldn't find "good" place for validation in diagnostics picker, there is no direct call to line_width opt in here. Also, function 'make_entry.gen_from_diagnostics' is used specifically for diagnostics, so I guess it's not the worst place for this validation.

jamestrew: The problem is that unlike with `error`, `utils.notify` will let the picker continue to run and it'll eventually throw an actual error that isn't as helpful. 

I think we can simplify this bit of code to
```lua
  if type(line_width) == "string" and line_width == "full" then
    line_width_opts = {}
  end
```

then in `lua/telescope/builtin/__diagnostics.lua:124`, do like
```lua
  if type(opts.line_width) == "string" and opts.line_width ~= "full" then
    utils.notify("builtin.diagnostics", {
      msg = string.format("'%s' is not a valid value for line_width", opts.line_width),
      level = "ERROR",
    })
    return
  end
```

blue-pitaya:
Hi, sorry for stalling this PR so much! I applied your suggestions.

jamestrew (approved):

jamestrew:
thanks!

rlpowell:
See https://github.com/nvim-telescope/telescope.nvim/issues/2020 for example use.  :)

PR #2455, opened 2023-04-08
feat(lsp): support custom action when there is no result

AckslD:
# Description

Following the thread [here](https://www.reddit.com/r/neovim/comments/12fburw/comment/jfh9aoe/?utm_source=share&utm_medium=web2x&context=3) I thought it would be nice to support a custom action in the case of no results from the LSP when eg finding definitions. For example a user can configure this to notify them which is helpful for slow LSPs since you might not know if it is still working or didn't find anything.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

I called:

```lua
require('telescope.builtin').lsp_definitions({
  on_no_result = function()
    vim.notify("could not find a definition")
  end
})
```
on something which the LSP couldn't find a definition and saw the notification.

**Configuration**:
* Neovim version (nvim --version): NVIM v0.9.0-dev-1312+g908494889
* Operating system and version: arch linux

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -886,13 +886,16 @@ builtin.treesitter()                          *telescope.builtin.treesitter()*
 
 
     Options: ~
-        {show_line}         (boolean)  if true, shows the row:column that the
-                                       result is found at (default: true)
-        {bufnr}             (number)   specify the buffer number where
-                                       treesitter should run. (default:
-                                       current buffer)
-        {symbol_highlights} (table)    string -> string. Matches symbol with
-                                       hl_group
+        {show_line}         (boolean)       if true, shows the row:column that
+                                            the result is found at (default:
+                                            true)
+        {bufnr}             (number)        specify the buffer number where
+                                            treesitter should run. (default:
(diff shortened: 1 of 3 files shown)


Conni2461:
we could also just use `error` for this, similar to currently the git_* builtins. in combination with: https://github.com/nvim-telescope/telescope.nvim/pull/2225 then users can script using pcall.

The alternative would be to either use autocmds or build an actual event system, depending on what we wanna do

AckslD:
Thanks for your response @Conni2461! Do you think raising `error` on no result before #2225 would be an acceptable change? It might arguable be a breaking change and not sure how that's generally handled in `telescope`?

Conni2461:
sorry for the late response, i am currently still swamped (so i took a day off for open source), i'll figure this out today

frantisekstanko:
This is a very cool feature which I miss often. It would be awesome if this was merged. Thank you!

AckslD:
at least for my part I just wanted this for notifications so closing this for #3158

PR #2458, opened 2023-04-09
fix: expr for insert mode

### LINKED ISSUE #2431: Map with expr for insert mode does not work with attach_mappings
### Description

With below script to call out `find_files`,  then press `C-w`, you'll just see `v:null` to be inserted into prompt.

                  map("i", "<C-w>", function(_prompt_bufnr)
                    if vim.g.UseCW == 1 then
                        return "<C-w>"
                    else
                        return "<C-o>"
                    end
                  end, { expr = true })

                  return true
                end,
              }

It should trigger `C-w` or `C-o`.

### Neovim version

```markdown
NVIM v0.8.2
Build type: Release
LuaJIT 2.1.0-beta3
```

macOS 12.6.1

master

### checkhealth telescope

```markdown
telescope: 
  - ERROR: No healthcheck found for "telescope" plugin.
```

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. press `<space>j`
3. press `<C-w>` on the telescope ui

### Expected behavior

_No response_

### Actual behavior

you'll see `v:null` to be inserted into prompt.

### Minimal config

```Lua
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
    vim.fn.system({
        "git",
        "clone",
        "--filter=blob:none",
        "--branch=stable",
        lazypath,
    })
end
vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
    {
        dependencies = {
            { "nvim-lua/plenary.nvim" },
        },
        config = function()
            vim.keymap.set('n', '<space>j', function()
                  map("i", "<C-w>", function(_prompt_bufnr)
                    if vim.g.UseCW == 1 then
                        return "<C-w>"
                    else
                        return "<C-o>"
                    end
                  end, { expr = true })

                  return true
                end,
              }
          end)
        end,
    },
})
```

Conni2461:
fix #2431

it only contains the fix for 2431, it does not completely refactor mappings. plan to do that in a separate branch. I also cant recall when `vim.keymap` was introduced and if it would break 0.7 support (we still have on 0.1.x). need to figure this out before merging ^^

### DIFF
--- a/lua/telescope/mappings.lua
+++ b/lua/telescope/mappings.lua
@@ -240,9 +240,14 @@ local telescope_map = function(prompt_bufnr, mode, key_bind, key_func, opts)
     key_func = actions[key_func]
   elseif type(key_func) == "table" then
     if key_func.type == "command" then
-      a.nvim_buf_set_keymap(prompt_bufnr, mode, key_bind, key_func[1], opts or {
-        silent = true,
-      })
+      vim.keymap.set(
+        mode,
+        key_bind,
+        key_func[1],
+        vim.tbl_extend("force", opts or {
+          silent = true,
+        }, { buffer = prompt_bufnr })
+      )
       return
     elseif key_func.type == "action_key" then
       key_func = actions[key_func[1]]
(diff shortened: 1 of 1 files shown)


max397574:
quite sure vim.keymap was added in 0.7

akinsho:
Random drive by comment (as I was just reading through this issue) but you might benefit from
https://github.com/tweekmonster/helpful.vim plugin, it allows you to check when things were added to nvim e.g. you could do `:HelpfulVersion vim.keymap`

PR #2459, opened 2023-04-09
Change lsp_document_symbols display

datamadsen:
# Description

Before this patch, lsp document symbols would be displayed in two columns:

1. A column with the 25 first characters in the symbol name.
2. A column with the symbol kind that could take up remaining characters.

That seems a little wrong-way-around and unnecessary cutting of symbol names, which is why with this patch there are still two columns, but now they are:

1. A column with the symbol kind (column size fitted to the longest name)
2. A column with the symbol name that takes up the remaining space.

It could have just stopped there, but it sort of lacked some icons. So added is now configuration options for adding a symbol icon either in front of or after the symbol's kind. These options are documented in `config.lua`, but basically it is possible to:

* Have an icon or not (text will remain).
* Put the icon on either side of the text.
* Add space and or other characters around the icon.
* Choose which preset to use with `lspkind.nvim`
* Make substitutions on symbol kind, so if LSP reports kind `A` it can be translated to `B` in case `lspkind.nvim` has no icon for `A` but the one for `B` makes sense to use instead.
* Define what character(s) to fall back to in case an icon was not found.
Fixes # (issue)

## Type of change

# How Has This Been Tested?

Tested through different kinds of configuration and then invoking the lsp_document_symbols picker. For instance:

## Default
```
lsp_symbol_icon = {}
```
(or just leave it out completely)

## Text and icon
```
lsp_symbol_icon = {
  mode = "text_icon",
  lspkind_preset = "codicons",
  icon_prefix = " ",
  icon_postfix = ""
}
```

## Just the icon
```
lsp_symbol_icon = {
  mode = "icon",
  icon_prefix = "",
  icon_postfix = ""
}
```

**Configuration**:
NVIM v0.8.3
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by runner@Mac-1675344370646.local

Features: +acl +iconv +tui

* Operating system and version: Darwin voyager.local 22.3.0 Darwin Kernel Version 22.3.0: Mon Jan 30 20:39:35 PST 2023; root:xnu-8792.81.3~2/RELEASE_ARM64_T8103 x86_64

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -291,6 +291,37 @@ lsp.document_symbols = function(opts)
     end
 
     opts.path_display = { "hidden" }
+
+    -- Add icon and find the max length of symbol so we can use that to size the kind column accordingly.
+    local symbol_type_max_width = 0
+    local symbol_type_length = function(location)
+      local config = require("telescope.config").values.lsp_symbol_icon
+      local mode = "text"
+      if config ~= false then
+        mode = config.mode
+      end
+
+      local lengths_for_modes = {
+        ["icon"] = location.icon:len(),
+        ["text"] = location.kind:len(),
+        ["icon_text"] = location.icon:len() + location.kind:len(),
+        ["text_icon"] = location.kind:len() + location.icon:len(),
+      }
+
+      return lengths_for_modes[mode];
+    end
+
+    for _, location in pairs(locations) do
+      location.icon = utils.get_lsp_symbol_icon(location.kind)
+
+      local len = symbol_type_length(location)
+      if len > symbol_type_max_width then
+        symbol_type_max_width = len
+      end
+    end
+
+    opts.symbol_type_width = symbol_type_max_width + 1
+
     pickers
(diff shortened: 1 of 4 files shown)


datamadsen:
# Obligatory Before and After Shots

## Before
before

## After
after

datamadsen:
How it looks after the commit where the order of columns was reversed back to name, kind:
after name, kind

datamadsen:
Eh I broke some stuff - sorry :D Will close the PR for now and re-open once I fix it.

Wiebesiek:
This PR is a great idea. Any progress?

gzbd:
I'm looking for such eye-candy improvements in lsp pickers. Can you tell what's broken in this PR?

datamadsen:
sorry guys - I was/am new neovim and did a little too much `gd` so I ended up editing some of the neovim source in order to make this PR happen, without realizing it. It was in some lsp part of the neovim source that I found out adding the icon to the symbol type if I recall correctly. Not sure that I will have time to pick this up again in the forseeable future.

PR #2488, opened 2023-05-02
Revert "chore: simplify vimgrep_arguments args (#2440)"

### LINKED ISSUE #2482: live_grep makes linux computer run out of memory and freeze completely
### Description

I know there are a number of other performance related issues open related to live_grep using ripgrep, but i could not find instances of people reporting freezing of the entire OS making them lose all their work.

I know my use case is very crazy with say 10GB of code and i can see from journalctl that the ripgrep process is what causes the machine to run out of 16GB of RAM and 8GB of swap (fedora 38 with latest updates).

I can reproduce 100% if i go to a large directory and simply pless "k" to go through the live_grep results and i can see from my task manager that pressing "k" even once or twice makes the swap fill up and i have a frozen computer.

Running ripgrep on its own with the exact same scenario chugs along fine using all 8 cores and finishes without crashing. A single instance of ripgrep can handle itself no matter how heavy the load (100GB of code)

Would it be possible to maybe automatically slow things down a bit when a heavy load is detected to avoid freezing the operating system? 

It seems like the memory issue is specifically related to cycling through results in the UI. I've had freezes happen even when there were 5 results in total and i quickly went through them.

My concrete question is then: Can the logic that is triggered when pressing "j" and "k" be adapted to refresh less or something when the queries get crazy? Maybe it could ask the OS how its doing before performing all the updates on every press of "j" or "k"?

### Neovim version

```markdown
NVIM v0.9.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

Fedora 38

latest 0.1.x

### checkhealth telescope

```markdown

- OK plenary installed.

- OK rg: found ripgrep 13.0.0
- OK fd: found fd 8.7.0

Telescope Extension: `file_browser` ~
- No healthcheck provided

Telescope Extension: `fzf` ~
- OK lib working as expected
```

### Steps to reproduce

1. Download e.g. 50GB of code (duplicate a large project like linux kernel a few times)
2. open live_grep
3. press "k" a few times
4. If on a linux system you can easily run out of RAM and swap if you quickly cycle through results, freezing the whole OS

### Expected behavior

Either slow down or kill neovim instead of freezing the OS.

### Actual behavior

Frozen operating system (Fedora 38)

Also had the same issues on Fedora 36

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

jamestrew:
This reverts commit cfe6df6257aa7ee691fb56305923e9fec6c83c97.

# Description

It appears that the commit above is causing issues in certain instances when using grepping pickers.
The `--vimgrep` option seems to be the culprit.

[BurntSushi](https://github.com/BurntSushi/ripgrep/issues/2505#issuecomment-1536422272) doesn't really recommend using `--vimgrep`:
> --vimgrep is itself a strange thing, and is IMO legacy. It would probably be better if text editor plugins used ripgrep's --json flag to get structured matches. This is what VS Code does, for example.

Issue is reproducible live grepping in this repo/directory: https://github.com/swagger-api/swagger-ui/tree/master/dist

Fixes #2482

## Type of change

# How Has This Been Tested?

- [x] Tested live grepping in the [swagger repo](https://github.com/swagger-api/swagger-ui) with/without `--vimgrep` option

**Configuration**:
* Neovim version (nvim --version): NVIM v0.10.0-dev-19+g339011f59
* Operating system and version: Linux archlinux 6.2.12-arch1-1

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -532,13 +532,16 @@ telescope.setup({opts})                                    *telescope.setup()*
         Hint: Make sure that color is currently set to `never` because we do
         not yet interpret color codes
         Hint 2: Make sure that these options are in your changes arguments:
-          ("--no-heading", "--with-filename", "--line-number", "--column") or
-          "--vimgrep"
+          "--no-heading", "--with-filename", "--line-number", "--column"
         because we need them so the ripgrep output is in the correct format.
 
         Default: {
           "rg",
-          "--vimgrep",
+          "--color=never",
+          "--no-heading",
+          "--with-filename",
+          "--line-number",
+          "--column",
           "--smart-case"
         }
 
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -626,20 +626,23 @@ append(
 
 append(
   "vimgrep_arguments",
-  { "rg", "--vimgrep", "--smart-case" },
+  { "rg", "--color=never", "--no-heading", "--with-filename", "--line-number", "--column", "--smart-case" },
   [[
     Defines the command that will be used for `live_grep` and `grep_string`
     pickers.
     Hint: Make sure that color is currently set to `never` because we do
     not yet interpret color codes
     Hint 2: Make sure that these options are in your changes arguments:
-      ("--no-heading", "--with-filename", "--line-number", "--column") or
-      "--vimgrep"
+      "--no-heading", "--with-filename", "--line-number", "--column"
     because we need them so the ripgrep output is in the correct format.
 
     Default: {
       "rg",
-      "--vimgrep",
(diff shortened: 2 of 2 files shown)


tjdevries:
Thanks! I haven't had a chance to reproduce everything, but I do remember something like this awhile ago. Thanks for the PR

PR #2492, opened 2023-05-04
Prefix color highlighting reset only on nvim 9.0

xim:
# Description

Just adds an if around what I believe to be the appropriate version to start doing that refresh.

Feel free to rewrite / ignore.

Fixes #2461

## Type of change

- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Just manually tested on my neovim version

**Configuration**:
* Neovim version (nvim --version): 7.2
* Operating system and version: Debian bookworm

# Checklist:

- [ ] My code follows the style guidelines of this project (stylua)
- [X] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (lua annotations)

PR #2510, opened 2023-05-14
Revert "Prefix color highlighting reset only on nvim 9.0"

Conni2461:
Reverts nvim-telescope/telescope.nvim#2492


Conni2461:
#2492 did not fix the issue and actually broke prompt highlighting.

Reasoning see: https://github.com/nvim-telescope/telescope.nvim/pull/2492#issuecomment-1546992062

PR #2522, opened 2023-05-23
fix: highlights start at index 0 (was broken on nightly)

folke:
# Description

There were two locations where icons were highlighted at index `1`.
Highlight indices start at `0`, so that was wrong.

What happened instead is that the highlight started inside a multi-byte character.

That used to work accidentally, but with the virtual text PR, this now breaks.

https://github.com/neovim/neovim/pull/20130

# Checklist:


Conni2461:
Thanks :)

folke:
Thank you for the quick merge! :)

PR #2525, opened 2023-05-24
feat!: allow full height, width by resolving 1 as a percentage rather than absolute val

### LINKED ISSUE #2508: Full `width` / `height` picker using `layout_config`
Using `layout_config.width` (or `height`), it's not possible to set a value which resolves to 100%.

- `0.x` to `0.9` works for window percentages
- `1` returns an error about needing to be a positive integer
- > `1` is an absolute value

So there's not ability to set full width.

Allow setting a full-view width & height relative to 100%

My workaround is to set `width` & `height` to high numbers, but it seems that the preview contents tries to span the full width and generally this just feels hacky

**Additional context**
I need fullwidth as I'm using `telescope-tmux` picker in `tmux` to spawn an `nvim` process in an `tmux` popup to switch sessions. I want this to be a full width picker.

Conni2461:
fixes #2508

this is technically a breaking change but i doubt anyone uses height/width 1 (absolute value). It makes more sense to treat 1 as 100%

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -2160,14 +2160,14 @@ Provides "resolver functions" to allow more customisable inputs for options.
 resolver.resolve_height()                 *telescope.resolve.resolve_height()*
     Converts input to a function that returns the height. The input must take
     one of five forms:
-    1. 0 <= number < 1
+    1. 0 <= number <= 1
        This means total height as a percentage.
-    2. 1 <= number
+    2. 1 < number
        This means total height as a fixed number.
     3. function
        Must have signature: function(self, max_columns, max_lines): number
     4. table of the form: { val, max = ..., min = ... }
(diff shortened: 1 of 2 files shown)


kriansa:
> i doubt anyone uses height/width 1

[Never say never](https://github.com/kriansa/dotfiles/blob/6a2447eba25f8ba632de0211da82ea2c60bb8f67/modules/nvim/include/.config/nvim/lua/plugins/file-finder.lua#L136) 😂

Conni2461:
yeah f, i find another solution, currently thinking using math.ceil rather than floor, that would allow `0.999` basically be treated as 1 == 100%

Conni2461:
revert: https://github.com/nvim-telescope/telescope.nvim/pull/2537

PR #2551, opened 2023-06-07
refactor(mappings): use vim.keymap and remove `__TelescopeKeymapStore`

Conni2461:


Conni2461:
i ran it for a couple of hours with no regressions, i think this should be good to go! after that i would actually even backport it to 0.1.x if nothing breaks and then 0.1.2

clason:
found a regression: `['<c-h>'] = 'which_key',` no longer works

Conni2461:
oh yeah because it used the global to find mappings, forgot about that

clason:
Guess I'll just have to start learning the mappings ;)

Conni2461:
i think we should use `nvim_buf_get_keymap` and we just set the name as a mappings desc

PR #2556, opened 2023-06-08
fix: help

Conni2461:
CC @clason 

fixes regression in https://github.com/nvim-telescope/telescope.nvim/pull/2551

currently anon functions aren't figured out like before

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -17,7 +17,6 @@ globals = {
   "TelescopeGlobalState",
   "_TelescopeConfigurationValues",
   "_TelescopeConfigurationPickers",
-  "__TelescopeKeymapStore",
 }
 
 -- Global objects defined by the C code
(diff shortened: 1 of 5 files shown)


Conni2461:
problem is with functions that aren't telescope actions we need to open the source file and check the content.

But with the new implementation we need to figure out the desc on mapping register, not sure if its worth to do that

Conni2461:
but it might actually be just a handful of cases so it might not even have an impact

Conni2461:
got a better idea, we could just serialize the `debug.getinfo` into the desc so we can still "scrape" the information on action open

Conni2461:
okay, think i got it back to the previous functionality. @clason do you have a sec and could test it

Conni2461:
nah i think we should keep it, even though i dont use it, it helps new users to get going, so it has a purpose.

The new hack isn't that bad ^^ and its only used to store information about anonymous functions which isn't that common anyway

clason:
And it works nicely :)

Conni2461:
thx for testing :)

PR #2557, opened 2023-06-09
fix(previewer): ft detect for filetypes defined as functions

Conni2461:
still need to resolve binary warning

CC @clason

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -218,6 +218,7 @@ previewers.file_maker = function(filepath, bufnr, opts)
       else
         vim.schedule(function()
           opts.ft = opts.use_ft_detect and putils.filetype_detect(filepath)
+          local possible_binary = false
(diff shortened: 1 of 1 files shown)


clason:
> still need to resolve binary warning

probably due to the timeout?

otherwise it seems to work fine

Conni2461:
binary should also be fixed

clason:
yep, everything seems to work now, thanks for the quick fix!

Conni2461:
thanks for testing and reporting. i gonna open an issue on neovim, for that problem i found later today :)

clason:
Yes, the combination `filename` + `contents` is not supported (yet). It's a reasonable feature request, though. (Not a bug, since that combination is not documented to work.)

Conni2461:
> Not a bug, since that combination is not documented to work

ah good point, yeah then there is no rush ^^

clason:
But the reason it's not implemented is that nobody has asked for it yet ;)

PR #2560, opened 2023-06-09
fix(previewer): call fthook after we determined the filetype

Conni2461:


Conni2461:
@Hippo0o could you retest

Hippo0o:
works :+1:

Issue #2548, opened 2023-06-03
Options to configure user-defined layout strategies


boltlessengineer:
Currently I can adjust only some values of few default strategies. I can't change width/height/position for each specific prompt/results/preview windows.  

For example, I can't adjust `bottom_pane`'s prompt width like `horizontal`. If someone wants to adjust the window rendering deeply like #2381, the only option is making a whole new default layout_strategy as a PR.

Just allowing users to pass custom layout_strategy function in config would be better. And I think telescope's API is already well documented for users to put their own `layout_strategy` funcion in their config.

Having this feature would also fix some other similar issues like #848

Issue #1054, opened 2021-07-27
Internal padding on preview window


smackesey:
The preview window shows text right up against the border. It would be an improvement to be able to add one or more spaces of internal padding. This could be done through the addition of another layout config option.

l-kershaw:
This is not currently possible in Telescope.

I also can't think of a good way of implementing it, but if you want to write a PR for it, then I'd be happy to give feedback 🙂

PR #2572, opened 2023-06-12
feat(pickers): fully customizable layout

MunifTanjim:
# Description

Introduces a `create_layout` config. It enables plugging in completely separate UI components.

This also enables features like this to stay outside Telescope core: https://github.com/nvim-telescope/telescope.nvim/pull/2381

Telescope core can essentially stop caring about layout customization. It'll only require `create_layout` to return a specific structure. And the rest can be managed externally.

Related Issues/PRs:
- https://github.com/nvim-telescope/telescope.nvim/issues/2548
- https://github.com/nvim-telescope/telescope.nvim/issues/2508
- https://github.com/nvim-telescope/telescope.nvim/issues/1054
- https://github.com/nvim-telescope/telescope.nvim/pull/2381

## Type of change

# How Has This Been Tested?

Kapture 2023-06-13 at 16 13 27

```lua
local Layout = require("nui.layout")
local Popup = require("nui.popup")

local telescope = require("telescope")
local TSLayout = require("telescope.pickers.layout")

telescope.setup({
  defaults = {
    create_layout = function(picker)
      local border = {
        results = {
          top_left = "┌",
          top = "─",
          top_right = "┬",
          right = "│",
          bottom_right = "",
          bottom = "",
          bottom_left = "",
          left = "│",
        },
        results_patch = {
          minimal = {
            top_left = "┌",
            top_right = "┐",
          },
          horizontal = {
            top_left = "┌",
            top_right = "┬",
          },
          vertical = {
            top_left = "├",
            top_right = "┤",
          },
        },
        prompt = {
          top_left = "├",
          top = "─",
          top_right = "┤",
          right = "│",
          bottom_right = "┘",
          bottom = "─",
          bottom_left = "└",
          left = "│",
        },
        prompt_patch = {
          minimal = {
            bottom_right = "┘",
          },
          horizontal = {
            bottom_right = "┴",
          },
          vertical = {
            bottom_right = "┘",
          },
        },
        preview = {
          top_left = "┌",
          top = "─",
          top_right = "┐",
          right = "│",
          bottom_right = "┘",
          bottom = "─",
          bottom_left = "└",
          left = "│",
        },
        preview_patch = {
          minimal = {},
          horizontal = {
            bottom = "─",
            bottom_left = "",
            bottom_right = "┘",
            left = "",
            top_left = "",
          },
          vertical = {
            bottom = "",
            bottom_left = "",
            bottom_right = "",
            left = "│",
            top_left = "┌",
          },
        },
      }

      local results = Popup({
        focusable = false,
        border = {
          style = border.results,
          text = {
            top = picker.results_title,
            top_align = "center",
          },
        },
        win_options = {
          winhighlight = "Normal:Normal",
        },
      })

      local prompt = Popup({
        enter = true,
        border = {
          style = border.prompt,
          text = {
            top = picker.prompt_title,
            top_align = "center",
          },
        },
        win_options = {
          winhighlight = "Normal:Normal",
        },
      })

      local preview = Popup({
        focusable = false,
        border = {
          style = border.preview,
          text = {
            top = picker.preview_title,
            top_align = "center",
          },
        },
      })

      local box_by_kind = {
        vertical = Layout.Box({
          Layout.Box(preview, { grow = 1 }),
          Layout.Box(results, { grow = 1 }),
          Layout.Box(prompt, { size = 3 }),
        }, { dir = "col" }),
        horizontal = Layout.Box({
          Layout.Box({
            Layout.Box(results, { grow = 1 }),
            Layout.Box(prompt, { size = 3 }),
          }, { dir = "col", size = "50%" }),
          Layout.Box(preview, { size = "50%" }),
        }, { dir = "row" }),
        minimal = Layout.Box({
          Layout.Box(results, { grow = 1 }),
          Layout.Box(prompt, { size = 3 }),
        }, { dir = "col" }),
      }

      local function get_box()
        local height, width = vim.o.lines, vim.o.columns
        local box_kind = "horizontal"
        if width < 100 then
          box_kind = "vertical"
          if height < 40 then
            box_kind = "minimal"
          end
        elseif width < 120 then
          box_kind = "minimal"
        end
        return box_by_kind[box_kind], box_kind
      end

      local function prepare_layout_parts(layout, box_type)
        layout.results = TSLayout.Window(results)
        results.border:set_style(border.results_patch[box_type])

        layout.prompt = TSLayout.Window(prompt)
        prompt.border:set_style(border.prompt_patch[box_type])

        if box_type == "minimal" then
          layout.preview = nil
        else
          layout.preview = TSLayout.Window(preview)
          preview.border:set_style(border.preview_patch[box_type])
        end
      end

      local box, box_kind = get_box()
      local layout = Layout({
        relative = "editor",
        position = "50%",
        size = {
          height = "60%",
          width = "90%",
        },
      }, box)

      layout.picker = picker
      prepare_layout_parts(layout, box_kind)

      local layout_update = layout.update
      function layout:update()
        local box, box_kind = get_box()
        prepare_layout_parts(layout, box_kind)
        layout_update(self, box)
      end

      return TSLayout(layout)
    end,
  }
})
```

**Configuration**:

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -146,6 +146,13 @@ telescope.setup({opts})                                    *telescope.setup()*
 
         Default: 'horizontal'
 
+                                         *telescope.defaults.create_layout*
+    create_layout: ~
+        Configure the layout of Telescope pickers.
+        See |telescope.pickers.layout| for details.
+
+        Default: 'nil'
+
                                          *telescope.defaults.layout_config*
     layout_config: ~
         Determines the default configuration values for layout strategies.
@@ -1945,6 +1952,155 @@ ordered from the lowest priority to the highest priority.
     end,
   }
 <
+
+
+================================================================================
+LAYOUT                                                *telescope.pickers.layout*
+
+The telescope pickers layout can be configured using the
+|telescope.defaults.create_layout| option.
+
+Parameters: ~
+  - picker : A Picker object.
+
+Return: ~
+  - layout : instance of `TelescopeLayout` class.
+
+Example: ~
+>
+local Layout = require "telescope.pickers.layout"
+
+require("telescope").setup {
+  create_layout = function(picker)
+    local function create_window(enter, width, height, row, col, title)
+      local bufnr = vim.api.nvim_create_buf(false, true)
+      local winid = vim.api.nvim_open_win(bufnr, enter, {
+        style = "minimal",
+        relative = "editor",
+        width = width,
+        height = height,
+        row = row,
+        col = col,
+        border = "single",
+        title = title,
+      })
+
+      vim.wo[winid].winhighlight = "Normal:Normal"
+
+      return Layout.Window {
+        bufnr = bufnr,
+        winid = winid,
+      }
+    end
+
+    local function destory_window(window)
+      if window then
+        if vim.api.nvim_win_is_valid(window.winid) then
+          vim.api.nvim_win_close(window.winid, true)
+        end
+        if vim.api.nvim_buf_is_valid(window.bufnr) then
+          vim.api.nvim_buf_delete(window.bufnr, { force = true })
+        end
+      end
+    end
+
+    local layout = Layout {
+      picker = picker,
+      mount = function(self)
+        self.results = create_window(false, 40, 20, 0, 0, "Results")
+        self.preview = create_window(false, 40, 23, 0, 42, "Preview")
+        self.prompt = create_window(true, 40, 1, 22, 0, "Prompt")
+      end,
+      unmount = function(self)
+        destory_window(self.results)
+        destory_window(self.preview)
+        destory_window(self.prompt)
+      end,
+      update = function(self) end,
+    }
+
+    return layout
+  end,
+}
+<
+
+TelescopeWindowBorder.config                    *TelescopeWindowBorder.config*
+
+
+    Fields: ~
+        {bufnr}        (integer)
+        {winid}        (integer|nil)
+        {change_title} (nil|function)  (self: TelescopeWindowBorder, title:
+                                       string, pos?:
+                                       "NW"|"N"|"NE"|"SW"|"S"|"SE"):nil
+
+
+TelescopeWindowBorder                                  *TelescopeWindowBorder*
+
+
+    Fields: ~
+        {bufnr} (integer|nil)
+        {winid} (integer|nil)
+
+
+TelescopeWindow.config                                *TelescopeWindow.config*
+
+
+    Fields: ~
+        {bufnr}  (integer)
+        {winid}  (integer|nil)
+        {border} (TelescopeWindowBorder.config|nil)
+
+
+TelescopeWindow                                              *TelescopeWindow*
+
+
+    Fields: ~
+        {border} (TelescopeWindowBorder)
+        {bufnr}  (integer)
+        {winid}  (integer)
+
+
+TelescopeLayout.config                                *TelescopeLayout.config*
+
+
+    Fields: ~
+        {mount}   (function)             (self: TelescopeLayout):nil
+        {unmount} (function)             (self: TelescopeLayout):nil
+        {update}  (function)             (self: TelescopeLayout):nil
+        {prompt}  (TelescopeWindow|nil)
+        {results} (TelescopeWindow|nil)
+        {preview} (TelescopeWindow|nil)
+
+
+TelescopeLayout                                              *TelescopeLayout*
+
+
+    Fields: ~
+        {prompt}  (TelescopeWindow)
+        {results} (TelescopeWindow)
+        {preview} (TelescopeWindow|nil)
+
+
+Layout:mount()                              *telescope.pickers.layout:mount()*
+    Create the layout. This needs to ensure the required properties are
+    populated.
+
+
+
+Layout:unmount()                          *telescope.pickers.layout:unmount()*
+    Destroy the layout. This is responsible for performing clean-up, for
+    example:
+     - deleting buffers
+     - closing windows
+     - clearing autocmds
+
+
+
+Layout:update()                            *telescope.pickers.layout:update()*
+    Refresh the layout. This is called when, for example, vim is resized.
+
+
 
 
 ================================================================================
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -230,12 +230,13 @@ action_set.scroll_previewer = function(prompt_bufnr, direction)
   local previewer = action_state.get_current_picker(prompt_bufnr).previewer
   local status = state.get_status(prompt_bufnr)
 
+  local preview_winid = status.layout.preview and status.layout.preview.winid
   -- Check if we actually have a previewer and a preview window
-  if type(previewer) ~= "table" or previewer.scroll_fn == nil or status.preview_win == nil then
+  if type(previewer) ~= "table" or previewer.scroll_fn == nil or preview_winid == nil then
     return
   end
 
-  local default_speed = vim.api.nvim_win_get_height(status.preview_win) / 2
+  local default_speed = vim.api.nvim_win_get_height(status.layout.preview.winid) / 2
   local speed = status.picker.layout_config.scroll_speed or default_speed
 
   previewer:scroll_fn(math.floor(speed * direction))
@@ -270,12 +271,12 @@ end
 --      Valid directions include: "1", "-1"
 action_set.scroll_results = function(prompt_bufnr, direction)
   local status = state.get_status(prompt_bufnr)
-  local default_speed = vim.api.nvim_win_get_height(status.results_win) / 2
+  local default_speed = vim.api.nvim_win_get_height(status.layout.results.winid) / 2
   local speed = status.picker.layout_config.scroll_speed or default_speed
 
   local input = direction > 0 and [[]] or [[]]
 
-  vim.api.nvim_win_call(status.results_win, function()
+  vim.api.nvim_win_call(status.layout.results.winid, function()
     vim.cmd([[normal! ]] .. math.floor(speed) .. input)
   end)
 
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -20,6 +20,7 @@ local entry_display = require "telescope.pickers.entry_display"
 local p_highlighter = require "telescope.pickers.highlights"
 local p_scroller = require "telescope.pickers.scroller"
 local p_window = require "telescope.pickers.window"
+local Layout = require "telescope.pickers.layout"
 
 local EntryManager = require "telescope.entry_manager"
 local MultiSelect = require "telescope.pickers.multi"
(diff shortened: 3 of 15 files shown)


MunifTanjim:
@tjdevries @Conni2461 Is this something that has the potential to get merged? Wanted to confirm before I spend more time polishing it up.

mklishevych:
Would be great if this gets merged.

MunifTanjim:
This is a layer above `telescope.pickers.layout_strategies`.

With `telescope.pickers.layout_strategies`, you can control height/width/border/padding for the layout. But you need to do that in a very imperative way. You can't just plug in a UI library that can manage the popups. And you also can't go crazy with the layout (i.e. you only get to have 3 popups, you can't just add a 4th to display something if you want to).

With `create_layout`, you get complete control over the layout for telescope. You get to create and arrange popups however you want, with whatever you want. And you give telescope the things that it needs to do its thing:

- `:mount()` method to initially create the layout
- `:update()` method to refresh the layout on resize
- `:unmount()` to destroy the layout
- `.prompt`, `.results`, `.preview` each with:
  - `.winid` - window id
  - `.bufnr` - buffer number
  - `.border`
    - `.winid` - border window id
    - `.bufnr` - border buffer number

Then the current system of telescope layout becomes the default implementation of `create_layout` and `telescope.pickers.layout_strategies` is used by that default implementation.

Benefits:
- Fully plugable with external ui libraries. e.g. `nui.nvim` or `vim.ui.popup` (when Neovim introduces it).
- Telescope core can be fully decoupled with layout management and theme-ing. Telescope will expect an interface, the return value of `create_layout` will be the implementation of the expected interface.

Conni2461:
I personally get what you are saying and i already looked at the PR some time ago but i struggled to formulate my worries. 
We already have two ways to configure telescope layouts, layout_configuration and themes on top of that and its already confusing and a pain to configure both for end users. Layout configs can only be defined inside telescope right now of you do the hack that @l-kershaw described above which i wanted to change some time ago. So i am worried about introducing another layer below that allows to do even more crazy things like swap out the complete windowing library.

This makes future maintenance definitely harder because now we have to take into account the wildest of configuration which could completely break other peoples configs.
Then i am worried that this generates tickets of ppl that miss configured it an now border:change_title throws an error or toggle previewer no longer works or idk some other thing i can't predict right now.

The alternative would be to say, hey you at least have to implement this interface, so we actually know that these functions are available and if not they are just stubs, do nothing. That way we might be able to easily extend telescope with future features. Like we still wanna provide a way to have telescope open as a split and i dont know if we make the task, a lot more difficult by doing the proposed solution.
We already do a interface like solution for all previewers that can be created: https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/previewers/previewer.lua

Do you have any thoughts? I dont wanna kill the attempt of giving more flexibility to user, especially because i see telescope more as a "highly extendable fuzzy finder over lists." than the 50+ builtins that most endusers are interested in ^^, but i am just worried about future maintenance because my time is basically already 0 :(

max397574:
I think that this feature would just bring the customizability to a next level by allowing users to create completely new layouts.
This can make maintainance harder because of issues people will open (which imo can just be closed with "there is a problem in your custom layout")
But this is also a chance to make telescope better and perhaps even more future proof. Because e.g. if there would be a new popup api it could be switched out in the default layout conditionally.

MunifTanjim:
> layout_configuration and themes on top of that and its already confusing and a pain to configure both for end users

Yep, both of those were very hard for me to understand. I'm not sure I yet understand them fully. That was the reason I created this draft.

I said in my previous comment that "This is a layer above `telescope.pickers.layout_strategies`." But that's not entirely correct. This is would be a complete replacement of that. But the default built-in `create_layout` implementation would still use `telescope.pickers.layout_strategies` internally, so it'll still be supported.

If you look at the example in the PR description, it doesn't use layout_configuration or themes at all.

> The alternative would be to say, hey you at least have to implement this interface, so we actually know that these functions are available and if not they are just stubs, do nothing.

That's the idea. `create_layout` will return the expected interface (that contains winid, bufnr and methods/functions that telescope needs for interacting with the layout).

> Like we still wanna provide a way to have telescope open as a split and i dont know if we make the task, a lot more difficult by doing the proposed solution.

With the proposed solution it'll be completely possible. I can add another example for `create_layout` that uses the `Split` from `nui.nvim` to open telescope in a split layout.

If sombody wants they can hand-craft the layout using `vim.api.nvim_open_win` and return the expected interface from `create_layout` and it would just work.

---

In terms of maintainability... I think this would require less work to maintain. The current system for customizing layout is too complex and hard to understand (at least to me). Because there are too many moving parts. It took me a long time to figure out how the UI is being created for telescope, which value affects which thing.

If we take the well defined interface route, Telescope can stop maintaining layout_configuration and themes. If somebody wants a new layout or different style, they can just write the `create_layout` themselves by implementing the interface (either manually using `vim.api.nvim_open_win` or using ui libraries like `nui.nvim` or the hypothetical `vim.ui.popup` whenever that arrives).

---

I hope that makes the intention behind the PR more clear. I'm not attached to the interface I did in this PR... I'm absolutely open to suggestions and change things as needed.

The only thing I wanted was to have an easy way to create the layout for telescope.

max397574:
I also think that this would perhaps make maintainance easier in the regards of custom layouts
because to be honest the current implementation is really complicated imo because it's so "high level"
I think the users who'd like to create custom layouts do also have the knowledge to do it with this new way. It's imo much easier like this, I always struggled with the old way.

max397574:
@MunifTanjim I can 100% agree with the things you're saying in your comment about the current approach and would love the idea to have this new feature.
One thing though. Could you perhaps give a (minimal) example of a layout that uses the native apis (nvim_open_win and/or splits?). Would really help users (like me) that never used `nui.nvim` before to understand how they could actually use this (ig I could also do this myself but it would just take me much more time to figure out what the nui functions exactly do)

MunifTanjim:
> One thing though. Could you perhaps give a (minimal) example of a layout that uses the native apis (nvim_open_win and/or splits?). Would really help users (like me) that never used nui.nvim before to understand how they could actually use this (ig I could also do this myself but it would just take me much more time to figure out what the nui functions exactly do)

Yep, I can absolutely add that example in the PR description. It'll just be a bit verbose (more codes) than using `nui.nvim`. All `nui.nvim` does is to provide a nicer high-level api on top of native `vim.api.nvim_*` functions. For example:
- `Popup(options)` is just a lot of calculations to prepare params for `nvim_open_win` 
- `:mount` is basically `nvim_open_win` with the prepared params
- `:unmount` is basically `nvim_win_close` and cleanups
- `:update` is `nvim_win_set_config`
- etc. etc.

But before that I actually want to have some kinda confirmation if something like this would be accepted. I just don't want to spend more time polishing the implementation and preparing more examples if the core maintainers don't agree/think that `create_layout` is acceptable/good idea.

Conni2461:
> I said in my previous comment that "This is a layer above telescope.pickers.layout_strategies." But that's not entirely correct. This is would be a complete replacement of that. But the default built-in create_layout implementation would still use telescope.pickers.layout_strategies internally, so it'll still be supported.

Yeah makes sense that it completly replaces our window calculation thing ^^

> That's the idea. create_layout will return the expected interface (that contains winid, bufnr and methods/functions that telescope needs for interacting with the layout).

I have to go over the current implementation again because currently i dont see that really happening. Your example currently just returns a nui.Layout, which for me at least is a black box. We think we need a metaclass that is provided by telescope which is well documented and custom implementations implement it like that

```lua
return telescope.Layout:new({
 winid = ...
 fn1 = ...
 fn2 = ...
 -- change_title
})
```

Change title then e.g. is not implemented so the interface uses a stub. That makes it easier and more transparent for us to add new functionallity down the road

MunifTanjim:
> I have to go over the current implementation again because currently i dont see that really happening. Your example currently just returns a nui.Layout, which for me at least is a black box

Yeah, it's just a POC. The final version, if it happens, would surely include explicit and well-documented metaclass.

The interface is more-or-less like this (https://github.com/nvim-telescope/telescope.nvim/pull/2572#issuecomment-1650486064):

> * `:mount()` method to initially create the layout
> * `:update()` method to refresh the layout on resize
> * `:unmount()` to destroy the layout
> * `.prompt`, `.results`, `.preview` each with:
>   * `.winid` - window id
>   * `.bufnr` - buffer number
>   * `.border`
>     * `.winid` - border window id
>     * `.bufnr` - border buffer number

Conni2461:
ping me once you need a more in depth review :) thanks for the PR and your work on improving the whole layout situation

Conni2461:
Ah one more thing, you dont have to document these classes right now, we can do that at the end once we actually settled on working interfaces :)

max397574:
one more thing that annoyed me about the current implementation was actually that the window and the borders are just created with plenary.popup which since I think latest release actually alows less customization of the window title than the builtin windows do
With this pr it would be possible to just set the borders to `nil` and use native windows right?
(or just create completely custom borders)

Conni2461:
yeah i guess border object can be optional and i dont even think we need to expose bufnr and winid here (on top of my head) but we still want an object so we can have functions on border like update title. Then its up to you to just not implement it and go with default borders all the way.

The resulting code in telescope would then be like:

```lua
if self.layout.previewer.border then
  self.layout.previewer.border:change_title(...)
end
```

max397574:
always expose the numbers 😂 
I'm sure I'll find a usecase

Conni2461:
no more like, its not required by the interface to be present. Telescope doesnt really need it so, you, who implements the interfaces dont even have to provide them, and you actually cant if you use builtin borders

jamestrew:
I like the general direction of this PR but this would be a breaking change for extensions and more advanced custom user conflgs right?

MunifTanjim:
> I like the general direction of this PR but this would be a breaking change for extensions and more advanced custom user conflgs right?

Right now with this PR, if it's considered a public interface - the only breaking change is what's stored in `telescope.state`. And that too can be avoided. I was under the impression that `telescope.state` is only used internally. If it's used externally too, I can store the same stuffs that was stored before inside that: https://github.com/nvim-telescope/telescope.nvim/blob/2d92125620417fbea82ec30303823e3cd69e90e8/lua/telescope/pickers.lua#L548-L559

@jamestrew 

TLDR: this PR requires absolutely 0 breaking changes. And it'll be completely opt-in.

max397574:
I saw that in the last commits you were mostly working on the annotations. I don't know if you planned to do this anyway but I think it would be reall helpfull to add comment above the type annotations (just also with `---`) so you can see them as docs in completion from lua-ls. This would make it much easier to get used to the whole thing without always having go to read helpfiles.
I think just a short line describing what the function does (even if it already has a meaningful name) would make a huge difference here.

lua/telescope/pickers/layout.lua
```diff
@@ -0,0 +1,122 @@
...
+local function wrap_instance(class, instance)
+  local self = instance
+  if not getmetatable(instance) then
+    self = setmetatable(instance, { __index = class })
+  end
+  return self
+end
+
+--luacheck: push no max line length
+---@alias telescope_window_border_config { bufnr: integer, winid?: integer, change_title?: (fun(self: TelescopeWindowBorder, title: string, pos?: "NW"|"N"|"NE"|"SW"|"S"|"SE"): nil) }
```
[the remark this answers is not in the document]
MunifTanjim: I'm not very familiar with the documentation style in this repo. I'm absolutely open to suggestion.

I've also added a minimal example that only uses native `vim.api.nvim_` functions to implement `create_layout`.

MunifTanjim:
The implementation here is finished. @max397574 @Conni2461 @jamestrew 

I could use some help with how to organize the documentation for this. I've added some rough draft for documentation. Some review on that would be helpful.

lua/telescope/pickers/layout.lua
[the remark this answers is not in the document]
max397574: > I'm not very familiar with the documentation style in this repo. I'm absolutely open to suggestion.

I'm also not familiar with them.
But for things more complex than simply choices or function e.g. `alias x 1|2|3` or `alias y fun(self: TelescopeWindowBorder, title: string, pos?: "NW"|"N"|"NE"|"SW"|"S"|"SE"): nil` I'd use a separate class

max397574:
there should definitely be a full example like in the description of this pr (including the `require"telescope.setup"`
I'm not even sure if this is easily doable but would it be possible to add an example where you create a custom picker with a custom layout?

MunifTanjim:
> there should definitely be a full example like in the description of this pr (including the require"telescope.setup" I'm not even sure if this is easily doable

A lot goes into handling various corner-cases and conveniences. For example, supporing percentage string as size/position and implementing `layout:update` (what adjusts the size/position for all the windows) to work with that would need a looot of code. That was the reason why I wrote `nui.nvim` 😂

I don't think including a full-fleged example with all the features is feasible in the documentation. I think Wiki would be a better place to include those.

max397574:
yes imo there should definitely be such a full example in the wiki
but with "full" example I just meant something like in the docs rn but with the full setup
because if you just paste that somewhere your code won't work, will it?

MunifTanjim:
> because if you just paste that somewhere your code won't work, will it?

Oh, I see what you meant. So just like this?

```lua
require("telescope").setup {
  create_layout = function(picker)
    -- ...
  end,
}
```

instead of

```lua
local fucntion create_layout(picker)
  -- ...
end
```

right?

lua/telescope/pickers/layout.lua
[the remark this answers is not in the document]
Conni2461: we are doing documentation generation with this repo: https://github.com/tjdevries/tree-sitter-lua/blob/master/HOWTO.md

Conni2461:
> TLDR: this PR requires absolutely 0 breaking changes. And it'll be completely opt-in.

there are breaking changes regarding internals because status, ... changes, but thats not a big deal, we have releases for a reason.

I mostly reviewed the PR and really like it, i am going to run and play around with it for a couple of days and then do another quick review, see if i no longer like something, or whatever and then i am confident with merging that i think, unless you still have open tasks on your list for this PR

lua/telescope/pickers.lua
```diff
@@ -31,6 +32,188 @@ local ns_telescope_matching = a.nvim_create_namespace "telescope_matching"
...
+---@field titlehighlight integer
+---@field width integer
+
+-- Create three windows:
+-- 1. Prompt window
+-- 2. Options window
+-- 3. Preview window
+--
+---@param picker Picker
+local function default_create_layout(picker)
```
Conni2461, line 55: i am thinking that we should maybe now move this in a new separate file

`lua/telescope/pickers/window.lua`

MunifTanjim: > there are breaking changes regarding internals because status, ... changes,

Hmm 🤔 I tried to keep the status object same... only added an extra property `layout`.

image

If I broke compatibility, it wasn't intentional. Let me know if I missed something, If breaking changes can be avoided, I'd like to avoid it.

> i am going to run and play around with it for a couple of days and then do another quick review, see if i no longer like something, or whatever and then i am confident with merging that i think

Sounds good! 😃 👍🏼

> unless you still have open tasks on your list for this PR

I don't have anything else to implement. Just need to make the documentation better.

Conni2461:
Okay i already found two issues with the default config, and i just list them below, if you cant instantly reproduce those i can share my current config
- start picker with previewer enabled, error: `.../packer/start/plenary.nvim/lua/plenary/window/border.lua:177: attempt to index field '_border_win_options' (a nil value)`
- start picker with previewer disabled and then toggle it: `...er/start/telescope.nvim/lua/telescope/actions/layout.lua:32: attempt to index field 'preview' (a nil value)`

lua/telescope/pickers.lua
MunifTanjim: > start picker with previewer disabled and then toggle it: ...er/start/telescope.nvim/lua/telescope/actions/layout.lua:32: attempt to index field 'preview' (a nil value)

Fixed it.

> start picker with previewer enabled, error: .../packer/start/plenary.nvim/lua/plenary/window/border.lua:177: attempt to index field '_border_win_options' (a nil value)

I can't repro this one. 🤔

Conni2461: > If I broke compatibility, it wasn't intentional. Let me know if I missed something, If breaking changes can be avoided, I'd like to avoid it.

oh i missed that compat part. thats really good :+1:

doc/telescope.txt
```diff
@@ -1946,6 +1953,148 @@ ordered from the lowest priority to the highest priority.
...
+        {bufnr}  (integer)
+        {winid}  (integer)
+
+
+TelescopeLayout.config                                *TelescopeLayout.config*
+
+
+    Fields: ~
+        {mount} (fun)
+
```
MunifTanjim, line 2066: There should be more fields here. But they're not being generated.

image

Conni2461: Ill look into it

Conni2461:
```lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'MunifTanjim/telescope.nvim', branch = "feat-custom-layout",
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup({
    defaults = {
      mappings = {
        i = {
          ["<C-h>"] = require "telescope.actions.layout".toggle_preview,
        },
      },
      dynamic_preview_title = true,
    },
  })
end
end
load_plugins()
require('packer').sync()
```

minimal lua rc, to reproduce the bug. You can just run `:Telescope find_files` and the bug should show up, i think `dynamic_preview_title = true` is the option that will end up throwing the error

doc/telescope.txt
Conni2461: it seems like the parser doesnt understand `fun(self: ...` (its probably the `:`

this works:

```diff
diff --git a/lua/telescope/pickers/layout.lua b/lua/telescope/pickers/layout.lua
index 73a25e5..295bee8 100644
--- a/lua/telescope/pickers/layout.lua
+++ b/lua/telescope/pickers/layout.lua
@@ -138,9 +138,9 @@ local Window = setmetatable({}, {
 })
 
 ---@class TelescopeLayout.config
----@field mount fun(self: TelescopeLayout):nil
----@field unmount fun(self: TelescopeLayout):nil
----@field update fun(self: TelescopeLayout):nil
+---@field mount function: (self: TelescopeLayout):nil
+---@field unmount function: (self: TelescopeLayout):nil
+---@field update function: (self: TelescopeLayout):nil
 ---@field prompt TelescopeWindow|nil
 ---@field results TelescopeWindow|nil
 ---@field preview TelescopeWindow|nil
```

Conni2461:
found the fix

```diff
diff --git a/lua/telescope/pickers.lua b/lua/telescope/pickers.lua
index cd9d602..d73a994 100644
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -60,8 +60,8 @@ local function default_create_layout(picker)
     return {
       winid = border.win_id,
       bufnr = border.bufnr,
-      change_title = border.change_title and function(...)
-        border.change_title(...)
+      change_title = border.change_title and function(_, t)
+        border:change_title(t)
       end,
     }
   end
```

MunifTanjim:
> found the fix
> 
> ```diff
> diff --git a/lua/telescope/pickers.lua b/lua/telescope/pickers.lua
> index cd9d602..d73a994 100644
> --- a/lua/telescope/pickers.lua
> +++ b/lua/telescope/pickers.lua
> @@ -60,8 +60,8 @@ local function default_create_layout(picker)
>      border.winid = border.win_id
>      return border {
>        winid = border.win_id,
>        bufnr = border.bufnr,
> -      change_title = border.change_title and function(...)
> -        border.change_title(...)
> +      change_title = border.change_title and function(_, t)
> +        border:change_title(t)
>        end,
>      }
>    end
> ```

Did something similar:

image

Conni2461:
already tried it, works great. i gonna look for more regressions over the coming days and i'' try to configure my own layout

lua/telescope/actions/set.lua
```diff
@@ -230,12 +230,13 @@ action_set.scroll_previewer = function(prompt_bufnr, direction)
...
 
+  local preview_winid = status.layout.preview and status.layout.preview.winid
   -- Check if we actually have a previewer and a preview window
-  if type(previewer) ~= "table" or previewer.scroll_fn == nil or status.preview_win == nil then
+  if type(previewer) ~= "table" or previewer.scroll_fn == nil or preview_winid == nil then
     return
   end
 
-  local default_speed = vim.api.nvim_win_get_height(status.preview_win) / 2
+  local default_speed = vim.api.nvim_win_get_height(status.layaout.preview.winid) / 2
```
jamestrew, line 239: found a typo on `layout` here

MunifTanjim: fixed

MunifTanjim:
Quick experiment with `nui.split`. Doesn't look that good, because of the `statusline`s. But it mostly works 🤷‍♂️

<details>

<summary>Expand Code Snippet</summary>

```lua
create_layout = function(picker)
  local NuiSplit = require("nui.split")

  function Split(options)
    local split = NuiSplit(options)
    split.border = {
      split = split,
      change_title = function(self, title)
        vim.schedule(function()
          vim.wo[self.split.winid].winbar = title
        end)
      end,
    }
    return split
  end

  local results = Split({
    win_options = {
      number = false,
      relativenumber = false,
      winbar = picker.results_title,
    },
  })

  local prompt = Split({
    win_options = {
      cursorline = false,
      number = false,
      relativenumber = false,
      winbar = picker.prompt_title,
    },
  })

  local preview = Split({
    win_options = {
      number = true,
      relativenumber = false,
      winbar = picker.preview_title,
    },
  })

  local box_by_kind = {
    vertical = Layout.Box({
      Layout.Box(preview, { grow = 1 }),
      Layout.Box(results, { grow = 1 }),
      Layout.Box(prompt, { size = 2 }),
    }, { dir = "col" }),
    horizontal = Layout.Box({
      Layout.Box({
        Layout.Box(results, { grow = 1 }),
        Layout.Box(prompt, { size = 2 }),
      }, { dir = "col", size = "50%" }),
      Layout.Box(preview, { size = "50%" }),
    }, { dir = "row" }),
    minimal = Layout.Box({
      Layout.Box(results, { grow = 1 }),
      Layout.Box(prompt, { size = 2 }),
    }, { dir = "col" }),
  }

  local function get_box()
    local height, width = vim.o.lines, vim.o.columns
    local box_kind = "horizontal"
    if width < 100 then
      box_kind = "vertical"
      if height < 40 then
        box_kind = "minimal"
      end
    elseif width < 120 then
      box_kind = "minimal"
    end
    return box_by_kind[box_kind], box_kind
  end

  local function prepare_layout_parts(layout, box_type)
    layout.results = TSLayout.Window(results)

    layout.prompt = TSLayout.Window(prompt)

    if box_type == "minimal" then
      layout.preview = nil
    else
      layout.preview = TSLayout.Window(preview)
    end
  end

  local box, box_kind = get_box()
  local layout = Layout({
    relative = "editor",
    position = "bottom",
    size = "60%",
  }, box)

  layout.picker = picker
  prepare_layout_parts(layout, box_kind)

  return TSLayout(layout)
end,
```

</details>

image

max397574:
perhaps you could try laststatus=3?

max397574:
Anything left blocking this pr?

MunifTanjim:
Nope. This is the last update https://github.com/nvim-telescope/telescope.nvim/pull/2572#issuecomment-1712834508

Conni2461:
I'll merge it tomorrow morning (CET), i've been testing it over the past two weeks and havent found more regressions.

Its too late for me to merge it prior to going to bed, just in case something/some use case breaks that i've missed ^^

Conni2461:
Thanks for your work here :)

i3d:
It would be _very_ nice to see enabling a new dimension of "layout" extensions by this PR so that layout design complexities can reasonably be remained in the layout design experts or enthusiasms where most of users can simply just plug and enjoy...

unphased:
Can we put this beautiful fused layout implementation example into a "fused" layout name so we can opt into it without having to merge our existing telescope setup configuration structure with the 189 line-of-code example shown above (and which will probably break on us within a few months down the road). Thanks, and looks absolutely fantastic. 

I would maybe even prefer to save the extra 3 or so lines and columns consumed by borders, by fusing even further down, and using differing bgcolors on the panes for distinguising them visually, but i'm probably in the minority there and clearly the customizability exists for me to go work out how to do that on my own, so i'm cool with that.

MunifTanjim:
The reason for adding `create_layout` config was so that `telescope.nvim` core doesn't have to maintain the layouts down the line. Inlcuding the fused layout in the core would defeat that purpose.

The config for fused layout it added in the wiki: https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes#layouts

So it would be better to just copy paste that in your config.

unphased:
Yeah I have been playing with it and it's quite readable and the responsive layout changing behavior as well as layout dimensioning (how big to make the preview, how big to make the telescope panel) are really nice and easy to tweak. 

It's so pretty that it's actually renewed my attention on my nerd font icon rendering being broken for a few of the codepoints whenever inside tmux... following up on that now...

PR #2583, opened 2023-06-24
Fix tagrelative option not considered in builtin.tags

### LINKED ISSUE #1309: builtin tag picker does not respect tagrelative option
### Description

I keep my ctags inside the `.git` folder of my repos to not clutter up the source tree.

For that I generate my tags with `ctags --tag-relative -f .git/tags ...` and add `.git/tags` to the `tags` option.

By default `tagrelative` is set in neovim and indeed `nvim` correctly reads the tag file, as confirmed by running `:tselect` with the cursor on `main`.

However for `:Telescope tags`, the relative path (`../main.c`)  is taken relative to the current working directory and not relative to the directory the tags file is in, and so neither the preview nor opening the file works.

### Neovim version

```
NVIM v0.5.1
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/bin/cc -D_FORTIFY_SOURCE=2 -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -DNDEBUG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/build/neovim/src/neovim-0.5.1/build/config -I/build/neovim/src/neovim-0.5.1/src -I/usr/include -I/build/neovim/src/neovim-0.5.1/build/src/nvim/auto -I/build/neovim/src/neovim-0.5.1/build/include
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "/usr/share/nvim"

```

ArchLinux

### checkhealth telescope

```markdown
health#telescope#check
  - OK: plenary installed.

  - OK: fd: found fd 8.2.1

```

### Steps to reproduce

```bash
mkdir .git
echo "int main() { return 0; }" > main.c
ctags --tag-relative --fields=afiKlmnsSzt -f ".git/tags" main.c
nvim -u minimal.lua
```

In nvim, run `:Telescope tags`

### Expected behavior

The `main.c` file is previewed correctly and opened once the tag is selected

### Actual behavior

The preview is empty and the wrong path is opened (`../main.c` instead of `main.c`).
2021-10-06-022649_2060x1360_scrot

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
  vim.opt.tags:prepend('.git/tags')
end
end
load_plugins()
require('packer').sync()
```

OscarCreator:
# Description

This makes buitin.tags respect tagrelative option. Instead of using `cat` command to show content of tags file `rg` or `grep` is used as they are able to print which file the line is from.

Fixes #1309

## Type of change

# How Has This Been Tested?

Create a tag which is **not** located in the directory you open nvim but in a subfolder. The tags in the tag file are then relative to the subfolder and not nvim's cwd. Run `:Telescope tags` when you have a file open in the subfolder, this should result in showing the tagpaths relative to nvim's cwd.

**Configuration**:
* Neovim version (nvim --version): 0.9.0
* Operating system and version: Ubuntu 22.04.2 LTS

# Checklist:

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1088,7 +1088,7 @@ function make_entry.gen_from_ctags(opts)
   return function(line)
     local tag_file
     -- split line by ':'
-    tag_file, line = string.match(line, '([^:]+):(.+)')
+    tag_file, line = string.match(line, "([^:]+):(.+)")
 
     local tag, file, scode, lnum
     -- ctags gives us: 'tags\tfile\tsource'
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1086,7 +1086,14 @@ function make_entry.gen_from_ctags(opts)
 
   local current_file_cache = {}
   return function(line)
-    if line == "" or line:sub(1, 1) == "!" then
+    if line == "" then
+      return nil
+    end
+    local tag_file
+    tag_file, line = string.match(line, '([^:]+):(.+)')
+
+    -- do not include tag file format
+    if line:sub(1, 1) == "!" then
       return nil
     end
 
@@ -1096,6 +1103,12 @@ function make_entry.gen_from_ctags(opts)
     if not tag then
       -- hasktags gives us: 'tags\tfile\tlnum'
       tag, file, lnum = string.match(line, "([^\t]+)\t([^\t]+)\t(%d+).*")
+    end
+
+    -- append tag file path
+    if vim.opt.tagrelative:get() then
+      local tag_path = Path:new(tag_file):parent():make_relative(cwd)
+      file = tag_path .. "/" .. file
     end
 
     if Path.path.sep == "\\" then
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1086,16 +1086,9 @@ function make_entry.gen_from_ctags(opts)
 
   local current_file_cache = {}
   return function(line)
-    if line == "" then
-      return nil
-    end
     local tag_file
+    -- split line by ':'
     tag_file, line = string.match(line, '([^:]+):(.+)')
-
-    -- do not include tag file format
-    if line:sub(1, 1) == "!" then
-      return nil
-    end
 
     local tag, file, scode, lnum
     -- ctags gives us: 'tags\tfile\tsource'
@@ -1107,8 +1100,8 @@ function make_entry.gen_from_ctags(opts)
 
     -- append tag file path
     if vim.opt.tagrelative:get() then
-      local tag_path = Path:new(tag_file):parent():make_relative(cwd)
-      file = tag_path .. "/" .. file
+      local tag_path = Path:new(tag_file):parent()
+      file = Path:new(tag_path .. "/" .. file):normalize(cwd)
     end
 
     if Path.path.sep == "\\" then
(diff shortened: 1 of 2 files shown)


lua/telescope/make_entry.lua
```diff
@@ -1086,7 +1086,14 @@ function make_entry.gen_from_ctags(opts)
...
   return function(line)
-    if line == "" or line:sub(1, 1) == "!" then
+    if line == "" then
+      return nil
+    end
+    local tag_file
+    tag_file, line = string.match(line, '([^:]+):(.+)')
+
+    -- do not include tag file format
+    if line:sub(1, 1) == "!" then
```
OscarCreator, line 1096: Should we filter this out when searching instead?

lua/telescope/make_entry.lua
```diff
@@ -1098,6 +1105,12 @@ function make_entry.gen_from_ctags(opts)
       tag, file, lnum = string.match(line, "([^\t]+)\t([^\t]+)\t(%d+).*")
     end
 
+    -- append tag file path
+    if vim.opt.tagrelative:get() then
+      local tag_path = Path:new(tag_file):parent():make_relative(cwd)
+      file = tag_path .. "/" .. file
```
OscarCreator, line 1111: Is a better way to append paths?

jamestrew (changes requested):
Hey
I'm pretty unfamiliar with ctags so thanks for taking the initiative on this.

Things are looking pretty good, just a couple of thoughts.

lua/telescope/make_entry.lua
OscarCreator: Done

lua/telescope/make_entry.lua
OscarCreator: This should now be fixed.

jamestrew (approved):

jamestrew:
thanks!

jamestrew:
Sorry had to revert this due to some performance issues.

OscarCreator:
Could you elaborate so I can fix it?

Conni2461:
> Sorry had to revert this due to some performance issues.

All good, i'll try to be more active and warn you about code i wouldn't change without "more proper testing". This was my bad, i am sorry

> Could you elaborate so I can fix it?

take a look at https://github.com/nvim-telescope/telescope.nvim/issues/2627

The problem is, for tag files with 100k entries calculating the resulting path for every of these entries from scratch isn't performant. Additionally doing `rg -H -N --no-heading --color never -v '^!|^$'` and parsing it out adds more memory and costs additional cpu time.

Its possible to solve this, but its a none trivial problem to solve.
My idea always was to dynamically generate a function for each file that correctly calculates the path only based on string manipulation because that would be fast enough. But i never gotten around to doing that ...

PR #2594, opened 2023-07-05
feat(lsp): added support for dynamic capabilities

folke:
# Description

Added support for dynamic registration of capabilities to all the lsp pickers.
I also refactored the relevant code a bit to make it less verbose.

Any code that wants to check lsp capabilities should use `client.supports_method`.
That function includes checks for dynamically registered capabilities.
Just checking `server_capabilities` only includes caps that were advertised during init.

Fixes https://github.com/LazyVim/LazyVim/discussions/990

Fixes `goto definition` for a number of lsp servers like `jdtls` and `solargraph`

## Type of change

# How Has This Been Tested?

I've testes all the methods using several LSP servers including the one `solargraph` where go to definition was broken.

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -393,44 +393,38 @@ lsp.dynamic_workspace_symbols = function(opts)
     :find()
 end
 
-local function check_capabilities(feature, bufnr)
-  local clients = vim.lsp.buf_get_clients(bufnr)
+local function check_capabilities(method, bufnr)
+  local clients = vim.lsp.get_active_clients { bufnr = bufnr }
 
-  local supported_client = false
   for _, client in pairs(clients) do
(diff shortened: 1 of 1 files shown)


folke:
The failing tests are not because of this change, but because homebrew can not be updated

tjdevries:
Thanks, makes sense to me. Some of that LSP code is quite old (not all the methods existed at the time). Thanks :)

ben-krieger:
Is there a reason that this commit hasn't been cherry-picked into a release (of which there have been a few since the PR was merged)? The README discourages using `master`, but as a user of `jdtls`, I need this fix for jumping to definition to work.

jamestrew:
@ben-krieger we aren't cherry picking features into 0.1.x following semver. This will come in 0.2.0 but no timeline for this.
Best bet is using the master branch pinned to this commit.

PR #2599, opened 2023-07-13
expand paths more smartly

jamestrew:
getting tired of unusual paths being over expanded by `vim.fn.expand` causing an assortment of issues (commonly in telescope-file-browser).

this might be overkill and have some edge cases where things being under expanded. 
i'll ride this branch for a few days to see how it goes.

partially closes https://github.com/nvim-telescope/telescope-file-browser.nvim/issues/289

### DIFF
--- a/lua/telescope/_.lua
+++ b/lua/telescope/_.lua
@@ -5,6 +5,7 @@ local log = require "plenary.log"
 
 local async = require "plenary.async"
 local channel = require("plenary.async").control.channel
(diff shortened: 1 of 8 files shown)


merlinvn:
Hi I don't know if I set it right or wrong but my setting with cwd="~/dotfiles" is not working with this merge.

jamestrew:
@merlinvn thanks for pointing that out. I reverted this commit.

PR #2618, opened 2023-07-22
fix: preserve queued keys at picker launch (#2274)

### LINKED ISSUE #2274: Async Opening Misses Keypresses
### Description

When opening telescope keypresses can be missed.

This is most notable for slower pickers (such as LSP stuff, especially if you just opened a file/workspace) but can be easily reproduced for all pickers.

### Neovim version

```markdown
% nvim --version
NVIM v0.8.1
Build type: Release
LuaJIT 2.1.0-beta3
```

NixOS

cabf991b1d3996fa6f3232327fc649bbdf676496

### checkhealth telescope

```markdown
  - OK: plenary installed.

```

### Steps to reproduce

1. Run `:Telescope lsp_workspace_symbols`
2. Immediately press another letter such as `i`.

### Expected behavior

The search box has an `i` in it.

### Actual behavior

The search box has `A` in it? Or is empty. Or sometimes is in normal mode instead of insert mode.

### Minimal config

```Lua
-- No config required.
```

drmikehenry:
# Description

Ensure that any keystrokes that are queued at picker launch are processed only
after the picker's mode (`insert` or `normal`) has been chosen, preserving
their intended meaning.

Previously the picker's mode was set by simulating keystrokes via `feedkeys()`.
In the absence of queued keystrokes, this works fine; but if the user is able
to queue keystrokes before the call to `feedkeys()`, those queued keystrokes
are processed before the simulated keystrokes that change the picker's mode.
Because of this unexpected ordering, the user's queued keystrokes may appear to
be ignored or may cause the picker to start in the wrong mode.

For example, consider the below normal-mode mapping:
```vim
:nnoremap <space>ff :Telescope find_files<CR>
```

Upon launching the picker via `<space>ff`, Neovim is already in normal mode.
To switch to insert mode in the picker, Telescope previously used a call to
`feedkeys("A")`, simulating a keypress of `A` to enter insert mode at the end
of the current line.  This `A` will not be processed until all previously
queued user keystrokes have been processed, causing issues.

In real-world use, problems occur when the user types `<space>ff` followed
quickly by characters intended as fuzzy match text.  This can be demonstrated
using `feedkeys()` as shown below.

```vim
:call feedkeys("\<space>ff" . "apple")
```

The user intended to search for `apple`, but the `a` is mis-interpreted as a
request to enter insert mode at end of line, after which `pple` is inserted;
subsequently, Telescope's simulated `A` is then appended, resulting in a search
string of `ppleA`.

Using `:startinsert!` (to enter insert mode as if by `A`) or `:normal! $` (to
enter normal mode and move to end-of-line) avoids interfering with the user's
queued keys.

Fixes #2274.

## Type of change

# How Has This Been Tested?

- [x] Tested interactively and via `:call feedkeys()` as explained above

**Configuration**:

      NVIM v0.9.1
      Build type: Release
      LuaJIT 2.1.0-beta3

        fall-back for $VIM: "/__w/neovim/neovim/build/nvim.AppDir/usr/share/nvim"

      $ lsb_release -a
      No LSB modules are available.
      Distributor ID: Ubuntu
      Description:    Ubuntu 22.04.2 LTS
      Release:        22.04
      Codename:       jammy

# Checklist:

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -430,16 +430,16 @@ function Picker:find()
   end
 
   if vim.tbl_contains({ "insert", "normal" }, self.initial_mode) then
-    local mode = vim.fn.mode()
-    local keys
+    -- Note that `feedkeys()` should not be used to change modes, as
+    -- the user may have queued keystrokes that would be processed first,
+    -- interfering with desired operation.
     if self.initial_mode == "normal" then
-      -- n: A<ESC> makes sure cursor is at always at end of prompt w/o default_text
-      keys = mode ~= "n" and "<ESC>A<ESC>" or "A<ESC>"
+      -- Ensure cursor is at always at end of prompt w/o default_text
+      vim.cmd "normal! $"
     else
       -- always fully retrigger insert mode: required for going from one picker to next
-      keys = mode ~= "n" and "<ESC>A" or "A"
+      vim.cmd "startinsert!"
     end
-    a.nvim_feedkeys(a.nvim_replace_termcodes(keys, true, false, true), "n", true)
   else
     utils.notify(
       "pickers.find",


drmikehenry:
I've removed the local variable`mode` that was rendered unnecessary by this pull request, and have removed the parentheses around `vim.cmd "normal! $"` and `vim.cmd "startinsert!"`.

jamestrew:
Very nice, thanks!

fdschmidt93:
Pardon my interruption, just randomly driving by. Looking at the diff this likely breaks why we originally simulated `feedkeys`

See https://github.com/nvim-telescope/telescope.nvim/pull/1917 (and a long list of preceding discussion).

Looking at the diff and a quick test, what for instance now broke is:
1. Launch picker with `initial_mode = "insert"`
2. Have some chain that functionality that picker and opens another one from within the first picker with `initial_mode = "normal"`
3. Be in insert mode instead

I'm not a 100% sure whether the diff handles all cases appropriately, but I'd kindly ask you guys to check.

For feedkeys to cause issues on my machine I have to smash keys _very_ fast. IIRC it's a you win some, you lose some kind of scenario, but I'd be very happy if there's a one-fits-all solution.

jamestrew:
@fdschmidt93 thanks for chiming in. I'm right, this PR does cause certain issues like you mentioned. I'm reverting the commit for now.

drmikehenry:
@fdschmidt93 Thanks for driving by :-)  I wouldn't want to cause any regressions.

Issue #2274 was closed by this PR; I suggest it should be re-opened since this PR was reverted.

On my machine, timing-related problems with the `feedkeys()` solution occur frequently throughout the day when I am switching to a familiar buffer name (such that I know in advance the full sequence of keystrokes and I get a lot of practice typing them), so I would ideally like to find another solution for #2274.

I'd like to test with your scenario, but I don't know what these steps mean:

> 1. Launch picker with initial_mode = "insert"
> 2. Have some chain that functionality that closes the picker and opens another one from within the first picker with initial_mode = "normal"
> 3. Be in insert mode instead

Is this something for which you could offer some quick replication steps for testing?

I hadn't been using `<cmd>` mappings in the past; historically I've always created mappings that exit insert or visual mode first, then invoke the command using `:TheCommand`.  `<cmd>` allows staying in the original mode; I hadn't been thinking about needing this functionality, but on searching the Telescope documentation there are some hints that Telescope can be launched from other modes (e.g., `telescope.builtin.grep_string()` can search for the string in the visual selection).

I read through a bunch of the linked issues but on a first reading I haven't understood all the considerations.  It's definitely more complicated than I'd imagined.

fdschmidt93:
> On my machine, timing-related problems with the feedkeys() solution occur frequently throughout the day when I am switching to a familiar buffer name (such that I know in advance the full sequence of keystrokes and I get a lot of practice typing them), so I would ideally like to find another solution for https://github.com/nvim-telescope/telescope.nvim/issues/2274.

I understand. I primarily wanted to give a heads up that fixing all issues simultaneously is very, very hard unfortunately (what I referred to by win some, lose some).

>  1. Launch picker with initial_mode = "insert"
>  2. Have some chain that functionality that closes the picker and opens another one from within the first picker with initial_mode = "normal"
>  3. Be in insert mode instead

This is something quite a few telescope extensions or user pickers are doing in "some" way.

Here is an example (that in regular use though probably wouldn't cause the issue).

```lua
-- Find an (multi-)select files in which you want to perform live_grep
builtin.find_files({
      local selections = current_picker:get_multi_selection()
      -- if no multi-selection, leverage current selection
      if vim.tbl_isempty(selections) then
        table.insert(selections, action_state.get_selected_entry())
      end
      local paths = vim.tbl_map(function(e)
        return e.path
      end, selections)
      actions.close(prompt_bufnr)
      builtin.live_grep({
        search_dirs = paths,
      })
    end)
    -- true: attach default mappings; false: don't attach default mappings
    return true
  end,
})
```

I hope the example is clear.

There's 3 things to consider.

1. The mode from which you start the first picker
2. The mode you exit the first picker
3. The mode you want to enter the second picker

In the above example, `initial_mode` of course is unset, but some pickers (depending on user functions) e.g. lend themselves well to be launched from visual mode.
For instance, you _could_ launch `builtin.live_grep` from visual mode to pass the visual selection as `default_text` (you might intermittently also already leave visual mode and start the picker from normal mode with the selection).
The problem is it's hard to cover all cases universally, while setting modes reliably. The original PR still has issues (e.g. key presses essentially "ahead of time"), but was quite predictable for 99% of typical usage.

drmikehenry:
@fdschmidt93 Thanks for the lengthy explanation and example.  I'm able to test with the example.

What about using the `i` flag with `feedkeys()` to insert the keys at the head of the typeahead buffer (so that the mode switch occurs before processing any already-typed user keystrokes):

```diff
diff --git a/lua/telescope/pickers.lua b/lua/telescope/pickers.lua
index 12a6ba1..d6615b4 100644
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -439,7 +439,7 @@ function Picker:find()
       -- always fully retrigger insert mode: required for going from one picker to next
       keys = mode ~= "n" and "<ESC>A" or "A"
     end
-    a.nvim_feedkeys(a.nvim_replace_termcodes(keys, true, false, true), "n", true)
+    a.nvim_feedkeys(a.nvim_replace_termcodes(keys, true, false, true), "ni", true)
   else
     utils.notify(
       "pickers.find",
```

In my limited testing, this works for my single calls to `find_files()` and for your example.

A possible issue with this approach would be if multiple calls to `Picker:find()` could be made before waiting for user interaction on the first picker; it's not clear to me if or when that would ever happen, but if that were a valid scenario then the chain of calls to `nvim_feedkeys()` would be effectively reversed.  If this is a problem, then maybe there would be some way to differentiate the single-picker scenario from the chained-picker scenario such that for the single-picker case, the `i` flag would be used to fix the user-typeahead problem for this particular edge case.

fdschmidt93:
To clarify, `"ni"` fixes your original issue (the one I can hardly if ever reproduce)? It indeed then seems very promising to solve it entirely then.

```lua
vim.keymap.set("v", "<C-x>", function()
    default_text = "normal",
    initial_mode = "normal",
    attach_mappings = function()
        require("telescope.builtin").live_grep { initial_mode = "insert", default_text = "insert" }
      end)
      return true
    end,
  }
end)
```

For the above function, it works for me as intended with your diff. Nice :)

> A possible issue with this approach would be if multiple calls to Picker:find() could be made before waiting for user interaction on the first picker

I don't think this realistically is a problem. If I understand it correctly, multiple calls to `Picker:find` would have to be made (a) in the context of a single mapping and (b) without user interaction. Since the next `Picker` kills the preceding one, this is not a valid (use) case.

drmikehenry:
Yes, using the `i` flag as shown in the diff I posted solves the fast-user-typing issue for me.  My test for replicating this scenario is:

```vim
nmap <space>ff :Telescope find_files<CR>
call feedkeys("\<space>ff" . "apple")
```

With the `i` flag, the above has `apple` as the (desired) search string; without the `i` flag, the search string is scrambled into `ppleA`.

A different test (that doesn't use `feedkeys()`) borrows for the one suggested by @kevincox in #2274:

```vim
nmap <space>ff :Telescope find_files<CR>
nmap <space>ffo :echo 'not used'<CR>
```

Then type `<space>ffa`.  With the `i` flag, the search string is `a` as expected; without the `i` flag, the search string is corrupted to `A`.

It's hard to replicate with just typing on my home machine; at work, my machine is slower and the problem occurs much more often (though still only randomly).

It sounds like the addition of the `i` flag might well fix things.  I'm at least going to run with that patch at work for a while to test things.  Thanks very much for your kind attention to this issue :-)

drmikehenry:
@fdschmidt93 This week on my work machine I've run with the above patch to include the `i` flag in the call to `nvim_feedkeys()`.  All of the too-fast-typing related problems have gone away.  The only undesirable behavior that the patch might have caused (as explained previously) shouldn't happen given that we don't expect back-to-back calls to `Picker:find()` without time for user interaction.  With slower typing and/or a faster machine, the typeahead buffer will be empty at picker launch and the addition of the `i` flag will have no effect; with faster typing and/or a slower machine, the current bad behavior is fixed by the use of the `i` flag.

fdschmidt93:
Thanks! I'm happy to try `"ni"` for feedkeys :)

drmikehenry:
@fdschmidt93 Excellent;  I've created another pull request to add the `i` flag: <https://github.com/nvim-telescope/telescope.nvim/pull/2625>

Feel free to make any changes you'd like, or let me know if you'd like me to adjust anything.  Thanks again for your help.

PR #2457, opened 2023-04-09
fix: preview windows parentheses

Conni2461:
fix #2446

Issue #2446, opened 2023-03-28
Telescope unable to preview and open files with sveltekit advanced layouts


davisthedev:
### Description

When using telescope search files or live grep within a sveltekit project telescope is unable to preview or open files contained within a sveltekit advanced layout group: [sveltekit docs](https://kit.svelte.dev/docs/advanced-routing).

### Neovim version

```markdown
NVIM v0.8.2
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compiled by runneradmin@fv-az28-353
```

### Operating system and version

Windows 11

### Telescope version / branch / rev

telescope 0.1.1

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
  - OK: nvim-treesitter installed.

## Checking external dependencies
  - OK: rg: found ripgrep 13.0.0 (rev af6b6c543b)
  - OK: fd: found fd 8.7.0

## ===== Installed extensions =====
```

### Steps to reproduce

Within a sveltekit project using the advanced routeing -> advanced layout -> (groups).
Open a search files or live grep telescope window and search for content that will be nested below the route like /routes/(group)/.
The preview window will not display any content.
Then use your bound key or <CR> to open the selected file.
The route will show as /routes(group)/ with the / preceding the group not displaying.
This will open a new buffer with no content.

### Expected behavior

That telescope is able to parse, display preview, and open files contained within these routes.

### Actual behavior

Telescope is unable to preview content and open files contained within these routes

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          { 'nvim-telescope/telescope-fzf-native.nvim', run = 'make' },
        },
      },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
      display = { non_interactive = true },
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
  require('telescope').load_extension('fzf')
end
if vim.fn.isdirectory(install_path) == 0 then
[...]

PR #2628, opened 2023-07-30
expand paths more selectively

jamestrew:
There's a few areas where I believe we're over expanding paths.

We've done some quick fixes with `vim.fn.escape` here and there but I want to explore rather than escaping certain characters then expanding, only expanding when the provided "path" is worth expanding (starts with either '%', '#' or '<'). 

~The places where I'm made changes, it should be safe to assume that at that point, a $ in a path mean the literal $ rather than some shell variable and I'm not even sure _if_ they even ever need expanding.~

Edit: ^ this should largely not be a concern anymore. Unlike with `vim.fn.escape`, `vim.fs.normalize` only expands `$` if it's actually a shell variable. 

closes https://github.com/nvim-telescope/telescope-file-browser.nvim/issues/289

also relevant: 
https://github.com/nvim-telescope/telescope.nvim/pull/2457

### DIFF
--- a/lua/telescope/_.lua
+++ b/lua/telescope/_.lua
@@ -5,6 +5,7 @@ local log = require "plenary.log"
 
 local async = require "plenary.async"
 local channel = require("plenary.async").control.channel
+local utils = require "telescope.utils"
 
 local M = {}
 
@@ -21,7 +22,7 @@ function AsyncJob.new(opts)
   self.stderr = opts.stderr or M.NullPipe()
 
   if opts.cwd and opts.cwd ~= "" then
-    self.uv_opts.cwd = vim.fn.expand(vim.fn.escape(opts.cwd, "$"))
+    self.uv_opts.cwd = utils.path_expand(opts.cwd)
     -- this is a "illegal" hack for windows. E.g. If the git command returns `/` rather than `\` as delimiter,
     -- vim.fn.expand might just end up returning an empty string. Weird
     -- Because empty string is not allowed in libuv the job will not spawn. Solution is we just set it to opts.cwd
--- a/lua/telescope/actions/history.lua
+++ b/lua/telescope/actions/history.lua
@@ -73,7 +73,7 @@ function histories.History:new(opts)
   if conf.history.limit then
     obj.limit = conf.history.limit
   end
-  obj.path = vim.fn.expand(conf.history.path)
+  obj.path = utils.path_expand(conf.history.path)
   obj.content = {}
   obj.index = 1
   obj.cycle_wrap = conf.history.cycle_wrap
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -115,12 +115,12 @@ files.live_grep = function(opts)
   end
   local search_dirs = opts.search_dirs
   local grep_open_files = opts.grep_open_files
-  opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or vim.loop.cwd()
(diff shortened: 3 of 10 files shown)


Conni2461:
I'd like to test run this for a couple of days because it touches some code that i never wanted to touch again (because also windows is affected, and i currently no longer have a easy way to test, without setting up a vm) ^^

Also i dont know if #2457 (#2446) is not fixed by this, because i think they also had trouble opening these files, and `action/set.lua` isn't touched by this PR

jamestrew:
For sure.

As for #2446, I'm pretty sure the issue goes beyond telescope based on some of the discussion + some basic testing I did on a windows machine. This PR is related to #2457 in the sense that it deals with avoiding strange expansions.

Conni2461:
https://github.com/nvim-telescope/telescope.nvim/pull/2457 was never intended to be merged, i just needed that so a windows guy could test out a theory ^^ it should be closed

AvarianKnight:
This fixes my [issue](https://github.com/nvim-telescope/telescope.nvim/issues/2675) with a few tweaks

`make_entry.lua` Line 152 and 313 need this check too, afterwards I can search through files with brackets as the path

Conni2461:
I've just rebased this branch, and am actually testing this now for 2 days straight, maybe we can get this merge then afterwards and get it in 0.1.x prior to 0.1.6

btw james i've wrote you on matrix, regarding that release :)

Conni2461:
Also also maybe we can see if we need @AvarianKnight remarks in `make_entry.lua`, can you look at it james?

Avarian do you use windows? if yes maybe you could test the branch with the paths mentioned in this issue: https://github.com/nvim-telescope/telescope.nvim/issues/2446

AvarianKnight:
> Avarian do you use windows? if yes maybe you could test the branch with the paths mentioned in this issue: #2446

On Windows this is still broken, Live Grep shows nothing and opening the file results it opening the wrong path `src\routes(app)\loading\+page.svelte`

In `buffer_previewer.lua:265` the path needs to have a conditional expand

In `set.lua:200`, it doesn't seem like `vim.fn.fnameescape` escapes parenthesis.
Doing a hacky fix `filename = Path:new(filename):normalize(vim.loop.cwd()):gsub("%(", "\\("):gsub("%)", "\\)")` allows the file to open properly

After doing these two thing every seems to work fine

jamestrew:
Let me look into this some more. This is actually at least my second attempt trying to use less `vim.fn.expand` but previous attempt(s) were too aggressive and broke things the other way by under-expanding.

Main issue is stemming from the fact that people can pass non-absolute paths to the `cwd` option for pickers. Things like `~/` and `$XDG_CONFIG_HOME/` need expanding.

If we abandon the goal of porting this fix to telescope 0.1.x, we can use `vim.fs.normalize` to account for this. I guess we can also use plenary path's `:expand()` method on a path to remain telescope 0.1.x compatible. But I think plenary's `Path` tend to be pretty expensive - I can do some light benchmarking to see the difference/effect.

Conni2461:
yeah lets leave this out, we can always just do multiple releases back to back :D i would personally prefer a solution also for 0.1.x and a solution that doesnt require plenary.path because we kinda wanna move away from that see #2552 :laughing: 

Also also i am not sure how stable plenary `:expand` actually is

jamestrew:
I went with the lazy approach of just borrowing the existing `vim.fs.normalize` function into telescope :sweat_smile:  and then replacing most `vim.fn.expand` references with our own custom expand function.

I'll do some testing on this branch over the next few days.

jamestrew:
I'm going to let this fly.

Issue #506, opened 2021-02-07
Grep not working


debianw:
![Screen Shot 2021-02-06 at 7 44 21 PM](https://user-images.githubusercontent.com/1002407/107134163-3d322d80-68b5-11eb-9da6-4df9eb96c0f2.png)

### Config
`nnoremap <leader>fg <cmd>Telescope live_grep<cr>`

### Description
Grep not working

**Expected Behavior**

**Actual Behavior**
Nothing shows in the preview screen. A Lua error shows up `[telescope] ^I...ig/nvim/plugged/telescope.nvim/lua/telescope/pickers.lua:451: in function <...ig/nvim/plugged/telescope.nvim/lua/telescope/pickers.lua:413>`

### Details

<details><summary>Reproduce</summary>

Example:

1. nvim .
2. :Telescope live_grep
3. type something

NVIM v0.5.0-dev+1064-gb96db38af
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Linux Ubuntu

<details><summary>Configuration</summary>
<p>

```viml
...
nnoremap <leader>ff <cmd>Telescope find_files<cr>
nnoremap <leader>fg <cmd>Telescope live_grep<cr>
nnoremap <leader>fb <cmd>Telescope buffers<cr>
nnoremap <leader>fh <cmd>Telescope help_tags<cr>

 call plug#begin()
   Plug 'neoclide/coc.nvim', {'branch': 'release'}
   Plug 'preservim/nerdtree'
   Plug 'morhetz/gruvbox'
   Plug 'vim-airline/vim-airline'
   Plug 'vim-airline/vim-airline-themes'
   Plug 'nvim-lua/popup.nvim'
   Plug 'nvim-lua/plenary.nvim'
   Plug 'nvim-telescope/telescope.nvim'
   Plug 'dracula/vim'
   Plug 'ryanoasis/vim-devicons'
call plug#end()
...
EOF
```

delphinus:
Do you have `rg` executable? `live_grep` uses `vimgrep_arguments` in config. See [Telescope Defaults](https://github.com/nvim-telescope/telescope.nvim#telescope-defaults).

SuperCuber:
Is there a way to live grep without installing external dependencies?

Conni2461:
@SuperCuber It might be possible to to use grep you just need to end up with a line that we can interpret. Our regex:
[...]

PR #2639, opened 2023-08-06
add missing dependency warning for ripgrep

jamestrew:
For live_grep and grep_string pickers.

Still semi-regularly see comments on old issues about missing rg as a dependency for live_grep https://github.com/nvim-telescope/telescope.nvim/issues/506.

I think adding a warning for these pickers when ripgrep (or some alternative referenced in a non-default `vimgrep_arguments`) is not found will probably save quite a few people some headache.

@Conni2461 maybe we can back port?

PS: redo of #2623

### DIFF
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -36,6 +36,22 @@ local escape_chars = function(string)
   })
 end
 
+local has_rg_program = function(picker_name, program)
+  if vim.fn.executable(program) == 1 then
+    return true
+  end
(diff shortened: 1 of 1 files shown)


Conni2461:
Yeah i can backport, once it hits master, no problem. 

I never was a fan of adding this because `finder.one_shot` or `finder.new_job` both check that (and print an error) but after was opened and maybe that was to confusing for ppl, idk.

So yeah adding this is fine, i'll backport with a cherry-pick

PR #2648, opened 2023-08-11
perf(utils): linear scan `is_uri`

jamestrew:
Benchmarked against 80k+ unique paths, both Windows and Linux (macos should be close enough to llinux).

Approximately 13% faster for linux paths and 55% faster for windows paths.

```
hyperfine 'luajit current_linux.lua' 'luajit fast_linux.lua' --warmup 10
Benchmark 1: luajit current_linux.lua
  Time (mean ± σ):      25.7 ms ±   0.8 ms    [User: 22.7 ms, System: 3.4 ms]
  Range (min … max):    24.0 ms …  28.8 ms    87 runs

Benchmark 2: luajit fast_linux.lua
  Time (mean ± σ):      22.7 ms ±   0.8 ms    [User: 19.4 ms, System: 3.6 ms]
  Range (min … max):    21.4 ms …  26.0 ms    99 runs

Summary
  luajit fast_linux.lua ran
    1.13 ± 0.05 times faster than luajit current_linux.lua

hyperfine 'luajit current_win.lua' 'luajit fast_win.lua' --warmup 10
Benchmark 1: luajit current_win.lua
  Time (mean ± σ):      36.3 ms ±   0.9 ms    [User: 34.1 ms, System: 2.7 ms]
  Range (min … max):    34.7 ms …  38.9 ms    69 runs

Benchmark 2: luajit fast_win.lua
  Time (mean ± σ):      23.5 ms ±   0.7 ms    [User: 21.1 ms, System: 2.9 ms]
  Range (min … max):    22.4 ms …  25.7 ms    95 runs

Summary
  luajit fast_win.lua ran
    1.55 ± 0.06 times faster than luajit current_win.lua
```

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -173,12 +173,32 @@ utils.is_path_hidden = function(opts, path_display)
     or type(path_display) == "table" and (vim.tbl_contains(path_display, "hidden") or path_display.hidden)
 end
 
-local URI_SCHEME_PATTERN = "^([a-zA-Z]+[a-zA-Z0-9.+-]*):.*"
-local WINDOWS_ROOT_PATTERN = "^[a-zA-Z]:\\"
 utils.is_uri = function(filename)
-  local is_uri_match = filename:match(URI_SCHEME_PATTERN) ~= nil
-  local is_windows_root_match = filename:match(WINDOWS_ROOT_PATTERN)
-  return is_uri_match and not is_windows_root_match
+  local char = string.byte(filename, 1)
+
+  -- is alpha?
+  if char < 65 or (char > 90 and char < 97) or char > 122 then
+    return false
+  end
+
+  for i = 2, #filename do
+    char = string.byte(filename, i)
+    if char == 58 then -- `:`
+      return i < #filename and string.byte(filename, i + 1) ~= 92 -- `\`
+    elseif
+      not (
+        (char >= 48 and char <= 57) -- 0-9
+        or (char >= 65 and char <= 90) -- A-Z
+        or (char >= 97 and char <= 122) -- a-z
+        or char == 43 -- `+`
+        or char == 46 -- `.`
+        or char == 45 -- `-`
+      )
+    then
+      return false
+    end
+  end
+  return false
 end
 
 local calc_result_length = function(truncate_len)
--- a/lua/tests/automated/utils_spec.lua
+++ b/lua/tests/automated/utils_spec.lua
@@ -1,7 +1,7 @@
 local utils = require "telescope.utils"
 
 describe("is_uri", function()
-  it("detects valid uris", function()
+  describe("detects valid uris", function()
     local uris = {
       [[https://www.example.com/index.html]],
       [[ftp://ftp.example.com/files/document.pdf]],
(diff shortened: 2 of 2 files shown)


juntuu:
Hi, just bumped into this while browsing.

I'm not at all familiar with the project, so can't say if the performance improvement would outweigh the complexity cost, and what would be the effect in real workloads.

However, if I'm not mistaken this change introduces two differences in the matching:

1. does not require a `:` in uri (`is_uri("hello") == true`)
2. the windows "drive letter" can be more than one letter (`is_uri("hello:\\") == false`)

Because the function is checking for uri and not windows path, I think 2. is more correct now. Uri should not contain `\` anyway.

But 1. seems like a bug.

I think the following (pseudocode) should fix 1. while keeping 2.:
```lua
if filename[1] not in [a-zA-Z] then
  return false
end
for i = 2, #filename do
  if filename[i] not in [a-zA-Z0-9.+-] then
    return false
  elseif filename[i] == ':' then
    return filename[i+1] ~= '\\'
  end
end
return false
```

max397574:
@juntuu that pseudocode definitely won't work since there is no chance the `elseif` statement ever is true
it would be true if the character is a colon but if that's the case it will already return false because of the if statement

juntuu:
> that pseudocode won't work

Oops, good catch. Thanks!

I flipped the conditions now, so the `:` is checked first.

jamestrew:
Good catches to both!
I guess under ideal conditions, something like `is_uri('hello')` or really anything that's not a full path shouldn't happen but for the sake of correctness, I like the suggested change.

No significant performance difference between my original implementation and the suggested change.

```
$ hyperfine 'luajit before_linux.lua' 'luajit after_linux.lua' --warmup 10 --runs 100
Benchmark 1: luajit before_linux.lua
  Time (mean ± σ):      27.2 ms ±   0.5 ms    [User: 22.6 ms, System: 4.5 ms]
  Range (min … max):    26.4 ms …  29.9 ms    100 runs

Benchmark 2: luajit after_linux.lua
  Time (mean ± σ):      27.5 ms ±   0.8 ms    [User: 22.8 ms, System: 4.7 ms]
  Range (min … max):    26.0 ms …  30.5 ms    100 runs

Summary
  luajit before_linux.lua ran
    1.01 ± 0.03 times faster than luajit after_linux.lua
```

```
$ hyperfine 'luajit before_win.lua' 'luajit after_win.lua' --warmup 10 --runs 100
Benchmark 1: luajit before_win.lua
  Time (mean ± σ):      26.9 ms ±   0.5 ms    [User: 23.6 ms, System: 3.6 ms]
  Range (min … max):    26.0 ms …  28.8 ms    100 runs

Benchmark 2: luajit after_win.lua
  Time (mean ± σ):      25.9 ms ±   0.7 ms    [User: 22.9 ms, System: 3.3 ms]
  Range (min … max):    24.8 ms …  28.9 ms    100 runs

Summary
  luajit after_win.lua ran
    1.04 ± 0.03 times faster than luajit before_win.lua
```

jamestrew:
I'm going to go ahead and merge this in. I think considering this is a pretty hot function, the performance gain, particularly on Windows is worth the diminished readability.

PR #2655, opened 2023-08-16
feat: check if cmd extension can be loaded

willothy:
# Description

As a last resort if no other cmd is found in `run_cmd`, check if the command can be loaded as an extension.

I know pcall is a bit hacky, but not needing to load every extension on startup is nice.

## Type of change

# How Has This Been Tested?

Install extension `telescope-frecency`, but don't load it with `load_extension`. Then, run `:Telescope frecency`. Before this change, this would error. Now, the command runs as expected.

### DIFF
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -194,6 +194,12 @@ local function run_command(args)
     return
   end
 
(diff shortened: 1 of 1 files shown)


Conni2461:
thanks LGTM :)

PR #2656, opened 2023-08-18
feat(jobs): emit job errors

jamestrew:
@Conni2461 thoughts on doing something like this (and possible for oneshot_job as well)?

I feel like 50% of to notification I get for telescope these days is people having issues with live_grep not working. Sometimes it's just a matter of checking they have it installed but other times it's almost possibly to tell what the issue is.

### DIFF
--- a/lua/telescope/finders/async_job_finder.lua
+++ b/lua/telescope/finders/async_job_finder.lua
@@ -1,5 +1,6 @@
 local async_job = require "telescope._"
 local LinesPipe = require("telescope._").LinesPipe
+local ErrorPipe = require("telescope._").ErrorPipe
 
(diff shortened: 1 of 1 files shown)


jamestrew:
I guess not all errors are not worth emitting however.
For example with a prompt like `foo\.bar`, upon typing the `\`, it would it emit a `regex parse error: incomplete escape sequence`.

Conni2461:
i dont know we wanna do that again, we already had that in the beginning and every wrong query just end up showing `nil:nil:nil:nil` at least for grep, because the parsing no longer worked. I think we just need a different consumer for everything written to stderr, like a logger or `vim.notify`

jamestrew:
yeah i don't think we should merge this.

PR #2682, opened 2023-09-02
fix(builtin.pickers): fix wrong picker resuming when using filtering

cristiansofronie:
# Description

When filtering builtin.pickers doesn't resume the correct picker or delete it.

## Type of change

# How Has This Been Tested?

Ran require('telescope.builtin').live_grep 10 times with different queries. Then search for specific entries and hit enter and verified that there resumed picker has the distinct text that I had selected. I also removed some of them from the beginning, middle and end. 

# Checklist:

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -1158,12 +1158,25 @@ end
 --- This action is not mapped by default and only intended for |builtin.pickers|.
 ---@param prompt_bufnr number: The prompt bufnr
 actions.remove_selected_picker = function(prompt_bufnr)
-  local current_picker = action_state.get_current_picker(prompt_bufnr)
-  local selection_index = current_picker:get_index(current_picker:get_selection_row())
+  local curr_picker = action_state.get_current_picker(prompt_bufnr)
+  local curr_entry = action_state.get_selected_entry()
   local cached_pickers = state.get_global_key "cached_pickers"
-  current_picker:delete_selection(function()
+
+  if not curr_entry then
+    return
+  end
+
+  local selection_index, _ = utils.list_find(function(v)
+    if curr_entry.value == v.value then
+      return true
+    end
+    return false
+  end, curr_picker.finder.results)
+
+  curr_picker:delete_selection(function()
     table.remove(cached_pickers, selection_index)
   end)
+
   if #cached_pickers == 0 then
     actions.close(prompt_bufnr)
   end
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -194,9 +194,21 @@ internal.pickers = function(opts)
       cache_picker = false,
       attach_mappings = function(_, map)
         actions.select_default:replace(function(prompt_bufnr)
-          local current_picker = action_state.get_current_picker(prompt_bufnr)
-          local selection_index = current_picker:get_index(current_picker:get_selection_row())
+          local curr_picker = action_state.get_current_picker(prompt_bufnr)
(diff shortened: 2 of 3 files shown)


jamestrew:
Sorry I haven't worked with or used the `pickers` picker ever so I'm not exactly clear on what the issue is.

Using default settings, resuming pickers via `pickers` and deleting pickers seem to work fine for me. Can you provide some steps to reproduce the issue?

cristiansofronie:
The current implementation takes the row of the current selection and indexes into the global cached_pickers table. This only works when the row of the selection corresponds to the right index in the cached_pickers table. When you use filtering things changes ordering and a random picker is resumed.

```lua
for i = 1, 10 do
  require('telescope.builtin').live_grep({ default_text = i })
end

require('telescope.builtin').pickers()
```

:luafile that and filter for a specific number and hit Enter. It will not select the correct picker.

cristiansofronie:
This should be a video with the issue

https://github.com/nvim-telescope/telescope.nvim/assets/53446505/c4dd35d6-b690-48ca-b909-08cebb57486d

jamestrew:
It looks like the video is corrupt - I'm not able to view it. And also putting the launching of pickers in a for loop seems to be causing other weird issues with input queuing.
Also you're not using the default ` cache_picker = { num_pickers = 1 }` right?

But trying your steps manually with `num_pickers = 10` seems to still work fine for me.

https://github.com/nvim-telescope/telescope.nvim/assets/66286082/2552b2b0-1387-4ca2-80f2-9d75c9deabd3

cristiansofronie:
Yes but you are not entering anything inside the prompt. You are not searching you are just using Telescope just for selecting. Type 2 inside the prompt. The live_grep with the prompt set to 2 should be restored when you hit enter

jamestrew:
Okok I see now sorry. Give me a moment to give your PR a look.

cristiansofronie:
The only thing the PR replaces is the method by which we find the index. Everything else is the same.

cristiansofronie:
Also, what method do you all use to quickly test new things. I use Packer with local packages, do you have some faster methods since this is really slow and forces me to use a broken version of Telescope while developing telescope itself.

jamestrew:
> Also, what method do you all use to quickly test new things. I use Packer with local packages, do you have some faster methods since this is really slow and forces me to use a broken version of Telescope while developing telescope itself.

I do a similar thing with lazy where my telescope installation points to a local copy. From there I might use two sessions of neovim (using tmux), one with a working copy of telescope and another for experimenting/testing.
I can make changes in the working copy then test the changes in the experimental nvim session either by relaunching neovim or using this `R` function to reload telescope.

```lua
if pcall(require, "plenary") then
  RELOAD = require("plenary.reload").reload_module

  R = function(name)
    RELOAD(name)
    return require(name)
  end
end
```

cristiansofronie:
Cool, thanks. I have seen something over at https://github.com/nvim-telescope/telescope.nvim/issues/2667 for having a minimal config when testing Telescope. It might be interesting to experiment with but I haven't tried it yet.

jamestrew:
thanks :+1: 

> Cool, thanks. I have seen something over at https://github.com/nvim-telescope/telescope.nvim/issues/2667 for having a minimal config when testing Telescope. It might be interesting to experiment with but I haven't tried it yet.

For these minimal configs, I just stick them into like a `/tmp/min.lua` file and launch neovim with `nvim -u /tmp/min.lua` (requires restarting a couple times to get plugins installed and working correctly). This lets us test things in a reproducible way without people's personal configs affecting things.

PR #2684, opened 2023-09-04
fix(telescope.state.get_existing_prompts): it should only return keys that are numbers

cristiansofronie:
# Description

telescope.state.get_existing_prompts currently returns all the keys from TelescopeGlobalState, this includes the 'global' key. This function is only used in Picker:close_existing_pickers where we iterate through the  prompt_bufnr and call actions.close on each prompt_bufnr so it should only return keys that are numbers

## Type of change

# How Has This Been Tested?

Run
:lua =require('telescope.state').get_existing_prompts()
When there are no pickers opened. It should return and empty table. With a picker present it should return the bufnr for that picker's prompt.

# Checklist:

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -1227,7 +1227,7 @@ end
 
 --- Close all open Telescope pickers
 function Picker:close_existing_pickers()
-  for _, prompt_bufnr in ipairs(state.get_existing_prompts()) do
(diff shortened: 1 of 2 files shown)


jamestrew:
looks good. maybe we can rename the function `get_existing_prompt_bufnrs` to be explicit.

cristiansofronie:
I did the name change and also used pairs to only iterate once over the table since tbl_keys and tbl_filter each use a pairs internally and that is slower

jamestrew (approved):

Issue #2689, opened 2023-09-06
dynamic_preview_title doesn't update in builtin.resume picker


davidsierradz:
### Description

Hi!

When calling any picker that uses `dynamic_preview_title`, closing it, then calling `builtin.resume`, I expect the preview title to keep getting updated the same as the cached picker, but instead the preview title is frozen on the last item of the cached picker when closed.

Thanks!

### Neovim version

```markdown
NVIM v0.10.0-dev-069fad6
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

NixOS 23.11.20230902.e569908 (Tapir)

### Telescope version / branch / rev

6b79d7a6a45adc1508a7afee5bc973173ec22f59

### checkhealth telescope

```markdown
──────────────────────────────────────────────────────────────────────────────
telescope: health#telescope#check
 
Checking for required plugins  
- OK plenary installed.
- WARNING nvim-treesitter not found. (Required for `:Telescope treesitter`.)
 
Checking external dependencies  
- OK rg: found ripgrep 13.0.0
- OK fd: found fd 8.7.0
 
===== Installed extensions =====
```

### Steps to reproduce

1. nvim -u minimal.lua
2. Open file picker (`<Space>j`)
3. Close file picker
4. Open resume picker (`<Space>z`)
5. Move through different files 

### Expected behavior

I expect the `dynamic_preview_title` to update the filename in the preview title in the resume picker

### Actual behavior

https://github.com/nvim-telescope/telescope.nvim/assets/9190258/43f92448-ef41-4470-a804-eefaf9475ddb

### Minimal config

```Lua
vim.g.mapleader = ' '

local lazypath = vim.fn.stdpath('data') .. '/lazy/lazy.nvim'
if not vim.uv.fs_stat(lazypath) then
  vim.fn.system({
    'git',
    'clone',
    '--filter=blob:none',
    'https://github.com/folke/lazy.nvim.git',
    '--branch=stable', -- latest stable release
    lazypath,
  })
end
vim.opt.rtp:prepend(lazypath)

require('lazy').setup({
  {
    'nvim-telescope/telescope.nvim',
    lazy = false,
    dependencies = { { 'nvim-lua/plenary.nvim', lazy = false } },
  },
}, {
  lockfile = os.getenv('HOME') .. '/nixos/hosts/base/config/nvim/lazy-lock-test.json',
  defaults = {
    lazy = false,
  },
})

require('telescope').setup({
[...]

PR #2696, opened 2023-09-08
fix(internal.resume): dynamic_preview_title was not respected and title became static

cristiansofronie:
In Picker:new there was
```lua
    if obj.preview_title == nil or #obj.all_previewers > 1 then
      obj.preview_title = obj.previewer:title(nil, config.values.dynamic_preview_title)
    else
      obj.fix_preview_title = true
    end
```
so if a preview_title was not provided at creation it became static forever. This is a problem when you resume a picker as internal.resume calls Picker:new again and the same preview_title is populated during runtime by the dynamic_preview_title. So now that field is not nil and will become static forever.

To fix it add an additional check. If the picker is resumed and fix_preview_title is not already set by the first creation do not set it now.

This fixes https://github.com/nvim-telescope/telescope.nvim/issues/2689

## Type of change

# How Has This Been Tested?

Manual testing. Set   dynamic_preview_title = true in defaults when you setup Telescope and then open the find_files picker close it and resume.

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -161,6 +161,7 @@ internal.resume = function(opts)
     picker.hidden_previewer = nil
     opts.previewer = vim.F.if_nil(opts.previewer, false)
   end
+  opts.resumed_picker = true
   pickers.new(opts, picker):find()
 end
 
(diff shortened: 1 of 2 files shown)


jamestrew (approved):
thanks :+1:

PR #2597, opened 2023-07-09
feat(git): support detached working trees

jamestrew:
closes #2595

# Description

Adds support for detached working tree for git related pickers.
Closely follows the pattern utilized by gitsigns.nvim in https://github.com/lewis6991/gitsigns.nvim/pull/600

eg use case
```lua
local opts = {
  worktrees = {
    {
      toplevel = vim.env.HOME,
      gitdir = vim.env.HOME .. "projects/dotfiles/.git",
    },
  },
}

require("telescope").setup({
  pickers = {
    git_files = opts,
    git_stash = opts,
    git_status = opts,
    git_commits = opts,
    git_bcommits = opts,
    git_branches = opts,
  },
})
```

## Type of change

Please delete options that are not relevant.

- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list relevant details about your configuration

- [ ] Test A
- [ ] Test B

**Configuration**:
* Neovim version (nvim --version):
* Operating system and version:

# Checklist:

- [ ] My code follows the style guidelines of this project (stylua)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (lua annotations)

Issue #435, opened 2021-01-15
find_command - support environment variables and aliases


roginfarrer:
I was trying to create a custom search command for my dotfiles. I have my dotfiles in a bare git repository, so it's not as simple as pointing `git_files` to a specific directory. My first attempt, and what I want to see work as expected, is the following:

```lua
function search_dotfiles()
  require('telescope.builtin').find_files {
    find_command = {'git', '--git-dir', '$HOME/.dotfiles/', '--work-tree', '$HOME', 'ls-tree', '--full-tree', '-r', '--name-only', 'HEAD'}
  }
end
```

`$HOME` is an environment variable set to `/Users/rfarrer/` on my machine. 

Furthermore, it would be cool if my shell aliases also worked. I have the command `config` mapped to `git --git-dir=$HOME/.dotfiles/ --work-tree=$HOME`, and it would be cool if that would work, as well:

```lua
function search_dotfiles()
  require('telescope.builtin').find_files {
    find_command = {'config', 'ls-tree', '--full-tree', '-r', '--name-only', 'HEAD'}
  }
end
```

If I try to use my `config` command or `$HOME`, Telescope fails to find any files (I think because it fails to expand those variables).

Conni2461:
About the first one: We would need to either expand each entry of the table, or we would need to pass in your whole env to plenary job because a plenary job, which uses luvit, has an empty env on calling. I don't like both tbh.

About the second one: If i am not wrong but i think thats impossible for the same reason why `:!config status` will also give you `command not found` we would need to source your `.bashrc`/`.zshrc` before we actually could run a command which can be slow.

But you could just solve that issue by your own. Put a bash script called `config` in your path (don't forget to make it executable):
```bash
#!/bin/sh

/usr/bin/git --git-dir=$HOME/.dotfiles/ --work-tree=$HOME $*
```

roginfarrer:
Thanks for the help. I thought I'd mention it since I could see this be a common use case. I can do this pretty easily in fzf, which I understand interacts with the terminal environment differently:

```vim
function! s:fzf_dotfiles()
  let command = 'config ls-tree --full-tree -r --name-only HEAD'
  call fzf#run(fzf#wrap({ 
    \ 'dir': '~/', 
    \ 'source': command }))
endfunction
```

Feel free to close this if you don't think it will be implemented.

Conni2461:
I think its not a trivial think to realize. If someone wants to prove me wrong, they are very welcome to do so.

[...]

PR #2705, opened 2023-09-18
fix(actions): detached worktrees not detected for git actions

Subjective:
# Description
- Fixes an issue where git actions (checkout, switch branch, etc.) do not work when within a custom detached working tree (first introduced in https://github.com/nvim-telescope/telescope.nvim/pull/2597).
- Moves helper functions for setting git command options in [__git.lua](https://github.com/nvim-telescope/telescope.nvim/compare/master...Subjective:telescope.nvim:fix-git-worktree-actions?expand=1#diff-ad7c3e26be2be9a1936821c7295f4cb83be559844a35c1331870e599909ef89f) to [lua/telescope/utils.lua](https://github.com/nvim-telescope/telescope.nvim/compare/master...Subjective:telescope.nvim:fix-git-worktree-actions?expand=1#diff-d386ae6bc0710be7aeece837417a83e1b28473f002e54449e103e8f44bf7abbc) so that they can be set directly in `utils.__git_command`.

## Type of change

# How Has This Been Tested?

Tested the following pickers with detached/attached working trees and git worktrees.

- git_files
- git_status
- git_stash
- git_commits
- git_bcommits
- git_branches

**Configuration**:
* Neovim version (nvim --version): NVIM v0.10.0-dev-d70667a
* Operating system and version: MacOS 13.4

# Checklist:


Subjective:
@jamestrew Since you originally created the PR for the detached git_worktrees feature, do you mind reviewing these changes?

Subjective:
Now that I think about it, none of the git actions are inheriting the `use_file_path`, and `use_git_root` options passed into the builtin git pickers.

https://github.com/nvim-telescope/telescope.nvim/blob/b543aaa2c9cf8123ed2fe7dbb6c211a9cd415124/doc/telescope.txt#L1018-L1027

From what I can tell, it doesn't seem like the `cwd` used by any of the actions is even guaranteed to match the `cwd` that was passed into the picker — currently it's being retrieved from Telescope's internal state:

```lua
local cwd = action_state.get_current_picker(prompt_bufnr).cwd
```

Subjective:
While I'm not yet entirely sure how telescope manages its internal state for pickers, it seems like it might be possible to store the options passed into any of the builtin get pickers via: https://github.com/nvim-telescope/telescope.nvim/blob/b543aaa2c9cf8123ed2fe7dbb6c211a9cd415124/lua/telescope/state.lua#L6-L9

If this is true, then they can retrieved with something like:
```lua
local opts = action_state.get_current_picker(prompt_bufnr).opts

local get_git_command_output = function(args, opts)
  return utils.get_os_command_output(utils.__git_command(args, opts), opts.cwd)
end

get_git_command_output({ "checkout", "-b", new_branch }, opts))
```

Instead of how it's done currently:
```lua
local cwd = action_state.get_current_picker(prompt_bufnr).cwd
utils.get_os_command_output(git_command { "checkout", "-b", new_branch }, cwd)

**Comment by jamestrew** [comment]
Hmm... yeah I see the issue and honestly it was a bit of an oversight on my part not handling the actions before merging that previous PR.

As far as getting access to the `gitdir` and `toplevel` fields inside actions, I have a couple of alternative solutions:
1. inherit and extend the base picker and just add `gitdir`/`toplevel` (and any other necessary fields) to it and use that for all git related pickers -- then I think we can just do like 
```lua
local gitdir = action_state.get_current_picker(prompt_bufnr).gitdir
```
2. bind the necessary fields to the finder but this might not work since some actions cause a refresh of the picker so the binding would get lost
3. wrap all git actions, passing in necessary fields so the actions have access to them as part of the closure
eg.
```lua
actions.git_create_branch = function(opts)
  return function(prompt_bufnr)
    -- has access to opts.gitdir, etc
    
  end
end

-- then map like this
map({ "i", "n" }, "<c-a>", actions.git_create_branch(opts))
```

This way we can avoid resolving `gitdir` & `toplevel` each time want to call a git command I think.

Option 1 might be easiest but I don't think it's something we do.

@Conni2461 Is there an idiomatic way of passing arbitrary option fields for pickers to actions?

psnelgrove-r7:
I've been working through setting up a bare git repository for managing my dotfiles. Been using the work around mentioned in the last comment of this issue.
 
https://github.com/nvim-telescope/telescope.nvim/issues/435 

Works reasonably well but it got me digging into why `git_files` wasn't working out of the box and I found #2597 which links to this issue.

What's a little confusing is where along the line `git_worktrees` was dropped and why.

jamestrew:
> I've been working through setting up a bare git repository for managing my dotfiles. Been using the work around mentioned in the last comment of this issue.

Some of the git pickers support a `git_command` option where you can pass these git env variables. But not all do. The aim of #2597 was to enable passing these git env vars to all git pickers (for multiple git worktrees if desired) at the top level telescope settings.

Problem with both these methods is that the actions related to the pickers still won't know about the env vars.

> I think there is a good reason to just store the original passed opts inside the picker and then use these later in actions. 

Yeah that's what I'd like to do if I'm interpreting this correctly. Like do something like this inside each of the git actions:
```lua
local gitdir = current_picker.gitdir
local toplevel = current_picker.toplevel
```
Is this currently possible? I don't think

iton0:
> I've been working through setting up a bare git repository for managing my dotfiles. Been using the work around mentioned in the last comment of this issue.
> 
> #435
> 
> Works reasonably well but it got me digging into why `git_files` wasn't working out of the box and I found #2597 which links to this issue.
> 
> What's a little confusing is where along the line `git_worktrees` was dropped and why.
> 
> EDIT: Ah I see that's it's in master and not the 1.x branch.

@psnelgrove-r7 how did you implement the work around in your telescope config?

PR #2715, opened 2023-09-23
feat(preview)!: add `opts.preview.highlight_limit` with default 1MB

### LINKED ISSUE #2247: can add a option to disable treesitter highlight when preview large file？
when i use telescope.nvim in large file, the goto lsp or current fuzzy find will wait some time, the probably reason is
treesitter，so can add a option to disable treesitter highlight when preview large file？

https://user-images.githubusercontent.com/43649186/204116585-4ee2b7f4-b2d8-46b9-aaab-f632e333d9b6.mp4

llllvvuu:
Fixes #2247 - highlighting is not async (yet) so blocks input

BREAKING CHANGE: Default value will cause files > 1MB to not be highlighted. Removing this default would make this a non-breaking change.

The following Wiki section should also be updated to point out this new option: https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes#disable-highlighting-for-certain-files

# How Has This Been Tested?

- [x] Used Telescope in my repo with huge files, no longer hangs
- [x] Tried `highlight_limit = false` and `highlight_limit = 25`, hangs again

# Checklist:

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -182,8 +182,8 @@ local handle_file_preview = function(filepath, bufnr, stat, opts)
       end
     end
 
+    local mb_filesize = stat.size / bytes_to_megabytes
     if opts.preview.filesize_limit then
-      local mb_filesize = math.floor(stat.size / bytes_to_megabytes)
       if mb_filesize > opts.preview.filesize_limit then
         if type(opts.preview.filesize_hook) == "function" then
           opts.preview.filesize_hook(filepath, bufnr, opts)
@@ -228,7 +228,10 @@ local handle_file_preview = function(filepath, bufnr, stat, opts)
         if opts.callback then
           opts.callback(bufnr)
         end
-        putils.highlighter(bufnr, opts.ft, opts)
+
+        if not (opts.preview.highlight_limit and mb_filesize > opts.preview.highlight_limit) then
+          putils.highlighter(bufnr, opts.ft, opts)
+        end
       else
         if type(opts.preview.timeout_hook) == "function" then
           opts.preview.timeout_hook(filepath, bufnr, opts)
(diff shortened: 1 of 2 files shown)


jamestrew (approved):
lgtm :+1: 
I think the file size limit is reasonable enough that it won't impact most use cases so not a major breaking change.

I'll let @Conni2461 be the final judge though.

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -182,8 +182,8 @@ local handle_file_preview = function(filepath, bufnr, stat, opts)
       end
     end
 
+    local mb_filesize = math.floor(stat.size / bytes_to_megabytes)
```
Conni2461, line 185: hmm, why did we every floor this?

It doesnt make sense to me, so e.g. 821792byte file ~ 0.78 mb, after the floor its 0.
seems weird. so i can never actually ignore all files >= 0.5mb e.g.

also if we now have a file with 1500000byte ~ 1.43mb, after the floor its 1. which is in fact not > highlight_limit an will be highlighted. so we actually start ignoring files > 2 not > 1. 

"The maximum file size in MB attempted to be highlighted." suggest that 1mb is the maximum filesize that is being highlighted, but 1.2mb files will still get highlighting.

I know the floor was already there (so not your mistake), but in my opinion it should not be floored.

Any thoughts?

Conni2461:
I don't care about a breaking change here ^^ we have a stable branch for that reason. I'd like to do more drastic changes soon anyway

lua/telescope/previewers/buffer_previewer.lua
llllvvuu: You're right, let's use this opportunity to change it.

I just updated the PR.

Conni2461:
thanks works great now :)

PR #2727, opened 2023-10-04
Fixing small typo in the Readme file.

andreangelucci:
# Description

Fixing a code snippet in the Readme file.

## Type of change

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ customizations globally, or individually per picker.
 ### Telescope setup structure
 
 ```lua
-require('telescope').setup{
+require('telescope').setup({
   defaults = {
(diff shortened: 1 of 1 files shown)


jamestrew (changes requested):
Thanks but the original is actually valid lua and closer in line with our lua style guide (stylua config). 

Technically, according to our stylua config it would be more like
```lua
require("telescope").setup {
    -- ...
}
```

Our lua formatting in the readme is all over the place so it's not a big deal but if you'd like to address the formatting for these lines and/or others, that'll be welcomed.

andreangelucci:
Thanks, James!

PR #2726, opened 2023-10-03
fix(actions): fixes folding. Fixes #699

folke:
# Description

Folding (with `foldexpr`) for files opened with Telescope is broken.

Fixes #699

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [x] set a foldxpr and open a file with telescope. foldexpr should work

**Configuration**:
* Neovim version (nvim --version):
* Operating system and version:

# Checklist:

- [x] My code follows the style guidelines of this project (stylua)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (lua annotations)

PR #2729, opened 2023-10-06
refactor(actions): make foldexpr fix window-local

llllvvuu:
See: https://github.com/nvim-telescope/telescope.nvim/pull/2726

FWIW the reason it works is: https://github.com/neovim/neovim/issues/25518


llllvvuu:
Just tested, you're right

PR #2751, opened 2023-10-22
feat(lsp-jump-type): `tab drop` as new `jump_type` option for `go-to` LSP pickers

HicaroD:
# Description

Sometimes when I'm coding I want to jump to a definition using the command `builtin.lsp_definitions`, for example, and it adds a new buffer on top of the buffer I'm working on. This behavior is something I don't like, so I started reading the documentation for changing that behavior and I found a parameter called [`jump_type`](https://github.com/nvim-telescope/telescope.nvim/blob/74ce793a60759e3db0d265174f137fb627430355/lua/telescope/builtin/init.lua#L444) and I set it to `tab`. However, there is another problem: when I jump to a definition, it adds a new tab buffer, but, if that tab buffer is already opened, it does not jump to it, it adds a new one. So, the behavior that I was looking for only exists with the command `tab drop`, so I added it to the list of options for the `jump_type` parameter of the following commands: `builtin.lsp_type_definitions`, `builtin.lsp_implementations`, `builtin.lsp_references` and `builtin.lsp_definitions`.

NOTE: I'm using native Neovim tabs, not the plugin [`bufferline`](https://github.com/akinsho/bufferline.nvim).

## Type of change

# How Has This Been Tested?

- Using [`nvim-lspconfig`](https://github.com/neovim/nvim-lspconfig) for providing LSP features, `tab drop` as `jump_type` option in the following Telescope LSP pickers: `builtin.lsp_type_definitions`, `builtin.lsp_implementations`, `builtin.lsp_references` and `builtin.lsp_definitions`, and `pyright` as my Python language server, I have some files that import things from each other, and I tried to use `builtin.lsp_definitions` feature for going to the definition of some random class defined in other file, and it opened a new tab buffer (like expected). After this, I left the definition tab buffer opened and used the same command for going to the definition, and it jumped to the existing buffer instead of adding a new one. Additionally, I tried some equivalent example in some Golang code using `gopls` as my language server and it worked perfectly as well. It worked well for the remaining LSP pickers.

**Configuration**:
  ```
  NVIM v0.9.2
  Build type: Release
  LuaJIT 2.1.1694285958
  ```

  ```
  OS: Manjaro Linux x86_64
  Kernel: 6.1.55-1-MANJARO
  ```

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1549,8 +1549,8 @@ builtin.lsp_references({opts})            *telescope.builtin.lsp_references()*
         {jump_type}            (string)   how to goto reference if there is
                                           only one and the definition file is
                                           different from the current file,
-                                          values: "tab", "split", "vsplit",
-                                          "never"
+                                          values: "tab", "tab drop", "split",
+                                          "vsplit", "never"
         {fname_width}          (number)   defines the width of the filename
                                           section (default: 30)
         {show_line}            (boolean)  show results text (default: true)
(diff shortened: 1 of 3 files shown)


Conni2461:
LGTM thanks :)

i've ran docgen so the vim documentation is generated.

PR #2760, opened 2023-10-26
fix: nil problem when jump to location when `tab drop` is set as `jump_type`

HicaroD:
# Description

I recently implemented a feature to add `tab_drop` as `jump_type` option for some LSP pickers in Telescope. I tested in two programming languages (Python and Go) and it worked fine. Today, I tried to write some Rust code and I was not able to use the "Go to definition" feature from  `nvim-lspconfig` due to problems with `nil` when trying to jump to the location.

That was being caused the file path (`flattened_results[1].uri`) being passed in the method `vim.uri_to_fname` was nil. That got me confuse because it works really well with Python (`pyright`) and Go (`gopls`), but not in Rust (`rust-analyzer`). After that, I realized that there is another attribute called `targetUri` (`flattened_results[1].targetUri`) that represents the target URI (as the name says). I tried to replace `uri` for `targetUri` and it worked well with Rust, but Go and Python didn't accept it and the nil problem came back now with Python and Go.

Therefore, I making that PR for addressing this issue and now it is working fine with all the languages that I tested (Dart with `dartls`, Go with `gopls`, Python with `pyright`, Rust with `rust-analyzer` and C++ with `clangd`)

## Type of change

# How Has This Been Tested?

- Used the feature for jumping to the definition in the following programmings languages and its language servers: Dart with `dartls`, Go with `gopls`, Python with `pyright`, Rust with `rust-analyzer` and C++ with `clangd`.

**Configuration**:

  ```
  NVIM v0.9.2
  Build type: Release
  LuaJIT 2.1.1694285958
  ```
  
  ```
  OS: Manjaro Linux x86_64
  Kernel: 6.1.55-1-MANJARO
  ```

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1549,8 +1549,8 @@ builtin.lsp_references({opts})            *telescope.builtin.lsp_references()*
         {jump_type}            (string)   how to goto reference if there is
                                           only one and the definition file is
                                           different from the current file,
-                                          values: "tab", "split", "vsplit",
-                                          "never"
+                                          values: "tab", "tab drop", "split",
+                                          "vsplit", "never"
         {fname_width}          (number)   defines the width of the filename
                                           section (default: 30)
         {show_line}            (boolean)  show results text (default: true)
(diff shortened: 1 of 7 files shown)


jamestrew (approved):
sorry for the late response
lgtm thanks!

Issue #2768, opened 2023-11-05
`lsp_definitions` doesn't work for builtins and deps in Deno LSP


solson:
### Description

Using `require("telescope.builtin").lsp_definitions()` fails even when `vim.lsp.buf.definition()` works in the Deno LSP for a dependency or builtin (like `console.log`). See recordings in expected/actual sections.

### Neovim version

```markdown
NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1693350652
```

### Operating system and version

nixos-unstable (Linux 6.1.60)

### Telescope version / branch / rev

telescope 4522d7e3ea75ffddabdc39957168a8a7060b5df0

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()

Checking for required plugins ~
- OK plenary installed.
- OK nvim-treesitter installed.

Checking external dependencies ~
- OK rg: found ripgrep 13.0.0
- OK fd: found fd 8.7.1

===== Installed extensions ===== ~

Telescope Extension: `frecency` ~
- OK sqlite.lua installed.
- OK nvim-web-devicons installed.

Telescope Extension: `smart_history` ~
- No healthcheck provided
```

### Steps to reproduce

1. [Install Deno](https://docs.deno.com/runtime/manual/getting_started/installation) (which includes the LSP).
2. Make a scratch directory with `minimal.lua` and the following.
3. `touch deno.json` (so the LSP sees this as a project root).
4. `echo 'console.log("hello world")' > foo.ts`
5. `nvim -nu minimal.lua`
6. `:e foo.ts`
7. `:Telescope lsp_definitions` (on `console` or `log`).

Alternately, to see the successful outcome, replace step 7 with `:lua vim.lsp.buf.definition()`, but not during the same execution because `vim.lsp.buf.definition` and `lsp_definitions` change each other's outcomes if run in succession in the same Neovim session.

### Expected behavior

When I use `vim.lsp.buf.definition()` for a dependency or builtin (like `console.log`), it works correctly: [asciinema](https://asciinema.org/a/DeWzczjGwRZfM5tb0g5Q0rCaf)

### Actual behavior

But when I use `require("telescope.builtin").lsp_definitions()`, it fails: [asciinema](https://asciinema.org/a/j85nsVQn4ZuUh1bMeExjYrcyU)

```
[...]

PR #2770, opened 2023-11-07
fix(lsp): pass server resp through client handlers (1)

jamestrew:
For LSP methods/pickers:
- references
- definition
- typeDefinition
- implementation

Passes the server results through client (and neovim's) lsp handler function.
This lets those handlers deal with some of the language server specific idiosyncrasies.

Also refactors shared code between these pickers.

Other pickers/methods will need similar treatment.

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -9,80 +9,6 @@ local utils = require "telescope.utils"
 
 local lsp = {}
 
-lsp.references = function(opts)
-  local filepath = vim.api.nvim_buf_get_name(opts.bufnr)
-  local lnum = vim.api.nvim_win_get_cursor(opts.winnr)[1]
-  local params = vim.lsp.util.make_position_params(opts.winnr)
-  local include_current_line = vim.F.if_nil(opts.include_current_line, false)
-  params.context = { includeDeclaration = vim.F.if_nil(opts.include_declaration, true) }
-
-  vim.lsp.buf_request(opts.bufnr, "textDocument/references", params, function(err, result, ctx, _)
-    if err then
-      vim.api.nvim_err_writeln("Error when finding references: " .. err.message)
-      return
-    end
-
-    local locations = {}
-    if result then
-      local results = vim.lsp.util.locations_to_items(result, vim.lsp.get_client_by_id(ctx.client_id).offset_encoding)
-      if not include_current_line then
-        locations = vim.tbl_filter(function(v)
-          -- Remove current line from result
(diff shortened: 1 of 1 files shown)


solson:
Thanks for the quick PR! When I re-try my reproduction steps from https://github.com/nvim-telescope/telescope.nvim/issues/2768, I run into this error:

```
Error executing vim.schedule lua callback: ...ker/start/telescope.nvim/lua/telescope/builtin/__lsp.lua:147: attempt to index local 'location' (a nil value)
stack traceback:
        ...ker/start/telescope.nvim/lua/telescope/builtin/__lsp.lua:147: in function 'handler'
        ...eovim-unwrapped-0.9.4/share/nvim/runtime/lua/vim/lsp.lua:1393: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

Which suggests that `local location = items[1].user_data` was nil.

jamestrew:
Ah.. it's probably a neovim version issue.
It looks like I'm relying on some bleeding edge changes on nightly (https://github.com/neovim/neovim/pull/25830)

I think we're planning to bump the minimum neovim version requirement to 0.10 for telescope v0.2.0 but there's no timeline for that. Either we come up with a different approach or can't merge this until then.

I'm not sure if I'm in love with my solution anyways. This relies on the custom handlers eventually calling neovim's builtin handler or supporting a `on_list` option like the builtin one does. And I don't think this is a requirement for lsp handlers.

josephcrawfordSRH:
@jamestrew any thoughts on when this may get released?  This fixes an issue I have been having where I cannot jump to the definition for typescript dependencies which are imported via a URL.

jamestrew:
See https://github.com/nvim-telescope/telescope.nvim/issues/2768#issuecomment-2375607071

telescope doesn't handle custom handlers and I don't see any great way of doing so as the linked comment explains.

I'm still thinking about this problem but I don't think this PR is the right approach now.

josephcrawfordSRH:
@jamestrew is there a way to make telescope not handle these, I am not quite sure why gd to jump to a definition would go through telescope at all, I can see gr (references) but not jump to definition.

jamestrew:
That would be dependent on your neovim config. Somewhere you must have gd mapping to telescope.

josephcrawfordSRH:
yep you are correct, it seems LazyVim is doing this by default

https://github.com/LazyVim/LazyVim/blob/13a4a84e3485a36e64055365665a45dc82b6bf71/lua/lazyvim/plugins/extras/editor/telescope.lua#L278-L293

PR #2782, opened 2023-11-15
chore: add `pull_request` trigger on `docgen`

jamestrew:
`docgen` doesn't get ran on PRs from forks.
Not sure if this is on purpose.

### DIFF
--- a/.github/workflows/docgen.yml
+++ b/.github/workflows/docgen.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches-ignore:
       - master
+  pull_request:
+    branches:
+      - master
 
 jobs:
   build-sources:


Conni2461:
I dont know if that fixes it. because i dont get why it isnt even working.

I always work on my fork and it works for me (i know "works for me" is infamous). example: https://github.com/nvim-telescope/telescope.nvim/pull/2256

jamestrew:
I'm gonna merge this in to give it a try.
I can't manually run docgen locally anymore and it seems like the docgen workflow isn't running at all for non-member contributors.

jamestrew:
It _seems_ to trigger now but now getting permission issues...

PR #2791, opened 2023-11-18
feat: Support Line Column in file pickers

### LINKED ISSUE #2650: Add an ability to open direct location
I usually deal with direct location output from the compiler which is a path to the file plus the location marker. e.g.

```
my/directory/file.rs:24:4
```

I want to be able to paste this directly to the telescope, press enter, and appear in the direct symbol location.

> P.S. This is possible in vs code with no additional setup for example

1. Open telescope search
2. Observe nothing found
3. Remove paths 
4. type :24<CR> to go to the line 

**Additional context**
This should be primarily an input feature — strip out everything that looks like location, search, and when open file focus specific symbol. That's why I think it should be a part of the telescope core and not be solved via plugins. 

If you are interested I can open PR for this

dmtrKovalenko:
## Type of change

## Description

Fixing the original issue: When selecting files in the telescope impossible to copy paths from the compiler, you need to clear the location from the path.

Fixes #2650

This fixes the problem by stripping out the location from the path (important: only for file pickers). And using the location to set the cursor when opening the file.

There is a new option `files_picker` for a `Picker` that needs to be set to true for all file pickers, for now, it basically enables location stripping which makes no sense in another type of pickers.

Watch the video:

https://github.com/nvim-telescope/telescope.nvim/assets/16926049/1ee3b507-4af9-4d0f-abe2-3c67254dfcd9

## Implementation notice

I didn't find another way to put the location stripping because it must happen right after the input and make sure that if the location is set the prompt = input without location. But after we calculated that during selection we can set `entry.lnum` and `entry.col` to the value and reuse all the existing functionality to jump to the cursor

## How it is different from https://github.com/nvim-telescope/telescope.nvim/pull/2312

The PR seems to be abandoned but I need this functionality so much so I updated and changed it myself, made the API closer to the current entry row/linenr selection state which is used in live_grep, implemented support for the column number, added automated tests, and highlighted the line and column in the previewer.

## How Has This Been Tested?

An added test suite that covers the algorithm of stripping the location from the path to the `utils_spec.lua` 

Manual test:

1. Open telescope to find files
2. Enter a:12:1
3. Observer that files with a in the name found
4. Open the file and observe that you are in the :12:1 position

**Configuration**:

NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1699180677
* Operating system and version: MacOS 13

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -386,7 +386,7 @@ files.find_files = function(opts)
       prompt_title = "Find Files",
       __locations_input = true,
       finder = finders.new_oneshot_job(find_command, opts),
-      previewer = conf.file_previewer(opts),
+      previewer = conf.grep_previewer(opts),
       sorter = conf.file_sorter(opts),
     })
     :find()
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -384,6 +384,7 @@ files.find_files = function(opts)
   pickers
     .new(opts, {
       prompt_title = "Find Files",
+      __locations_input = true,
       finder = finders.new_oneshot_job(find_command, opts),
       previewer = conf.file_previewer(opts),
       sorter = conf.file_sorter(opts),
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -319,6 +319,8 @@ function Picker:new(opts)
     cache_picker = config.resolve_table_opts(opts.cache_picker, vim.deepcopy(config.values.cache_picker)),
 
     __scrolling_limit = tonumber(vim.F.if_nil(opts.temp__scrolling_limit, 250)),
+
+    __allow_locations_input = opts.__locations_input or false,
   }, self)
 
   obj.create_layout = opts.create_layout or config.values.create_layout or default_create_layout
@@ -626,6 +628,24 @@ function Picker:find()
       local start_time = vim.loop.hrtime()
 
       local prompt = self:_get_next_filtered_prompt()
+      if self.__allow_locations_input == true then
+        local filename, line_number, column_number = utils.__separate_file_path_location(prompt)
+
+        if line_number or column_number then
+          state.set_global_key("prompt_location", { row = line_number, col = column_number })
+          self:refresh_previewer()
+        elseif state.get_global_key "prompt_location" then
+          state.set_global_key("prompt_location", nil)
+          self:refresh_previewer()
+        end
+
+        -- it is important to continue behaving as if there is no location in prompt
+        prompt = filename
+      elseif state.get_global_key "prompt_location" then
+        -- in case new picker that does not support locations is opened clear the location
+        -- without refreshing previewer
+        state.set_global_key("prompt_location", nil)
+      end
 
       -- TODO: Entry manager should have a "bulk" setter. This can prevent a lot of redraws from display
       if self.cache_picker == false or self.cache_picker.is_cached ~= true then
@@ -1046,6 +1066,13 @@ function Picker:set_selection(row)
   end
 
   local entry = self.manager:get_entry(self:get_index(row))
+
+  local prompt_location = state.get_global_key "prompt_location"
+  if entry and prompt_location then
+    entry.lnum = prompt_location.row or 0
+    entry.col = prompt_location.col or 0
+  end
+
   state.set_global_key("selected_entry", entry)
 
   if not entry then
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -320,7 +320,7 @@ function Picker:new(opts)
 
     __scrolling_limit = tonumber(vim.F.if_nil(opts.temp__scrolling_limit, 250)),
 
-    __allow_locations_input = opts.__locations_input or false,
+    __locations_input = vim.F.if_nil(opts.__locations_input, false),
   }, self)
 
   obj.create_layout = opts.create_layout or config.values.create_layout or default_create_layout
@@ -628,7 +628,7 @@ function Picker:find()
       local start_time = vim.loop.hrtime()
 
       local prompt = self:_get_next_filtered_prompt()
-      if self.__allow_locations_input == true then
+      if self.__locations_input == true then
         local filename, line_number, column_number = utils.__separate_file_path_location(prompt)
 
         if line_number or column_number then
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -633,17 +633,14 @@ function Picker:find()
 
         if line_number or column_number then
           state.set_global_key("prompt_location", { row = line_number, col = column_number })
-          self:refresh_previewer()
         elseif state.get_global_key "prompt_location" then
           state.set_global_key("prompt_location", nil)
-          self:refresh_previewer()
         end
 
         -- it is important to continue behaving as if there is no location in prompt
         prompt = filename
       elseif state.get_global_key "prompt_location" then
         -- in case new picker that does not support locations is opened clear the location
-        -- without refreshing previewer
         state.set_global_key("prompt_location", nil)
       end
 
@@ -1130,6 +1127,7 @@ function Picker:set_selection(row)
     return
   end
 
+  self:refresh_previewer()
   if old_entry == entry and self._selection_row == row then
     return
   end
@@ -1137,8 +1135,6 @@ function Picker:set_selection(row)
   -- TODO: Get row & text in the same obj
   self._selection_entry = entry
   self._selection_row = row
-
-  self:refresh_previewer()
 
   vim.api.nvim_win_set_cursor(self.results_win, { row + 1, 0 })
 end
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -1067,7 +1067,13 @@ function Picker:set_selection(row)
   local prompt_location = state.get_global_key "prompt_location"
   if entry and prompt_location then
     entry.lnum = prompt_location.row or 0
-    entry.col = prompt_location.col or 0
+    if prompt_location.col and prompt_location.col > 0 then
+      entry.col = prompt_location.col
+      entry.colend = prompt_location.col + 1
+    else
+      entry.col = 1 -- we do + 1 here because previewer does -1
+      entry.colend = 0
+    end
   end
 
   state.set_global_key("selected_entry", entry)
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -524,10 +524,11 @@ previewers.vimgrep = defaulter(function(opts)
       local lnum, lnend = entry.lnum - 1, (entry.lnend or entry.lnum) - 1
 
       local col, colend = 0, -1
+      -- Both col delimiters should be provided for them to take effect.
+      -- This is to ensure that column range highlighting was opted in, as `col`
+      -- is already used to determine the buffer jump position elsewhere.
       if entry.col and entry.colend then
         col, colend = entry.col - 1, entry.colend - 1
-      elseif entry.col and entry.col > 0 and not entry.colend then
-        col, colend = entry.col - 1, entry.col
       end
 
       for i = lnum, lnend do
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -488,29 +488,6 @@ end
 previewers.cat = defaulter(function(opts)
   opts = opts or {}
   local cwd = opts.cwd or vim.loop.cwd()
-  local function jump_to_line(bufnr, winid)
-    pcall(vim.api.nvim_buf_clear_namespace, bufnr, ns_previewer, 0, -1)
-    local location = global_state.get_global_key "prompt_location"
-
-    if location and location.row > 0 then
-      local highlight_range = location.col and location.col > 0 and { location.col - 1, location.col } or { 0, -1 }
-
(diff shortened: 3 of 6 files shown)


lua/telescope/pickers.lua
```diff
@@ -330,9 +332,9 @@ function Picker:new(opts)
     end
     obj.previewer = obj.all_previewers[obj.current_previewer_index]
     if
-      obj.preview_title == nil
-      or #obj.all_previewers > 1
-      or opts.resumed_picker and opts.fix_preview_title ~= true
+        obj.preview_title == nil
+        or #obj.all_previewers > 1
+        or opts.resumed_picker and opts.fix_preview_title ~= true
```
dmtrKovalenko, line 337:
```suggestion
      obj.preview_title == nil
      or #obj.all_previewers > 1
      or opts.resumed_picker and opts.fix_preview_title ~= true
```

lua/telescope/pickers.lua
```diff
@@ -626,6 +628,24 @@ function Picker:find()
...
+          self:refresh_previewer()
+        end
+
+        -- it is important to continue behaving as if there is no location in prompt
+        prompt = filename
+      elseif state.get_global_key "prompt_location" then
+        -- in case new picker that does not support locations is opened clear the location
+        -- without refreshing previewer
+        state.set_global_key("prompt_location", nil)
+      end
```
Conni2461, line 648: this doesnt really address this comment: https://github.com/nvim-telescope/telescope.nvim/pull/2312#issuecomment-1479645530

i agree with tj here, that i dont really want builtin specific code in `pickers.lua`, because i dont think we should couple specific builtin code with the framework.

The framework should rather expose callbacks that allow us to do these things.

why can't we use `_on_input_filter_cb` its used in `_get_next_filtered_prompt` and we can then return the new prompt with `{ prompt = "asdf" }`

lua/telescope/pickers.lua
```diff
@@ -1046,6 +1066,13 @@ function Picker:set_selection(row)
   end
 
   local entry = self.manager:get_entry(self:get_index(row))
+
+  local prompt_location = state.get_global_key "prompt_location"
+  if entry and prompt_location then
+    entry.lnum = prompt_location.row or 0
+    entry.col = prompt_location.col or 0
+  end
```
Conni2461, line 1074: not sure we should inject that here into the entry, i'd rather we just use that prompt_location in the actions, similar to what you did with the previewer, because with this right now you could have just used the `entry.lnum` and `entry.col` in previewer code, fyi

Why do you do it like this?

dmtrKovalenko: To reuse the code of putting the cursor to the specific location, thus we will need to duplicate them but I’m find with that either

lua/telescope/pickers.lua
dmtrKovalenko: I can look into that as well

dmtrKovalenko: Hey @Conni2461 after a bit of investigation I realised why it would be a bad idea to put it into some callback — the biggest problem is the global state that we need to mutate and when the new picker is open we need to clear this state (or at least reset on reopening). I don't like an idea of stacking callbacks (we need to keep the functionality) and mutating global state inside of the stacked callbacks.

My proposal would be different. Define a new callback — `get_prompt_metadata` that will be executed in the picker and get the metadata from the prompt, then use this metadata to set the entry.lnum and entry.col on selection and also push this into the previewer.state. 

Thus we will never end up in a way when the global state is mutated. But every file-related picker will need to provide this `get_prompt_metadata` to achieve location parsing. 

If you are ok with this design please let me know — waiting for the green light.

Conni2461: hmm, let me think about it. I am also not that happy with a new kinda specific callback (cant think of another reason why we would need that)

dmtrKovalenko: I'll just put 3 things that are requirements for this feature (in my opinion): 

1. You need to mutate the prompt when the location is valid
2. You need to save the location and if it global state clear it for non supported picker
3. In my ideal world all of my pickers even external https://github.com/nvim-telescope/telescope-frecency.nvim will support this feature, or for instance I want to implement search by sapling files would support this feature, so having it as one flag option on the picker is preferred

Conni2461: please be patient. Its been 3 days (, i reviewed it, i started looking into different implementations but i am also really busy. This project is my free time and i which is generally limited to weekends.

I'll merge a solution for this feature, but i need to think about this a couple more days. I'll not rush a solution because the majority of vim users tend to run the default branch, which makes breaking changes a pain.

I think you can run your fork a couple more days because there werent any mayor commits on master anyway.

dmtrKovalenko: Sure, happy to understand that it is not lost.

Conni2461: i think i have a implementation i slighly prefer: https://github.com/nvim-telescope/telescope.nvim/commit/64fc6c58f984b17ae243da90a75f3bafd23227df

i am not entirely happy and we need some util functions so people can easily set the callback but i think that makes it less coupled inside the picker.

so its not done, but i also havent spend more than 30minutes on it.

Its pretty likely that i dont have time to look at this prior to saturday, just a heads up

dmtrKovalenko: Thank you for doing that, what confuses me in this is that the “filter_input_cb” sets totally global state which is super unclear side effect as for me. Just a thought, happy to help with any implementation you choose.

just let me know what I can do, thank you for your help on this!!

Conni2461: yeah what i dont like about your implementation is that picker now does things that are totally only related to file specific things.

the picker itself, ui, sorting, etc, has nothing to do with files it operates on lists.

> telescope.nvim is a highly extendable fuzzy finder over lists. Built on the latest awesome features from neovim core. Telescope is centered around modularity, allowing for easy customization.

its also not a global global thing. i was thinking about introducing this local_state, which gets cleaned up after shutdown and allows users to save information, that is then accessible in the different stages of the picker (preview and actions). i am not happy that the default action/previewer have knowledge about this key/value pair, thats why i think its not done yet, but i think its a step up, because the picker no longer is responsible for it.

xiantang (approved):

dmtrKovalenko:
@Conni2461 it is already a month — could I somehow help to get this merged? If you have any ideas of how I can improve the implementation please briefly write them here and I'll try to add them. 

Or do you want to simply move forward with your version https://github.com/nvim-telescope/telescope.nvim/commit/64fc6c58f984b17ae243da90a75f3bafd23227df?

Conni2461:
Okay lets merge this as an "experimental" feature ... I still wanna mention that I dont like the application Interface, and i wanna 100% rework it but this isn't possible without some mayor effort with some kinda telescope hooking system which takes too long right now, conceptional and then the actual implementation of that. (Problems summed up in this comment thread: https://github.com/nvim-telescope/telescope.nvim/pull/2791#discussion_r1399037013 and in the original proposal https://github.com/nvim-telescope/telescope.nvim/pull/2312). 

TLDR: telescope core has nothing to do with files and having the core system be tied to this is a system design decision i dont like.

I need a couple of changes that bascially mark this as experimental after that i would merge this, even though it doesnt makes me happy. Comments regarding these changes, follow below

lua/telescope/builtin/__files.lua
```diff
@@ -384,6 +384,7 @@ files.find_files = function(opts)
   pickers
     .new(opts, {
       prompt_title = "Find Files",
+      files_picker = true,
```
Conni2461, line 387: lets rename this to `__cursor_support` (i think, important thing is the `__` prefix)

lua/telescope/pickers.lua
```diff
@@ -319,6 +319,8 @@ function Picker:new(opts)
     cache_picker = config.resolve_table_opts(opts.cache_picker, vim.deepcopy(config.values.cache_picker)),
 
     __scrolling_limit = tonumber(vim.F.if_nil(opts.temp__scrolling_limit, 250)),
+
+    allow_location_input = opts.files_picker or false,
```
Conni2461, line 323: same name on both sides, and prefix with `__`

lua/telescope/builtin/__files.lua
Conni2461: or `__location_input` dont care.

lua/telescope/previewers/buffer_previewer.lua
```diff
@@ -489,6 +488,29 @@ end
 previewers.cat = defaulter(function(opts)
   opts = opts or {}
   local cwd = opts.cwd or vim.loop.cwd()
+  local function jump_to_line(bufnr, winid)
```
Conni2461, line 491: we inject `entry` already with `lnum` and `col`. why not just use a picker that already has support for jumping to the line and highlighing the line. aka grep_previewer, isnt that good enough?

lua/telescope/utils.lua
```diff
@@ -567,4 +567,45 @@ utils.list_find = function(func, list)
...
+--- @param path string
+utils.separate_file_path_location = function(path)
+  local location_numbers = {}
+  -- Split the 2 last `:` separated parts and if they are numbers treat
+  -- as line and columnd numbers
+  for i = #path, 1, -1 do
+    if path:sub(i, i) == ":" then
+      -- If this is the last `:` we ignore it for a case when user
+      -- just entered it with intention to enter a line or column number
+      -- to not display "No results" immediately.
```
Conni2461, line 581: i dont think we need these comments

lua/telescope/utils.lua
```diff
@@ -567,4 +567,45 @@ utils.list_find = function(func, list)
   end
 end
 
+--- Takes the path and parses optional cursor location `$file:$line:$column`
+--- If line or column not present `0` returne returned.
+--- @param path string
+utils.separate_file_path_location = function(path)
```
Conni2461, line 573: can we also prefix this function with `__`, i am happy to have it as a utils function because then its tested but i am not sure i want extension authors to use that function

lua/telescope/previewers/buffer_previewer.lua
dmtrKovalenko: I am here more afraid about potential issues and int the behavior because I might miss because the code is somewhat different and has some handlers that I am not really sure I understand 100%. 

If you 100% sure it will work we can try that out as well

dmtrKovalenko:
@Conni2461 I addressed all the changes you requested except the grep previewer because I don't feel 100% sure it is working in all edge casees

Conni2461:
I pushed 3 more commits, addressed https://github.com/nvim-telescope/telescope.nvim/pull/2791#discussion_r1438706606 and https://github.com/nvim-telescope/telescope.nvim/pull/2791#discussion_r1438707304

can you retest if everything still works for you? because i havent found more issues and it should be mergeable now.

dmtrKovalenko:
I pulled the changes and everything works like a charm.  **Very big thanks for your help on this.**.

I have another small question related to this feature: would it be ok if I'll make a PR to https://github.com/nvim-telescope/telescope-frecency.nvim to add the `__locations_input` option and to make it also handle the locations? I recently tried to switch from the default pickers and would be happy to make it integration. Can we make this option public for some 3d party pickers?

Conni2461:
I found a regression and fix that one today and pushed a fix so it actually be good to go, ill merge tomorrow morning.

Regarding frecency, i am not the current maintainer of that plugin so i am not sure delphinus wants this, this is up to him, but if they dont want this, you should be able to just enable this in your config. Because the opts are usually passed into `pickers.new` like this `pickers.new(opts, default_stuff)`. `__locations_input` doesnt have to be part of `default_stuff`, for it to work

I have not tested this but this should probably work
```lua
require("telescope").extensions.frecency.frecency{
  __locations_input = true,
  previewer = require("telescope.config").values.grep_previewer({}),

}
```

dmtrKovalenko:
That's wonderful!

Conni2461:
Thanks for the contribution :)

xiantang:
great feature. I updated my telescope instantly.

Ryel:
Thank you @dmtrKovalenko and @Conni2461 !!!

josephcrawfordSRH:
How does one go about enabling this, is it in the release that is installed in lazyvim or would I have to install a specific pre-release version?  Highly interested in being able to jump to line numbers when jumping files, awesome work getting this in place!

ctretyak:
how to turn it on? It doesn't work by default in 0.1.8 and cannot find anything in help

jamestrew:
@ctretyak https://github.com/nvim-telescope/telescope.nvim/pull/2791#issuecomment-1962187594

josephcrawfordSRH:
@jamestrew Is the same thing possible with telescope to say jump to function defenition so instead of doing `:21` to jump to the line can you also do `@methodName`?  This is something that I miss from my vscode days but still not enough for me to switch back :)

jamestrew:
@josephcrawfordSRH 
see this https://github.com/nvim-telescope/telescope.nvim/issues/1042

I think adding `@` support to `find_files` could makes sense but there isn't any plan to do so currently.

I'm also little wary about turning `find_files` into a vscode/sublime style catch all finder.
We've added support for jumping to line/col of a file with this PR, I'm now seeing more requests for jumping directly to symbols with `@`.

If we add `@` support, next people will ask for `#` and `>` (workspace symbol & commands respectively) maybe :sweat_smile: 

I'm not saying these are bad ideas. But at some point, it's probably a separate new picker.

josephcrawfordSRH:
@jamestrew thanks for getting back to me, sorry for my delay in response I was on PTO.  I completely understand your points above and was more curious if I was missing something but it sounds like I am not.  I do hope that something like this could be implemented in the future either in this picker or a different picker :)

PR #2808, opened 2023-12-10
make the 0 args command could be repeated by `@:`

gh-liu:
make the 0-args command could be repeated by `@:`


jamestrew (approved):
lgtm thanks

Issue #2816, opened 2023-12-24
FEAT REQUEST: Only store non-empty prompts in the cache (for last search)


krshrimali:
**Is your feature request related to a problem? Please describe.**
An example flow to show-case the problem:

- I searched for something using telescope. It's stored in the cache_picker.
- I opened the telescope search again (not the last search, just something like `find_files`).
- I didn't want to search anything, as I wanted to open last search but I mistakenly opened `find_files` picker.
- I close the prompt.
- I open last search prompt expecting it to show me my last valid search (definition of valid is to be discussed)
- But it shows me the last empty search I did.

This is a problem for _some_, but not for all probably if they really wanted to see the last search.

**Describe the solution you'd like**

- **Solution - 1: Have a flag to change the behaviour of last search**

  In order to not break the existing flow, I would like this to be an option in the config: (I am really bad at phrasing, so please take the following as an example only for config change). If the flag is set to `true`, the last search picker would only show the last non-empty prompt and not literally the last prompt.

```
cache_non_empty_only = true/false
```

- **Solution - 2: Have another option to just show last non empty search**

  This would be like, adding another option to use this non-empty picker, where if they do `Telescope last_non_empty_search` - it would show the last non-empty prompt.

**Describe alternatives you've considered**
I've not really been able to consider any alternatives for this.

**Additional context**
- There is a PR I made to show the change (though, haven't added the config change option to avoid breaking change) to be made to support only non-empty prompt in the last search. PR link: https://github.com/nvim-telescope/telescope.nvim/pull/2817
- The PR is only to show and discuss my idea, and I would totally understand if the maintainers decide against this feature request, happy to close the PR post discussion. :)

PR #2817, opened 2023-12-24
Only store into cache picker if the prompt is non empty; support last non-empty search instead

krshrimali:
# Description

Attempts to fix https://github.com/nvim-telescope/telescope.nvim/issues/2816, partially though - haven't added a config flag to avoid breaking change.

Please note that this PR is for discussion first, and I can work on testing/cleaning up later once the issue is addressed and approved to be worked upon.

## Type of change

# How Has This Been Tested?

It's been tested locally with the flow given in the issue.

**Configuration**:

# Checklist:

### DIFF
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -506,7 +506,7 @@ append(
   {
     num_pickers = 1,
     limit_entries = 1000,
-    ignore_empty_prompt = false,
+    ignore_empty_prompt = true,
   },
   [[
     This field handles the configuration for picker caching.
@@ -528,9 +528,9 @@ append(
       - limit_entries:        The amount of entries that will be saved for each
                               picker.
                               Default: 1000
-      - ignore_empty_prompt:  If true, the picker will not be cached if the
+      - ignore_empty_prompt:  If false, the picker will be cached even if the
                               prompt is empty (i.e., no text has been typed at the time of closing the prompt).
-                              Default: false
+                              Default: true
     ]]
 )
 
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -520,17 +520,18 @@ append(
 
     Fields:
       - num_pickers:          The number of pickers to be cached.
-                              Set to -1 to preserve all pickers of your session.
-                              If passed to a picker, the cached pickers with
-                              indices larger than `cache_picker.num_pickers` will
-                              be cleared.
+                              Set to -1 to preserve all pickers of your
+                              session. If passed to a picker, the cached
+                              pickers with indices larger than
+                              `cache_picker.num_pickers` will be cleared.
                               Default: 1
-      - limit_entries:        The amount of entries that will be saved for each
-                              picker.
+      - limit_entries:        The amount of entries that will be saved for
+                              each picker.
                               Default: 1000
-      - ignore_empty_prompt:  If false, the picker will be cached even if the
-                              prompt is empty (i.e., no text has been typed at the time of closing the prompt).
-                              Default: true
+      - ignore_empty_prompt:  If true, the picker will not be cached if
+                              the prompt is empty (i.e., no text has been
+                              typed at the time of closing the prompt).
+                              Default: false
     ]]
 )
 
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -506,6 +506,7 @@ append(
   {
     num_pickers = 1,
     limit_entries = 1000,
+    ignore_empty_prompt = false,
   },
   [[
     This field handles the configuration for picker caching.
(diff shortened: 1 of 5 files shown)


lua/telescope/pickers.lua
```diff
@@ -1544,11 +1544,17 @@ function pickers.on_close_prompt(prompt_bufnr)
...
       picker.cache_picker.selection_row = picker._selection_row
-      picker.cache_picker.cached_prompt = picker:_get_prompt()
+
+      -- only cache if prompt is not empty
+      if not (curr_prompt == nil or curr_prompt == "") then
+        picker.cache_picker.cached_prompt = curr_prompt
+        table.insert(cached_pickers, 1, picker)
+      end
+
       picker.cache_picker.is_cached = true
```
jamestrew, line 1557: Should these `picker.*` values only set inside the if statement?

jamestrew (commented):
I think I'm in agreeance about only caching non-empty-prompt pickers but I'm not personally a regular user of telescope's history/resume functionality and I don't have a lot of familiarity with this part of the code either. I'll need to take some time or have others chime in.

lua/telescope/pickers.lua
krshrimali: Good catch! Done :) Also tested locally, it works as expected. Could you please help approve workflows for this PR so that I can check if there is any workflow failing? Thank you!

krshrimali:
Hi, @jamestrew - Apologies for the ping, just wondering, would you know the right person to ping for this PR to be reviewed? 🤔 It would help if I could get some thoughts on this, and get this landed in master if everything sounds good to the reviewers. 💯

jamestrew:
@Conni2461 any chance you can take a look at this?
otherwise I will try to get around to it eventually

Conni2461:
i'll think about this, let me come back to you in the next 2 days

krshrimali:
Hi, @Conni2461 - Happy new year! I understand this PR might have been lost because of holidays and other PRs, but could you help take a look if you get some time? Thank you!

Conni2461:
yeah thanks for the ping ill try to take a look at it this weekend. I usually cant make enough time throughout the week for "high profile" changes like this

krshrimali:
> yeah thanks for the ping ill try to take a look at it this weekend. I usually cant make enough time throughout the week for "high profile" changes like this

Ah, not sure if it was a subtle pun ("high profile"), but anw, I'm not in rush with it and was a gentle reminder. I understand it's difficult to get to all the PRs in limited time. Thanks!

krshrimali:
@Conni2461 / @jamestrew - Friendly ping! :) I understand it's not high profile, but I hope y'all don't mind a friendly ping as well. Thank you!

Conni2461:
Yeah, i think the code looks good :+1: but i think it doesnt hurt to add the config value. I would argue that true as a default (so meaning a breaking change here) is good because i link having this as default (only store if prompt is not empty) but some might disagree.

So because making this configuration doesnt add a whole lot of more lines we can do that.

that means, adding a new key to `cache_picker` in `config.lua:505` adding a comment in the documentation regarding this and then using that config value in the new if statement.

Thanks for your contribution :)

lua/telescope/config.lua
```diff
@@ -506,6 +506,7 @@ append(
   {
     num_pickers = 1,
     limit_entries = 1000,
+    ignore_empty_prompt = false,
```
krshrimali, line 509: @Conni2461 - Apologies if I got your comment wrong before, but I set it to `false` to avoid any breaking changes. (Though, I agree that having this as `true` would make more sense, but this might differ for others given it might be used by many the existing way already).

lua/telescope/pickers.lua
```diff
@@ -1544,11 +1544,24 @@ function pickers.on_close_prompt(prompt_bufnr)
...
+          table.insert(cached_pickers, 1, picker)
+
+          picker.cache_picker.is_cached = true
+        end
+      else
+        picker.cache_picker.cached_prompt = curr_prompt
+        table.insert(cached_pickers, 1, picker)
+
+        picker.cache_picker.is_cached = true
+      end
```
krshrimali, line 1564: As much as I hate the code duplication here, and I did consider the following:

```lua
if not picker.cache_picker.ignore_empty_prompt or (curr_prompt and curr_prompt ~= "") then
    picker.cache_picker.cached_prompt = curr_prompt
    table.insert(cached_pickers, 1, picker)
    picker.cache_picker.is_cached = true
end
```

But the above^ snippet was very hard to read/understand as too many conditions in an if condition can make it harder to debug later on in case of any bug. I'm open to suggestions here but thought I'll share what I have considered with this piece of code. :)

krshrimali:
@Conni2461 - Thank you so much for taking a look, I appreciate your feedback and thoughts around it. Have made the required changes, whenever you get some time, could you please help take a look? I've tested it locally and things worked fine but please feel free to test it yourself as well if you'd like. :) Thank you again!

lua/telescope/pickers.lua
Conni2461: i dont think `not picker.cache_picker.ignore_empty_prompt or (curr_prompt and curr_prompt ~= "")` is unreadable. i actually would argue its more readable than having the duplicated code

lua/telescope/config.lua
Conni2461: yeah but defaults should refect what we think are sensible defaults. so i prefer true, even its a breaking change. we have a release branch for exactly that.

Conni2461:
i found a usecase we might wanna consider also caching even with empty prompt. e.g. if we have done a live_grep search and then used the refined_action (`<ctrl-space>`) we get an empty prompt but if the user now closes telescope we still might wanna cache it because it still have live_grep filtered results. Does that make sense?

Marskey:
> i found a usecase we might wanna consider also caching even with empty prompt. e.g. if we have done a live_grep search and then used the refined_action (`<ctrl-space>`) we get an empty prompt but if the user now closes telescope we still might wanna cache it because it still have live_grep filtered results. Does that make sense?

this confused me, in cache picker it is shows 'live_grep' but actually represents  a static finder that stored the result of live_grep last query.

so maybe refined_action should not be store in cache

Conni2461:
we should think about refined actions and caching in a separate issue because the results are cached once the picker is closed, if you have refined your input then the previous list is already dropped and now longer accessible, refine here acts an replacement/filter of the list of results found.

Ill merge this feature as it is and open a new Issue, to discuss this further

Conni2461:
thanks for the contribution :D 

new issues about refine + caching https://github.com/nvim-telescope/telescope.nvim/issues/2945

PR #2818, opened 2023-12-24
Also add untracked files in git viewer

Norlock:
# Description

If you add new files to your project you have to track them first before telescope shows them. This is incovenient so I added a flag --others to display those files immediately.

## Type of change

- New feature (non-breaking change which adds functionality), (or bug fix)

# How Has This Been Tested?

I added files and checked if they were immediately visible

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__git.lua
+++ b/lua/telescope/builtin/__git.lua
@@ -41,7 +41,8 @@ git.files = function(opts)
   -- By creating the entry maker after the cwd options,
   -- we ensure the maker uses the cwd options when being created.
   opts.entry_maker = vim.F.if_nil(opts.entry_maker, make_entry.gen_from_file(opts))
-  opts.git_command = vim.F.if_nil(opts.git_command, git_command({ "ls-files", "--exclude-standard", "--cached" }, opts))
+  opts.git_command =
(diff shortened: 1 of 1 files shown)


jamestrew:
Wouldn't it be better to just use `find_files` if you also want to see untracked files? rg/fd both respect gitignore by default and I'm assuming both would be faster. Maybe I'm missing something.

Either way, there exists a `git_command` option for `git_files` to let you set arbitrary flags and it will let you do this.
```lua
require("telescope").setup({
  pickers = {
    git_files = {
      git_command = { "ls-files", "--exclude-standard", "--cached", "--others" },
    },
  },
})
```

Norlock:
No i don't want to use find files because sometimes I use files that are hidden like .github/workflow or other files which I want to lookup with git files. However I can't think of any circumstance where I would be bothered by seeing the files I recently added.

jamestrew:
You can still just use `find_files` with rg for much faster results:
`require("telescope.builtin").find_files({ find_command = { "rg", "--files", "--color", "never", "--hidden", "-g", "!.git" } })`.
This will respect gitignore but still show hidden files (except .git files).

If there's other use cases for which you want to stick with git_files, you can use that `git_command` option but I don't want to touch the default git ls-files options too lightly.

PR #2819, opened 2023-12-24
Close which key on prompt close

smackesey:
# Description

Add autocmd to make `which_key` window close on prompt exit.

Currently `actions.which_key` supports a `close_with_action` option (default true). When this is set, the `which_key` window will close after any Telescope action is triggered. This makes sense. However, when it is false, the `which_key` window remains open even after Telescope
closes. This seems like a bug.

This PR fixes this by setting an autocommand when `close_with_action` is false to close on prompt exit.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Manually tested this by running `Telescope` and triggering `which_key` with `close_with_action` set to both true and false, observed expected behavior in both cases.

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1394,27 +1394,32 @@ builtin.buffers({opts})                          *telescope.builtin.buffers()*
         {opts} (table)  options to pass to the picker
 
     Options: ~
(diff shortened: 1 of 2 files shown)


jamestrew (approved):
lgtm thanks!

Issue #2761, opened 2023-10-30
Better preview for built-in diagnostics picker


filakhtov:
**Is your feature request related to a problem? Please describe.**
At the moment the built-in diagnostics picker displays a short problem description in the results window and a source code of the file where the problem occurred in the previewer. This has a few limitations:

- only the first line of the diagnostics message is displayed in the results window
- when a buffer has unsaved changes, the preview window can’t accurately display the source of the problem either

This can be seen well on the following two screenshots respectively:

![image](https://github.com/nvim-telescope/telescope.nvim/assets/5741553/4d97a29a-8c22-473a-82a2-0a9530d7362d)
the source of the error is located on the line that is way beyond the buffer that is displayed.

and

![image](https://github.com/nvim-telescope/telescope.nvim/assets/5741553/305f6098-5628-4727-adbd-f0a3f58f95c3)
 Only the first line of the diagnostics is visible, out of the:
```
error: expected identifier, found `<eof>`
  --> src/lib.rs:13:5
   |
13 | pub fn
   |     ^^ expected identifier


error: could not compile `etl` (lib) due to previous error
```
which is rather inconvenient.

**Describe the solution you'd like**
I would like to propose an idea of displaying a full diagnostic message in the preview window, instead of the contents of the file where the problem occurs. I am happy to contribute, but would need a bit of guidance as to the best way of implementing this.

**Describe alternatives you've considered**
- [lsp_lines](https://git.sr.ht/~whynothugo/lsp_lines.nvim), but it makes things look ugly and consuming a lot of vertical space at all times in the buffer area itself;
- Setting higher limit of the message width for telescope results, but this consumes a lot of space in the results window and doesn't really help in quick at a glance navigation;
- [trouble.nvim](https://github.com/folke/trouble.nvim), but that's yet another plugin with a lot of the same things I can get from Telescope;

**Additional context**
Open to any other thoughts and idea on this. Does everyone else thinks this is actually feasible?

rudiejd:
You can actually do this right now in your config:

```lua
        local telescope = require('telescope')
        telescope.setup({
            defaults = {
                qflist_previewer =  function (opts)
                    return require('telescope.previewers').new_buffer_previewer {
                            title = "Diagnostics",
                            dyn_title = function(_, entry)
                                return entry.title
                            end,

                            get_buffer_by_name = function(_, entry)
                                return "diagnostics_" .. tostring(entry.nr)
                            end,

[...]

PR #2820, opened 2023-12-24
feat(builtin.diagnostics): add previewer option to diagnostics picker

rudiejd:
# Description

https://github.com/nvim-telescope/telescope.nvim/issues/2761#issuecomment-1868556399

Fixes  https://github.com/nvim-telescope/telescope.nvim/issues/2761

One question here: are there any in-flight changes to add the previewer for each picker to the pickers configuration? If not, I think that would be a better approach than doing a one-off fix here, and I'll modify this PR to do that. 

## Type of change

# How Has This Been Tested?

- Tested locally by calling as follows

```lua        
vim.keymap.set('n', '<leader>px', function()
            return builtin.diagnostics({
                line_width = "full",
                previewer = require('telescope.previewers').new_buffer_previewer {
                    title = "Diagnostics",
                    dyn_title = function(_, entry)
                        return entry.title
                    end,

                    get_buffer_by_name = function(_, entry)
                        return "diagnostics_" .. tostring(entry.nr)
                    end,

                    define_preview = function(self, entry)
                        vim.api.nvim_buf_set_lines(self.state.bufnr, 0, -1, false, { entry.text })
                    end,
                }
            })
        end, {})
```
**Configuration**:
```
NVIM v0.10.0-dev-1746+g5651c1ff2
Build type: RelWithDebInfo
LuaJIT 2.1.1700008891
```
```
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.3 LTS
Release:        22.04
Codename:       jammy
```
# Checklist:

### DIFF
--- a/lua/telescope/builtin/__diagnostics.lua
+++ b/lua/telescope/builtin/__diagnostics.lua
@@ -165,7 +165,7 @@ diagnostics.get = function(opts)
         results = locations,
         entry_maker = opts.entry_maker or make_entry.gen_from_diagnostics(opts),
       },
-      previewer = conf.qflist_previewer(opts),
+      previewer = opts.previewer or conf.qflist_previewer(opts),
       sorter = conf.prefilter_sorter {
         tag = "type",
         sorter = conf.generic_sorter(opts),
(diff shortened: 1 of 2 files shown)


jamestrew:
Thanks for looking into this but I don't believe this is necessary.
See my comment in the related issue: https://github.com/nvim-telescope/telescope.nvim/issues/2761#issuecomment-1869184027

PR #2841, opened 2024-01-02
feat(builtin.files): Add support for 'bfs' as find_command

cryptomilk:
See https://tavianator.com/2023/bfs_3.0.html

# Description

This adds support for bfs, which currently is the fastest 'find' implementation.

## Type of change

# How Has This Been Tested?

Install 'bfs' on your system and run `builtin.find_files`. You can run find_files on a directory with a lot of files like your home directory and in other terminal check if it executes bfs, like:

```
$ ps aufx | grep -A3 nvim
asn      31995  0.0  0.0  11004  8256 pts/1    Sl+  19:41   0:00  |   |   \_ /usr/bin/nvim
asn      31998  4.1  0.7 488616 485288 ?       Rsl  19:41   0:03  |   |       \_ /usr/bin/nvim --embed
asn      32311 65.1  0.0 194172  9880 ?        Sl   19:43   0:00  |   |           \_ bfs -type f -nocolor
```

**Configuration**:
* Operating system and version: openSUSE Tumbleweed

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -268,6 +268,9 @@ files.find_files = function(opts)
         return opts.find_command(opts)
       end
       return opts.find_command
+    elseif 1 == vim.fn.executable "bfs" then
+      -- Use https://github.com/tavianator/bfs
+      return { "bfs", "-type", "f", "-nocolor" }
     elseif 1 == vim.fn.executable "rg" then
       return { "rg", "--files", "--color", "never" }
     elseif 1 == vim.fn.executable "fd" then
--- a/lua/telescope/health.lua
+++ b/lua/telescope/health.lua
@@ -21,7 +21,18 @@ local optional_dependencies = {
     },
   },
   {
-    finder_name = "find-files",
+    finder_name = "find-files (bfs)",
+    package = {
+      {
+        name = "bfs",
+        binaries = { "bfs" },
+        url = "[tavianator/bfs](https://github.com/tavianator/bfs)",
+        optional = true,
+      },
+    },
+  },
+  {
+    finder_name = "find-files (fd)",
     package = {
       {
         name = "fd",


jamestrew:
I'm not sure about adding builtin support for `bfs` into telescope at this time. I think for the vast majority of people, this would just be an additional `executable` check.

We'd need to incorporate it into the other options for `find_files` here as well (which could probably use a refactor)
https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/builtin/__files.lua#L306-L376

Happy to reconsider at a future date and curious to see where this project goes but my thought is I'd probably like to see windows support and more adoption/battle testing before adding it as the go-to find command.

That said, you're welcome to use `bfs` yourself for `find_files`. 
```lua
require('telescope').setup {
  -- your defaults
  pickers = {
    find_files = {
      find_command = { "bfs", "-type", "f", "-nocolor" }
    }
  }
}
```

cryptomilk:
I've tried setting it via the config and it doesn't work. Looking at the code, you can only set it via `opts` passed to `find_files()` it doesn't seem to evaluate the config.

jamestrew:
That snippet I shared works for me. Any options set in `.setup({ ..., pickers = { ... } })` gets passed as `opts` to those pickers. I can help debugging if you wanna share your telescope config.

But yeah you can also just do `require('telescope.builtin').find_files({ find_command = { "bfs", "-type", "f", "-nocolor" }})` when you call it.

cryptomilk:
My config is here: https://git.cryptomilk.org/users/asn/dotfiles.git/tree/nvim/.config/nvim

However looking at the code, I can't see how the config you mention gets picked up. The find_command is supplied to the picker via:

`finder = finders.new_oneshot_job(find_command, opts),`

Searching the code for `find_command` doesn't show anything that would indicate, that `config.pickers.find_files.find_command` is read. I'm looking at the master branch of telescope.nvim.

Conni2461:
it gets applied here: https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/builtin/init.lua#L545-L593

It is also described here on how to define picker default configs: `:help telescope.setup`

just make sure its not parts of the `defaults` table. This is the important part

```lua
require('telescope').setup{
  defaults = {
    -- config_key = value,
    -- ..
  },
  pickers = {
    -- picker_name = {
    --   ...
    -- }
    -- Now the picker_config_key will be applied every time you call this
    -- builtin picker
  },
  extensions = {
    -- Your extension configuration goes here:
    -- extension_name = {
    --   extension_config_key = value,
    -- }
    -- please take a look at the readme of the extension you want to configure
  }
}
```

Issue #2849, opened 2024-01-05
Wrong cursor position after actions.close


Ajnasz:
### Description

The cursor position is off by 1 character after calling `actions.close()` https://github.com/nvim-telescope/telescope.nvim/blob/87e92ea31b2b61d45ad044cf7b2d9b66dad2a618/lua/telescope/actions/init.lua#L389

It causes issue when we try to change the text at the cursor after an entry has been selected in treesitter.

My suggested solution is to remove the `+ 1` from the line above.

### Neovim version

```markdown
NVIM v0.10.0-dev-2037+ge09adfdcf
Build type: RelWithDebInfo
LuaJIT 2.1.1703358377
```

### Operating system and version

Debian Bookworm

### Telescope version / branch / rev

master

### checkhealth telescope

```markdown
telescope: health#telescope#check

Checking for required plugins ~
- OK plenary installed.
- WARNING nvim-treesitter not found. (Required for `:Telescope treesitter`.)

Checking external dependencies ~
- OK rg: found ripgrep 13.0.0
- WARNING fd: not found. Install [sharkdp/fd](https://github.com/sharkdp/fd) for extended capabilities
```

### Steps to reproduce

Open nvim with the provided minimal config

Add some text to the buffer, eg: `##`

Place the cursor between the two characters (`#|#`), then type the `:OpenPicker`, then select the "Insert text" entry by pressing enter.

### Expected behavior

I expected to add the `inserted text` between the two characters, so to have `#inserted text#` text in the buffer

### Actual behavior

The text added after the cursor, `##inserted text`

If you use the user command `:InserthTheText`, it will add the text to the correct place.

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          { 'nvim-telescope/telescope-fzf-native.nvim', run = 'make' },
        },
      },
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
[...]

Issue #1366, opened 2021-10-21
Opening file with telescope messes with last position jump on `BufEnter`


jesseleite:
### Description

For those of us using Neovim's recommended autocmd for last position jump on `BufEnter` (see `:help last-position-jump`), Neovim will restore last cursor position when re-opening a file.

This works perfectly when opening a file using `:edit somefile`. However, if I open `somefile` using a Telescope fuzzy finder, it will open the file and move the cursor one character to the left.

I never noticed this before, but it's messing with a custom autocmd I'm trying to write where I automatically highlight some text on BufEnter.

### Neovim version

NVIM v0.5.1
Build type: Release
LuaJIT 2.1.0-beta3
Compilation: clang -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -DNDEBUG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/tmp/neovim-20210927-1441-demfsz/neovim-0.5.1/build/config -I/tmp/neovim-20210927-1441-demfsz/neovim-0.5.1/src -I/usr/local/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX11.sdk/usr/include -I/usr/local/opt/gettext/include -I/tmp/neovim-20210927-1441-demfsz/neovim-0.5.1/build/src/nvim/auto -I/tmp/neovim-20210927-1441-demfsz/neovim-0.5.1/build/include
Compiled by brew@BigSur

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/Cellar/neovim/0.5.1/share/nvim"

Run :checkhealth for more info

### Operating system and version

macOS 11.2.1

### checkhealth telescope

```markdown
health#telescope#check
========================================================================
## Checking for required plugins
  - OK: plenary installed.
[...]

PR #2850, opened 2024-01-06
fix: restore cursor to correct position on picker `close`

### LINKED ISSUE #2319: strange cursor move when telescope closes
### Description

When telescope closes, the cursor will strangely move right by 1 column. That will influence some other plugins like https://github.com/AckslD/nvim-neoclip.lua . Neoclip uses telescope to provide a selection of yank history, and paste the selected item where it launched. This strange move makes neoclip paste to a wrong position.(https://github.com/AckslD/nvim-neoclip.lua/issues/96).
Also, if you launch `require('telescope.builtin').help_tag`, choose one and close the help window, the cursor will move right.

### Neovim version

```markdown
NVIM v0.8.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "/usr/share/nvim"

```

archlinux

master

### checkhealth telescope

```markdown
  - OK: plenary installed.

  - OK: fd: found fd 8.6.0

## Telescope Extension: `fzf`
  - OK: lib working as expected
  - OK: file_sorter correctly configured
  - OK: generic_sorter correctly configured

## Telescope Extension: `macroscope`

## Telescope Extension: `neoclip`
```

### Steps to reproduce

- nvim -nu minimal.lua
-  type something
- place cursor to somewhere you remember
- :Telescope help_tags, and choose any
- close help window
- cursor will move right

### Expected behavior

_No response_

### Actual behavior

described above

### Minimal config

```Lua
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        requires = {
          'nvim-lua/plenary.nvim',
        },
      },
    },
    config = {
      package_root = package_root,
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
end
end
load_plugins()
require('packer').sync()
```

jamestrew:
I *think* we have several factors contributing to flaky (off by one error) behaviors in the cursor position on picker close.

Mainly, I think we should be scheduling `nvim_win_set_cursor`. Without doing so, we've had to "arbitrarily" +1 on the columns to compensate: https://github.com/nvim-telescope/telescope.nvim/blob/87e92ea31b2b61d45ad044cf7b2d9b66dad2a618/lua/telescope/actions/set.lua#L214-L216

This is made worse by a couple of other factors

1. ripgrep gives columns in 1-index whereas `nvim_win_set_cursor` expects 0-index.
This wasn't previously an issue probably due to not scheduling `nvim_win_set_cursor` but I think we were just depending on UB.
2. not accounting for original mode on `select_default` (like this https://github.com/nvim-telescope/telescope.nvim/blob/87e92ea31b2b61d45ad044cf7b2d9b66dad2a618/lua/telescope/actions/init.lua#L388-L390)

I recognize this is a very sensitive part of our code that we've had repeat issues with so I still need to do some testing and verify this closes a few issues without any obvious or previously experienced regressions

closes #2319 and semi-related https://github.com/nvim-telescope/telescope.nvim/issues/2319#issuecomment-1806854277
closes https://github.com/nvim-telescope/telescope.nvim/issues/2849

- [ ] check 0/1 index of columns in lsp/treesitter/etc pickers

TODO: related?
https://github.com/nvim-telescope/telescope.nvim/issues/1366
https://github.com/nvim-telescope/telescope.nvim/issues/559


Conni2461:
let me run this for a couple of days before merging :D but thanks for addressing this issue :)

Conni2461:
i rebased the branch and continue to run it till friday, if there arent any issue till then, i think this is good to go thanks :)

jamestrew:
I found an regression with `:Telescope help_tags` related to opening an existing buffer at a different location.

1. `:Telescope help_tags` -> search for `bufhidden` -> this opens the `options.txt` doc buffer
2. `:Telescope help_tags` -> search for `hidden` -> same `options.txt` but doesn't move, still on `bufhidden`

jamestrew:
> I found an regression with `:Telescope help_tags` related to opening an existing buffer at a different location.
> 
>     1. `:Telescope help_tags` -> search for `bufhidden` -> this opens the `options.txt` doc buffer
> 
>     2. `:Telescope help_tags` -> search for `hidden` -> same `options.txt` but doesn't move, still on `bufhidden`

The problem is that we're scheduling cursor position inside `actions.close()` so when a picker calls `actions.close()` then does something that moves the cursor under the current window, the previously scheduled cursor position moves the cursor again.

That why I had to schedule this so it happens after the cursor position is reset
https://github.com/nvim-telescope/telescope.nvim/pull/2850/files#diff-f192865bab9995c5a69e48b85b5d4534094c885a1556b10bb1e9407b0e0570d0R387-R390

For this `help_tags` issue, I can fix it by doing the same by scheduling this
https://github.com/nvim-telescope/telescope.nvim/blob/da8b3d485975a8727bea127518b65c980521ae22/lua/telescope/builtin/__internal.lua#L778-L784

Ideally, we just defer/schedule everything after `actions.close()` to happen after cursor position is set.
I'll try that since I don't like the idea of throwing `vim.schedule` case by case on things after `actions.close()`. This really puts the responsibility on each picker actions way too much.

jamestrew:
Yeah idno how to cleanly fix this without API changes and big refactors.

PR #2856, opened 2024-01-10
chore: add type annotations

jamestrew:
trying something out.
I like types and I think it can help with our documentation deficiencies as well (eg. confusions around why you can pass `default_text` to `find_files` and why it's not documented, where to document it, etc, but this might need some enhancements to tree-sitter-lua since the newer luals luacats annotation support is limited).

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -797,12 +797,8 @@ options you want to use. Here's an example with the live_grep picker:
   })
 <
 
-builtin.live_grep({opts})                      *telescope.builtin.live_grep()*
-    Search for a string and get results live as you type, respects .gitignore
-
-
-    Parameters: ~
-        {opts} (table)  options to pass to the picker
+TelescopeLiveGrepOpts                                  *TelescopeLiveGrepOpts*
+    TelescopeBasePickerOpts
 
     Options: ~
         {cwd}                 (string)          root dir to search from
@@ -812,7 +808,7 @@ builtin.live_grep({opts})                      *telescope.builtin.live_grep()*
         {grep_open_files}     (boolean)         if true, restrict search to
                                                 open files only, mutually
                                                 exclusive with `search_dirs`
-        {search_dirs}         (table)           directory/directories/files to
+        {search_dirs}         (string[])        directory/directories/files to
                                                 search, mutually exclusive
                                                 with `grep_open_files`
         {glob_pattern}        (string|table)    argument to be used with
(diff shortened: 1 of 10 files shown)


lua/telescope/pickers.lua
```diff
@@ -215,13 +215,107 @@ local pickers = {}
...
+---@field create_layout fun()?
+---@field get_window_options fun()?
+---@field resumed_picker boolean?
+---@field fix_preview_title boolean?
+---@field private __hide_previewer boolean?
+---@field on_complete ((fun(picker: TelescopeBasePicker): nil)[])?
+
 --- Picker is the main UI that shows up to interact w/ your results.
 -- Takes a filter & a previewer
+---@class TelescopeBasePicker
```
jamestrew, line 269: this feels pretty ridiculous lol

doc/telescope.txt
```diff
@@ -797,12 +797,8 @@ options you want to use. Here's an example with the live_grep picker:
...
 <
 
-builtin.live_grep({opts})                      *telescope.builtin.live_grep()*
-    Search for a string and get results live as you type, respects .gitignore
-
-
-    Parameters: ~
-        {opts} (table)  options to pass to the picker
+TelescopeLiveGrepOpts                                  *TelescopeLiveGrepOpts*
+    TelescopeBasePickerOpts
```
jamestrew, line 801: don't like this

jamestrew:
@Conni2461 what do you think?

jamestrew:
> they aren't compatible with emmylua annotations which are used within docgen (its uses a lua treesitter parser with emmylua support).

Yeah I actually made an issue on tree-sitter-lua about (me) potentially adding or converting the ts grammar to support luals annotations.

I know this is something neovim core eventually would like for their docgen purposes as well. Somebody just needs to do it.

jamestrew:
FYI I replaced telescope-file-browser's docgen with an alternative I've been working on.
HEAVILY based on neovim core's docgen script, integrating LuaCATS much more.

Looking to make a similar conversion for telescope core to enable changes like those in this PR.

https://github.com/nvim-telescope/telescope-file-browser.nvim/pull/399

clason:
> FYI I replaced telescope-file-browser's docgen with an alternative I've been working on.
HEAVILY based on neovim core's docgen script, integrating LuaCATS much more.


This Is The Way. (There are now tree-sitter parsers for both luadoc and emmylua, so we don't have to choose between type checking and docgen.)

If this project is going to be actively developed in the future, strict annotations are a _must_ IMO.

clason:
(But maybe it makes sense to first pick a tool -- as a target, luals and emmyluals are broadly compatible -- and then start from scratch?)

jamestrew:
I did also raise a PR a while back to switch out the docgen tool for telescope core too. https://github.com/nvim-telescope/telescope.nvim/pull/3227

I haven't kept up much with how neovim core is still doing docgen. It was all LuaCATS parsing with LPEG back when I converted the neovim docgen script into https://github.com/jamestrew/docgen.nvim
Main motivation for my docgen.nvim was
1. no dependencies (didn't want to require building treesitter parsers like we currently do with tj's tree-sitter-lua)
2. look and feel of generated help docs should be very close if not identical to neovim core
3. easy for plugins to use (main area where things start to differ from neovim core as their implementation has/had some bespoke rules/setup particular to neovim core)

I can take a look at giving it a refresh based on what neovim core is at today.

clason:
If there's something core can expose through a refactor to make this easier for plugins, consider raising an issue! Plugin ecosystem health is something the team is taking quite seriously these days.

clason:
@jamestrew is this superseded by #3227 ?

jamestrew:
something like a second attempt https://github.com/nvim-telescope/telescope.nvim/pull/3670

PR #2861, opened 2024-01-11
fix!: fallback to vim.o.winblend

### LINKED ISSUE #2702: Respect winblend by default
Telescope doesn't respect `winblend` value

Default value of `winblend = vim.o.winblend`, if none is set in `opts.defaults.winblend`.

- `opts.defaults.winblend = vim.o.winblend`: doesn't get updated if you set a new value after loading Telescope
- Wrapping every picker in something like
```lua
local function picker(name)
  return function()
    require('telescope.builtin')[name](require('telescope.themes').get_ivy({
      winblend = vim.o.winblend,
    }))
  end
end
```
Works fine, but then why have the defaults in the first place

**Additional context**
Not a big deal obviously, but `:Lazy`,`:LspInfo`,`:NullLsInfo` and [ThePrimeagen/harpoon](https://github.com/ThePrimeagen/harpoon)'s quick menu all do this by default, so I guess this would be the expected behaviour

Conni2461:
Fixes #2702

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -214,9 +214,10 @@ telescope.setup({opts})                                    *telescope.setup()*
                                               *telescope.defaults.winblend*
     winblend: ~
         Configure winblend for telescope floating windows. See |winblend| for
-        more information.
+        more information. Type can be a number or a function returning a
+        number
 
-        Default: 0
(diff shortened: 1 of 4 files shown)


Conni2461:
i guess i merge without feedback from anyone who reported the issue

PR #2876, opened 2024-01-18
docs: update CONTRIBUTING.md

jamestrew:
Better reflect the thoughts in https://github.com/nvim-telescope/telescope.nvim/issues/1228

I feel bad for people submitting PRs for new pickers only to pretty much be insta-closed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,10 @@
 # Contributing
 
-## Submitting a new feature
-
-Thanks for taking the time to submit code to Telescope if you're reading this! We love having new contributors and love seeing the Neovim community come around this plugin and keep making it better. If you are submitting a new PR with a feature addition, please make sure that you add the appropriate documentation. For examples of how to document a new picker for instance, check the `lua/telescope/builtin/init.lua` file to see how we write function headers for all of the pickers there. To learn how we go about writing documentation for this project, keep reading below!
+Thanks for taking the time to submit code to Telescope if you're reading this!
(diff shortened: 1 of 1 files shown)


Conni2461:
I dont know i dont completely agree with whats written here. Why was this self merged without feedback from other contributors?

jamestrew:
Sorry Conni we can revert this or tweak it should you want to. This hasn't been updated in something like 3 years (except syntax/grammar) and 2 years since the "move out all builtins" discussion started. I felt at the very least it was more accurate than what we already have.


Regarding self-merges, honestly I don't wish to self-merge anything but if I'm being honest, timely feedback is not always reliably there I'm sorry to say. And this is completely understandable considering we're all just volunteering our spare time.

I'm just trying to find the balance between keeping things moving for the sake of the project but also my own motivation level while still respecting the feedback process.

What do have any thoughts how to find this balance?

Conni2461:
> I felt at the very least it was more accurate than what we already have.

updating this wasnt wrong. 

> Regarding self-merges, honestly I don't wish to self-merge anything but if I'm being honest, timely feedback is not always reliably there I'm sorry to say. And this is completely understandable considering we're all just volunteering our spare time.

we should take this private but i live in the eu and this was merged within minutes while i was asleep so idk. So yeah we are currently the only 2 ppl left who more or less actively work here, so i suggest we have a little chat, sometime soon, because i am aware that i am not as active as i should be especially on pings. does that work for you?

jamestrew:
sounds good :+1:

Issue #2730, opened 2023-10-06
Support the `^wr` map in the input window


coot:
**Describe the solution you'd like**
It would be nice if the input box would support `^wr` the same way command line supports it (input the word under the cursor from the buffer which telescope was called).

jamestrew:
Do you mean `<c-r><c-w>` as in inserting the word under the cursor in the command line?

konosubakonoakua:
any new progress or advice?

jamestrew:
Created a mapping for it `<c-r><c-w>` on master.

PR #2878, opened 2024-01-19
feat(picker): allow appending original window cword to prompt

jamestrew:
Store `<cword>` on picker:find and create new action `insert_original_cword` (maybe just `insert_cword`?) to insert/append the cword of the original window/buffer under telescope into the prompt buffer.

Allows for similar behaviors to `c_CTRL-R_CTRL-W`.
closes https://github.com/nvim-telescope/telescope.nvim/issues/2730


Conni2461:
we could map this on default

PR #2879, opened 2024-01-20
docs(contrib): update contribution guide

jamestrew:
better reflect stance on feature additions specifically around new pickers


Conni2461:
thanks :) that perfectly sums up my thoughts

PR #2880, opened 2024-01-20
feat(builtin): initial support for cscope

sgruszka:
# Description

There are some benefits when using cscope for big C projects:

 * LSP definition/references do not show entries that are conditionally compiled using macros.
 * Finding all references is faster than ripgrep.
 * Cscope database can be fine tuned to skip files developer is not interested in.

In general cscope can be used as complement to LSP and ripgrep, to improve navigation on large C code base. I'm using all three to navigate in linux kernel and cscope is helpful.

TODO:
 * Implement jump if just one entry was found.
 * Improve showing functions called by this function.
 * Add configuration options similar to LSP picker.

## Type of change
- This change requires a documentation update (Maybe ?)

# How Has This Been Tested?
- Used when working on linux kernel with cscope database build there using 'make cscope'. 

**Configuration**:
NVIM v0.10.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

Ubuntu 22.04.3 LTS

# Checklist:

- [?] My code follows the style guidelines of this project (stylua)
- [x ] I have performed a self-review of my own code
- [?] I have commented my code, particularly in hard-to-understand areas
- [x ] I have made corresponding changes to the documentation (lua annotations)

### DIFF
--- a/lua/telescope/builtin/__cscope.lua
+++ b/lua/telescope/builtin/__cscope.lua
@@ -1,0 +1,112 @@
+local pickers = require "telescope.pickers"
+local finders = require "telescope.finders"
+local make_entry = require "telescope.make_entry"
+local utils = require "telescope.utils"
+
+local conf = require("telescope.config").values
+
+local cscope = {}
+
+local cscope_find = function(opts, type_num)
+  local dir = opts.cwd
+  local word = vim.F.if_nil(opts.search, vim.fn.expand "<cword>")
+  -- local search = opts.use_regex and word or escape_chars(word)
+  local search = word
+
+  local output =
+     utils.get_os_command_output({ "cscope", "-d", "-L", "-" .. tostring(type_num), search }, dir)
+
+  if #output == 0 then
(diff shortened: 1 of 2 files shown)


Conni2461:
I think should be an extension rather than a new builtin. We (telescope) are currently not interested in extending the list of builtins, see https://github.com/nvim-telescope/telescope.nvim/issues/1228 and for that reason offer a extension interface which can be used. that way you can host your builtins in a separate repository.

Also cscope was removed from nvim, some time ago (see `:help cscope`), so i currently dont see a reason to include it here. Still thanks for opening a PR.

Once you published an extension, feel free to add it to this list: https://github.com/nvim-telescope/telescope.nvim/wiki/Extensions

sgruszka:
thanks for feedback, will do the extension

rbmarliere:
Make sure to check out [cscope_maps](https://github.com/dhananjaylatkar/cscope_maps.nvim).

PR #2885, opened 2024-01-23
feat(find_files): prefer fd over rg

vegerot:
feat(find_files): prefer fd over rg

Summary:
`fd` and `rg --files` are pretty much the same thing, but `fd` is purpose-built
for finding files

### DIFF
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -268,10 +268,10 @@ files.find_files = function(opts)
         return opts.find_command(opts)
       end
       return opts.find_command
-    elseif 1 == vim.fn.executable "rg" then
-      return { "rg", "--files", "--color", "never" }
     elseif 1 == vim.fn.executable "fd" then
       return { "fd", "--type", "f", "--color", "never" }
+    elseif 1 == vim.fn.executable "rg" then
+      return { "rg", "--files", "--color", "never" }
     elseif 1 == vim.fn.executable "fdfind" then
       return { "fdfind", "--type", "f", "--color", "never" }
     elseif 1 == vim.fn.executable "find" and vim.fn.has "win32" == 0 then


jamestrew:
There was some benchmarks a while back, a few times if I recall correctly, that showed that `rg --files` is generally a bit faster than `fd` hence the order.

If you'd still like to use `fd`, you can use the `find_command` option for `find_files` and explicitly use `fd` and its flags.

jamestrew:
Sorry I can't seem to find it. It's probably buried in the comments of few old PR/issues.
But quick test I did on the linux repo with ~80000 files
```
~/projects/linux master ❯ rg --files | wc -l
80854
~/projects/linux master ❯ fd -tf | wc -l
80854
~/projects/linux master ❯ hyperfine 'rg --files' 'fd -tf' --warmup 10 --runs 100
Benchmark 1: rg --files
  Time (mean ± σ):      28.8 ms ±   3.5 ms    [User: 230.8 ms, System: 66.4 ms]
  Range (min … max):    23.8 ms …  39.0 ms    100 runs

Benchmark 2: fd -tf
  Time (mean ± σ):      35.9 ms ±   4.2 ms    [User: 395.1 ms, System: 105.8 ms]
  Range (min … max):    26.6 ms …  46.7 ms    100 runs

Summary
  rg --files ran
    1.24 ± 0.21 times faster than fd -tf
```

It's about 24% faster over 100 runs. I think that's statically significant. At least on my setup.

jamestrew:
hmm I did try the benchmark on my home dir as well and got `fd` faster (`1.09 +- 0.09`) which is much less convincing. Both rg and fd are on the latest version for me.

I don't know. 
a) I'm not statistician clearly lol 
b) feels very case by case regardless

Ultimately, neither will be a bottleneck for Telescope I think. Put differently, rg over fd is not the reason some consider Telescope slow.

I guess the one factor that matters the most is that I don't know that most people will be on the latest version of fd.

mehalter:
Yeah the biggest thing I have found as reason to switch it in core Telescope is just that the `rg` developer recommends using `fd` if it's available. `rg --files` is just there for people who was to simply list files without having to install a second tool. So it would make sense to prefer the tool that is built for listing files over a tool that added it in "just because" and put any further file listing features or improvements as "out of scope". So chances are improvements for that specific feature will not be all that much.

mehalter:
So it just seems more of a future proofing decision as well as a seemingly more sane default in my opinion, but not a huge deal either way

jamestrew:
Yeah that's interesting.
I would switch the default to fd if most people were on v9 if there really is such a significant perf difference like you say (and I think it was the case when this discussion was had previously). 
But I'm guessing most people don't update auxiliary tools that often, if they even could (ie stuck with apt's latest which I think is still on v7.4 or something).

Something to keep in mind for the future though. Thanks for the insight.

mehalter:
Yeah sounds good! Thanks for hearing me out!

PR #2892, opened 2024-01-25
fix(mappings): allowing passing `desc` to mappings

### LINKED ISSUE #2888: Allow setting description for anonymous functions
How to set descriptions for mappings in attach_mappings? I cannot get the help menu to display descriptions for decorated functions (function within another anon function). I have tried to pass .desc in map options but that did not work either. 

This would be a very helpful addition to my toolkit because I can determine the number of args to pass to an action depending on whether it is a multiselection or not and bypass the getting the picker from prompt_bufnr and manually closing the window and then doing the task. It is causing a lot of boilerplate even for things like simply rm -r selected files in telescope-file-browser just to ensure that the descriptions show up after pressing.

```
function T:selected(bufnr, many)
  local picker = self.action_state.get_current_picker(bufnr)

  if picker then
    self.actions.close(bufnr)

    if many then
      return picker:get_multi_selection()
    end

    return self.action_state.get_selected_entry()
  end
end

function T:module()
  local mod = mtset({}, {
    __newindex = function(MOD, key, value)
      if key:match "^multi_" then
        rawset(MOD, key, function(bufnr)
          list.each(T:selected(bufnr, true), value)
        end)
      else
        rawset(MOD, key, function(bufnr)
          return value(T:selected(bufnr))
        end)
      end
    end,
  })

  return mod
end
```

If I were to use module defined above then it would result in 'anonymous' for 'CR' (Enter/default action)

jamestrew:
Currently the `map` function in `attach_mappings` doesn't respect the `desc` option for `vim.keymap.set` which we use under the hood.
This results in anonymous functions to show up nameless, as `<anonymous>`,  in which-key regardless of whether a `desc` is passed.
```lua
    map("i", "<C-f>", function() print("im anonymous") end, { desc = "hi" }) -- shows up as anonymous
end
```

closes #2888

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1939,7 +1939,7 @@ ordered from the lowest priority to the highest priority.
 
       map({"i", "n"}, "<C-r>", function(_prompt_bufnr)
         print "You typed <C-r>"
-      end)
+      end, { desc = "desc for which key"})
 
       -- needs to return true if you want to map default_mappings and
       -- false if not
--- a/lua/telescope/mappings.lua
+++ b/lua/telescope/mappings.lua
@@ -113,7 +113,7 @@
 ---
 ---       map({"i", "n"}, "<C-r>", function(_prompt_bufnr)
 ---         print "You typed <C-r>"
----       end)
+---       end, { desc = "desc for which key"})
 ---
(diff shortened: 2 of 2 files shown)


Conni2461:
Yeah makes sense thanks :)

PR #2897, opened 2024-01-27
feat(action): add `type` into qf entry when calling `entry_to_qf` function

gh-liu:
When calling `send_to_qflist` for `builtin.diagnostics`, this `text` field of the qf entry could be set to describe the error type.

before:
Screenshot 2024-01-27 at 12 37 24

after:
Screenshot 2024-01-27 at 12 36 00

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -912,6 +912,7 @@ local entry_to_qf = function(entry)
     lnum = vim.F.if_nil(entry.lnum, 1),
     col = vim.F.if_nil(entry.col, 1),
     text = text,
+    type = entry.qf_type,
   }
 end
 
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1121,12 +1121,12 @@ end
 function make_entry.gen_from_diagnostics(opts)
   opts = opts or {}
 
+  local type_diagnostic = vim.diagnostic.severity
   local signs = (function()
     if opts.no_sign then
       return
     end
     local signs = {}
-    local type_diagnostic = vim.diagnostic.severity
     for _, severity in ipairs(type_diagnostic) do
       local status, sign = pcall(function()
         -- only the first char is upper all others are lowercalse
@@ -1183,6 +1183,13 @@ function make_entry.gen_from_diagnostics(opts)
     }
   end
 
+  local errlist_type_map = {
+    [type_diagnostic.ERROR] = "E",
+    [type_diagnostic.WARN] = "W",
+    [type_diagnostic.INFO] = "I",
+    [type_diagnostic.HINT] = "N",
+  }
+
   return function(entry)
     return make_entry.set_default_entry_mt({
       value = entry,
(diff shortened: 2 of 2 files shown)


lua/telescope/actions/init.lua
```diff
@@ -906,13 +906,19 @@ local entry_to_qf = function(entry)
...
+  local qf_entry = {
     bufnr = entry.bufnr,
     filename = from_entry.path(entry, false, false),
     lnum = vim.F.if_nil(entry.lnum, 1),
     col = vim.F.if_nil(entry.col, 1),
     text = text,
   }
+
+  if entry.type then
+    qf_entry.type = entry.type
```
gh-liu, line 918: As [`:h setqflist()`](https://github.com/neovim/neovim/blob/38bb0e1da868fd7593387319376d0e167616638b/runtime/doc/builtin.txt#L6683) metions, the `type` is a single character, maybe it should be the first letter of `entry.type`.

gh-liu: [This](https://github.com/neovim/neovim/blob/38bb0e1da868fd7593387319376d0e167616638b/runtime/lua/vim/diagnostic.lua#L1987) is how the `neovim core` handles the `type` field, the distinction between the two lies in mapping the `Hint` to `N`.

Conni2461: yeah we should handle it the same way although i think this should be part of `builtin/__diagnostic`, the entry that is being produced there. because its specific to diagnostics and not to the mapping from telescope entry -> qf list entry.

So we can add a new entry there, like hint which is already single line and correctly mapped. does that make sense?

gh-liu: > yeah we should handle it the same way although i think this should be part of `builtin/__diagnostic`, the entry that is being produced there. because its specific to diagnostics and not to the mapping from telescope entry -> qf list entry.
> 
I couldn't agree more.

I took a quick look at the [code](https://github.com/gh-liu/telescope.nvim/blob/4bd510943836edcf588c4dedff814dbaf372a273/lua/telescope/make_entry.lua#L1121), and noticed that the `entry.type` is used in `display` and `complete_tag`, or possibly elsewhere, I am wondering if the `entry.type` should be align with the `type` field of `qf entry`. If not, maybe a new field could be added for this purpose, such as `qf_type`.

Conni2461: i prefer a new key that that `entry.qf_type` is good enough

gh-liu: Done.

lua/telescope/actions/init.lua
```diff
@@ -906,13 +906,19 @@ local entry_to_qf = function(entry)
...
     bufnr = entry.bufnr,
     filename = from_entry.path(entry, false, false),
     lnum = vim.F.if_nil(entry.lnum, 1),
     col = vim.F.if_nil(entry.col, 1),
     text = text,
   }
+
+  if entry.qf_type then
+    qf_entry.type = entry.qf_type
+  end
```
Conni2461, line 919: cant we just do that in the table

```
{
  bufnr = entry.bufnr,
  filename = from_entry.path(entry, false, false),
  lnum = vim.F.if_nil(entry.lnum, 1),
  col = vim.F.if_nil(entry.col, 1),
  text = text,
  type = entry.qf_type,
}
```

gh-liu: Done.

Conni2461:
Thanks :)

PR #2909, opened 2024-02-03
feat(builtin.current_buffer_fuzzy_find): jump to first matched char

antonk52:
# Description

Currently the in buffer fuzzy search always jumps to the first character of the selected line. This PR enhances it to jump to the first same character on the line as the first character from the search. See demo below.

Motivation: I assume people use this helper as an enhanced version of <kbd>/</kbd> or <kbd>?</kbd>. Alternatively this is as useful as <kbd>cmd/ctrl</kbd><kbd>f</kbd> from modern GUI editors and IDEs. Both jump to the match instead of the beginning of the line.

## Type of change

# How Has This Been Tested?

1. Open any file
2. Run `:Telescope current_buffer_fuzzy_find`
3. Search for something that is not on the first character on the line that you want to go to
4. Expect cursor to land on the first character in your search on this line

**Before** 

https://github.com/nvim-telescope/telescope.nvim/assets/5817809/a237499f-2859-4ae6-9250-d8da7c22ee9c

**After**

https://github.com/nvim-telescope/telescope.nvim/assets/5817809/3a91cc57-0c21-41b5-90a5-fd3032ed9579

**Configuration**:
* Neovim version (nvim --version): VIM v0.9.5
* Operating system and version: MacOS 14.2.1

# Checklist:

- [x] NA ~~I have made corresponding changes to the documentation (lua annotations)~~

### DIFF
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -527,8 +527,6 @@ files.current_buffer_fuzzy_find = function(opts)
     opts.line_highlights = line_highlights
   end
 
-  local column = 0
-
   pickers
     .new(opts, {
       prompt_title = "Current Buffer Fuzzy",
@@ -539,26 +537,21 @@ files.current_buffer_fuzzy_find = function(opts)
       sorter = conf.generic_sorter(opts),
       previewer = conf.grep_previewer(opts),
       attach_mappings = function()
-        action_set.select:enhance {
-          pre = function(bufnr)
-            local selection = action_state.get_selected_entry()
-            if not selection then
-              return
-            end
-            local current_picker = action_state.get_current_picker(bufnr)
-            local searched_for = require("telescope.actions.state").get_current_line()
-            local highlighted = current_picker.sorter:highlighter(searched_for, selection.ordinal)
-            column = highlighted[1]
-          end,
-          post = function()
-            local selection = action_state.get_selected_entry()
-            if not selection then
-              return
-            end
-
-            vim.api.nvim_win_set_cursor(0, { selection.lnum, column or 0 })
-          end,
-        }
+        actions.select_default:replace(function(prompt_bufnr)
+          local selection = action_state.get_selected_entry()
+          if not selection then
+            utils.__warn_no_selection "builtin.current_buffer_fuzzy_find"
+            return
+          end
+          local current_picker = action_state.get_current_picker(prompt_bufnr)
+          local searched_for = require("telescope.actions.state").get_current_line()
+          local highlighted = current_picker.sorter:highlighter(searched_for, selection.ordinal)
+          local column = math.min(unpack(highlighted)) or 0
+          actions.close(prompt_bufnr)
+          vim.schedule(function()
+            vim.api.nvim_win_set_cursor(0, { selection.lnum, column })
+          end)
+        end)
 
         return true
       end,
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -527,6 +527,8 @@ files.current_buffer_fuzzy_find = function(opts)
     opts.line_highlights = line_highlights
   end
 
+  local column = 0
+
   pickers
     .new(opts, {
       prompt_title = "Current Buffer Fuzzy",
@@ -538,26 +540,23 @@ files.current_buffer_fuzzy_find = function(opts)
       previewer = conf.grep_previewer(opts),
       attach_mappings = function()
         action_set.select:enhance {
+          pre = function(bufnr)
+            local selection = action_state.get_selected_entry()
+            if not selection then
+              return
+            end
+            local current_picker = action_state.get_current_picker(bufnr)
+            local searched_for = require("telescope.actions.state").get_current_line()
+            local highlighted = current_picker.sorter:highlighter(searched_for, selection.ordinal)
+            column = highlighted[1]
+          end,
           post = function()
             local selection = action_state.get_selected_entry()
             if not selection then
               return
             end
 
-            local searched_for = require("telescope.actions.state").get_current_line()
-            local first_search_char = string.sub(searched_for, 1, 1)
-
-            local column_as_is = string.find(selection.text, first_search_char)
-            local reversed_char
-            if first_search_char == string.upper(first_search_char) then
-              reversed_char = string.lower(first_search_char)
-            else
-              reversed_char = string.upper(first_search_char)
-            end
-            local column_reversed = string.find(selection.text, reversed_char)
-            local column = math.min(column_as_is or math.huge, column_reversed or math.huge)
-
-            vim.api.nvim_win_set_cursor(0, { selection.lnum, column })
+            vim.api.nvim_win_set_cursor(0, { selection.lnum, column or 0 })
           end,
         }
 
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -546,7 +546,13 @@ files.current_buffer_fuzzy_find = function(opts)
           local current_picker = action_state.get_current_picker(prompt_bufnr)
           local searched_for = require("telescope.actions.state").get_current_line()
           local highlighted = current_picker.sorter:highlighter(searched_for, selection.ordinal)
-          local column = math.min(unpack(highlighted)) or 0
+          local column = math.min(unpack(highlighted))
+          if column then
+            column = column - 1
+          else
+            column = 0
+          end
+
           actions.close(prompt_bufnr)
           vim.schedule(function()
             vim.api.nvim_win_set_cursor(0, { selection.lnum, column })
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -546,12 +546,7 @@ files.current_buffer_fuzzy_find = function(opts)
           local current_picker = action_state.get_current_picker(prompt_bufnr)
           local searched_for = require("telescope.actions.state").get_current_line()
           local highlighted = current_picker.sorter:highlighter(searched_for, selection.ordinal)
-          local column = math.min(unpack(highlighted))
-          if column then
-            column = column - 1
-          else
-            column = 0
-          end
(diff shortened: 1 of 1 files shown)


lua/telescope/builtin/__files.lua
```diff
@@ -544,7 +544,20 @@ files.current_buffer_fuzzy_find = function(opts)
...
+
+            local column_as_is = string.find(selection.text, first_search_char)
+            local reversed_char
+            if first_search_char == string.upper(first_search_char) then
+              reversed_char = string.lower(first_search_char)
+            else
+              reversed_char = string.upper(first_search_char)
+            end
+            local column_reversed = string.find(selection.text, reversed_char)
+            local column = math.min(column_as_is or math.huge, column_reversed or math.huge)
```
jamestrew, line 558: sorry can you explain the reason for this upper/lower/reversed stuff? It's not quite obvious to me why it's needed.

antonk52: Sure thing. Ideally I'd want it to jump to first character that gets highlighted from the matched option. I could not figure out how to do that. Instead I am checking what is the leftmost character in the line that it is the same or reversed case and we will jump to it. 

You can search for `fbr` or `Fbr`, note lower and capital case f/F in the beginning. Both will match the same letters on this line

```
local fuzzy_buffer = function(buffer)
      ^     ^    ^
```

So this logic is to account for reversed case of the first character in user's search.

Let me know I managed to to explain this well

jamestrew: Hmm.. I think the trouble is that the fuzzy finder scoring and highlighting is dependent of the sorter that's being used.

With `telescope-fzf-native`, `Fbr` will not match `fuzzy_buffer`.
image

But with the default telescope sorter (fzy algorithm), it will match.
image

> Ideally I'd want it to jump to first character that gets highlighted from the matched option

I agree.
Maybe an alternative approach: we can get the highlight positions through the current picker (`action_state.get_current_picker`) with something like `picker.sorter:highlighter(searched_for, selection.ordinal)`.
You can see `Picker:highlight_one_row` in `pickers.lua` for how we apply highlighting to each entry.
This would need an `action_set.select:replace` or something like that however since using the `post` on `enhance`, the picker is already closed.

antonk52: I will play around with it to see if I can get it to work. I will report back once I take a stub at it. Thank you for the code pointers.

antonk52: Thanks for the code pointers @jamestrew ! I have updated the PR. I now has a simpler implementation and more importantly works correctly ie jumps to what was highlighted instead of the first matching character. I have tested it and updated the "after" video in the PR description. I think it is ready to be reviewed and merged if you are happy with the implementation.

lua/telescope/builtin/__files.lua
```diff
@@ -538,13 +540,23 @@ files.current_buffer_fuzzy_find = function(opts)
       previewer = conf.grep_previewer(opts),
       attach_mappings = function()
         action_set.select:enhance {
```
jamestrew, line 542:
```suggestion
        actions.select_default:replace(function(prompt_bufnr)
          local selection = action_state.get_selected_entry()
          if not selection then
            utils.__warn_no_selection "builtin.current_buffer_fuzzy_find"
            return
          end
          local current_picker = action_state.get_current_picker(prompt_bufnr)
          local searched_for = require("telescope.actions.state").get_current_line()
          local highlighted = current_picker.sorter:highlighter(searched_for, selection.ordinal)
          local column = highlighted[1] or 0
          actions.close(prompt_bufnr)

          vim.schedule(function()
            vim.api.nvim_win_set_cursor(0, { selection.lnum, column })
          end)
        end)
```
Minor touch up so we can do this all is one rather than split across `pre` and `post` operations.

lua/telescope/builtin/__files.lua
```diff
@@ -538,13 +540,23 @@ files.current_buffer_fuzzy_find = function(opts)
...
+          pre = function(bufnr)
+            local selection = action_state.get_selected_entry()
+            if not selection then
+              return
+            end
+            local current_picker = action_state.get_current_picker(bufnr)
+            local searched_for = require("telescope.actions.state").get_current_line()
+            local highlighted = current_picker.sorter:highlighter(searched_for, selection.ordinal)
+            column = highlighted[1]
```
jamestrew, line 551: In my testing, it looks like the column numbers in `highlighted` isn't necessarily ordered. I think we need to get the smallest value more deterministically.

I think there's a off-by-one error as well. I think we need to `- 1` the value as done so in the highlighting step as well.
https://github.com/nvim-telescope/telescope.nvim/blob/0f865f17af4f9bc1587a0132414cdfd32d91852e/lua/telescope/pickers.lua#L483

jamestrew (changes requested):

lua/telescope/builtin/__files.lua
antonk52: Interesting, during testing(see "after" in the PR description) it was jumping to the correct character. I will test again once I address the comments.

antonk52:
Hi, I have addressed the comments
- replace entire select behaviour instead of enhancing
- use minimal value from `highlighted` instead of the first one

I tried reproducing the off by one error the you mentioned but the current implementation works for me. I tested it on both 0.9.5 and a fresh master build. Do you have a specific case where this did not work for you? Do you use a non default sorter?

jamestrew:
> I tried reproducing the off by one error the you mentioned but the current implementation works for me. I tested it on both 0.9.5 and a fresh master build. Do you have a specific case where this did not work for you? Do you use a non default sorter?

Hmm... that's interesting
I tried with both my config and a minimal config and I'm still getting an off-by-one.
Tried 0.9.5 as well as nightly.


https://github.com/nvim-telescope/telescope.nvim/assets/66286082/65d9db06-928b-48e3-9da9-299f3e0ecbee



minimal config:
```lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    lazypath,
  })
end

-- install plugins
local plugins = {
  {
    "antonk52/telescope.nvim",
    branch = "feat/buf-fuzzy-column",
    dependencies = {
      "nvim-lua/plenary.nvim",
      "nvim-tree/nvim-web-devicons",
    },
    config = function()
      require("telescope").setup({})
    end,
  },
}

  root = root .. "/plugins",
})
```

antonk52:
I have updated the PR to handle non 0 based column index returned from `highlighted`.

Thank you for the repro steps. I tried it out and turns out the last time I tested it I used a shortcut bound to my version of the picker that I was using prior to opening the PR instead of the command. My bad but glad it is resolved now.

jamestrew (approved):

jamestrew:
I noticed that if there's no prompt given and/or there's no highlighting, `math.min` isn't given any values and errors so I pushed a fix for that.

The rest looks good to me. Thanks!

b0ae989c:
For [telescope-zf-native.nvim](https://github.com/natecraddock/telescope-zf-native.nvim) users: this commit crashes Telescope when we do `:Telescope current_buffer_fuzzy_find. And we have the following error message:
```txt
E5108: Error executing lua: ...plugins/telescope.nvim/lua/telescope/builtin/__files.lua:549: bad argument #1 to 'unpack'
 (table expected, got nil)
stack traceback:
        [C]: in function 'unpack'
        ...plugins/telescope.nvim/lua/telescope/builtin/__files.lua:549: in function 'run_replace_or_original'
        ...epro/plugins/telescope.nvim/lua/telescope/actions/mt.lua:65: in function 'key_func'
        ....repro/plugins/telescope.nvim/lua/telescope/mappings.lua:269: in function <....repro/plugins/telescope.nvim/l
ua/telescope/mappings.lua:268>
```
This is a bug on `telescope-zf-native.nvim`, not on Telescope.

To reproduce this error:
```lua
local root = vim.fn.fnamemodify('./.repro', ':p')
for _, name in ipairs({ 'config', 'data', 'state', 'cache' }) do
  vim.env[('XDG_%s_HOME'):format(name:upper())] = root .. '/' .. name
end
local lazypath = root .. '/plugins/lazy.nvim'
  vim.fn.system({
    'git',
    'clone',
    '--filter=blob:none',
    'https://github.com/folke/lazy.nvim.git',
    lazypath,
  })
end
local plugins = {
  'folke/tokyonight.nvim',
  {
    commit = 'fac5da839e23e7a4c17a332a640541cd59ebfbd5',
    lazy = false,
    dependencies = {
      'nvim-lua/plenary.nvim',
      'natecraddock/telescope-zf-native.nvim',
    },
    config = function()
      require('telescope').setup({
        extensions = { ['zf-native'] = { generic = { enable = true } } },
      })
      require('telescope').load_extension('zf-native')
    end,
  },
}
require('lazy').setup(plugins, { root = root .. '/plugins' })
vim.cmd.colorscheme('tokyonight')
```

b0ae989c:
A temporary fix is to implement
```lua
local column = highlighted and math.min(unpack(highlighted) or 1) - 1 or 0
```
at the line https://github.com/nvim-telescope/telescope.nvim/blob/fac5da839e23e7a4c17a332a640541cd59ebfbd5/lua/telescope/builtin/__files.lua#L549

b0ae989c:
With some more tests, I find out that this commit causes the same issue with [telescope-fzf-native.nvim ](https://github.com/nvim-telescope/telescope-fzf-native.nvim).

dmtrKovalenko:
Just tried this and for some reason it puts the cursor to the end of the highlight, feels a bit strange

b0ae989c:
Personally I don't mind this issue since I don't use the builtin fuzzy finder often. But from a design point of view, using highlighting results to compute cursor positions doesn't look like a good idea.

dmtrKovalenko:
It's just different from using the native search behavior. But we always know where the start of the match happens, aren't we? Why not put the cursor at the start of the match?

antonk52:
I've pushed a follow up PR that I tested with the builtin telescope finder and fzf extension of jumping to the last highlighted character #2922

jamestrew:
Thanks everyone for the quick response. I've merged in the follow up PR to fix this.

PR #2910, opened 2024-02-04
fix: do not change the jumplist when using to a line in a file

mdedonno1337:
# Description

While using some Telescope commands that jump directly to a line in a file (like `tags`), the function first opens the file, then call `norm! gg` and search for the target line.
This will include the top of the file in the jumplist.

Since the top of the file is not jumped-to by user, it's not a position that the user cares about; it would make sense to not include this position in the jumplist.

# Reproduce the bug

To see this behaviour, open a file, call the `Telescope tags` command, select and confirm any tag that is in the middle of a big file (i.e. when the target tag is jumped-to, the top of the file (line 1) should ideally not be visible), then press `Ctrl-O` to go back in the first file you edited just before the `Telescope tags` invocation.
This will not set you back in the first file, but at the top of the file containing the target tag.

# Type of change

# How Has This Been Tested?

I've used this patch for a while in my normal workflow, to ensure no obvious problems arise.

# Configuration
NVIM v0.9.5
AppImage

Debian 12

### DIFF
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -591,7 +591,7 @@ files.tags = function(opts)
                 return "\\" .. x
               end)
 
-              vim.cmd "norm! gg"
+              vim.cmd "keepjumps norm! gg"
               vim.fn.search(scode)
               vim.cmd "norm! zz"
             else
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -293,7 +293,7 @@ local search_cb_jump = function(self, bufnr, query)
   end
   vim.api.nvim_buf_call(bufnr, function()
     pcall(vim.fn.matchdelete, self.state.hl_id, self.state.winid)
-    vim.cmd "norm! gg"
+    vim.cmd "keepjumps norm! gg"
     vim.fn.search(query, "W")
     vim.cmd "norm! zz"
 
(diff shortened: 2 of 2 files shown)


jamestrew (approved):
thank you thank you

LEXUGE:
Seems like due to this change the `builtins.current_buffer_fuzzy_find` will not append the jump list upon jumping. I don't think this is the expected behavior as the jump is user-initiated in this case.

jamestrew:
@LEXUGE I think this PR will address this. Are you able to confirm?
https://github.com/nvim-telescope/telescope.nvim/pull/3205

LEXUGE:
I will test later when I am free

dhananjaylatkar:
`edit` is not called when item is selected in `builtins.current_buffer_fuzzy_find`.

Following change is required to fix buffer_fuzzy_find jumplist -

```diff
diff --git a/lua/telescope/builtin/__files.lua b/lua/telescope/builtin/__files.lua
index 2a3f2ce..4de397e 100644
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -564,6 +564,7 @@ files.current_buffer_fuzzy_find = function(opts)

           actions.close(prompt_bufnr)
           vim.schedule(function()
+            vim.cmd [[normal! m']]
             vim.api.nvim_win_set_cursor(0, { selection.lnum, first_col })
           end)
         end)
```

jamestrew:
@LEXUGE should be fixed in https://github.com/nvim-telescope/telescope.nvim/pull/3219

PR #2911, opened 2024-02-05
feat(pickers): custom results height for centered layout

ribru17:
# Description

I really like the `center` layout strategy, but I wish I had more control of the height of the individual sections. This PR adds a `results_height` option to the strategy that will set the height of the results window, and fill the rest of the height with the `preview` window. This can be especially nice for things like `bcommits` where the results window really doesn't need to be that large, and you mostly want to see your commit previews in all of their glory.

## Type of change

# How Has This Been Tested?

- [x] Setting the new option to a value less than 1 and greater than 0 (so it should be a percentage value)
- [x] Setting the option to an absolute height value

**Configuration**:
Build type: Release
LuaJIT 2.1.1702233742
* Operating system and version: Arch Linux x86_64, 6.7.1-arch1-1

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -2289,6 +2289,7 @@ layout_strategies.center()                         *telescope.layout.center()*
 
     `picker.layout_config` unique options:
       - preview_cutoff: When lines are less than this value, the preview will be disabled
+      - results_height: Optional height specifically for results window
 
 
 layout_strategies.cursor()                         *telescope.layout.cursor()*
--- a/lua/telescope/pickers/layout_strategies.lua
+++ b/lua/telescope/pickers/layout_strategies.lua
@@ -431,6 +431,7 @@ layout_strategies.center = make_documented_layout(
   "center",
   vim.tbl_extend("error", shared_options, {
     preview_cutoff = "When lines are less than this value, the preview will be disabled",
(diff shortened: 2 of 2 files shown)


Conni2461:
something doesnt work here, this is default dropdown theme

image

this is dropdown theme with `results_height` capped to 5

image

and this is dropwdown theme with `results_height` capped to 5 and `previewer = false`

image

am not against this, but this needs a little bit longer in the oven, i am also not quite sure what is causing this, on top of my head. i would need to debug a little bit longer here. Do you experience the same bugs or am i just using it wrong?

ribru17:
Thanks for looking into this, sorry for not testing as thoroughly. It seems I was always using `prompt_position = 'bottom'`, using `'top'` was what was messing things up, as the prompt position would also need to be updated in this case. It now gets updated when the position is set to `'top'`, and after some more testing I think this fixes all the issues; let me know what you think.

ribru17:
I gave this a rebase, I think it's good to merge :+1:

franklouwers:
What's the status of this PR? It should fix #3133

ribru17:
@Conni2461 friendly ping :) I believe this is good to merge still, I have addressed your comments above :+1:

Issue #2113, opened 2022-08-07
`lsp_dynamic_workspace_symbols` should prioritize matches for files in current working directory


macintacos:
**Is your feature request related to a problem? Please describe.**

I am using the `sumneko_lua` LSP with the `lua-dev` plugin to get completion/documentation/etc. for the neovim API, among other things. I noticed while using the `lsp_dynamic_workspace_symbols` picker that when searching for a symbol (for example "horizontal") the matches for my current file are placed way below the matches for other random symbols that I'm looking for. Here's an example:

<img width="965" alt="CleanShot 2022-08-07 at 14 10 40@2x" src="https://user-images.githubusercontent.com/8128941/183305162-1789ff5e-b517-4e08-9b51-a109c3d58684.png">

If there any way to adjust this to have it prefer matches for files in the current working directory? I don't mind the other matches, I just want the sorting to more heavily weight my own files instead of symbols in the broader workspace.

macintacos:
Thanks @ecerulm! That's totally fair, however, I don't necessarily want to get rid of the other options, I just want the sorting algorithm to be skewed more towards showing symbols that match in the currently working directory before all others.

Although I'll be honest, I stopped using Telescope for this while ago, so I have no idea what the sorting algorithm is like now.

PR #2923, opened 2024-02-15
feat(lsp_workspace_symbols): add option `ignore_folders` (fixes #2920)

### LINKED ISSUE #2920: FR: add directory filter option for `lsp_workspace_symbol`
When using the `lsp_workspace_symbol` picker, it really searches *the whole* workspace. In the case of tsserver, for instance, this includes everything in `node_modules` as well. While that may sometimes be useful, more often than not, this adds so much noise that finding symbols of the actual project (without the dependency code) becomes really difficult.

Add a filter option, where you can name folders to ignore.

I tried overriding the lsp handler for `workspace/symbols`, but that had no effect, I assume telescope is getting the results in some other way than using `vim.lsp.buf.workspace_symbol()`? 

Another hacky approach I also tried was removing workspace folders, running workspace symbol search, and then adding workspace folders back, which at least somewhat works for `vim.lsp.buf.workspace_symbol()`, but does not seem to affect the results listed by Telescope.

**Additional context**
There is a loosely related issue #2113 about prioritizing those matches.

chrisgrieser:
# Description

Adds an option `ignore_folders` for the built-in picker `lsp_workspace_symbols`. This option is a string array of folder names which is used to filter the results from the workspace symbols. 

Motivation for this option is that workspace symbols includes *everything* in the workspace, such as everything in `node_modules`. This unnecessarily clogs up the search results (adding a few thousand results), making `lsp_workspace_symbols` almost unusable in larger projects.

Fixes #2920 (and partially #2113).

## Type of change

# How Has This Been Tested?
I have tested with various LSPs that I use. To reproduce my test:

1. open a js/ts project
2. have `tsserver` running
3. Run the following code:

```lua
require("telescope.builtin").lsp_workspace_symbols {
	ignore_folders = { "node_modules" },
}
```

Furthermore, the option defaults to an empty list, making this an opt-in feature. That means no user is affected who does not actively use this feature.

**Configuration**:
* Operating system and version: macOS 14.1.2 (arm)

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -333,6 +333,20 @@ lsp.workspace_symbols = function(opts)
     end
 
     local locations = vim.lsp.util.symbols_to_items(server_result or {}, opts.bufnr) or {}
+
+    -- Filter specific folders from the results. Only works with full folder
+    -- names, and not glob patterns for performance reasons in big projects.
+    locations = vim.tbl_filter(function(l)
+      local dir = vim.fs.dirname(l.filename)
+      local parent_folders = vim.split(dir, "/")
+      for _, folder in ipairs(parent_folders) do
+        if vim.tbl_contains(opts.ignore_folders or {}, folder) then
+          return false
+        end
+      end
+      return true
+    end, locations)
+
     locations = utils.filter_symbols(locations, opts, symbols_sorter)
     if locations == nil then
       -- error message already printed in `utils.filter_symbols`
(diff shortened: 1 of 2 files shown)


jamestrew:
`:h telescope.defaults.file_ignore_patterns` should already handle this.

chrisgrieser:
You are right. 🙈

Wasn't aware that `file_ignore_patterns` also works for lsp pickers, assumed it only works for pickers like `find_files`. Yeah, this makes this PR unnecessary.

PR #1850, opened 2022-04-18
feat: Add entry_index for entry_makers

Conni2461:
cherry picked and rebased from https://github.com/nvim-telescope/telescope.nvim/pull/1473/commits/8b433235d0b7763376fbeaca3b9de1c1824dc252 because of unrelated changes

I will add some documentation here as well

PR #2925, opened 2024-02-17
feat(builtin.live_grep): live_grep entry_maker can receive display as…

m4ttm:
# Description

Added the ability to provide the live_grep entry_maker display function as an opt. Trying to override the whole entry maker causes problems with the grep preview, this is much easier for the purpose of customizing the display text according to the scenario.

# How Has This Been Tested?

Tested overriding the display option in my configuration to provide class and function name in the display text for Solidity.

**Configuration**:
* Neovim version (nvim --version): v0.9.5

# Checklist:

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -312,7 +312,7 @@ do
     mt_vimgrep_entry = {
       cwd = vim.fn.expand(opts.cwd or vim.loop.cwd()),
 
-      display = function(entry)
+      display = opts.make_entry_display or function(entry)
         local display_filename = utils.transform_path(opts, entry.filename)
 
         local coordinates = ":"


jamestrew:
Can you clarify the motivation behind this option?
I'm confused about why you'd want something like this
> provide class and function name in the display text

m4ttm:
> Can you clarify the motivation behind this option? I'm confused about why you'd want something like this
> 
> > provide class and function name in the display text

Similar to how the results in IntelliJ can show more detailed information than just the file name. It's a useful feature for me and a custom display function works perfectly for my use case.

[https://resources.jetbrains.com/help/img/idea/2023.3/find_tool_window_preview_area.png](url) 

I like to be able to see which class and function a result belongs too, I use this on unfamiliar code when auditing smart contracts and having the additional results information available at first sight is useful.

Conni2461:
this is already possible, overriding a single part of entry_maker, using `entry_index`, although its still not documented: https://github.com/nvim-telescope/telescope.nvim/pull/1850

jamestrew:
I think I will close this. Even without the `entry_index` option, enabling overriding of `display` for your purpose feels like a bit of a misuse of `live_grep`.

Either way, should be able to achieve this with something like 
```lua
local opts = {
  entry_index = {
    display = function(entry, _opts)
      return string.format("MY PREFIX - %s", entry.text), true
    end,
  },
}
require("telescope.builtin").live_grep(opts)
```

PR #2930, opened 2024-02-19
fix(picker): actually set prompt win to text wrap

jamestrew:
Despite setting `wrap=true` for the prompt window earlier, after setting `filetype=TelescopePrompt` in the main loop, `wrap` gets set to false.

I haven't determined the root cause of this. I can't reproduce this with a simple neovim api only config. I suspect there's something funny going on with the async main loop.

Maybe this isn't worth fixing though. I think we've been using prompt window `wrap=false` for a while now and I don't see any issues with it.


Conni2461:
thanks :D

probably not worth to investigate further, but with these changes it looks better with extremely long prompts and when you delete those.

jamestrew:
I think this might be causing a weird regression https://github.com/nvim-telescope/telescope.nvim/issues/2953

Issue #2940, opened 2024-02-23
telescope.actions.select_drop does nothing on selection


Giovanni0114:
### Description

as in title, when i'm trying to configure telescope to open in new buffer rather then in splitted window, it does not work

### Neovim version

```markdown
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1692716794
```

### Operating system and version

Linux 6.7.4-arch1-1 # x86_64 GNU/Linux 

### Telescope version / branch / rev

branch master         commit ba64547

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()

Checking for required plugins ~
- OK plenary installed.
- OK nvim-treesitter installed.

Checking external dependencies ~
- OK rg: found ripgrep 14.1.0
- OK fd: found fd 9.0.0

===== Installed extensions ===== ~

Telescope Extension: `terms` ~
- No healthcheck provided

Telescope Extension: `themes` ~
- No healthcheck provided
```

### Steps to reproduce

1. set the mapping below to telescope config
```lua
 mappings = {
      i = { ["<CR>"] = require("telescope.actions").select_drop, },
    },
```
2. run telescope command for configured picker (":lua require("telescope.builtin").man_pages()" for example)
3. hit enter

### Expected behavior

the selection will open in new buffer in current window

### Actual behavior

nothing happend, selection is ignored

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          { 'nvim-telescope/telescope-fzf-native.nvim', run = 'make' },
        },
      },
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
      display = { non_interactive = true },
    },
  }
end

_G.load_config = function()
  require("telescope").setup {
    pickers = {
      man_pages = {
        mappings = {
          i = {
		["<CR>"] = require("telescope.actions").select_drop, 
          },
        },
      },
    },
 }
[...]

PR #2944, opened 2024-02-24
fix: drop for help_tags and man_pages

Conni2461:
close https://github.com/nvim-telescope/telescope.nvim/issues/2940


Conni2461:
this doesnt even cover a fraction of the underlying problem

PR #2958, opened 2024-02-28
fix(actions): pass pattern name to QuickFixCmd* autocmds

### LINKED ISSUE #2955: QuickFixCmdPost Conflict Resulting in Error E776
### Description

I have noticed that [the code](https://github.com/nvim-telescope/telescope.nvim/blob/2e1e382df42467029b493c143c2e727028140214/lua/telescope/actions/init.lua#L938) in telescope.nvim may lead to a problem: 
```lua
  vim.api.nvim_exec_autocmds("QuickFixCmdPost", {})
```
If the user's configuration or other plugins (e.g., [go.nvim](https://github.com/ray-x/go.nvim/blob/d748e79011437d5b080006b896b3c296656641d7/lua/go/commands.lua#L126)) define similar code like this:
```lua
au QuickFixCmdPost    l* nested lwindow
```
When editing a file starting with the letter "l" and triggering the `actions.smart_send_to_qflist` function, an error "Error detected while processing QuickFixCmdPost Autocommands for 'l*': E776: No location list" occurs.

### Neovim version

```markdown
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.0-beta3
```

Arch Linux / macOS 10.15.7

2e1e382

### checkhealth telescope

```markdown

- OK plenary installed.

- OK rg: found ripgrep 13.0.0
- OK fd: found fd 8.4.0

Telescope Extension: `fzf` ~
- OK lib working as expected

Telescope Extension: `yank_history` ~
- No healthcheck provided
```

### Steps to reproduce

1. Open a Go file starting with the letter "l".
2. Move the cursor to any variable or function.
3. Trigger the `Telescope lsp_references include_current_line=true` command using gr or other shortcuts.
4. Use `q` or other shortcuts to trigger `actions.smart_send_to_qflist + actions.open_qflist` on the popped-up window.

### Expected behavior

_No response_

### Actual behavior

```
Error detected while processing QuickFixCmdPost Autocommands for "l*":
E776: No location list
Press ENTER or type command to continue
```

### Minimal config

```Lua
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--branch=stable",
    lazypath,
  })
end

require("lazy").setup({
    {
        config = function()
            require('config.telescope')
        end,
        enabled = true,
    },
    {
        'ray-x/go.nvim',
        config = function()
            require('go').setup()
        end
    },
})

require('telescope').setup{
  defaults = {
    mappings = {
      i = {
        ["<C-h>"] = "which_key"
      }
    }
  },
  pickers = {
  },
  extensions = {
  }
}
```

jamestrew:
closes #2955

I'm not actually sure if this is a good fix and whether the usage of `pattern` here is appropriate.
Maybe executing these QuickFixCmd autocmds are incorrect.

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -927,7 +927,7 @@ local send_selected_to_qf = function(prompt_bufnr, mode, target)
   local prompt = picker:_get_prompt()
   actions.close(prompt_bufnr)
 
-  vim.api.nvim_exec_autocmds("QuickFixCmdPre", {})
+  vim.api.nvim_exec_autocmds("QuickFixCmdPre", { pattern = "Telescope" })
   if target == "loclist" then
     vim.fn.setloclist(picker.original_win_id, qf_entries, mode)
   else
@@ -935,7 +935,7 @@ local send_selected_to_qf = function(prompt_bufnr, mode, target)
     vim.fn.setqflist(qf_entries, mode)
     vim.fn.setqflist({}, "a", { title = qf_title })
   end
-  vim.api.nvim_exec_autocmds("QuickFixCmdPost", {})
+  vim.api.nvim_exec_autocmds("QuickFixCmdPost", { pattern = "Telescope" })
 end
 
 local send_all_to_qf = function(prompt_bufnr, mode, target)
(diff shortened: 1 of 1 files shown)


itaranto:
This fix works for me.

Why isn't this merged already?

clason:
> I'm not actually sure if this is a good fix and whether the usage of pattern here is appropriate.

It's not: `:h QuickFixCmdPre` says
>  The pattern is matched against the command being run.  When |:grep| is used but 'grepprg' is set to "internal" it still matches "grep".

This means the "command" is from a closed builtin list and does not apply to `:Telescope` (which in any case would not work when a picker is triggered through the Lua API). So what this change does is actually disable these autocommands... (Which _may_ be a fix, but then we can just remove them altogether.)

PR #2959, opened 2024-02-28
feat(lsp_jump_type): jump type can be passed as a function

Blackmorse:
# Description

For builtin LSP actions, there is a parameter called `jump_type`, which allows choosing how `goto_reference` will be executed when only one entity is found in another file: "tab", "tab drop", "split", "vsplit", "never".

In this PR, I've added the ability to pass a function as the `jump_type` parameter. This function will be executed instead of using the predefined commands.

### Motivation:
I have the following use case: I want to have the ability to choose the window in which a reference will be opened when navigating to it via `lsp_definitions`. Currently, when multiple definitions are found, I can achieve this by using a window-picker after opening the telescope. However, when only one definition is found, I don't have this option.

With this PR, I'll be able to solve my problem by using the following invocation:

```lua
require('telescope.builtin').lsp_definitions(
    {
      get_selection_window = function()
        local window = require('window-picker').pick_window({ include_current_win = true });
        return window
      end,
      jump_type = function ()
        local window = require('window-picker').pick_window({ include_current_win = true });
        vim.api.nvim_set_current_win(window)
      end
    })
```

## Type of change

# How Has This Been Tested?

Thoroughly tested manually with aforementioned config

- [x] Opening while many references has been found is working as before
- [x] Opening while one reference has been found execution window-picker and then opening reference it selected window

**Configuration**:
```
nvim --version
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1703358377
```
* Operating system and version: MacOS Monterey 12.6.8

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1558,7 +1558,9 @@ builtin.lsp_references({opts})            *telescope.builtin.lsp_references()*
                                           only one and the definition file is
                                           different from the current file,
                                           values: "tab", "tab drop", "split",
-                                          "vsplit", "never"
+                                          "vsplit", "never" or a function,
+					  which will be executed before the
+					  jump
         {show_line}            (boolean)  show results text (default: true)
         {trim_text}            (boolean)  trim results text (default: false)
         {file_encoding}        (string)   file encoding for the previewer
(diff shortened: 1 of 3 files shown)


Blackmorse:
@jamestrew , indeed, repurposing `get_selection_window` option will be nice solution. However, we can't rely on fact that `picker` and `entry` params are not widely used - they could be used by someone and 'repurposing' will break their code. 
In fact it is two different features and I suggest to keep them as a separate features

Blackmorse:
Hi @jamestrew,

It's been a while. Could you merge this, or do you have any suggestions for alternatives to them implemetation of this feature? I find this addition essential for my coding patterns, so it would be great to finalize this request.

Thanks!

Blackmorse:
@jamestrew , any thoughts?

PR #2971, opened 2024-03-03
No regex highlighting in previewers

jan-xyz:
# Description

I was digging into why my colorscheme wasn't displaying the correct diff colours for Telescope buffers, and I found that it was using the regex highlighter instead of tree-sitter based highlighting. This lead to it using the deprecated `diffAdded` and `diffRemoved` highlight groups. Another fix would've been to add a link for these deprecated groups, and that's still an OK option for me. I just wasn't sure if it's maybe better to use the tree-sitter highlighter anyways? Why was it forced to regex in those cases?

## Type of change

(I'm not sure if this is a bug-fix or breaking change, is tree-sitter a bad default that might break people?)

# How Has This Been Tested?

- [x] I checked all affected previewers, the git but also the `help` and `man` previewers manually.

**Configuration**:
* Operating system and version: macOS 14.3.1

# Checklist:

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -711,7 +711,7 @@ previewers.help = defaulter(function(opts)
         winid = self.state.winid,
         preview = opts.preview,
         callback = function(bufnr)
-          putils.regex_highlighter(bufnr, "help")
+          putils.highlighter(bufnr, "help")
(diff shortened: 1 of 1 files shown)


jamestrew (approved):
Yeah LGTM. I think using treesitter if available/desired and using regex highlighting as a fallback makes sense.

Thanks!

PR #2975, opened 2024-03-05
fix(builtin.git): check if we're in git repo

Kek5chen:
# Description

**Change: Check in git repo before assuming so and then propagating an error up to nvim.**

When running any git builtin like, git_commits or something outside of a git repo, this would produce an error() that would not be caught and propagated up to the user. This will check if we're even in a git repo before running any git related pickers.

**Note: This is my first commit to telescope and this has been bothering me for a while, but I'm not sure if the `return {}` I placed is the correct way to handle it. If anyone can double check it, i'd appreciate it.** _It did work when testing. > \_ >_

Before:
image

After:
image

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__git.lua
+++ b/lua/telescope/builtin/__git.lua
@@ -494,6 +494,15 @@ local function apply_checks(mod)
     mod[k] = function(opts)
       opts = vim.F.if_nil(opts, {})
 
+      local status, ret = utils.get_os_command_output({ "git", "status" })
+      if ret ~= 0 then
+        utils.notify("apply_checks", {
+          msg = (vim.fn.expand "%:p:h") .. " is not a git directory",
+          level = "WARN",
+        })
+        return {}
+      end
+
       set_opts_cwd(opts)
       v(opts)
     end


jamestrew:
Telescope is actually the one explicitly throwing the error:
https://github.com/nvim-telescope/telescope.nvim/blob/dc1ea28cc2a02cd1cebb9d80e967807d668f78f8/lua/telescope/builtin/__git.lua#L448

But yeah this could probably be cleaned up a bit to use `utils.notify`.
If you're interested, it would be great if you can just tweak that bit of code there. And maybe we can pass down the name of the picker to the `try_worktrees` function and use that as the first param to `utils.notify` to be more clear about the source of the warning from a user perspective.
We can build the picker name by prefixing the `k` from `for k, v in pairs(mod)` with `git_` from the `apply_checks` function.

Conni2461:
This was already done in the past and this was a breaking change that got reverted in the end because you can script around `error` using `pcall` but you cant do the same with `vim.notify`. There is https://github.com/nvim-telescope/telescope.nvim/pull/2225 which was an idea from me to make this configurable

jamestrew:
True. Yeah let's not change this. Thanks for the catch Conni.

jamestrew:
The snippet in the wiki isn't the "true" fallback approach.
The real thing is https://github.com/nvim-telescope/telescope.nvim/issues/410#issuecomment-765656002

One reason the `pcall` approach might be preferred is that Telescope doesn't only check if the current directory is in a git repo but also checks against detached worktrees as defined in your personal config https://github.com/nvim-telescope/telescope.nvim/pull/2597 (although this isn't fully stable yet).

Issue #2939, opened 2024-02-21
Make it so that `lsp_incoming_calls` takes you to the line of the function call itself, similar to `vim.lsp.buf.incoming_calls()`


tristan957:
**Is your feature request related to a problem? Please describe.**
The `lsp_incoming_calls` picker doesn't take you to the line of the call while `vim.lsp.buf.incoming_calls()` does.

**Describe the solution you'd like**
Take me to the line of parent function that calls the function I am analyzing.

**Describe alternatives you've considered**
Looked at the options of the picker and nothing stood out. `vim.lsp.buf.incoming_calls()` is an alternative.

tristan957:
https://github.com/nvim-telescope/telescope.nvim/pull/2935 seems related.

jamestrew:
I think this is actually a bug. It only seems to only one item from each file as well and not in line with `vim.lsp.buf.incoming_calls()`.

PR #2985, opened 2024-03-14
fix(lsp): incoming/outgoing_call range locations

jamestrew:
Previous implementation uses the incorrect lsp result fields for generating call site locations resulting in incomplete and incorrect results.

closes https://github.com/nvim-telescope/telescope.nvim/issues/2939


Conni2461:
this did work previously, so it might be related to the neovim version?! so not sure regarding to the backport

tristan957:
I really appreciate you taking a look at this issue :).

Issue #2697, opened 2023-09-08
builtin.find_files & builtin_git_files not showing files in results view


mattcarlotta:
### Description

When invoking the `telescope.builtin.find_files` or `telescope.builtin.git_files`, the preview window won't display file search results until an arrow key is pressed.

### Neovim version

```markdown
NVIM v0.10.0-dev-1051+gd0d4160dd1
Build type: RelWithDebInfo
LuaJIT 2.1.1693350652
```

### Operating system and version

Endeavour OS rolling

### Telescope version / branch / rev

telescope 0.1.2

### checkhealth telescope

```markdown
==============================================================================
telescope: health#telescope#check

- WARNING vim.health.report_start() is deprecated, use vim.health.start() instead. :help |deprecated|
  This feature will be removed in Nvim version 0.11

Checking for required plugins ~
- WARNING vim.health.report_ok() is deprecated, use vim.health.ok() instead. :help |deprecated|
  This feature will be removed in Nvim version 0.11
- OK plenary installed.
- OK nvim-treesitter installed.

Checking external dependencies ~
- OK rg: found ripgrep 13.0.0
- OK fd: found fd 8.7.0

===== Installed extensions ===== ~

```

### Steps to reproduce

1. In the cmdline type: `:Telescope find_files`
2. File results will **appear** empty
3. Type in a file name to search
4. The preview window will display the file, but the results file list will still **appear** empty

### Expected behavior

The results view should contain a list of files on initial window open and should update the list as a search term is inputted.

### Actual behavior

Command:
![screenshot-20230908-202329Z-current](https://github.com/nvim-telescope/telescope.nvim/assets/22607722/e8ddc144-082a-4be7-9cc1-6946a5e13c30)

Initial window view:
![screenshot-20230908-202335Z-current](https://github.com/nvim-telescope/telescope.nvim/assets/22607722/4dbb8168-4c01-40c6-a4d1-4d5a1aacebeb)

Adding a search term:
![screenshot-20230908-202340Z-current](https://github.com/nvim-telescope/telescope.nvim/assets/22607722/a7d14c82-8ccd-4aeb-8174-decdf19d1f9b)

Arrow key pressed (up or down, doesn't matter):
![screenshot-20230908-202345Z-current](https://github.com/nvim-telescope/telescope.nvim/assets/22607722/0ab117a0-3945-4474-b438-655a35daf25b)

Removing a character from the search term (results list appears empty):
[...]

PR #2986, opened 2024-03-15
fix(picker): ensure that the selected result matches the screen display in ascending sort order

0x7a7a:
# Description

The current stable version (v0.9.5) and the nightly version have different results for the cursor position on the displayed content, I don't know exactly why.

After debugging, I found that this line of code causes the cursor to move down(row+1) when `sorting_strategy=ascending`,this result is consistent in v0.9.5 and nightly.
https://github.com/nvim-telescope/telescope.nvim/blob/67c598fdd4fca113224281c85721c4d8a6df055e/lua/telescope/pickers.lua#L436

However, when the result is displayed, the highlighted row and the cursor position are not the same, so it cannot be displayed in nightly version(by default, the second result is displayed at the top, in the same position as the cursor.).

So I think putting the cursor to the first line (i.e. highest score) again in `Picker:get_result_completor` fixes the problem and doesn't affect older versions.

Fixes # (issue)
#2697 

## Type of change

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -1458,6 +1458,8 @@ function Picker:get_result_completor(results_bufnr, find_id, prompt, status_upda
       local visible_result_rows = vim.api.nvim_win_get_height(self.results_win)
       vim.api.nvim_win_set_cursor(self.results_win, { self.max_results - visible_result_rows, 1 })
       vim.api.nvim_win_set_cursor(self.results_win, { self.max_results, 1 })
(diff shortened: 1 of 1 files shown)


0x7a7a:
This is more obvious if the picker is another theme.
Ex:
```lua
pickers = {
  find_files = {
    hidden = true,
    theme = 'dropdown',
    previewer = false,
    follow = true,
  },
}
```

jamestrew:
thanks!

PR #2989, opened 2024-03-18
docs(builtin.colorscheme): add `colors` option desc

jamestrew:


jamestrew:
feels like both the docgen CI and running docgen locally is getting less and less reliable :/

Issue #2368, opened 2023-02-07
jump_type of split will open a [No Name] ("new") buffer in addition to splitting window with definition


yudi-tan:
### Description

I was trying to setup my telescope + lsp config and i'd configured it such that my lsp_definitions call has `jump_type` set to `split`. However, when doing so, i realized that while it does open the file with the definition of that function in a new split window buffer, it also opens an additional [No Name] (aka new) buffer. In other words, 2 buffers are opened, and the "new" buffer is very unnecessary. I'm not sure if this is a configuration issue on my end but i dug into the implementation here https://github.com/nvim-telescope/telescope.nvim/blob/master/lua/telescope/builtin/__lsp.lua#L190 and it does seem like it calls the "new" and "vnew" commands if `jump_type` is set to `split` or `vsplit`. 

I'm quite new to vim so i'm not 100% sure if i understood that logic correctly. Is it intended that every time we go-to-definition in a new split window it should open an additional [No Name] buffer? I could `set nohidden` and that fixes this issue but i don't really wanna set `nohidden`. 

Here's my config:

```
local function lsp_keymaps(bufnr)
  local opts = { noremap = true, silent = true }

  -- Declaration and definition
  vim.api.nvim_buf_set_keymap(bufnr, "n", "gd", "<cmd>lua require('telescope.builtin').lsp_definitions({jump_type='vsplit'})<CR>", opts)
  vim.api.nvim_buf_set_keymap(bufnr, "n", "hd", "<cmd>lua require('telescope.builtin').lsp_definitions({jump_type='split'})<CR>", opts)
end

M.on_attach = function(client, bufnr)
  lsp_keymaps(bufnr)
end
```
where i am passing in this to `null-ls`.

My alternative right now is to `vim.api.nvim_buf_set_keymap(bufnr, "n", "gd", ":vsplit | lua vim.lsp.buf.definition()<CR>", opts)` instead which works but was just wondering why using Telescope's `lsp_definition` with `jump_type='vsplit') will result in a new [No Name] buffer?

### Neovim version

```markdown
nvim --version
NVIM v0.8.3
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@HMBRW-A-001-M1-004.local
```

### Operating system and version

macOS 12.6

### Telescope version / branch / rev

master

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()
========================================================================
## Checking for required plugins
  - OK: plenary installed.
[...]

PR #2990, opened 2024-03-18
fix(builtin.lsp): extra No Name buffer opening with jump_type

jamestrew:
closes https://github.com/nvim-telescope/telescope.nvim/issues/2368

Just needed to pass the filename to these window/tab opening commands so that they don't needlessly create an empty buffer.


Frederick888:
I can confirm this fixed my issue. Thank you!

jamestrew:
@divaltor Thanks for catching that and mentioning it. I just put in a fix.

PR #2999, opened 2024-03-22
fix(utils.path_expand): improve windows support

jamestrew:
In order to maintain plenary compatibility (for now, and slightly for other edge case reasons), avoid converting backslashes to forward slashes.

Aside: would be nice to add windows as part of the ci testing.

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -17,14 +17,31 @@ local utils = {}
 
 local iswin = vim.loop.os_uname().sysname == "Windows_NT"
 
---- `vim.fs.normalize` co-opted for 0.1.x (neovim 0.7) compat
---- TODO: get rid of this and use  `vim.fs.normalize` directly for future releases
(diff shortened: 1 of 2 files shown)


sibouras:
hi this pr fixes `Telescope buffers` but it doesn't fix `Telescope git_status path_display={'shorten'}`(didn't mention it previously because i just found out its bugged)

`Telescope git_status path_display={'smart'}` was showing absolute path even before 3b8399c but 
`Telescope git_status path_display={''}` was working and with the new pr its bugged,

here's a gif to better show what's happening

https://github.com/nvim-telescope/telescope.nvim/assets/64098483/9a02ce11-f1fd-47fd-b333-09e4b8bd251a

jamestrew:
@sibouras Thanks, I think I'll merge this for now and tactile the `path_display` issue in a follow up. I need to look a little deeper into it.
Your video helps a lot though.

Issue #2014, opened 2022-06-16
Ability to make file name more visible in file path


GuillaumeLagrange:
**Is your feature request related to a problem? Please describe.**
When I use git_files or find_files on large project with a lot of directories, it is sometimes a bit hard to spot the filename from the rest of the path.

**Describe the solution you'd like**
I would really love to have a built-in option to bolden/put a colot emphasis on the file name in the results so I can quickly spot out the file name.

**Describe alternatives you've considered**
I've considered
```
            -- Format path as "file.txt (path\to\file\)"
            path_display = function(opts, path)
              local tail = require("telescope.utils").path_tail(path)
              return string.format("%s (%s)", tail, path)
            end,
```
But I am not sure how to handle coloring/bolding output in lua, I am a bit unfamiliar with the language.

**Additional context**
I would love to open a PR for this if you would have this feature, I may need a bit of help to get started though.

Thanks for the awesome work you guys have been doing !

Conni2461:
`path_display` is currently not integrated with our hightlighter. Shouldn't be hard to do, will add it later.

After that you should be able to do

```lua
-- Format path as "file.txt (path\to\file\)"
path_display = function(opts, path)
  local tail = require("telescope.utils").path_tail(path)
  return string.format("%s (%s)", tail, path), { { { 1, #tail }, "Constant" } }
end,
```

vignesh0025:
I too require this feature implemented. Thanks

DasOhmoff:
Yes please. I would love this too!

XobSod:
Bumping +1 ;)

JoosepAlviste:
This feature would be very nice to have indeed!

I quite like the formatting of the file picker in VSCode where the path of the file is darker:

<img width="769" alt="image" src="https://user-images.githubusercontent.com/9450943/219456061-79057567-8e83-45c9-a9bc-2fdf7d921cc5.png">

But in Telescope, the file name (first section before the dash) is not very easy to discern from the full file path:

<img width="1243" alt="image" src="https://user-images.githubusercontent.com/9450943/219456270-134a26d1-b07d-41c2-abb9-9e4625f56c72.png">

tjex:
agreed, this would be ace!

towry:
<img width="1450" alt="截屏2023-05-10 08 38 55" src="https://github.com/nvim-telescope/telescope.nvim/assets/8279858/0f68ae46-d901-4807-a1f7-ed77829a3771">

[...]

PR #3010, opened 2024-03-24
Adds option to reverse path display with highlighting

alycklama:
# Description

Long paths make it sometimes difficult to see the file name. You can create a custom formatter, but I thought it would be great if we can all have a format similar to Intelij, where the file name is in front.

## Type of change

# How Has This Been Tested?

I've tested my code with the following configurations:

**Simple config:**
```lua
path_display = {
  filename_first = {
    reverse_directories = false
  }
},
```

# Screenshots

## Find Files
image

## Live Grep
image

## Quickfix
image

## Find Word
image

**Configuration**:

* Neovim version (nvim --version): NVIM v0.9.5 | Build type: Release | LuaJIT 2.1.1710088188
* Operating system and version: MacOs 13.6.4 (22G513)

# Checklist:

- [ x ] My code follows the style guidelines of this project (stylua)
- [ x ] I have performed a self-review of my own code
- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] I have made corresponding changes to the documentation (lua annotations)


lua/telescope/utils.lua
```diff
@@ -147,6 +147,27 @@ utils.filter_symbols = function(results, opts, post_filter)
   end
 end
 
+utils.path_reverse = (function(filepath, file_sep, dir_open_sep, dir_close_sep)
+  local dirs = vim.split(filepath, Path.path.sep)
```
alycklama, line 151: > This wouldn't be compatible on Windows with how we currently preserve backslash path separators for Windows.

I checked the code and saw the use of `Path.path.sep`. I've replaced the hardcoded "/" in my previous PR. Would this solve your remark above? @jamestrew

alycklama:
I've added the highlighting code of @chrisgrieser, as it was the most simple and straightforward solution for highlighting:
https://github.com/nvim-telescope/telescope.nvim/issues/2014#issuecomment-1873229658

I tried the code posted here, put I guess it has not yet been implemented:
https://github.com/nvim-telescope/telescope.nvim/issues/2014#issuecomment-1166467071

alycklama:
I think I found a nice way to do highlighting in a flexible way. I'm lacking a bit of time right now, so I made a very simple proof-of-concept:

# make_entry.lua

```lua
    mt_file_entry.display = function(entry)
      local hl_group, icon
      local display = utils.transform_path(opts, entry.value)

      display, hl_group, icon = utils.transform_devicons(entry.value, display, disable_devicons)
       
      local path_style = { { 10, 20, }, hl_group }
      local style = { { { 0, #icon, }, hl_group }, path_style }

      if hl_group then
        return display, style
      else
        return display
      end
    end
```

When `utils.transform_path` returns both the transformed_path, as well as the `path_style`, you can easily merge the styles, resulting in the following:

image

What do you think? I'll work on a real example this week so you can see how it would look like.

alycklama:
I've pushed a new commit as a proof-of-concept:
https://github.com/nvim-telescope/telescope.nvim/pull/3010/commits/106ac899b659c05a167790e05facc6e9d525c006

image

I've a hardcoded value (4) in there that I want to get rid of, but I need to learn more about unpacking tables dynamically in Lua. That way I could return a relative style position in the functions and shift it to the right absolute position where I need it.

alycklama:
I've removed the hardcoded value (4) with a function to shift the offset dynamically:
https://github.com/nvim-telescope/telescope.nvim/pull/3010/commits/ac33dd0dd3e8c1db3ebd347fe6b0138ed019a287

alycklama:
Hereby an example how this could work for other places where the `transform_path` function is called:

image

lua/telescope/utils.lua
jamestrew: Thanks, I think this looks good :+1:

alycklama:
So far, it seems to be working quite well. Does this align with what you envision? @jamestrew 

## Find Files
image

## LSP definitions
image

## Live Grep
image

## Quickfix
image

jamestrew:
Thanks @alycklama
At first glance, this all looks pretty good. Give me a few days to give it a proper look through.

alycklama:
> Thanks @alycklama At first glance, this all looks pretty good. Give me a few days to give it a proper look through.

Take all the time you need!

In the meantime I'm trying to fix the tests, but I'm having some issues running them. Any idea what the easiest way would be? I've created a local Docker container to install Ubuntu 22.04. However, I'm not able to run `make test`:

```bash
nvim --headless --noplugin -u scripts/minimal_init.vim -c "PlenaryBustedDirectory lua/tests/automated/ { minimal_init = './scripts/minimal_init.vim' }"
Error detected while processing command line:
E492: Not an editor command: PlenaryBustedDirectory lua/tests/automated/ { minimal_init = './scripts/minimal_init.vim' }
```

To workaround this, I've manually created a `~/.config/nvim/init.lua` file with Lazy to install plenary. This works, because I can now use the PlenaryBustedDirectory Ex Command from the command line. However, it still looks like plenary.path is missing.

image

Any tips regarding testing?

jamestrew:
Yeah I have another PR to improve the CI and local testing workflow better.
In the meantime, probably easiest thing is to temporarily edit the `minimal_init.vim` file like this
```vim
set rtp+=.
set rtp+=../plenary.nvim/ " changed this to your plenary installation path
" set rtp+=../tree-sitter-lua/ not needed for testing

runtime! plugin/plenary.vim
runtime! plugin/telescope.lua

" I think you can comment this out as well
" outputs unnecessary warnings
" runtime! plugin/ts_lua.vim

let g:telescope_test_delay = 100
```

alycklama:
@jamestrew I've updated the initial post, fixed the tests. However, the nightly OSX build seems to fail in the preparation phase. Any idea what could be wrong? If that's fixed then this PR is ready for review.

lua/telescope/make_entry.lua
```diff
@@ -158,14 +158,26 @@ do
...
+          offset = #icon + 1
+        end
+
+        return offset
+      end
+
       if hl_group then
-        return display, { { { 0, #icon }, hl_group } }
+        local style = { { { 0, get_offset() }, hl_group } }
+        style = utils.merge_styles(style, path_style, get_offset())
```
jamestrew, line 177: Is this `get_offset` function necessary? Seems redundant to me. Something similar to what been done to `gen_from_vimgrep` should work no?

jamestrew:
> @jamestrew I've updated the initial post, fixed the tests. However, the nightly OSX build seems to fail in the preparation phase. Any idea what could be wrong? If that's fixed then this PR is ready for review.

Yeah don't worry about that. I think I'm pretty good with these changes. It's slightly unfortunate that entry makers using `entry_display.create` won't be able to utilize the highlighting but I think there's plans to rework the whole `entry_display` stuff so I think we'll handle the remaining entry makers in a follow up.

lua/telescope/make_entry.lua
alycklama: You are correct. I cleaned it up and pushed my changes.

alycklama:
> > @jamestrew I've updated the initial post, fixed the tests. However, the nightly OSX build seems to fail in the preparation phase. Any idea what could be wrong? If that's fixed then this PR is ready for review.
> 
> Yeah don't worry about that. I think I'm pretty good with these changes. It's slightly unfortunate that entry makers using `entry_display.create` won't be able to utilize the highlighting but I think there's plans to rework the whole `entry_display` stuff so I think we'll handle the remaining entry makers in a follow up.

Ok, that's great to hear! If there are discussions going on regarding the plans around `entry_display`, I'd be more than happy to jump in and read through them.

lkhphuc (approved):
Tested on neovim nightly and worked great. Thanks.

alycklama:
> Tested on neovim nightly and worked great. Thanks.

Good to hear, thanks!

I've rebased the PR and resolved the conflict.

lkhphuc:
I found a case where the path is not properly highlight when run: `Telescope buffers`
All other pickers (find files, git files, oldfiles, grep word, etc) are highlighted correctly though.

Screenshot 2024-04-18 at 16 33 48

alycklama:
> I found a case where the path is not properly highlight (applied the `Comment` hlgroup?) when run: `Telescope buffers` All other pickers (find files, git files, oldfiles, grep word, etc) are highlighted correctly though.

I've pushed a commit fixing this issue. Can you check again?

image

lkhphuc:
That worked. Thank you. 
On that note, do you think this would be nicer / possible to add the cursor position right after the file's name, instead of after the path i.e `invariant.md:1 docs/typeclasses` in your screenshot above?
I think that would make more sense, and the highlight color would match too.

alycklama:
> That worked. Thank you. On that note, do you think this would be nicer / possible to add the cursor position right after the file's name, instead of after the path i.e `invariant.md:1 docs/typeclasses` in your screenshot above? I think that would make more sense, and the highlight color would match too.

That's a good idea, but unfortunately, the line numbers are not part of the display path. They are rendered independently, and it would require a rather big rewrite internally to accomplish this.

I'd prefer to keep this PR as it is, and refactor the internals in a new PR to accommodate your suggestion.

alycklama:
@jamestrew Is there anything still needed to be done for this PR?

jamestrew:
I fixed some formatting (and rebased to try something with our docgen CI but didn't work, whatever :shrug:).

I'll need to think deeper about some of the loose ends like `entry_display.create` and line number stuff but I think we have a good starting point, enough to merge this.

Thanks!

Tronikelis:
Hi, was testing these changes and I saw that with `truncate` parameter, it truncates the filename and not the filepath, see image below

<details>
<summary>Image</summary>

image

</details>


Config:

```
				path_display = {
					"truncate",
					filename_first = {
						reverse_directories = false,
					},
				},
```

This is probably not the expected behavior right?

alycklama:
> Hi, was testing these changes and I saw that with `truncate` parameter, it truncates the filename and not the filepath, see image below of find files picker
> 
> Config:
> 
> ```
> 				path_display = {
> 					"truncate",
> 					filename_first = {
> 						reverse_directories = false,
> 					},
> 				},
> ```
> 
> This is probably not the expected behavior right?

The path_display options are mutually exclusive (as far as I understand). You need to remove "truncate" and only use the "filename_first" option. Does that solve your issue?

alycklama:
I didn't realize the options are composable. Straight from the documentation:

```text
        path_display can be set to an array with a combination of:
        - "hidden"          hide file names
        - "tail"            only display the file name, and not the path
        - "absolute"        display absolute paths
        - "smart"           remove as much from the path as possible to only show
                            the difference between the displayed paths.
                            Warning: The nature of the algorithm might have a negative
                            performance impact!
        - "shorten"         only display the first character of each directory in
                            the path
        - "truncate"        truncates the start of the path when the whole path will
                            not fit. To increase the gap between the path and the edge,
                            set truncate to number `truncate = 3`
        - "filename_first"  shows filenames first and then the directories
```

I don't think filename_first is going to be composable in any way. The reason for that is it's the only option with the filename in front.

What do you guys want to do with this? I can document that this option is not composable with any of the other options for now. If we want it to be make it composable, then the internals need to be refactored. You can't do any logic on a string. So maybe returning a tuple instead `(filename, other directories)` could be an option.

Tronikelis:
> What do you guys want to do with this?

What vscode does and what I expected to happen

<details>
<summary>image</summary>


image


</details>

But it would be awesome if the current truncate behavior was not changed, i.e. the path **end** is visible, not the **start**

Basically, I would want it to be truncated like it currently is:

<details>
<summary>[truncate start] Ideally</summary>

image

</details>

<details>
<summary>[truncate end] how vscode does it, but not my ideal solution</summary>

image


</details>

jamestrew:
In terms of general composability, I think moving the `filename_first` logic to the end of the `transform_path` function could help in most cases I believe.

For flipping which side of the path (start vs end) is truncated, we would need to tweak the `truncate` logic to check for whether `filename_first` exists in `path_display`. The actual `truncate` function has a `direction` param which I'm guessing here controls the start vs end.

alycklama:
I created a new PR fixing this issue:
https://github.com/nvim-telescope/telescope.nvim/pull/3065

Thanks for reporting! @Tronikelis

Tronikelis:
@alycklama awesome 🚀

Leiyi548:
@alycklama Telescope git_files not useful

alycklama:
> @alycklama Telescope git_files not useful

Could you be a bit more specific? What is not useful and why is it not useful?

A quick check on my side makes me believe everything works as intended:

image

Leiyi548:
> > Telescope git_files not useful
> 
> @Leiyi548 Could you be a bit more specific? What is not useful and why is it not useful?
> 
> A quick check on my side makes me believe everything works as intended:
> 
> image
This is my config

image
image

alycklama:
@Leiyi548 Would you be able to post a screenshot of what you see as well?

Leiyi548:
> @Leiyi548 Would you be able to post a screenshot of what you see as well?

image

alycklama:
@Leiyi548 I can't reproduce it. Are you sure you are on the latest commit? Please check and report back.

Leiyi548:
> @Leiyi548 I can't reproduce it. Are you sure you are on the latest commit? Please check and report back.

maybe I use neovim in windows11?

alycklama:
> maybe I use neovim in windows11?

From your screenshot it seems you're not using the latest code, hence it'll never show correctly. I asked you earlier if you could check which commit you are at. If you can't answer this question, then unfortunately, I won't be able to help you further.

Leiyi548:
> > maybe I use neovim in windows11?
> 
> From your screenshot it seems you're not using the latest code, hence it'll never show correctly. I asked you earlier if you could check which commit you are at. If you can't answer this question, then unfortunately, I won't be able to help you further.

image

jamestrew:
@Leiyi548 please open a bug report for this and provide info there.

FernandoBasso:
I don't see `filename_first` in the `:help telescope`, neither any mention to `reverse_directories`. NVIM v0.10.1 on Arch Linux.

jamestrew:
It's only available on the master branch currently.

kristiyanraykov:
Would be great if this was added as extension, or if someone created a branch from "0.1.x" and cherry-picked those commits so we can use this without switching to the "unstable" master branch.

PR #3011, opened 2024-03-24
build: rework ci tests to include windows

### LINKED ISSUE #3000: add windows to test ci
could really help with all the windows path issues.
I definitely want to start writing more unit tests for all the path stuff as well anyways.
gonna loosely assign myself but I'm a total noob with github actions so I'll have some reading and exploring to do.

jamestrew:
- add windows runner as part of test ci and fix any windows sensitive tests
- tweak Makefile & minimal_init so running tests locally isn't dependent on local directory structure (I wanted to do this with docgen as well but it's so flaky and impossible on nightly neovim anyways)

close #3000 

ps. sorry if you got spammed by failed ci notifs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,48 +9,33 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-22.04, macos-12, windows-2022]
+        rev: [nightly, v0.9.0]
         include:
           - os: ubuntu-22.04
-            rev: nightly/nvim-linux64.tar.gz
-            manager: sudo apt-get
-            packages: -y ripgrep
-          - os: ubuntu-22.04
-            rev: v0.9.0/nvim-linux64.tar.gz
-            manager: sudo apt-get
-            packages: -y ripgrep
+            install-rg: sudo apt-get update && sudo apt-get install -y ripgrep
           - os: macos-12
-            rev: nightly/nvim-macos.tar.gz
-            manager: brew
-            packages: ripgrep
-          - os: macos-12
-            rev: v0.9.0/nvim-macos.tar.gz
-            manager: brew
-            packages: ripgrep
+            install-rg: brew update && brew install ripgrep
+          - os: windows-2022
+            install-rg: choco install ripgrep
+
     steps:
-      - uses: actions/checkout@v3
-      - run: date +%F > todays-date
-      - name: Restore from todays cache
-        uses: actions/cache@v3
+      - uses: actions/checkout@v4
+
+      - uses: rhysd/action-setup-vim@v1
         with:
-          path: _neovim
-          key: ${{ runner.os }}-${{ matrix.rev }}-${{ hashFiles('todays-date') }}
+          neovim: true
+          version: ${{ matrix.rev }}
 
       - name: Prepare
         run: |
-          ${{ matrix.manager }} update
-          ${{ matrix.manager }} install ${{ matrix.packages }}
-          test -d _neovim || {
-            mkdir -p _neovim
-            curl -sL "https://github.com/neovim/neovim/releases/download/${{ matrix.rev }}" | tar xzf - --strip-components=1 -C "${PWD}/_neovim"
-          }
-          mkdir -p ~/.local/share/nvim/site/pack/vendor/start
-          git clone --depth 1 https://github.com/nvim-lua/plenary.nvim ~/.local/share/nvim/site/pack/vendor/start/plenary.nvim
-          git clone --depth 1 https://github.com/nvim-tree/nvim-web-devicons ~/.local/share/nvim/site/pack/vendor/start/nvim-web-devicons
-          ln -s $(pwd) ~/.local/share/nvim/site/pack/vendor/start
+          ${{ matrix.install-rg }}
+          rg --version
+
+          git clone --depth 1 https://github.com/nvim-lua/plenary.nvim ../plenary.nvim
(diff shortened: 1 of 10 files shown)


jamestrew:
macos nightly failure: https://github.com/rhysd/action-setup-vim/issues/30

.github/workflows/ci.yml
```diff
@@ -11,46 +11,56 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
+
       - run: date +%F > todays-date
+
+      - uses: rhysd/action-setup-vim@v1
+        with:
+          neovim: true
+          version: ${{ matrix.rev }}
```
Conni2461, line 46: i am not the biggest fan of using third party actions for this simple stuff, imo it "doesnt give us" that much and we are dependent on that action to be maintained when issues arise. your linked issue is the perfect example. we cant fix something like this ourselves. but yeah we can still do that

.github/workflows/ci.yml
```diff
@@ -11,46 +11,56 @@ jobs:
...
       - name: Restore from todays cache
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
-          path: _neovim
-          key: ${{ runner.os }}-${{ matrix.rev }}-${{ hashFiles('todays-date') }}
+          path: deps
+          key: ${{ runner.os }}-deps-${{ matrix.version }}-${{ hashFiles('todays-date') }}
+          restore-keys: |
+            ${{ runner.os }}-deps-${{ matrix.version }}-
```
Conni2461, line 54: can we not do this? if cache fails for deps i prefer if we download them again or even better, lets not cache plenary and webicons, we can spare these 4,5 seconds of the clone. it doesnt need to be cached at all. We always wanna run against the last plenary version

Conni2461:
Yeah generally speaking we can add a windows test runner. Currently no checks did run. is this correct? 

image

> tweak Makefile & minimal_init so running tests locally isn't dependent on local directory structure (I wanted to do this with docgen as well but it's so flaky and impossible on nightly neovim anyways)

This is the only thing i actually dont like. We never wanna download dependencies for the user who runs the tests locally. We only wanna do this in an controlled environment, github runner. This was never an issue in the past 2.5 years so i dont see a reason to change that, users who actually have a different directory structure, should probably just manually edit the minimal rc. Thats far simpler than dealing with a "complex" Makefile which is a pain (see telescope-fzf-native.nvim, which only has to compile a singe c source file ...)

Makefile
```diff
@@ -1,8 +1,37 @@
...
+$(DEPS_DIR):
+	@mkdir -p $@
+
+plenary: | $(DEPS_DIR)
+	$(call git_clone_or_pull,$(PLENARY_DIR),https://github.com/nvim-lua/plenary.nvim)
+
+test-deps: plenary | $(DEPS_DIR)
+	$(call git_clone_or_pull,$(DEVICONS_DIR),https://github.com/nvim-tree/nvim-web-devicons)
+
+test: plenary
```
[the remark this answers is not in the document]
jamestrew: I just recently had a PR from a contributor who couldn't run `make test` since their directory structure wasn't compatible with the existing `minimal_init.lua` so I'm not sure if I agree with the sentiment that this doesn't benefit anyone.
And local tests now depend on whatever version of plenary you currently have (eg. if someone's pinned plenary on some old commit). I think this is less consistent with the testing we do in the CI. 

This gets even worse for docgen. Even if we didn't have the constraint of the pinned version of tree-sitter-lua not working on nightly, this is a dependency contributors have to manage anyways with if they want to docgen locally. Which nobody does currently largely due to this limitation even before docgen broke on nightly.

.github/workflows/ci.yml
jamestrew: Yeah I know what you mean.
For what it's worth, there are some other notable vim/neovim plugins using this action and this issue of the neovim nightly issue on macos got resolved in like 3 days. Our macos nightly job has been broken just as long and still is.

Makefile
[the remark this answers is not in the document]
Conni2461: nobody ever ran docgen locally. It was always documented in CONTRIBUTING and people always had questions especially if CI for some reason didnt ran. thats why i started just checking out finished PRs and running it myself.

> I think this is less consistent with the testing we do in the CI.

Doesnt matter, we have CI that says it either works or not. Same for stylua, or luacheck. if you have an old version for that, CI also fails. Should we now introduce nix to work around this issue (for the 0.1% of linux ppl that use nix). I dont think so.

You can never satisfy everyone and having less stuff to maintain and not being relying on a internet connection just to run tests, seems just weird. Each make test run requires this now, which is just unnecessary.

Makefile
```diff
@@ -1,8 +1,37 @@
...
 lint:
 	luacheck lua/telescope
 
+# pinned tree-sitter-lua commit no longer works on nightly (0.10)
 docgen:
 	nvim --headless --noplugin -u scripts/minimal_init.vim -c "luafile ./scripts/gendocs.lua" -c 'qa'
+
+clean:
+	@rm -rf $(DEPS_DIR)
+
```
Conni2461, line 37: tailing newline

.github/workflows/ci.yml
Conni2461: downloading cache also takes a couple of seconds iirc, so yeah lets just drop this

.github/workflows/ci.yml
Conni2461: yeah sure. just saying not happy with stuff like that, because we cant act. Not that we shouldnt do it at all. windows support probably justifies it.

Conni2461:
thanks :)

PR #3013, opened 2024-03-25
test(utils): add unit test for `transform_path`

jamestrew:
Adds os-agnostic unit tests for `transform_path` testing and fixes a couple of general edge cases and Windows specific path separator related bugs.

I was able to run these tests against my CI changes in https://github.com/nvim-telescope/telescope.nvim/pull/3011 to confirm they pass for all OSs.

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -15,7 +15,7 @@ local get_status = require("telescope.state").get_status
 
 local utils = {}
 
-local iswin = vim.loop.os_uname().sysname == "Windows_NT"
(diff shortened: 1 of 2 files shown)


lua/telescope/utils.lua
```diff
@@ -53,7 +53,7 @@ utils.path_expand = function(path)
 
   path = path:gsub("%$([%w_]+)", vim.loop.os_getenv)
   path = path:gsub("/+", "/")
-  if iswin then
+  if utils.iswin then
```
jamestrew, line 56: convert all `/` into a `\`?
at least this is closer to `vim.fn.expand`.

PR #3014, opened 2024-03-27
Command History index selection

peter-lyr:
# Description

`:Telescope command_history<cr>`, when the selection is too long, the cmd last ran will not be the origin command. then the cmd will not running


peter-lyr:
like this:

image

Issue #3019, opened 2024-03-29
Make spellrepall work with spell_suggest


ilan-schemoul:
**Is your feature request related to a problem? Please describe.**
With "z=" you can use :spellrepall to repeat last spell fixing. But when using telescope spell_suggest and then :spellrepall says "No previous spell replacement" instead of reapplying last suggestion

**Describe the solution you'd like**
spellrepall working

**Describe alternatives you've considered**
Add a method to telescope equivalent to spellrepall

jamestrew:
If anyone has any ideas of implementing this, I'm all ears.
In my quick search, I didn't find anything like a register we can populate for `:spellrepall` or some other way to make this work without doing something hack-ier than what we already do (basically `ciw` and inserting the replacement word).

ilan-schemoul:
We can just use z{index}= instead of the hacky ciw #3022 implements this and it works.
Random thoughts :
Lua really is a blast I didn't know the language and I made like 6 PRs these past days.
Also I feel like spell suggestion don't get a lot of attention from vim/neovim users but it's quite important

PR #3022, opened 2024-03-29
Selecting a spell suggestion now uses z{index}= 

ilan-schemoul:
Selecting a spell suggestion now uses z{index}= instead of ciw.
This fixes `spellrepall` command not working
z{index}= is the same than typing z= and then typing the index with the native spell plugin

# Description

#3019 

# How Has This Been Tested?

Open a file such as a.md with "Hey yelterdai yelterdai was good" 
`Telescope spell_suggest` on the first mistyped yesterday choose one element then type `:spellrepall` and it will apply to the second element (and all others if applicable) 

**Configuration**:
NVIM v0.10.0-dev-2723+g981301d11
windows 11 WSL ubuntu 22
# Checklist:

### DIFF
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -1374,7 +1374,7 @@ internal.spell_suggest = function(opts)
 
           action_state.get_current_picker(prompt_bufnr)._original_mode = "i"
           actions.close(prompt_bufnr)
-          vim.cmd("normal! ciw" .. selection[1])
+          vim.cmd("normal! " .. selection.index .. "z=")
           vim.cmd "stopinsert"
         end)
         return true


jamestrew:
I don't think this works depending of the `spellcapcheck` setting and whether you're on the first word of the sentence.
`z=` seems to check for matches against `spellcapcheck` but we're just calling `spellsuggest` straight up (basically `capital=false`) so if `spellcapcheck` matches, the results from `z=` and `vim.fn.spellsuggest` would differ.

So we'd need to check the `<cword>` against `spellcapcheck` pattern and pass `capital=true` based on that result for this approach to work effectively. Trouble is, `spellcapcheck` needs more context that just the `<cword>` to work properly I think...

ilan-schemoul:
Oh so you mean
"againn you did it" I do "z=" suggestion is "Again" I do telescope spell_checks and it offers "again" and you want to avoid that user select "again" and is left with "Again" because of my fix ?
I mean personally I don't care for my use case but I see your point. If you have a suggestion I can implement I can try otherwise I guess I'lll have to maintain my fork (damn tinkering with my neovim config now requires me to maintain like 5 forks :( )

jamestrew:
Yeah but it's worse than that cuz it's not just a difference of capitalized "Again" versus "again". Maybe some of the first few words match but the order and the list of words could be completely different depending on the word.

For example, if I have "hillo world" and I `z=` on "hillo", option 9 is "Hill o" but option 9 in `spellsuggest("hillo")` is "halloo". Completely different words.

I don't have any great solutions for this. I think for `z=`, internally vim check the progressively checks prior lines if it matches `spellcapcheck` and gives the suggestion based on this.
I don't really want to transpile this C code into Telescope...

But fyi
> otherwise I guess I'lll have to maintain my fork

you don't need to maintain a fork of telescope for this or other similar minor changes. You can just copy/paste the code for the picker into your own config and tweak things to your hearts content.

ilan-schemoul:
Hmm yeah it's a tricky thing. It could be documented that to the :spellr command working you need to `set nospellcapcheck=no` and then copy the snippet in your config
And yeah the c function that make the spell check with the full context so it can check with the cap option etc. is complex so ofc reimplementation is a nogo
Ideally there should be a neovim function which could be named `spellsuggestwithcontext()` that would take the position in the buffer of a word as an argument so it can provide suggestion contextually.

ilan-schemoul:
By the way for the record this is what I added to my config (+ the vim cmd `set spellcapcheck=no`)
```lua
vim.api.nvim_create_user_command('CustomTelescopeSpellSuggest', function()
  require('telescope.builtin').spell_suggest({
      local utils = require("telescope.utils")

        if selection == nil then
          utils.__warn_no_selection("builtin.spell_suggest")
          return
        end

        action_state.get_current_picker(prompt_bufnr)._original_mode = "i"
        actions.close(prompt_bufnr)
        vim.cmd("normal! " .. selection.index .. "z=")
        vim.cmd("stopinsert")
      end)
      return true
    end,
  })
end, {})
```

jamestrew:
Glad you got something working for yourself.
I'll close this PR until we come with a more robust solution. Thanks for trying though.

PR #3024, opened 2024-03-31
Fix over-normalization, only normalize when a filename starts with "~"

### LINKED ISSUE #3023: The mark picker over-normalises filenames, e.g., oil.nvim buffer
### Description

After #2859, all mark filenames are normalized. For example, marked filenames are changed from `{plugin}:///{path}` to `{plugin}:/{path}`, e.g., fugitive and oil buffers. This causes a problem where selecting the buffer does not open the intended marked buffer. 

### Neovim version

```markdown
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1703358377
```

macOS 14.4.1

master

### checkhealth telescope

```markdown

- OK plenary installed.

- OK rg: found ripgrep 13.0.0
- OK fd: found fd 9.0.0

```

### Steps to reproduce

1. `nvim -u minimal.lua`
2.  `:Oil`
3.  `mA`
4.  `:Telescope marks`
5.  select mark `A`

### Expected behavior

open the originally marked oil buffer

### Actual behavior

a new buffer `oil:/{path}` opened instead of `oil:///{path}`

### Minimal config

```Lua
-- save as repro.lua
-- run with nvim -u repro.lua
-- DO NOT change the paths

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "runtime", "cache" }) do
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    lazypath,
  })
end

-- install plugins
local plugins = {
  {
    "stevearc/oil.nvim",
    config = function()
      require("oil").setup({
        -- add any needed settings here
      })
    end,
  },
  {
    dependencies = { 'nvim-lua/plenary.nvim' }
  }
}
  root = root .. "/plugins",
})
```

gzbfgjf2:
# Description

Fixes #3023

This commit maintains the normalization of `~/{path}` as intended by #2859, but stops the normalization process for other paths to avoid unintended effects.

## Type of change

# How Has This Been Tested?

Original marked oil and fugitive buffers are opened as expected.

**Configuration**:
```
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1703358377
```
```
macOS 14.4.1
```
# Checklist:

- [ n/a] I have commented my code, particularly in hard-to-understand areas
- [ n/a] I have made corresponding changes to the documentation (lua annotations)

### DIFF
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -1114,9 +1114,7 @@ internal.marks = function(opts)
       -- same format to :marks command
       local line = string.format("%s %6d %4d %s", mark, lnum, col - 1, name)
       local filename = v.file or bufname
-      if filename:sub(1, 1) == '~' then
-        filename = vim.fs.normalize(filename)
-      end
+      filename = utils.path_expand(filename)
       local row = {
         line = line,
         lnum = lnum,
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -1113,11 +1113,15 @@ internal.marks = function(opts)
       local name = cnf.name_func(mark, lnum)
       -- same format to :marks command
       local line = string.format("%s %6d %4d %s", mark, lnum, col - 1, name)
+      local filename = v.file or bufname
+      if filename:sub(1, 1) == '~' then
+        filename = vim.fs.normalize(filename)
+      end
       local row = {
         line = line,
         lnum = lnum,
         col = col,
-        filename = vim.fs.normalize(v.file or bufname),
+        filename = filename,
       }
       -- non alphanumeric marks goes to last
       if mark:match "%w" then
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -39,6 +39,10 @@ utils.path_expand = function(path)
     path = { path, { "string" } },
   }
 
+  if utils.is_uri(path) then
+    return path
+  end
+
   if path:match "^[%%#<]" then
     path = vim.fn.expand(path)
   end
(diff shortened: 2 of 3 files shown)


jamestrew:
Thanks, this actually brings up an interesting question around how `vim.fs.normalize` should handle URIs. Similar functions like Python's `os.path.normpath` also will mangle URIs so I wouldn't be surprised if the Neovim core team deemed this behavior as "expected". But at the same time, getting a mangled path doing `vim.fs.normalize(vim.api.nvim_buf_get_name(0))` seems pretty undesirable as well.

I think my approach in this case is to abandon the use of `vim.fs.normalize` here and instead we use our own `utils.path_expand`. Then we'll check for `utils.is_uri` inside `utils.path_expand`earlier on and do an early return if it is a URI. How does that sound?

gzbfgjf2:
> Thanks, this actually brings up an interesting question around how `vim.fs.normalize` should handle URIs. Similar functions like Python's `os.path.normpath` also will mangle URIs so I wouldn't be surprised if the Neovim core team deemed this behavior as "expected". But at the same time, getting a mangled path doing `vim.fs.normalize(vim.api.nvim_buf_get_name(0))` seems pretty undesirable as well.
> 
> I think my approach in this case is to abandon the use of `vim.fs.normalize` here and instead we use our own `utils.path_expand`. Then we'll check for `utils.is_uri` inside `utils.path_expand`earlier on and do an early return if it is a URI. How does that sound?

Sounds great! Could I proceed with implementing this using the approach you've suggested?

jamestrew:
Yes please, thanks!

gzbfgjf2:
Closed in favor of #3029.

PR #3036, opened 2024-04-08
refactor: add types to lsp-related pickers

mikavilpas:
# Description

This change
- fixes the lua_ls language server not providing some LSP features documentation and autocompletion when I used `require("telescope.builtin").lsp_definitions({ jump_type = "vsplit" })` in my config (link to my config: https://github.com/mikavilpas/dotfiles/commit/0f6c0e58d8af0e0350ff57256d8139d4f09dc744)
- adds working documentation (lsp hover) and autocompletion (lsp) for parameters

## Type of change

# How Has This Been Tested?

As the change only changes the lua annotations, I have tested that the LSP features are now available. Details below:

## The hover documentation I got before this change

image
I also got no other features such as autocompletion.

## The effect after applying the change

Hover documentation now works
image

Autocompletion also works for sum types
image

**Configuration**:
* Neovim version (nvim --version): `NVIM v0.10.0-dev-2799+ge016f5bee`
* Operating system and version: osx 14.4.1

# Checklist:

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -413,97 +413,115 @@ builtin.jumplist = require_on_exported_call("telescope.builtin.__internal").jump
 --
 --
 
+---@class TelescopeBuiltinLspReferencesOpts
+---@field include_declaration? boolean include symbol declaration in the lsp references (default: true)
+---@field include_current_line? boolean include current line (default: false)
+---@field jump_type? "tab" | "tab drop" | "split" | "vsplit" | "never" how to goto reference if there is only one and the definition file is different from the current file, values: "tab", "tab drop", "split", "vsplit", "never"
+---@field show_line? boolean: show results text (default: true)
+---@field trim_text? boolean: trim results text (default: false)
+---@field file_encoding? string: file encoding for the previewer
+
 --- Lists LSP references for word under the cursor, jumps to reference on `<cr>`
----@param opts table: options to pass to the picker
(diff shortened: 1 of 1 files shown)


mikavilpas:
Just saw there already is https://github.com/nvim-telescope/telescope.nvim/pull/2856 although the contents seem to have a different focus:
- my PR is only about this one file, the other one is for many other files as well
- the PRs seem to have a different syntax. I'm not sure if they have the same result or not.

I wonder what would be the best thing to do here. 🙂

Conni2461:
unfortunately we decided to build our vimdocumentation based on lua emmy comments, see https://github.com/nvim-telescope/telescope.nvim/blob/master/CONTRIBUTING.md#documentation-with-treesitter

its currently autogenerated and merging this would result in us loosing this feature, which is imo more important than having types for lsp (at this point in time). I know its nice but we basically have redo parts of the grammar and the docgen algorithm or we have to drop docgen.

mikavilpas:
I see - this must be the reason I was not getting any help from the LSP. I thought it was a syntax error and "easily fixable" 🙂 

I noticed there has already been some interest in adding support for a new grammar https://github.com/tjdevries/tree-sitter-lua/issues/52. If you're open to that idea in the future, it might be a promising option.

But I think that might take weeks or months. I think it's not worth it to keep this open at this point. I'll close these PRs, and in case that future materializes, I can always reopen/redo them.

PR #3047, opened 2024-04-13
feat(actions): New `select_vertical_if_wide_enough` action

bharam:
# Description

The action opens the selected item in a new vertical split if the current tab page has only one window with a listed buffer, and the window is wide enough (width > 160). Otherwise, defaults to horizontal split.

This is especially helpful when working with multiple tmux panes on a small laptop screen.

Fixes # (issue)

This action reduces the mental effort required to decide whether to open a selected item (e.g., help_tags) in a vertical or horizontal split, potentially eliminating the inconvenience of rearranging Vim windows to make panes more legible. 

## Type of change

# How Has This Been Tested?

- [x] Set `help_tags` picker mapping to the new action as below: 
```lua
require("telescope").setup({
	-- ...
	pickers = {
	    help_tags = {
	        mappings = {
	            i = { ["<CR>"] = actions.select_vertical_if_wide_enough, },
	            n = { ["<CR>"] = actions.select_vertical_if_wide_enough, }
	        }
	    }
	}
}
```

When there is only one window in the current tab page with an editor width greater than 160, the help opens in a vertical split. If there are multiple windows open or the editor width is less than or equal to 160, the help will open horizontally.

**Configuration**:

# Checklist:

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -301,6 +301,33 @@ actions.select_vertical = {
   end,
 }
 
+--- Perform 'vertical' or `horizontal` action based on the number of windows opened and
+--- its width, usually something like<br>
+---`:vnew <selection>` or `:new <selection>`
+---
+--- i.e. open the selection in a new vertical split if there is only one window with width > 160,
+--- and open in a new horizontal split otherwise
+---@param prompt_bufnr number: The prompt bufnr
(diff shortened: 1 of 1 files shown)


jamestrew:
The logic to decide horizontal vs vertical and the colwidth limit of 160 feels a little too arbitrary to include this into telescope core.

I think this is best left as just a part of your personal config.

PR #3058, opened 2024-04-18
dev: POC proposal on_selection_callback for builtins

haug1:
# Description

It would be nice to be able to re-use the core handling of the builtin functions only adding a simple callback function when there's a selected result. In case of selected result, the consumer could optionally override the default handling.

## Type of change

# How Has This Been Tested?

I'm not really attempting to merge this code as is, but it works for POC purposes. With some guiding feedback and confirmed interest I could flesh out a more proper PR.


haug1:
[Yes, I'm dumb.](https://github.com/nvim-telescope/telescope.nvim/wiki/Configuration-Recipes#performing-an-arbitrary-command-by-extending-existing-find_files-picker)

PR #3065, opened 2024-04-20
Moves filename_first to last position to fix issues with composing other options

alycklama:
This fixes the issue where some of the other options that are composable didn't display correctly. For now, `filename_first` needs to be the last option.

image

image


jamestrew:
Thanks!

PR #3071, opened 2024-04-23
feat(buffer-fuzzy-find): add fuzzy refine

kevintraver:
Add **fuzzy refine** mapping to **current buffer fuzzy** find.


kevintraver:
Closing because this might not make sense as a default, and because _fuzzy_refine_ can be added by using attach_mappings in ones individual config.

PR #3074, opened 2024-04-24
feat: support path_style in path_display option

### LINKED ISSUE #3073: Allow returning of path_style from path_display function
I like `filename_first` option of `path_display`, however I also want to be able to modify the path. So, I need to pass a function to `path_display`. However, I am not able to replicate the way `filename_first` looks, because I cannot pass `TelescopeResultsComment` hl group.

Allow returning optional second value from `path_display` function.

sarmong:
# Description

Allows returning optional second argument from `path_display` option that is the `path_style`.

Fixes #3073 

## Type of change

# How Has This Been Tested?'

Added test for general function handling.

**Configuration**:

# Checklist:

### DIFF
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -359,7 +359,8 @@ append(
   path_display can also be set to 'hidden' string to hide file names
 
   path_display can also be set to a function for custom formatting of
-  the path display. Example:
+  the path display. You can also return the second optional value
+  to set highlight groups. Example:
 
       -- Format path as "file.txt (path\to\file\)"
       path_display = function(opts, path)
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -359,14 +359,38 @@ append(
   path_display can also be set to 'hidden' string to hide file names
 
   path_display can also be set to a function for custom formatting of
-  the path display. You can also return the second optional value
-  to set highlight groups. Example:
+  the path display with the following signature
+
+  Signature: fun(opts: table, path: string): string, table?
+
+  The optional table is an list of positions and highlight groups to
+  set the highlighting of the return path string.
+
+  Example:
 
       -- Format path as "file.txt (path\to\file\)"
       path_display = function(opts, path)
         local tail = require("telescope.utils").path_tail(path)
         return string.format("%s (%s)", tail, path)
       end,
+
+      -- Format path and add custom highlighting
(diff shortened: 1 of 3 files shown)


lua/telescope/config.lua
```diff
@@ -359,7 +359,8 @@ append(
   path_display can also be set to 'hidden' string to hide file names
 
   path_display can also be set to a function for custom formatting of
-  the path display. Example:
+  the path display. You can also return the second optional value
+  to set highlight groups. Example:
```
jamestrew, line 363: I think we could be clearer about the type/shape of this optional second value.
Maybe something like this

```suggestion
  path_display can also be set to a function for custom formatting of
  the path display with the following signature

  Signature: fun(opts: table, path: string): string, table?

  The optional table is an list of positions and highlight groups to
  set the highlighting of the return path string.

  Example:

      -- Format path as "file.txt (path\to\file\)"
      path_display = function(opts, path)
        local tail = require("telescope.utils").path_tail(path)
        return string.format("%s (%s)", tail, path)
      end,

      -- Format path and add custom highlighting
      path_display = function(opts, path)
        local tail = require("telescope.utils").path_tail(path)
        path = string.format("%s (%s)", tail, path)

        local highlights = {
          {
            {
              0, -- highlight start position
              #path, -- highlight end position
            },
            "Comment", -- highlight group name
          },
        }

        return path, highlights
      end
```

lua/tests/automated/utils_spec.lua
```diff
@@ -297,4 +297,10 @@ describe("transform_path", function()
       new_relpath "init.lua lua/telescope"
     )
   end)
+
+  it("handles function passed to path_display", function()
+    assert_path(function(_, path)
```
jamestrew, line 302: I don't think the `assert_path` function supports function type `path_display`s yet.
So this new test case won't actually do anything. If you can add support for it, I'd appreciate it.

jamestrew (changes requested):

lua/tests/automated/utils_spec.lua
sarmong: I added it, but forgot to commit...

lua/telescope/config.lua
sarmong: Nice, thanks!

jamestrew (approved):
lgtm thanks!

PR #3082, opened 2024-04-30
Notify on empty quickfix and loclist

kevintraver:
Notify when there is no items in quickfix or loclist


jamestrew:
Thanks, makes sense to me!

PR #3091, opened 2024-05-07
Fix jump regression in LSP references action handler

adriangoransson:
# Description

This change restores the possibility to exclude the current line when invoking the `lsp_references` picker.

- `opts.include_current_line` is by default unset, so the previous equality check would fail unless the option was set explicitly.
- `vim.api.nvim_win_get_cursor()` returns both line and column, so it can't be directly compared with the returned line number.
- The actual comparison was expecting quickfix-like items, when in actuality, we were dealing with raw LSP location objects.

## Type of change

# How Has This Been Tested?

- Triggering `lsp_references` on a symbol which is only referenced in one other location jumps directly to that location.
- Triggering `lsp_references` on a symbol which is referenced in multiple other locations lists reference locations. References on the cursor line where the picker was triggered are excluded.
- Triggering `lsp_references` with `opts = { include_current_line = true }` yields the same behavior as it does today.

**Configuration**:
	> NVIM v0.9.5
	> Build type: Release
	> LuaJIT 2.1.1713773202

	> macOs 14.4.1 (23E224)

# Checklist:

- [ ] ~I have made corresponding changes to the documentation (lua annotations)~ N/A, I think?

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -124,17 +124,17 @@ local action_handlers = {
 }
 
 ---@param action string
----@param results table
+---@param locations table
 ---@param items table
 ---@param opts table
 ---@return table results, table items
-local apply_action_handler = function(action, results, items, opts)
+local apply_action_handler = function(action, locations, items, opts)
   local handler = action_handlers[action]
   if handler then
-    return handler(results, items, opts)
+    return handler(locations, items, opts)
   end
 
-  return results, items
+  return locations, items
 end
 
 ---@param action string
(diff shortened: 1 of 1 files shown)


jamestrew:
Thanks!
For the `offset_encoding` part, I think it's probably better to call `vim.lsp.util.locations_to_items` earlier, before calling `apply_action_handler`, then pass the return value of it to `apply_action_handler`.

This is what we used to do before my (terrible :sweat_smile:) refactor. See here: https://github.com/nvim-telescope/telescope.nvim/blob/b744cf59752aaa01561afb4223006de26f3836fd/lua/telescope/builtin/__lsp.lua#L25-L67

adriangoransson:
> For the `offset_encoding` part, I think it's probably better to call `vim.lsp.util.locations_to_items` earlier, before calling `apply_action_handler`, then pass the return value of it to `apply_action_handler`.

Thanks for the help (and for your work on the plugin! 🌟). I think I managed to solve it in an okay way, by letting action handlers receive and process _both_ the lsp locations as well as the qf items. It requires some discipline to keep the processing "synced" between the two, but it keeps the calling function (that doesn't yet know which representation it is going to use) simpler.

I added a couple of refactoring commits to the PR as well, but can drop those if you wish.

jamestrew:
thanks lgtm!

PR #3093, opened 2024-05-09
:bug: Windows: allow binary without extension

ras0q:
# Description

In Windows, if only `rg` without extension is installed and `rg.exe` is not for some reason, checkhealth had failed previously.
This PR allows binaries without extensions to be recognised in Windows.

**FIXME**: In Windows environment, `io.popen(binary .. " --version")` does not return any outputs. How can I fix it?

```plaintext
- OK rg: found 
- OK fd: found 
```

## Type of change

# How Has This Been Tested?

- [x] confirm that binaries without extensions are recognised in Windows

**Configuration**:
* Neovim version (nvim --version): v0.9.5
* Operating system and version: Windows 11

# Checklist:

### DIFF
--- a/lua/telescope/health.lua
+++ b/lua/telescope/health.lua
@@ -102,8 +102,9 @@ M.check = function()
           )
         end
       else
-        local eol = version:find "\n" or 1
-        ok(("%s: found %s"):format(package.name, version:sub(0, eol - 1) or "(unknown version)"))
+        local eol = version:find "\n"
+        local ver = eol and version:sub(0, eol - 1) or "(unknown version)"
+        ok(("%s: found %s"):format(package.name, ver))
       end
     end
   end
--- a/lua/telescope/health.lua
+++ b/lua/telescope/health.lua
@@ -45,10 +45,12 @@ local required_plugins = {
 local check_binary_installed = function(package)
   local binaries = package.binaries or { package.name }
   for _, binary in ipairs(binaries) do
-    if is_win then
+    local found = vim.fn.executable(binary) == 1
+    if not found and is_win then
       binary = binary .. ".exe"
+      found = vim.fn.executable(binary) == 1
     end
-    if vim.fn.executable(binary) == 1 then
+    if found then
       local handle = io.popen(binary .. " --version")
       local binary_version = handle:read "*a"
       handle:close()
@@ -100,7 +102,7 @@ M.check = function()
           )
         end
       else
-        local eol = version:find "\n"
+        local eol = version:find "\n" or 1
         ok(("%s: found %s"):format(package.name, version:sub(0, eol - 1) or "(unknown version)"))
       end
     end


lua/telescope/health.lua
```diff
@@ -100,7 +102,7 @@ M.check = function()
           )
         end
       else
-        local eol = version:find "\n"
+        local eol = version:find "\n" or 1
         ok(("%s: found %s"):format(package.name, version:sub(0, eol - 1) or "(unknown version)"))
```
jamestrew, line 106:
```suggestion
        local eol = version:find "\n"
        local ver = eol and version:sub(0, eol - 1) or "(unknown version)"
        ok(("%s: found %s"):format(package.name, ver))
```

> FIXME: In Windows environment, io.popen(binary .. " --version") does not return any outputs. How can I fix it?

Not really a great solution but this should at least closer be to what this bit of code is trying to accomplish regarding failing to get version number.

jamestrew (commented):
Thanks, that's kinda weird that we need to check for `vim.fn.executable("rg.exe")` even when `rg` is installed.

ras0q:
Sorry,  I deleted the branch by mistake. I've reopened the PR.

ras0q:
Thank you for review!


> we need to check for `vim.fn.executable("rg.exe")` even when `rg` is installed.

I believe I wrote it to check for `rg.exe` only if `rg` was not found, is this code correct?

lua/telescope/health.lua
```diff
@@ -45,10 +45,12 @@ local required_plugins = {
 local check_binary_installed = function(package)
   local binaries = package.binaries or { package.name }
   for _, binary in ipairs(binaries) do
-    if is_win then
+    local found = vim.fn.executable(binary) == 1
+    if not found and is_win then
       binary = binary .. ".exe"
+      found = vim.fn.executable(binary) == 1
```
jamestrew, line 51:
```suggestion
      found = vim.fn.executable(binary .. ".exe") == 1
```
maybe instead of overriding `binary`, we can do this. Then the `io.popen()` will be without `.exe` and hopefully it will actually work and return the version number?

jamestrew:
> I believe I wrote it to check for rg.exe only if rg was not found, is this code correct?

yes, I think this is ok.
before I merge, I'd like it if you can try my suggestion. Maybe this can help with the issue of not getting the version number.

lua/telescope/health.lua
ras0q: In my environment, this code don't change the version result.
This line will be executed if `rg` does not exist.
Therefore,  when `rg` does not exist and `rg.exe` exists , does executing `rg --version` output unexpected error?

jamestrew: ok, let's just keep things as is then. thanks for trying.

jamestrew:
thanks!

PR #3096, opened 2024-05-11
fix(builtin.oldfiles): fix cwd_only for windows

chribi:
Remove backslash escaping in oldfiles picker, which was needed before the code was changed to use substring comparison instead of regex matching. (Commit b3ff5f33)

# Description

When using the `cwd_only` option for the `builtin.oldfiles` picker, no results were returned under windows.
The broken code replaces all backslashes by `\\`. This is not needed anymore, as the cwd-check now uses a substring match instead of matching via `matchstrpos()`.
Fixes # (issue)

## Type of change

# How Has This Been Tested?

- Switch to some directory under windows
- edit some files
- call `:Telescope oldfiles cwd_only=true`
- No results appear

**Configuration**:
* Operating system and version: Windows 10

# Checklist:


jamestrew:
Thanks!

PR #3097, opened 2024-05-12
fix: enable_preview true

### LINKED ISSUE #2966: Adding a previewer option to colorscheme picker
Currently the option to customise colorscheme is the 'enable_preview' parameter and it controls whether selecting a colorscheme from list changes colorscheme. And changing colorscheme causes both the previewer as well as the currently open buffers to change colorscheme. I would like the addition of a 'proper' previewer option so that I can set enable_preview=true and previewer=false and selecting a colorscheme changes the colorscheme of currenly open buffers while the previewer is kept disabled. Currently when I run :Telescope colorscheme previewer=false enable_preview=true, somehow both options are coupled together so that the enable_preview functionality (changing colorscheme of currently open buffers) is connected to previewer functionality.

'Decouple' previewer and enable_preview options to colorscheme picker so that they can be changed independently.

No alternatives as far as I thought.

**Additional context**
Nothing.

I know it's a bit confusing and I can provide more context if you'd like. Thanks for reading through this message.

MovieMaker93:
# Description

This fix aims to resolve the issue about the "enable_preview" feature that wasn't working. 
Now if you call:
`:Telescope colorscheme enable_preview=true`
while scrolling through the different color scheme options, the live preview is working (change color scheme accordingly)

Fixes #2966

## Type of change

# How Has This Been Tested?

I tried two different scenarios:

- :Telescope colorscheme enable_preview=true --> When scrolling through the different color schemes, the theme changes accordingly
- :Telescope colorscheme enable_preview=false -->  When scrolling through the different color schemes, the theme doesn't change

**Configuration**:
* Neovim version (nvim --version): 0.9.4
* Operating system and version: ubuntu 22.04

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -1000,45 +1000,20 @@ internal.colorscheme = function(opts)
     local bufnr = vim.api.nvim_get_current_buf()
     local p = vim.api.nvim_buf_get_name(bufnr)
 
-    -- don't need previewer
-    if vim.fn.buflisted(bufnr) ~= 1 then
-      local deleted = false
-      local function del_win(win_id)
-        if win_id and vim.api.nvim_win_is_valid(win_id) then
-          utils.buf_delete(vim.api.nvim_win_get_buf(win_id))
-          pcall(vim.api.nvim_win_close, win_id, true)
+    -- show current buffer content in previewer
+    previewer = previewers.new_buffer_previewer {
+      get_buffer_by_name = function()
+        return p
+      end,
+      define_preview = function(self)
+        if vim.loop.fs_stat(p) then
+          conf.buffer_previewer_maker(p, self.state.bufnr, { bufname = self.state.bufname })
+        else
+          local lines = vim.api.nvim_buf_get_lines(bufnr, 0, -1, false)
+          vim.api.nvim_buf_set_lines(self.state.bufnr, 0, -1, false, lines)
         end
-      end
-
-      previewer = previewers.new {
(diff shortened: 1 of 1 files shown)


jamestrew (changes requested):
I think the existing changes look good.
One thing I forgot to mention about that we should clean up is the existing code to set the previewer.

I _think_ the `if vim.fn.buflisted(bufnr) ~= 1 ...` part of the code path was a prior attempt to enable previews without a previewer, but it either never worked or no longer works. So we can just completely remove this part. It should be redundant now either way. You can also probably remove the `vim.cmd.colorscheme(entry.value)` inside the `define_preview` function in the `new_buffer_previewer` block as well as it's also redundant now.

MovieMaker93:
> I think the existing changes look good. One thing I forgot to mention about that we should clean up is the existing code to set the previewer.
> 
> I _think_ the `if vim.fn.buflisted(bufnr) ~= 1 ...` part of the code path was a prior attempt to enable previews without a previewer, but it either never worked or no longer works. So we can just completely remove this part. It should be redundant now either way. You can also probably remove the `vim.cmd.colorscheme(entry.value)` inside the `define_preview` function in the `new_buffer_previewer` block as well as it's also redundant now.

Okay, thanks for the feedback. I've removed both the **vim.fn.buflisted check** and the **cmd.colorscheme** command. In addition, just to be sure, I've tested it again, and it seems to work fine. Let me know if everything is correct on your end.

jamestrew (approved):

jamestrew:
Thanks! Looks good and works for me as well.

PR #3109, opened 2024-05-16
fix: adapt to Nvim deprecations in 0.10

### LINKED ISSUE #3108: Adapt to deprecated Neovim API
In Neovim 0.10 (just released), the following two widely-used API functions are deprecated:
* `vim.tbl_islist()` -> `vim.islist()`
* `vim.tbl_flatten(...)` -> vim.iter(...):flatten():totable()`

This means that nightly users will get a warning on every telescope action.

clason:
Fixes #3108

Note that `islist()` checks for a specific type of table: indexed by integers without holes (so `{ 1 = 'a', 3 = 'b' }` is _not_ a list); the intention is to ensure that `ipairs` and `#` (length) are well-defined. I did not check whether that is indeed intended in each case here; if not, consider simply replacing by `type(t) == 'table'`.

On a related note, the new `flatten()` only works correctly for lists; `{1,2,nil,3}` will _not_ be flattened to `{1,2,3}`. I don't know if that is an issue here, either. **EDIT** looks like it.
**EDIT2:** will be fixed upstream.

(Long-term, I would recommend moving to the new `vim.iter` API that allows efficiently chaining iterators. But this is of course a major refactoring effort.)

### DIFF
--- a/lua/telescope/builtin/__files.lua
+++ b/lua/telescope/builtin/__files.lua
@@ -12,7 +12,7 @@ local log = require "telescope.log"
 
 local Path = require "plenary.path"
 
-local flatten = vim.tbl_flatten
+local flatten = utils.flatten
 local filter = vim.tbl_filter
 
 local files = {}
--- a/lua/telescope/builtin/__git.lua
+++ b/lua/telescope/builtin/__git.lua
@@ -51,7 +51,7 @@ git.files = function(opts)
       prompt_title = "Git Files",
       __locations_input = true,
       finder = finders.new_oneshot_job(
-        vim.tbl_flatten {
+        utils.flatten {
           opts.git_command,
           show_untracked and "--others" or nil,
           recurse_submodules and "--recurse-submodules" or nil,
@@ -193,7 +193,7 @@ git.bcommits = function(opts)
 
   local title = "Git BCommits"
   local finder = finders.new_oneshot_job(
-    vim.tbl_flatten {
+    utils.flatten {
       opts.git_command,
       opts.current_file,
     },
@@ -234,7 +234,7 @@ git.bcommits_range = function(opts)
 
   local title = "Git BCommits in range"
   local finder = finders.new_oneshot_job(
-    vim.tbl_flatten {
+    utils.flatten {
       opts.git_command,
       line_range,
     },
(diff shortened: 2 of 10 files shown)


Conni2461:
thanks :) i am about to test these changes if they look good we can merge

clason:
Hold on for a second; the `vim.tbl_flatten()` change is not equivalent, and we will probably revert/bump the deprecation. If that happens, I'll drop those changes for now and just do `islist`.

If you can in the meantime check whether you really _want_ to check for a list (instead of generic table or array -- which has a separate function now), that would be very helpful.

clason:
Also, I'm not happy about having to duplicate the shims, but I wanted to avoid pulling in (heavy!) modules where they weren't before.

jamestrew (commented):
I think for everything but `plugin/telescope.lua`, it shouldn't make a difference to included utils. By the time these modules are included, utils would have been included.

But otherwise lgtm, thanks!

clason:
Ok, removed the duplication (except in /plugin), and fixed one more deprecation I found. @Conni2461 I think this is ready to merge.

Conni2461:
thanks :) i dont think we have to run with this for that long, i wanna do a 0.2 release sometime in the next 3 months (depending on how much freetime ill have) with support for 0.10 and newer.

clason:
Yes, I think 0.10 will be a good new baseline (and waiting a bit for the inevitable 0.10.1 makes sense as well ;))

clason:
Are you on the Nvim 0.10 **release** or some old nighly?

rmccord7:
Having issues on nightly 

image

built_in grep_string throwing this error.

E5108: Error executing lua: ...local/bin/build/nvim/share/nvim/runtime/lua/vim/iter.lua:236: flatten() requires a list-like table
stack traceback:
	[C]: in function 'error'
	...local/bin/build/nvim/share/nvim/runtime/lua/vim/iter.lua:236: in function 'flatten'
	...l/share/nvim/lazy/telescope.nvim/lua/telescope/utils.lua:23: in function 'flatten'
	...im/lazy/telescope.nvim/lua/telescope/builtin/__files.lua:233: in function 'v'
	...im/lazy/telescope.nvim/lua/telescope/builtin/__files.lua:641: in function 'v'
	.../nvim/lazy/telescope.nvim/lua/telescope/builtin/init.lua:580: in function <.../nvim/lazy/telescope.nvim/lua/telescope/builtin/init.lua:539>
	...2/rmccord/.config/nvim/lua/rm/plugins/telescope/util.lua:37: in function <...2/rmccord/.config/nvim/lua/rm/plugins/telescope/util.lua:36>

clason:
Will be fixed soon. Stick to releases, especially for the first few days/weeks after a release.

augustocdias:
@clason I'm on 0.10

PR #3113, opened 2024-05-17
fixup: adapt to deprecations in Nvim 0.10

clason:
vim.iter.flatten() is not a full replacement (does not support arrays
with `nil`); a fix may be backported but for now only use it on nightly
(0.10 has no warnings yet).


clason:
@Conni2461 sorry about that; I'm spread a bit thin these days so I keep papering over holes...

Issue #3118, opened 2024-05-19
Fix :Telescope lsp_references Failing to Jump Directly to Single Reference


odysseus0:
### Description

When there is only one reference, some time using the `:Telescope lsp_references` does not properly jump to the reference.

### Neovim version

```markdown
NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1713773202
```

### Operating system and version

macOS 14.5 23F79 arm64

### Telescope version / branch / rev

telescope 0.1.6

### checkhealth telescope

```markdown
telescope: health#telescope#check

Checking for required plugins ~
- OK plenary installed.
- WARNING nvim-treesitter not found. (Required for `:Telescope treesitter`.)

Checking external dependencies ~
- OK rg: found ripgrep 14.1.0
- OK fd: found fd 10.1.0

===== Installed extensions ===== ~
```

### Steps to reproduce

1. create a new file `main.go` and paste in the following code.

```go
package main

import "fmt"

func main() {
	hello()
}

func hello() {
	fmt.Println("Hello, World!")
}
```

2. Open the Go file with the specified Neovim configuration.
3. Move the cursor on top of `hello()` on the line `func hello() {`
4. Type in command `:Telescope lsp_references`

### Expected behavior

Cursor should jump to the only reference `hello()` inside

```go
func main() {
	hello()
}
```

### Actual behavior

The cursor won't move.

### Minimal config

```Lua
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
	vim.fn.system({
		"git",
		"clone",
		"--filter=blob:none",
		"https://github.com/folke/lazy.nvim.git",
		"--branch=stable",
		lazypath,
	})
end
vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
	{
		"nvim-telescope/telescope.nvim",
		tag = "0.1.6",
		dependencies = { "nvim-lua/plenary.nvim" },
	},
	{
		"neovim/nvim-lspconfig",
		config = function()
			require("lspconfig").gopls.setup({})
		end,
	},
	{ "williamboman/mason.nvim", build = ":MasonUpdate", opts = {} },
	{
		"williamboman/mason-lspconfig.nvim",
		dependencies = { "williamboman/mason.nvim" },
		opts = { ensure_installed = { "gopls" } },
	},
})
```

odysseus0:
Submitted a PR myself here #3120.

[...]

PR #3099, opened 2024-05-14
refactor(lsp): simplify `list_to_jump`

jamestrew:
clean up `list_to_jump` to work mostly with `vim.lsp.util.locations_to_items` output from the get-go

PR #3120, opened 2024-05-20
fix(lsp): correct handling of `find_references` (#3118)

odysseus0:
## Description

This PR resolves a bug reported in #3118 in the [`list_or_jump`](https://github.com/nvim-telescope/telescope.nvim/blob/0c12735d5aff6a48ffd8111bf144dc2ff44e5975/lua/telescope/builtin/__lsp.lua#L144) function of the [`find_references`](https://github.com/nvim-telescope/telescope.nvim/blob/0c12735d5aff6a48ffd8111bf144dc2ff44e5975/lua/telescope/builtin/__lsp.lua#L210) feature for the default option "excluding the current position". Previously, the function incorrectly assumed that the `locations` and `items` tables were in the same order, which is not the case as [`vim.lsp.util.locations_to_items`](https://github.com/neovim/neovim/blob/b16b287b8f36a8e25b27f3009d7fdd0d39348894/runtime/lua/vim/lsp/util.lua#L1720) sorts `items`. This resulted in incorrect exclusions, affecting the functionality when handling multiple references (wrong exclusion) or a single reference (not jumping at all).

## Changes Introduced

- Implemented a sorting mechanism for `locations` based on the order of `items` returned by [`vim.lsp.util.locations_to_items`](https://github.com/neovim/neovim/blob/b16b287b8f36a8e25b27f3009d7fdd0d39348894/runtime/lua/vim/lsp/util.lua#L1720).
- The sorting is based on a stable identifier generated from both `locations` and `items`, ensuring that they match correctly by index.

## Rationale

The chosen method leverages the output from [`vim.lsp.util.locations_to_items`](https://github.com/neovim/neovim/blob/b16b287b8f36a8e25b27f3009d7fdd0d39348894/runtime/lua/vim/lsp/util.lua#L1720) to ensure that `locations` are reordered to match the sorted `items`. This approach treats the Neovim utility as a black box and only assumes the stability of location and item type definitions, such as [`lsp.Location`](https://github.com/neovim/neovim/blob/b16b287b8f36a8e25b27f3009d7fdd0d39348894/runtime/lua/vim/lsp/_meta/protocol.lua#L24) and [`lsp.LocationLink`](https://github.com/neovim/neovim/blob/b16b287b8f36a8e25b27f3009d7fdd0d39348894/runtime/lua/vim/lsp/_meta/protocol.lua#L1697), so that it won't depend on the exact implementation of the utility. 

The alternative approach would be replicating the sorting logics in `vim.lsp.util.locations_to_items`, but that would break the abstraction of neovim utility and increases risk of the implementation breaking due to internal changes in the vim.lsp.util.locations_to_items function.

This update ensures that the `list_or_jump` function behaves as expected, particularly in scenarios where the current location should be excluded from `find_references` results.

## Testing

Use the same config in #3118.

Below is a minimal example that tests the functionality.

```go
package main

import "fmt"

func main() {
	hello()

	a := 1
	fmt.Println(a)
	fmt.Println(a + 1)
}

func hello() {
	fmt.Println("Hello, World!")
}
```

It now correctly jumps to the reference in the above example for single reference `hello`. Before it does not work.

It also properly lists both references for `a` for the case of multiple references.

## Checklist

- [x]  My code follows the style guidelines of this project (stylua)
- [x]  I have performed a self-review of my own code
- [x]  I have commented my code, particularly in hard-to-understand areas
- [x]  I have made corresponding changes to the documentation (lua annotations)

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -150,29 +150,38 @@ local function reorder_locations_by_items(locations, items)
     item_index_map[identifier] = index
   end
 
-  -- Map locations to their corresponding item index
-  local index_to_location = {}
+  -- Map locations to their corresponding item index and sort them
+  local sorted_locations = {}
   for _, location in ipairs(locations) do
-    local identifier = location.uri:gsub("^%w+://", "")
-      .. ":"
-      .. location.range.start.line
-      .. ":"
-      .. location.range.start.character
+    local uri, line, character
+    if location.uri then -- Handle Location
+      uri = location.uri:gsub("^%w+://", "")
+      line = location.range.start.line
+      character = location.range.start.character
+    else -- Handle LocationLink
+      uri = location.targetUri:gsub("^%w+://", "")
+      line = location.targetSelectionRange.start.line
+      character = location.targetSelectionRange.start.character
+    end
+    local identifier = uri .. ":" .. line .. ":" .. character
     local item_index = item_index_map[identifier]
+
     if item_index then
-      index_to_location[item_index] = location
+      sorted_locations[item_index] = location
     else
-      error("No corresponding item found for location with identifier: " .. identifier)
+      vim.api.nvim_err_writeln("Error: No corresponding item found for location with identifier: " .. identifier)
     end
   end
 
-  -- Collect sorted locations based on the item index order
-  local sorted_locations = {}
-  for _, location in ipairs(index_to_location) do
-    table.insert(sorted_locations, location)
+  -- Ensure sorted locations are dense
+  local compacted_locations = {}
+  for i = 1, #sorted_locations do
+    if sorted_locations[i] then
+      table.insert(compacted_locations, sorted_locations[i])
+    end
   end
 
-  return sorted_locations
+  return compacted_locations
 end
 
 ---@param action string
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -137,6 +137,44 @@ local apply_action_handler = function(action, locations, items, opts)
   return locations, items
 end
 
+--- Reorders locations to match the order of corresponding items.
+--- `items` are expected to be the result of a previous call to `vim.lsp.util.locations_to_items(locations)`.
+---@param locations lsp.Location[]|lsp.LocationLink[]
+---@param items vim.lsp.util.locations_to_items.ret[]
+---@return lsp.Location[]|lsp.LocationLink[]
+local function reorder_locations_by_items(locations, items)
+  -- Create a map from item identifiers to their index positions
+  local item_index_map = {}
+  for index, item in ipairs(items) do
+    local identifier = item.filename .. ":" .. (item.lnum - 1) .. ":" .. (item.col - 1)
+    item_index_map[identifier] = index
+  end
+
+  -- Map locations to their corresponding item index
+  local index_to_location = {}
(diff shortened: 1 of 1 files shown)


odysseus0:
Just got started using Neovim last Friday. Still a newbie. Thank you guys so much for making these killer plugins.

odysseus0:
Wondering what else I need to do to get the PR merged

Conni2461:
4 days open, give me at least time to look at it and try to understand the problem why it doesnt work. this code has been modified much more often than I would like and there is always the next PR. i actually need to think about this and do some testing. be patient, i will look at it. Thanks for understanding :)

adriangoransson:
> 4 days open, give me at least time to look at it and try to understand the problem why it doesnt work. this code has been modified much more often than I would like and there is always the next PR. i actually need to think about this and do some testing. be patient, i will look at it. Thanks for understanding :)

To give you some context, these might be useful:

- Initial "fix" that also contained the bug this PR aims to solve (thanks @odysseus0 for cleaning up my mess!) in https://github.com/nvim-telescope/telescope.nvim/pull/3091
- subsequent refactor in https://github.com/nvim-telescope/telescope.nvim/pull/3099
- which was later reverted in https://github.com/nvim-telescope/telescope.nvim/pull/3101 because it depended on nightly.

My take is that maybe #3099 maybe should be merged again, because just a few days after it was reverted 0.10 was released with the changes it depended on.

Conni2461:
Again it will just take a couple more days to get this issue fixed. #3099 will not be merged again, we still have 0.9 support.

odysseus0:
Don't mean to rush you at all! Just trying to figure out if there is something else I can or need to do on my end! Thank you so much for all your work.

jamestrew:
Sorry @Conni2461 I can take this one considering I sort of started this mess.

I'm thinking I want to implement something closer to #3099 but since we don't have access to the `user_data` field to get `Location|LocationLink` info, we can just build this data ourselves at least as a carryover till we drop 0.9 support.

```lua
--- convert `item` type back to something we pass to `vim.lsp.util.jump_to_location`
---@param item vim.lsp.util.locations_to_items.ret
---@param offset_encoding string|nil utf-8|utf-16|utf-32
---@return lsp.Location # not exactly the same, only important parts for `jump_to_location`
local function item_to_location(item, offset_encoding)
  local location = {
    uri = vim.uri_from_fname(item.filename),
    range = {
      start = {
        line = item.lnum,
      },
    },
  }

  -- TODO: real implementation will need to use `item.line` and offset_encoding to
  -- get the real start.character value
  location.range.start.character = item.col
  return location
end

-- ...
-- ...
-- many lines later where to use `jump_to_location

      local location = item_to_location(item, offset_encoding)
      vim.lsp.util.jump_to_location(location, offset_encoding, opts.reuse_win)
    else
      pickers
        .new(opts, {
```

I think this could be simpler than sorting `Location|LocationLink` items to match the order of `items`. And it let's us generally simplify the `action_handler` code.

I can flesh out this idea myself a little bit later tonight.

jamestrew:
here it is: https://github.com/nvim-telescope/telescope.nvim/pull/3137

odysseus0:
🤣 jesus I am being such an idiot. If I can generate the unique 1-1 mapping between them, that means I have enough information to just reconstruct the whole thing. Yes this approach makes more sense.

Conni2461:
would close this in favor of #3137 its the simpler implementation.

still thanks for looking into the issue :)

PR #3131, opened 2024-05-22
feat(builtin.git_diff): new picker for diff hunks

antonk52:
# Description

This PR adds a picker for git diff hunks. The new picker may seem similar to git_status picker at first. However, it allows to view multiple hunks in a single file as separate entries; when a hunk is selected it jumps the cursor to the changes line instead of a first line of a file. You can also supply additinal arguments such as `--ignore-all-space` which can make hunks seem even more granular and easier to review when navigating via telescope. I have been testing this picker in my configuration for the past few months and think that other telescope users will find this useful.

I understand that at the time of first publishing this PR is in a pretty raw shape as it requires documentation updates and is currently using vim.fn.systemlist instead of an entry maker as entries cannot be inferred from a single line of output. I could not figure out a way to make the way the git diff output is parsed to work similar to how other pickers work. I if you think this is a welcome change I am happy to make appropriate changes to get this PR to a mergeable state. 

Demo:

https://github.com/nvim-telescope/telescope.nvim/assets/5817809/d0966aaf-ba35-465f-b210-12a10eaf5cf0

Look forward to see what you think about this feature.
## Type of change

# How Has This Been Tested?

- [x] Has hunks
  - Make changes in a git repository
  - run `:Teelscope git_diff`
  - expect to see a picker with diff hunks (multiple entries for single file with multiple hunks)
  - when an entry is selected it jumps to changed lines
- [x] Clean git repositry
  - run `:Teelscope git_diff`
  - no picker, error message ""

**Configuration**:
* Neovim version (nvim --version): v0.10
* Operating system and version: macos 14.4

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -4046,6 +4046,12 @@ previewers.git_file_diff()              *telescope.previewers.git_file_diff()*
 
 
 
+previewers.git_hunk_diff()              *telescope.previewers.git_hunk_diff()*
+    A previewer that shows the current diff hunk. Used in git_diff.
+
+
+
+
 previewers.display_content()          *telescope.previewers.display_content()*
     A deprecated way of displaying content more easily. Was written at a time,
     where the buffer_previewer interface wasn't present. Nowadays it's easier
--- a/lua/telescope/builtin/__git.lua
+++ b/lua/telescope/builtin/__git.lua
@@ -429,6 +429,112 @@ git.status = function(opts)
     :find()
 end
 
+git.diff = function(opts)
+  if opts.is_bare then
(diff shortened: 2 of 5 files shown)


jamestrew:
I'm sorry but we're not really accepting any new pickers at the moment. We're kinda struggling with maintaining all the ones we already have. See https://github.com/nvim-telescope/telescope.nvim/issues/1228

I would recommend packaging this as an extension to telescope though. See https://github.com/nvim-telescope/telescope.nvim/blob/c2ce039188e22288449e43b77710036e69a45a70/developers.md#bundling-as-extension

nishantpillai5:
@antonk52 Would love this as an extension

PR #3137, opened 2024-05-26
fix(lsp.reference): single reference direct jump

jamestrew:
follow up to https://github.com/nvim-telescope/telescope.nvim/pull/3099
> clean up `list_to_jump` to work mostly with `vim.lsp.util.locations_to_items` output from the get-go

and closes https://github.com/nvim-telescope/telescope.nvim/issues/3118

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -97,51 +97,59 @@ lsp.outgoing_calls = function(opts)
   calls(opts, "to")
 end
 
----@type { [string]: fun(results: table, items: table, opts: table): table, table }
-local action_handlers = {
-  ["textDocument/references"] = function(results, items, opts)
-    if not opts.include_current_line then
-      local retresults = {}
-      local retitems = {}
-
-      for i, item in pairs(items) do
-        if
-          not (
-            item.filename == vim.api.nvim_buf_get_name(opts.bufnr)
-            and item.lnum == vim.api.nvim_win_get_cursor(opts.winnr)[1]
-          )
-        then
-          table.insert(retresults, results[i])
-          table.insert(retitems, items[i])
-        end
-      end
-
-      return retresults, retitems
-    end
-
-    return results, items
-  end,
-}
-
----@param action string
----@param locations table
----@param items table
----@param opts table
----@return table results, table items
-local apply_action_handler = function(action, locations, items, opts)
-  local handler = action_handlers[action]
-  if handler then
-    return handler(locations, items, opts)
-  end
-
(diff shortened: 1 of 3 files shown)


Conni2461:
i am going to review this tomorrow

lua/telescope/builtin/__lsp.lua
```diff
@@ -97,51 +97,74 @@ lsp.outgoing_calls = function(opts)
...
+---| "textDocument/definition"
+---| "textDocument/typeDefinition"
+---| "textDocument/implementation"
+
+-- luacheck: push ignore
+---@type { [telescope.lsp.list_or_jump_action]: fun(items: vim.lsp.util.locations_to_items.ret[], opts: table): vim.lsp.util.locations_to_items.ret[] }
+-- luacheck: pop
 local action_handlers = {
-  ["textDocument/references"] = function(results, items, opts)
+  ["textDocument/references"] = function(items, opts)
```
Conni2461, line 134: it doesnt make sense to have an "action_handers" table here, it has one entry, why would we want to construct a table for that.

we could just do this in `apply_action_handler`
```lua
if action == "textDocument/references" then
  -- stuff from function
end
return locations, items
```

Conni2461: its a fine idea for an abstraction but we dont need it here, we have 1 entry, we can still introduce an abstraction if we have like 3, 4, 5 entries

lua/telescope/builtin/__lsp.lua
```diff
@@ -97,51 +97,74 @@ lsp.outgoing_calls = function(opts)
...
----@param action string
+---@param action telescope.lsp.list_or_jump_action
 ---@param title string prompt title
 ---@param params lsp.TextDocumentPositionParams
 ---@param opts table
 local function list_or_jump(action, title, params, opts)
+  opts.reuse_win = vim.F.if_nil(opts.reuse_win, false)
+
+  local curr_filepath = vim.api.nvim_buf_get_name(opts.bufnr)
+
```
Conni2461, line 167: please remove empty line

Conni2461:
works for both 0.9 and 0.10, can be merged after two nits are addressed, thanks. i like that solution more than #3120

jamestrew:
thanks for reviewing!

jamestrew:
@lawrenceho that's unfortunate. I'll see what we can do to remedy this.

lawrenceho:
Thank you @jamestrew. By the way, I can create a PR for this by adopting the same approach used in `vim.uri_from_bufnr`.

PR #3139, opened 2024-05-26
feat: adds-option-entryseparatorhl

### LINKED ISSUE #3117: telescope results line color / highlight
The ability to change the color of the line next to the line numbers _(the red one in the screenshot)_ without effecting the normal text.

A new highlight group to change the color.

I changed `TelescopeResultsNormal`, but that also affects other text.

**Additional context**
Screenshot 2024-05-18 at 20 39 49

MovieMaker93:
# Description

Adds option to configure the highlight of the result separator with new hl group: **TelescopeResultsEntrySeparator**

Fixes #3117

## Type of change

# How Has This Been Tested?

I've tested it by changing both the foreground and background of the new hlgroup: **TelescopeResultsEntrySeparator**
First test by changing the background:
`:hi TelescopeResultsEntrySeparator guibg=#7f1425`
image
Second test by changing the foreground:
`:hi TelescopeResultsEntrySeparator guifg=#7f1425`
image

**Configuration**:
* Neovim version (nvim --version): NVIM v0.10.0
* Operating system and version: win 11

# Checklist:


lua/telescope/pickers.lua
jamestrew, line 1: I don't think these are relevant, can you verify? Using `TelescopeResultsEntrySeparator` should work without this.

jamestrew (changes requested):

lua/telescope/pickers.lua
MovieMaker93: Yes, you are right; it works anyway without that.

lua/telescope/pickers/entry_display.lua
```diff
@@ -122,6 +122,8 @@ entry_display.create = function(configuration)
       end
     end
 
+    configuration.separator_hl = vim.F.if_nil(configuration.separator_hl, "TelescopeResultsEntrySeparator")
+
     if configuration.separator_hl then
```
Conni2461, line 127: we no longer need this if, also drop the empty line

Conni2461:
this now has a new behavior

now

image

before

image

MovieMaker93:
> this now has a new behavior
> 
> now
> 
> image
> 
> before
> 
> image
> 
> i'm not sure i like it

I don't know if I like it more. Is there a cleaner way to allow users to change the separator highlight while keeping the actual default behaviour?

Conni2461:
> Is there a cleaner way to allow users to change the separator highlight while keeping the actual default behaviour?

but its not, we are losing the selected row highlighting because the selection is rendered before, i am not sure.

None the less for me it looks like a breaking change

jamestrew:
I think the selection highlight is applied later after but the way extmark handles highlight combining is what's leading to this outcome.

There is a `hl_mode` option for `nvim_buf_set_extmark` that might be able to control this but it currently only works for virt_text extmarks. I don't have a good solution for this. Maybe it's not currently doable in a non-breaking change manner.

This doesn't affect me personally but I could see this being a pretty annoying breaking change to some.

Issue #3128, opened 2024-05-21
[request] close buffer bind


gmankab:
hello

now to close buffer i open it and do `:bd`

can you please add a bind to the `:Telescope buffers` that will close selected buffers without opening

Issue #2643, opened 2023-08-08
delete_buffer action does not close the current buffer


tjvc:
### Description

The `delete_buffer` action does not close the current buffer if it is the only open buffer in the current tab.

### Neovim version

```markdown
NVIM v0.10.0-dev-774+g37c58226a-Homebrew
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

macOS 13.5

### Telescope version / branch / rev

telescope 0.1.2

### checkhealth telescope

```markdown
telescope: health#telescope#check

Checking for required plugins ~
- OK plenary installed.
- WARNING nvim-treesitter not found. (Required for `:Telescope treesitter`.)

Checking external dependencies ~
- OK rg: found ripgrep 13.0.0
- OK fd: found fd 8.6.0

===== Installed extensions ===== ~

Telescope Extension: `fzf` ~
- OK lib working as expected
- OK file_sorter correctly configured
- OK generic_sorter correctly configured
```

### Steps to reproduce

1. Open one or more buffers
2. Ensure that a single buffer is open in the current tab
3. Open the Telescope buffers picker
4. Try deleting the current buffer (mapped to `<C-D>` in the provided config)
5. See that the buffer is not deleted

### Expected behavior

The same as `:bdelete <buffer number>` on the current buffer: the buffer is deleted and another open buffer (if any) or the `[No Name]` buffer is displayed. 

### Actual behavior

The buffer is removed from the picker list but remains open.

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          { 'nvim-telescope/telescope-fzf-native.nvim', run = 'make' },
        },
      },
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
      display = { non_interactive = true },
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
  require('telescope').load_extension('fzf')
[...]

PR #3145, opened 2024-05-30
feat(builtin.buffers): enhance and bind `delete_buffer` action

jamestrew:
Binds the `delete_buffer` action to `<C-x>` by default for the `buffers` pickers.

closes https://github.com/nvim-telescope/telescope.nvim/issues/3128

Also enhances the `delete_buffer` action to have a "cleaner" behavior for deleting the current buffer (the buffer under telescope). Previously, when deleting the current buffer, the buffer would be removed from the buflist but closing the picker would put you right back into the buffer you just deleted.
From `:h bdelete`:
> If buffer [N] is the current buffer, another buffer will be displayed instead.  This is the most recent entry in the jump list that points into a loaded buffer.

Since the telescope prompt buffer is technically the current buffer, this "another buffer will be displayed behavior" wasn't actually triggering.
With this newest change, Telescope will manually get the buffer of the most recent entry in the jumplist and change the window behind telescope to this new buffer.

closes https://github.com/nvim-telescope/telescope.nvim/issues/2643

### DIFF
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -972,6 +972,9 @@ internal.buffers = function(opts)
       previewer = conf.grep_previewer(opts),
       sorter = conf.generic_sorter(opts),
       default_selection_index = default_selection_idx,
+      attach_mappings = function(_, map)
+        map({ "i", "n" }, "<C-x>", actions.delete_buffer)
+      end,
     })
     :find()
 end
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -973,7 +973,7 @@ internal.buffers = function(opts)
       sorter = conf.generic_sorter(opts),
       default_selection_index = default_selection_idx,
       attach_mappings = function(_, map)
-        map({ "i", "n" }, "<C-x>", actions.delete_buffer)
+        map({ "i", "n" }, "<M-d>", actions.delete_buffer)
       end,
     })
     :find()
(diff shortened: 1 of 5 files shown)


lua/telescope/builtin/__internal.lua
```diff
@@ -972,6 +972,9 @@ internal.buffers = function(opts)
       previewer = conf.grep_previewer(opts),
       sorter = conf.generic_sorter(opts),
       default_selection_index = default_selection_idx,
+      attach_mappings = function(_, map)
+        map({ "i", "n" }, "<C-x>", actions.delete_buffer)
+      end,
```
Conni2461, line 977: I am not the biggest fan of mapping custom picker actions by default, thats why its currently still not mapped. Mappings are always opinionation but we can do that but should document that at least in `builtins/init.lua`, with a simple mention

Conni2461 (approved):
LGTM, besides docs nit

Conni2461:
Thanks :)

jamestrew:
Thanks, I agree about setting default mappings but it's such a common question I see here/reddit/discord so.
Switched it to `<M-d>` since I forgot that `<C-x>` is already a default mapping.

igorlfs:
Hey! This commit introduced a bug :(

I can no longer use `:Telescope buffers`. I get a stacktrace that points to 
```telescope.nvim/lua/telescope/mappings.lua:330: Attach mappings must always return a value. `true` means use default mappings, `false` means only use attached mappings```. So I guess this should be retuning true?

PR #3148, opened 2024-05-31
fix(layout_strategies): odd flex layout default behavior

### LINKED ISSUE #3138: Odd default behavior with flex layout when resizing horizontally
### Description

When using the flex layout with default settings, odd behavior is observed when resizing the window horizontally. The preview pane is hidden at certain widths.

### Neovim version

```markdown
NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1713484068
```

Linux 6.8.1

master (c2ce039)

### checkhealth telescope

```markdown
- OK plenary installed.

- OK rg: found ripgrep 14.1.0
- OK fd: found fd 9.0.0

Telescope Extension: `fzf` ~
- OK lib working as expected
```

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. Resize Neovim window so it's at least 130x50 characters
3. Execute `:Telescope live_grep`
4. Start reducing the width of the window gradually

### Expected behavior

The preview pane should remain visible while resizing horizontally. The layout should flip to vertical once the width is reduced below 120.

### Actual behavior

There is a period while resizing horizontally where the preview pane is not visible. This happens between widths of 120 and 100. When the window width is reduced below 100, the preview appears again in a vertical layout. This doesn't make sense because if you started with a vertical layout, you wouldn't expect increasing the width of the window to show less of the UI.

Video:
https://github.com/nvim-telescope/telescope.nvim/assets/86718942/f8e272a0-b719-4752-9adc-fa08f085c099

### Analysis

The reason this happens is because of a mismatch between the default `flip_columns` and `flip_lines` values ([which are 100 and 20](https://github.com/nvim-telescope/telescope.nvim/blob/c2ce039188e22288449e43b77710036e69a45a70/lua/telescope/pickers/layout_strategies.lua#L774)) for the flex layout and the `preview_cutoff` values for the horizontal and vertical layouts ([which are 120 and 40](https://github.com/nvim-telescope/telescope.nvim/blob/c2ce039188e22288449e43b77710036e69a45a70/lua/telescope/config.lua#L78)). When the window width is increased to 100, the horizontal layout is switched to, but the horizontal layout won't display a preview pane because the window width is below the preview cutoff for the horizontal layout.

My suggested change would be to set the default `flip_columns` and `flip_lines` values in the flex layout code to 120 and 40 respectively. This would ensure the vertical layout is only switched away from when the window is wide enough for the horizontal layout to display correctly. Alternatively, the default `clip_columns` and `flip_lines` values could be taken from the `preview_cutoff` values for the horizontal and vertical layouts automatically.

Note that the desired behavior can be achieved just by adding the following code to telescope's config, but as I consider the default behavior a bit strange, I thought it might be better to have that changed so this config isn't needed.

```lua
        defaults = {
          layout_strategy = 'flex',
          layout_config = {
            flex = {
              flip_columns = 120,
              flip_lines = 40,
            },
          },
        },
```

### Minimal config

```Lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])
local package_root = "/tmp/nvim/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
	require("packer").startup({
		{
			"wbthomason/packer.nvim",
			{
				requires = {
					"nvim-lua/plenary.nvim",
					{ "nvim-telescope/telescope-fzf-native.nvim", run = "make" },
				},
			},
		},
		config = {
			package_root = package_root,
			compile_path = install_path .. "/plugin/packer_compiled.lua",
		},
	})
end
_G.load_config = function()
	require("telescope").setup({ defaults = { layout_strategy = "flex" } })
	require("telescope").load_extension("fzf")
end
	vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
vim.cmd([[autocmd User PackerComplete ++once echo "Ready!" | lua load_config()]])
```

ronandalton:
# Description

This PR fixes an issue with the flex layout by changing the default values for `flip_columns` and `flip_lines` to match the `preview_cutoff` values for the horizontal and vertical layouts. This should improve the user experience for users who have their default layout set to 'flex'. See the issue description linked below for more information.

Fixes #3138

## Type of change

# How Has This Been Tested?

The change implemented in this PR was tested using the following two tests:

- [ ] Test A: Set default `layout_strategy` to `'flex'` with no additional configuration. Run `:Telescope live_grep` and resize window. Observe that layout changes to vertical when `width < 120` given `height >= 40`.
- [ ] Test B: Set default `layout_strategy` to `'flex'` and `layout_config.horizontal.preview_cutoff` and `layout_config.vertical.preview_cutoff` to custom values. Run `:Telescope live_grep` and resize window. Observe that layout changes to vertical when the window size matches the `preview_cutoff` values set.

**Configuration**:
* Neovim version (nvim --version): NVIM v0.10.0
* Operating system and version: Linux 6.8.1

# Checklist:

### DIFF
--- a/lua/telescope/pickers/layout_strategies.lua
+++ b/lua/telescope/pickers/layout_strategies.lua
@@ -771,10 +771,10 @@ layout_strategies.flex = make_documented_layout(
     horizontal = "Options to pass when switching to horizontal layout",
   }),
   function(self, max_columns, max_lines, layout_config)
-    local flip_columns = vim.F.if_nil(layout_config.flip_columns, 100)
-    local flip_lines = vim.F.if_nil(layout_config.flip_lines, 20)
+    local flip_columns = vim.F.if_nil(layout_config.flip_columns, layout_config.horizontal.preview_cutoff)
+    local flip_lines = vim.F.if_nil(layout_config.flip_lines, layout_config.vertical.preview_cutoff)
 
-    if max_columns < flip_columns and max_lines > flip_lines then
+    if max_columns < flip_columns and max_lines >= flip_lines then
       self.__flex_strategy = "vertical"
       self.layout_config.flip_columns = nil
       self.layout_config.flip_lines = nil


Conni2461:
yep that is actually a problem, good catch and thanks for the fix :)

PR #3151, opened 2024-06-01
fix(action): edit action when path starts with "../"

dhananjaylatkar:
# Description
Path:normalize() removes "../" from front which breaks enter action so, do not normalize if path starts with "../"

Fixes # (issue)

## Type of change

# How Has This Been Tested?
Tested by providing path starting with `../` and enter action works.

**Configuration**:
NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1713484068

Ubuntu 22.04

# Checklist:

### DIFF
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -196,7 +196,11 @@ action_set.edit = function(prompt_bufnr, command)
     -- check if we didn't pick a different buffer
     -- prevents restarting lsp server
     if vim.api.nvim_buf_get_name(0) ~= filename or command ~= "edit" then
-      filename = Path:new(filename):normalize(vim.loop.cwd())
+      if not vim.startswith(filename, "../") then
+        -- Path:normalize() removes "../" from front which breaks jump
+        -- so, do not normalize if path starts with "../"
+        filename = Path:new(filename):normalize(vim.loop.cwd())
+      end
       pcall(vim.cmd, string.format("%s %s", command, vim.fn.fnameescape(filename)))
     end
   end


Conni2461:
What issues does this resolve? When do we have a ../ path?

dhananjaylatkar:
Hi @Conni2461 -

In [cscope_maps.nvim](https://github.com/dhananjaylatkar/cscope_maps.nvim), paths from result of `cscope` are converted to relative path. When result is in previous directory, "../" will be added as suffix.

you can refer to relative path api here - [get_rel_path()](https://github.com/dhananjaylatkar/cscope_maps.nvim/blob/main/lua/cscope_maps/utils/init.lua#L13)

jamestrew:
Is the `path` field a valid path here?
https://github.com/dhananjaylatkar/cscope_maps.nvim/blob/c5947e34c03b59e597656d2b24923f9ce63bcad8/lua/cscope/pickers/telescope.lua#L10-L31

We should be using `entry.path` as the `filename` path if it's not nil and it's the preferred way to communicate the path of the file to open.

dhananjaylatkar:
Are you referring to https://github.com/dhananjaylatkar/cscope_maps.nvim/blob/c5947e34c03b59e597656d2b24923f9ce63bcad8/lua/cscope/pickers/telescope.lua#L29 ?

I pass `path = filename` where `filename` is result from `cscope` and it is valid path always.

jamestrew:
Yes, and that `path` value should always be a fully qualified path.

dhananjaylatkar:
Hi @Conni2461 @jamestrew 

Hope all your queries are answered. Can you let me know it is ok to merge this PR?

lua/telescope/actions/set.lua
```diff
@@ -196,7 +196,11 @@ action_set.edit = function(prompt_bufnr, command)
     -- check if we didn't pick a different buffer
     -- prevents restarting lsp server
     if vim.api.nvim_buf_get_name(0) ~= filename or command ~= "edit" then
-      filename = Path:new(filename):normalize(vim.loop.cwd())
+      if not vim.startswith(filename, "../") then
```
jamestrew, line 199: If `filename` here is coming from `entry.path` and `entry.path` is a fully qualified path, then it can never start with `../`.

For this change to be necessary, I feel that something wrong has already happened in the entry maker.

dhananjaylatkar: @jamestrew 
Here is an example -
1. `cwd = /home/d/code/myproject`
2. `linux` source is located at `/home/d/code/linux`
3. I have configured cscope db in both `myproject` and `linux`
4. When query is made for definition of some `symbol` from `myproject`, I get results from `linux` directory.
5. This result is something like `/home/d/code/linux/some/header/file.h` (there could be more results)
6. Instead of passing this abs path, I pass `../linux/some/header/file.h` and avoid seeing my home directory repeatedly.
7. This path is "sanitized" by telescope making it `/linux/some/header/file.h`
8. Now this is invalid path.

jamestrew: I understand the use case. You're free to display a relative path (in the `entry.display` function). You just need to pass an absolute path for `entry.path`.

Conni2461:
I'd rather not merged this, this is really hot code, and i am always hesitant to change this code if there isnt a good reason for change and if i currently see this correctly, there isnt one

jamestrew:
> I'd rather not merged this, this is really hot code, and i am always hesitant to change this code if there isnt a good reason for change and if i currently see this correctly, there isnt one

I agree. This change shouldn't be necessary. I think there's room for improvement in our `make_entry` docs but I believe all that's needed is for the `filename` or `path` values in the entries to be set to the absolute path.
I think you can simply do
```lua
   path = Path:new(entry["filename"]):absolute(),
```
in your entry_maker where `local Path = require('plenary.path')`.

You're free to continue displaying relative paths in the `display` field.

dhananjaylatkar:
You should improve `:normalize` function. If it can't be improved to support all path types then better use something else.

This issue is only seen in the telescope. fzf-lua and quickfix work as expected.

jamestrew:
I think it's matter of convention, for the path to be an abs path, since basically the inception of telescope many years ago. There isn't a ton of motivation to change this.

Maybe something like this, which should achieve the same results, I wouldn't be opposed to too much
```lua
    if vim.api.nvim_buf_get_name(0) ~= filename or command ~= "edit" then
      local file = Path:new(filename)
      filename = file:is_absolute() and file:normalize(vim.loop.cwd()) or file.filename
      pcall(vim.cmd, string.format("%s %s", command, vim.fn.fnameescape(filename)))
    end
```

dhananjaylatkar:
I'm okay with your change. It is basically the same thing as my change.

It would be great if telescope can support all types of relative paths. I don't want to add special handling just for one picker.

PR #3152, opened 2024-06-01
feat: refactors transform_path and allow `smart` to be used in conjunction with `filename_first`

alycklama:
# Description
I've cleaned up `transform_path`:
- Extracted code into functions: `path_filename_first`, `path_truncate`, etc.
- By returning early for `smart` we would not be able to run `filename_first` afterwards.

## Smart pre-PR
image

## Smart post-PR
The directory can now be separated with the following configuration:

```
path_display = {
  "smart",
  "filename_first",
},
```

image

## Type of change

- Feature: allowing `smart` to be used in conjunction with `filename_first`

# How Has This Been Tested?

- [x] Local tests are green
- [x] Ran Telescope in Neovim locally

**Configuration**:
* Neovim version (nvim --version):

> NVIM v0.10.0
> Build type: Release
> LuaJIT 2.1.1716656478

* Operating system and version:

OSX Sonoma 14.4.1

# Checklist:

- [x] My code follows the style guidelines of this project (stylua)
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (lua annotations)

PR #3153, opened 2024-06-02
Feature: highlighting of path coordinates

alycklama:
# Description

Adds highlighting of path coordinates by using the existing highlight group `TelescopeResultsLineNr`.

## Type of change

### Live Grep

image

### Grep Word

image

### Buffers

image

### LSP definitions

image

### LSP references

image

### LSP type definitions

image

### Jumplist

image

### Loclist

image

# How Has This Been Tested?

This has been tested manually (see screenshots)

- [x] Live Grep
- [x] Grep Word
- [x] Buffers
- [x] LSP definitions
- [x] LSP references
- [x] Jumplist
- [x] Loclist

**Configuration**:

> NVIM v0.10.0
> Build type: Release
> LuaJIT 2.1.1716656478

> Sonoma 14.4.1

# Checklist:

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -462,11 +462,16 @@ function make_entry.gen_from_quickfix(opts)
   local hidden = utils.is_path_hidden(opts)
 
   local make_display = function(entry)
-    local display_filename, path_style = utils.transform_path(opts, entry.filename)
+    local display_filename, style = utils.transform_path(opts, entry.filename)
     local display_string = string.format("%s:%d:%d", display_filename, entry.lnum, entry.col)
     if hidden then
       display_string = string.format("%4d:%2d", entry.lnum, entry.col)
     end
+
+    -- Adds styling to the coordinates
+    style = utils.merge_styles(
+      { { { #display_filename, #display_string }, "TelescopeResultsComment" } }, style, 0
+    )
 
     if show_line then
       local text = entry.text
@@ -474,10 +479,16 @@ function make_entry.gen_from_quickfix(opts)
         text = vim.trim(text)
       end
       text = text:gsub(".* | ", "")
+
+      -- Accounts for the added ":" after the display_string
+      style = utils.merge_styles(
+        { { { #display_string, #display_string + 1 }, "TelescopeResultsComment" } }, style, 0
+      )
+
       display_string = display_string .. ":" .. text
     end
 
-    return display_string, path_style
+    return display_string, style
   end
 
   local get_filename = get_filename_fn()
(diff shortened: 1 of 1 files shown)


jamestrew:
I'm actually planning to move the line+col numbers over next to the file path. I wasn't planning on the current behavior being a permanent thing. See https://github.com/nvim-telescope/telescope.nvim/issues/3156

I'm thinking of adding a wrapper around `transform_path` that handles concatenating the line/col numbers. This would consolidate all the manual handling of line/col number handling across all the entry makers and allow a more consistent `disable_coordinates` option across all pickers that offer line/col numbers in the entries.

I want to give Conni a chance to approve this before making these moves. https://github.com/nvim-telescope/telescope.nvim/pull/3152

alycklama:
> I'm actually planning to move the line+col numbers over next to the file path. I wasn't planning on the current behavior being a permanent thing. See #3156
> 
> I'm thinking of adding a wrapper around `transform_path` that handles concatenating the line/col numbers. This would consolidate all the manual handling of line/col number handling across all the entry makers and allow a more consistent `disable_coordinates` option across all pickers that offer line/col numbers in the entries.

I agree. One thing to add, not related to this change.

In retrospective, I believe that "filename_first" should not be a `path_transform` option. Conceptually, it's no longer a path once you put the filename in front (e.g. `hello-world.txt /tmp`). Ideally, all options to transform a path should still return a path. How to display the path could then become a new option outside of `transform_path`. The advantage of this approach is that you can compose any path transformer with any other path transformer. As they all operate on a path, the user can decide how to compose the options sequentially. The `reverse_directories` option part of `filename_first` can become a new `path_transform` option: `reverse`. Showing or hiding the line numbers can then also be part of the function that decides how to render the path (filename in front and/or with line numbers).

I believe it would simplify things a bit.

Conni2461:
> I'm thinking of adding a wrapper around transform_path that handles concatenating the line/col numbers.

Would make sense. But this should probably be a new interface that lives alongside the only one, `transform_path_lnumcol`

> In retrospective, I believe that "filename_first" should not be a path_transform option.

I disagree. Transform path was never required to return a valid path, maybe the name is badly choosen. but it always was a way to take a path + config and return a modified version of that, shortened, or only tail, or absolute, or with filename first. And we also have a way for users to modify the path as they want.

Intoducing a new function wouldnt change that at all, we still need that ONE function (currently transform_path) for our or extension entry makers to use. If you want you can introduce a new function that actually returns a valid path, but this will be a local function (i dont see a reason why we need to expose something like that), if you feel like that improves readability

PR #3154, opened 2024-06-04
fix(lsp.reference): Only filter the current line of the current filname

Lonzoghf:
fix(lsp.reference): Only filter the current line of the current filname when include_current_line=false

# Description

Lsp references with the same number of lines in different files will be filtered when include_current_line=false

## Type of change

# How Has This Been Tested?

- Lsp references have the same number of lines in different files (with gopls)

**Configuration**:
NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1713773202

MacOs (14.5)

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -140,8 +140,9 @@ end
 local apply_action_handler = function(action, items, opts)
   if action == "textDocument/references" and not opts.include_current_line then
     local lnum = vim.api.nvim_win_get_cursor(opts.winnr)[1]
+    local cur_filename = vim.api.nvim_buf_get_name(opts.bufnr)
     items = vim.tbl_filter(function(v)
-      return not (v.filename and v.lnum == lnum)
+      return not (v.filename == cur_filename and v.lnum == lnum)
     end, items)
   end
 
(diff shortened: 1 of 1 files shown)


jamestrew (commented):
Thanks, good catch. 
We're actually getting the current file path a little bit further done on line 158 (now 159).
```lua
local curr_filepath = vim.api.nvim_buf_get_name(opts.bufnr)
```

Maybe we can just stick that into `opts` (`opts.curr_filepath = vim.api.nvim_buf_get_name(opts.bufnr)`) and reuse that value in the `tbl_filter` function rather than re-fetching that.

Lonzoghf:
> Thanks, good catch. We're actually getting the current file path a little bit further done on line 158 (now 159).
> 
> ```lua
> local curr_filepath = vim.api.nvim_buf_get_name(opts.bufnr)
> ```
> 
> Maybe we can just stick that into `opts` (`opts.curr_filepath = vim.api.nvim_buf_get_name(opts.bufnr)`) and reuse that value in the `tbl_filter` function rather than re-fetching that.

Thank you, I have resubmitted the modifications

jamestrew:
Thanks!

Issue #3168, opened 2024-06-14
lsp_definitions should respect file_ignore_patterns when deciding list or jump


dssste:
**Is your feature request related to a problem? Please describe.**
lsp_definitions would jump to the location if there was only one result, but it would not do so if such scenario was due to file_ignore_patterns, resulting in a list of a single location without automatic jumping.

**Describe the solution you'd like**
Maybe it should check (in list_or_jump I think?) the result after filtering, instead of before filtering? Not 100% sure about the implementation details behind this.

**Describe alternatives you've considered**
Or just leave it as is. I don't think you would encounter this in any sensible code base.

**Additional context**
I encountered this when I was to filter out the generated code in Unity's UIToolkit. Personally I think they introduced a solution to a non-existing problem, but wcyd. The example below should be reproducible in any language where multiple definitions can happen.
![Capture2](https://github.com/nvim-telescope/telescope.nvim/assets/17314075/e78ab02d-889c-40fd-bc03-57dbe5d8b833)
![Capture](https://github.com/nvim-telescope/telescope.nvim/assets/17314075/bdc150cc-1d71-4c4b-a183-9e3c813ca96c)

PR #3173, opened 2024-06-15
fix(builtin.lsp): respect `file_ignore_patterns` opt for direct jumps

jamestrew:
Adds support for respecting `file_ignore_patterns` option before any jumps are made in the following pickers utilizing the `list_or_jump` helper.
Without this, you can get a picker with only one result, so it should've jumped rather than opening the picker.

closes https://github.com/nvim-telescope/telescope.nvim/issues/3168


dssste:
Thank you. Tried it and it works.

PR #3175, opened 2024-06-17
fix(builtin.live_grep, builtin.treesitter): not adding jumps to jumplist

xz47sv:
# Description

Whenever a picker, that can jump to a different location in the same buffer does so it should have `push_cursor_on_edit` otherwise the jump won't be recorded.

Maybe this affects other pickers too, but I only noticed it on grep and treesitter since I use those two a lot.

## Type of change

# Checklist:


jamestrew:
I think this makes sense.
Can you also add it for `grep_string` while we're at it?

xz47sv:
added

jamestrew:
Thanks!

Issue #3157, opened 2024-06-07
filename_first doentst works with buffer list 


Wordluc:
### Description

When using Telescope.nvim to display buffers, the buffer list does not show the file names first. This issue occurs even though the path_display option is set to show file names first (for files/grep, this setting works correctly).

![image](https://github.com/nvim-telescope/telescope.nvim/assets/40872910/6b822b22-52ff-48b9-882b-502c0d76a737)

### Neovim version

```markdown
NVIM v0.9.5
Build type: RelWithDebInfo
LuaJIT 2.1.1703942320
```

### Operating system and version

Windows 11

### Telescope version / branch / rev

master

### checkhealth telescope

```markdown
Checking for required plugins ~
- OK plenary installed.
- OK nvim-treesitter installed.

Checking external dependencies ~
- OK rg: found ripgrep 14.1.0 (rev e50df40a19)
- WARNING fd: not found. Install [sharkdp/fd](https://github.com/sharkdp/fd) for extended capabilities
```

### Steps to reproduce

set the options
`	path_display = {
			filename_first = {
				reverse_directories = false
			}
}`
after (restart nvim), the options is applied of the other search but buffer

### Expected behavior

see each each buffer with the name first

### Actual behavior

some buffer have the name first other not

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          { 'nvim-telescope/telescope-fzf-native.nvim', run = 'make' },
        },
      },
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
      display = { non_interactive = true },
    },
  }
end
_G.load_config = function()
  require('telescope').setup(require("telescope").setup {
	defaults = {
		path_display = {
			filename_first = {
				reverse_directories = false
			}
		},
		file_ignore_patterns = { "./^.git/*", "./node_modules/*", "node_modules", "^node_modules/*", "node_modules/*" },
	} 
})
[...]

PR #3176, opened 2024-06-19
Fix problem with firstname options and buffer view

Wordluc:
Fix issue: https://github.com/nvim-telescope/telescope.nvim/issues/3157#issue-2340997131

### DIFF
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -940,11 +940,13 @@ internal.buffers = function(opts)
       default_selection_idx = 2
     end
 
-    local element = {
-      bufnr = bufnr,
-      flag = flag,
-      info = vim.fn.getbufinfo(bufnr)[1],
-    }
+    local info = vim.fn.getbufinfo(bufnr)[1]
+		info.name = info.name:gsub("[/]", "\\")
+		local element = {
+			bufnr = bufnr,
+			flag = flag,
+			info = info,
+		}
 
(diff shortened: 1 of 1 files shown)


jamestrew:
Hmm.. I think this is an issue that plagues various other places involving plenary's `Path` class and windows paths with mixed path separators.
This is probably best fixed upstream in plenary.

Wordluc:
@jamestrew why the problem is plenary? the problem maybe is  vim.fn.getbufinfo(bufnr)[1], that return the "wrong" result

radlinskii:
This resolves the issue on my machine with Windows 11, together with `path_display` set to `filename_first` works very well

jamestrew:
This messes up unix paths.
We could do a workaround but my preference is to fix this in plenary.

> why the problem is plenary? the problem seems to be vim.fn.getbufinfo(bufnr)[1], that return the "wrong" result

I say this because we use plenary for all the cross-platform path stuff and the path from `getbufinfo` eventually gets fed into `plenary.Path`. If we fix the issue in plenary, we can fix this + like 3 other nearly identical issues.

I'll try to work on this week.

thedeany:
Is there any news on this issue? I searched in Plenary's issues, but I didn't see one. Also didn't see one referenced in this PR. Thanks in advance!

alerque:
This has accumulated a lot of merges and style changes irrelevant to the actual fix. Can I suggest this get rebased with only relevant changes to make it easier to review?

Issue #3183, opened 2024-06-25
Error showing LSP definitions


mfelsche:
### Description

I get regular errors when opening lsp results upon using goto-definition.

My nvim is giving this error message with traceback: 
```
Error executing vim.schedule lua callback: vim/shared.lua:275: Expected table, got nil
stack traceback:
	[C]: in function 'assert'
	vim/shared.lua:275: in function 'tbl_isempty'
	...nvim/lazy/telescope.nvim/lua/telescope/builtin/__lsp.lua:156: in function 'filter_file_ignore_patters'
	...nvim/lazy/telescope.nvim/lua/telescope/builtin/__lsp.lua:198: in function 'handler'
	/usr/local/share/nvim/runtime/lua/vim/lsp.lua:1393: in function 'cb'
	vim/_editor.lua:263: in function <vim/_editor.lua:262>
```

### Neovim version

```markdown
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1699801871
```

### Operating system and version

FreeBSD 14.1

### Telescope version / branch / rev

58ec6ec

### checkhealth telescope

```markdown
telescope: 

- ERROR No healthcheck found for "telescope" plugin.
```

### Steps to reproduce

1. install and configure rust-analyzer via nvim-lspconfig or mason or whatever means. Any other LSP should work as well.
2. Use goto-definition on some type. e.g. here:

```rust
fn main() {
   let foo: Option<usize> = None;
}
```
3. experience the error described above.

### Expected behavior

telescope should open and show the definition list of the given type.

### Actual behavior

telescope errors with the traceback and error message pasted above.

### Minimal config

```Lua
-- sorry, I am unable to quickly get a minimal config in place
-- for reproducing the issue, i use LazyVim and installed rust-analyzer as an LSP of choice
```

oozzal:
Latest commit is broken. Reverting to commit: af78ba3b7da51ea5780bdb2c4b990ae2ea2ce716 should work.

jamestrew:
Oh wow this bug confused me since it looked like shouldn't be possible and wasn't able to reproduce initially but turns out `vim.F.if_nil` behavior was changed in neovim 0.10.

I will put a fix for <0.10.

PR #3184, opened 2024-06-25
fix(lsp): `if_nil` <0.10 compat

jamestrew:
closes https://github.com/nvim-telescope/telescope.nvim/issues/3183


muhajirdev:
is this released?

Issue #3156, opened 2024-06-07
Display line number with filename when using `filename_first`


mehalter:
**Is your feature request related to a problem? Please describe.**

When I'm using the relatively new `filename_first` `path_display` option, specifically with the buffer picker, it can be a bit jarring when the line number is displayed after the path rather than after the filename.

![2024-06-07_11:07:27_screenshot](https://github.com/nvim-telescope/telescope.nvim/assets/1591837/7d71a8d8-0826-4d83-ab46-dc0457cfbd2a)

The line number is really "operating" on the filename and not on the directory. Also because the directory highlighting most of the time is a bit more dimmed than the filename and the line number they feel a bit separated for the two parts that are being shown as the "most important"

**Describe the solution you'd like**

Add a configuration option similar to the `reverse_directories` to relocate the line number displaying to be next to the filename.

rather than being something like

```
utils.lua lua/telescope:686
```

It would be displayed as

```
utils.lua:686 lua/telescope
```

**Describe alternatives you've considered**

I believe there is a solution to be able to implement this as a custom `make_entry` function but this takes some deep understanding of how to write these functions for a small change. Also I think this could be something that many people may find benefit in which makes it seem like it would be a good option to include

jamestrew:
No I agree, the current behavior wasn't meant to be a long term thing.

leet0rz:
Sorry to ask here but I filename_first does not seem to be working for me, I recently updated to the latest telescope. This is the correct code right?:
```lua
require("telescope").setup {
    defaults = {
        path_display = {
            filename_first = {
                reverse_directories = false,
            }
        },
        sorting_strategy = "ascending",
        layout_strategy = "horizontal",
        layout_config = {
            horizontal = {
                prompt_position = "top",
            }
        },
    },
}
```

Thanks.

Issue #2607, opened 2023-07-18
Add `disable_coordinates` for buffer picker


net:
I'm not really sure what the value in seeing the line number for open buffers is, and they make the file name much more difficult to read.

It would also be nice if the buffer numbers on the left side were also hidden, since anyone using the buffer picker probably doesn't care about the buffer number.

A really clean buffer picker would also probably just show the modified flag (+) and none of the other ones, but the flags at least serve some purpose.

❤️

PR #3200, opened 2024-07-09
feat(path_display): move lnum/col next to filename for `filename_first`

jamestrew:
Combines `transform_devicons`, `transform_path` and the string formatting of line and column numbers when available as `create_path_display`.

This offers a few benefits:
- allows moving the lnum/col values to be next to the filename rather than and the end of the rest of the path when using `path_display = { 'filename_first' }`. Closes https://github.com/nvim-telescope/telescope.nvim/issues/3156
- simplifies the merging of styles when devicon hl groups are present into one function rather than split across many entry makers
- consolidates lnum/col values concatenation to one function rather than split across many entry makers so that any entry makers that support lnum/col will automatically support the `disable_coordinates` option that some support. closes https://github.com/nvim-telescope/telescope.nvim/issues/2607


jamestrew:
I've done some testing with some of the common pickers (find_files, live_grep, grep_string, buffers, diagnostics, lsp_[references/definition]).
I'm going to keep riding this branch for a bit.

Daydreamer-riri:
This function is great! Looking forward to the merger. Thanks for your great work!

milanglacier:
Tried this branch and it looks great. `find_files`, `buffers`, `old_files`, `lsp_references`, `lsp_definitions` all works great.

Daydreamer-riri (approved):

Daydreamer-riri (approved):

Conni2461:
i am going to review this soonish :clown_face:, because it looks like a useful refactoring. for now i would merge https://github.com/nvim-telescope/telescope.nvim/pull/3450 because it resolved an issue and is a free win, but i think a more generalized refactor is the better solution :+1: 

Still figuring out how i wanna want to do this telescope development again thats why i want to prefer "simpler" PRs first

PR #3211, opened 2024-07-18
Fetch results from multiple LSPs

enriclluelles:
# Description

Fixes https://github.com/nvim-telescope/telescope.nvim/pull/2342

Right now, when you have two LSPs connected to one buffer and you want to, let's say, go to a definition, there's a race condition / non-deterministic behavior in which Telescope uses the answer of the first LSP that responds

As an example (and the actual reason for me submitting this change), if you have a ruby application that uses sorbet, and you're running the Sorbet LSP and RubyLSP for the same buffer (which is pretty useful), what you'll see for `telescope.builtin.lsp_definitions` will be:

Kapture 2024-07-18 at 12 32 12

The first two times, it used the locations from RubyLSP, and the third one from Sorbet

With this change applied, this is what happens:

Kapture 2024-07-18 at 12 36 46

You've got all the results from both LSPs consistently

This leaves us with the issue of what happens if both LSPs returning overlapping locations, so deduplication could be in order
But since this isn't my case and I suspect not a lot of people use multiple LSPs that return the same stuff, I prefer to get your feedback first and maybe implement it later

Also, we should probably remove the `list_or_jump` implementation since this makes it obsolete I think, but again, I prefer to get feedback first

Apologies if the code is not up to standard, I'm a complete Lua illiterate

## Type of change

# How Has This Been Tested?

Manually, as my knowledge of lua is very limited and mostly learned by tinkering with neovim plugins and config

**Configuration**:
* Neovim version (nvim --version): v0.10.0
* Operating system and version: macOS 14.5 build 23F79

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -184,7 +184,7 @@ local function list_or_jump_all(action, title, funname, params, opts)
     for client_id, result_or_error in pairs(results_per_client) do
       local error, result = result_or_error.error, result_or_error.result
       if error then
-        vim.api.nvim_err_writeln("Error when executing " .. action .. " : " .. err.message)
+        vim.api.nvim_err_writeln("Error when executing " .. action .. " : " .. error.message)
         return
       end
 
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -173,7 +173,7 @@ end
 ---@param funname string: name of the calling function
 ---@param params lsp.TextDocumentPositionParams
 ---@param opts table
-local function list_or_jump_all(action, title, funname, params, opts)
+local function list_or_jump(action, title, funname, params, opts)
   opts.reuse_win = vim.F.if_nil(opts.reuse_win, false)
   opts.curr_filepath = vim.api.nvim_buf_get_name(opts.bufnr)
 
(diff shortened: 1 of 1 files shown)


jamestrew:
Is there a reason to have a new function called `list_or_jump_all` as opposed to just altering `list_or_jump` if the latter is no longer used?

lua/telescope/builtin/__lsp.lua
```diff
@@ -168,6 +168,97 @@ local function filter_file_ignore_patters(items, opts)
...
+  opts.curr_filepath = vim.api.nvim_buf_get_name(opts.bufnr)
+
+  vim.lsp.buf_request_all(opts.bufnr, action, params, function(results_per_client)
+    local items = {}
+    local first_encoding
+
+    for client_id, result_or_error in pairs(results_per_client) do
+      local error, result = result_or_error.error, result_or_error.result
+      if error then
+        vim.api.nvim_err_writeln("Error when executing " .. action .. " : " .. err.message)
```
jamestrew, line 187:
```suggestion
        vim.api.nvim_err_writeln("Error when executing " .. action .. " : " .. error.message)
```

enriclluelles:
> Is there a reason to have a new function called `list_or_jump_all` as opposed to just altering `list_or_jump` if the latter is no longer used?

No, I guess I didn't realize it wasn't used anymore, renaming it to leave only `list_or_jump`

enriclluelles:
@jamestrew any chance you can run the CI workflows? I think there's nothing left for me to do

lua/telescope/builtin/__lsp.lua
```diff
@@ -177,24 +177,36 @@ local function list_or_jump(action, title, funname, params, opts)
...
+  vim.lsp.buf_request_all(opts.bufnr, action, params, function(results_per_client)
+    local items = {}
+    local first_encoding
+
+    for client_id, result_or_error in pairs(results_per_client) do
+      local error, result = result_or_error.error, result_or_error.result
+      if error then
+        vim.api.nvim_err_writeln("Error when executing " .. action .. " : " .. error.message)
+        return
+      end
```
jamestrew, line 189: I'm contemplating whether we should collect all the errors first, then raise them later on.

Like suppose some lsp server returns an error but another returns valid results. In this case, I'm thinking it might be more user friendly to still let telescope handle the rest of the results, and then show a warning for the error.
If there is only errors, we can throw an error.

enriclluelles: gotcha, I thought about it while changing the code but then I looked to see if lua had a `next` or `continue` instruction. And since I didn't find it, I moved on

I'll give it a stab

enriclluelles:
@jamestrew done

jamestrew:
thanks!

PR #3219, opened 2024-07-24
fix(current_buffer_fuzzy_find): push cursor on edit

jamestrew:
Setting `push_cursor_on_edit = true` wasn't actually doing anything as the picker replaces `select_default`.

Instead, call `normal! m'` manually in the replacement `select_default`.


LEXUGE:
Can confirm it works!

PR #3227, opened 2024-07-27
docs: switch out docgen tool

jamestrew:
Initial conversion of gendoc to https://github.com/jamestrew/docgen.nvim

main differences from tree-sitter-lua's doc gen:
- uses luacats annotations instead of emmylua allowing for tighter integration between lua_ls type checking and documentation annotations
- all descriptions are parsed as markdown using tree-sitter-markdown so improved parsing
- vimdoc rendering is heavily based off of the code used by neovim core so things will look very similar to neovim core docs

### DIFF
--- a/.github/workflows/docgen.yml
+++ b/.github/workflows/docgen.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            url: https://github.com/neovim/neovim/releases/download/v0.9.5/nvim-linux64.tar.gz
+            url: https://github.com/neovim/neovim/releases/download/stable/nvim-linux-x86_64.tar.gz
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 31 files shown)


clason:
@jamestrew I took the liberty of rebasing the PR on latest `master` and simplifying the workflow: it now only checks whether the docs are updated and fails with a diff output -- this is much more robust and not a big issue since `make docgen` is now easy to run locally.

I'm very tempted to merge this quickly since it looks like a massive improvement (and will unblock the plenary migration work). Consider transferring the docgen repo to this org.

(Note to self: new workflow name means I'll have to update the branch protection.)

clason:
Follow-up: #3506 (needs a docgen.nvim change).

jamestrew:
tweaking a couple of things upstream

clason:
Great work!

PR #3234, opened 2024-08-02
Tags kind

### LINKED ISSUE #3233: Support for ctags kinds on picker builtin.tags
When I do `:Telescope tags` I have a hard time figuring out the matching tags, I mean if I search for "parse" I want to see in the picker for each match if  it's "parse kind:method" , "Parse kind:class", or "Parse kind:field" 

my ./tags file already contains the 'kinds`, I generate the `tags` file with

```
git ls-files .  |  $(brew --prefix universal-ctags)/bin/ctags --fields=+zK -L - -f tags
```

and the `tags` file contents are like this
```
parse	app/src/main/java/com/craftinginterpreters/lox/Parser.java	/^    List<Stmt> parse() {$/;"	kind:method	class:Parser
```

so I would like the picker for `tags` to show (file, kind, tagname) instead of just showing the (file, tagname)

The alternatives is to stop using ctags and use LSP since the pickers for lsp `lsp_document_symbol`, `lsp_workspace_symbols` and `lsp_dynamic_workspace_symbols` how the  "symbol type" (which is equivalent to the ctags kind) 

But LSP comes with it's own problems and limitations, for example it looks like the java `eclipse.jdt.ls`/`jdtls` language server cannot do queries for workspace symbols other that `class` so I can't search for methods in the whole workspace. 

In fact, I was trying ctags as a way to overcome this limitation with `jdtls`. 

**Additional context**

ecerulm:
# Description

The current tags picker will not show the kind

Current tags picker: 
image

With this changes the picker will show the kind (`method`, `field`, `package`, `class` , etc): 

image

It augments the regex match in the existing picker to get the `extensions_fields` that may be present on each line of the `tags` file (see [TAG FILE FORMAT](https://docs.ctags.io/en/latest/man/ctags.1.html#tag-file-format)).

That `extension_fields` may contain a `kind:xxx` field, if it does it will be picked up and added to the entry otherwise the `unknownkind` will be used.
 
To generate a `tags` file that includes kinds use 

```
git ls-files .  |  $(brew --prefix universal-ctags)/bin/ctags --fields=+zK -L - -f tags
```

where the `--fields=+zK` is the part that adds the kinds. 

Fixes #3233

## Type of change

# How Has This Been Tested?

- [x] manual test, using a `tags` file generated with `git ls-files .  |  $(brew --prefix universal-ctags)/bin/ctags --fields=+zK -L - -f tags`

**Configuration**:
```
NVIM v0.11.0-dev-492+g0cdeb06db
Build type: RelWithDebInfo
LuaJIT 2.1.1720049189
```
* Operating system and version: macOS Sonoma 14.5

# Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,56 +9,33 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-22.04, macos-12, windows-2022]
+        rev: [nightly, v0.7.2, v0.8.3, v0.9.0]
         include:
           - os: ubuntu-22.04
-            rev: nightly/nvim-linux64.tar.gz
-            manager: sudo apt-get
-            packages: -y ripgrep
-          - os: ubuntu-22.04
-            rev: v0.7.2/nvim-linux64.tar.gz
-            manager: sudo apt-get
-            packages: -y ripgrep
-          - os: ubuntu-22.04
-            rev: v0.8.1/nvim-linux64.tar.gz
-            manager: sudo apt-get
-            packages: -y ripgrep
+            install-rg: sudo apt-get update && sudo apt-get install -y ripgrep
           - os: macos-12
-            rev: nightly/nvim-macos.tar.gz
-            manager: brew
-            packages: ripgrep
-          - os: macos-12
-            rev: v0.7.2/nvim-macos.tar.gz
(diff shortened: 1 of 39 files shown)


ecerulm:
Something wrong with the way I forked/ pushed this , closing

PR #3235, opened 2024-08-02
Add kind to builtin.tags picker

### LINKED ISSUE #3233: Support for ctags kinds on picker builtin.tags

ecerulm:
# Description

The current tags picker will not show the kind

Current tags picker: 
image

With this changes the picker will show the kind (`method`, `field`, `package`, `class` , etc): 

image

It augments the regex match in the existing picker to get the `extensions_fields` that may be present on each line of the `tags` file (see [TAG FILE FORMAT](https://docs.ctags.io/en/latest/man/ctags.1.html#tag-file-format)).

That `extension_fields` may contain a `kind:xxx` field, if it does it will be picked up and added to the entry otherwise the `unknownkind` will be used.
 
To generate a `tags` file that includes kinds use 

```
git ls-files .  |  $(brew --prefix universal-ctags)/bin/ctags --fields=+zK -L - -f tags
```

where the `--fields=+zK` is the part that adds the kinds. 

Fixes #3233

## Type of change

# How Has This Been Tested?

- [x] manual test, using a `tags` file generated with `git ls-files .  |  $(brew --prefix universal-ctags)/bin/ctags --fields=+zK -L - -f tags`

Tested the with the following keymap to trigger telescope

```
vim.keymap.set("n", ";c", function()
	require("telescope.builtin").tags({ only_sort_tags = true, show_line = false, path_display = {"filename_first"} })
end, opts)
```

**Configuration**:
```
NVIM v0.11.0-dev-492+g0cdeb06db
Build type: RelWithDebInfo
LuaJIT 2.1.1720049189
```
* Operating system and version: macOS Sonoma 14.5

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1016,8 +1016,6 @@ builtin.tags({opts})                                *telescope.builtin.tags()*
                                     false)
         {fname_width}    (number)   defines the width of the filename section
                                     (default: 30)
-        {show_kind}      (boolean)  if true, show the tag kind on the picker
-                                    (default: true)
 
 
 builtin.current_buffer_tags({opts})  *telescope.builtin.current_buffer_tags()*
@@ -1039,8 +1037,6 @@ builtin.current_buffer_tags({opts})  *telescope.builtin.current_buffer_tags()*
                                     false)
         {fname_width}    (number)   defines the width of the filename section
                                     (default: 30)
-        {show_kind}      (boolean)  if true, show the tag kind on the picker
-                                    (default: true)
 
 
 builtin.git_files({opts})                      *telescope.builtin.git_files()*
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1016,6 +1016,8 @@ builtin.tags({opts})                                *telescope.builtin.tags()*
                                     false)
         {fname_width}    (number)   defines the width of the filename section
                                     (default: 30)
+        {show_kind}      (boolean)  if true, show the tag kind on the picker
+                                    (default: true)
 
 
 builtin.current_buffer_tags({opts})  *telescope.builtin.current_buffer_tags()*
@@ -1037,6 +1039,8 @@ builtin.current_buffer_tags({opts})  *telescope.builtin.current_buffer_tags()*
                                     false)
         {fname_width}    (number)   defines the width of the filename section
                                     (default: 30)
+        {show_kind}      (boolean)  if true, show the tag kind on the picker
+                                    (default: true)
 
 
 builtin.git_files({opts})                      *telescope.builtin.git_files()*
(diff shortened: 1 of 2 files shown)


ecerulm:
> I'm wondering if it's possible to dynamically choose to show/hide the extension depending on whether it is there

I'll guess you mean to hide the **kind** if all values are "unknownkind"? 

> Ideally, we add an extra column upon generating the first row (the first call to the gen_from_ctags callback) only if a valid extension_fields exists, generating and caching the displayer. I'm not sure if that's something that can be determined with just one row from a ctags point of view.

The things is that I don't think looking at the first row works. In a regular project you can have tags without kind. 

> If that's too much hassle, maybe just adding a new option to the picker to dictate this behavior.

I can certainly add a config to show/hide kind. I can also only add the `tag.kind = kind` if `kind` is no  `unknownkind` I can try that and see how does show up on screen. 

I'll update the PR with the an option `show_kind` and add it to the docs. 
I'll also experiment with only adding `tag.kind` if present and see how nvim-telescope reacts to that

jamestrew:
> I'll guess you mean to hide the kind if all values are "unknownkind"?

Yeah exactly. I think it would be jarring to existing users of the picker if there's suddenly a entirely new column of just "unknownkind".

> I can certainly add a config to show/hide kind. I can also only add the tag.kind = kind if kind is no unknownkind I can try that and see how does show up on screen.
> 
> I'll update the PR with the an option show_kind and add it to the docs.
> I'll also experiment with only adding tag.kind if present and see how nvim-telescope reacts to that

Thanks, sounds good!

ecerulm:
@jamestrew, this new version does not add "unknownkind" if the kind is not present in the tags file. That means that the column is omitted if all the entries for kind are nil. 

Also added a option `show_kind` (by default true) to control if the kind should be shown or not even it's present on the tags file. 

Can be tested with: 
```
vim.keymap.set("n", ";c", function()
	require("telescope.builtin").tags({ only_sort_tags = true, show_line = true, path_display = {"filename_first"}, show_kind=true})
end, opts)
```

jamestrew:
Thanks! Looks good to me.

PR #3251, opened 2024-08-13
Updated docs to clarify which telescope.defaults options can be overriden by pickers

### LINKED ISSUE #3238: Docs doesn't make it clear that pickers and extensions accept same opts as defaults
### Description

I've been diving into telescope recently, as I've been using it for some time but never configured it thoroughly. I read the entire :h manual inside Neovim.

I hard a hard time understanding what options the individual pickers or extensions passed into setup() could take. It took me a while to understand that a picker's config table accepts the same keys as "defaults" table **plus** its own config keys which are shown in telescope.builtin. Writing it down I understand that it's kinda implied that defaults can be overriden, but it took me some time to fully get how it works.

Do you think this warrants a little clarification ?

In my humble opinion, it'd be worth modifying the telescope.setup section and the telescope.builtin section, here is my opinion on the parts the documentation is lacking clarity, as well as my humble proposal. I'd love to hear your feedbacks on this.

:h telescope.setup()
```md
telescope.setup({opts})                                    *telescope.setup()*
    Setup function to be run by user. Configures the defaults, pickers and
    extensions of telescope.

    Usage:
    >
    require('telescope').setup{
      defaults = {
        -- config_key = value,
        -- ..
      },
      pickers = {
        -- picker_name = {
        --   ...
        -- }
        -- ADD: Each picker table accepts the same opts as the defaults table.
        -- ADD: On top of those, the pickers each have additional opts, which
        -- ADD: you can learn about in telescope.builtin.
        -- ADD: Options set here will be applied every time you call this builtin picker
        -- DEL: Now the picker_config_key will be applied every time you call this
        -- DEL: builtin picker
      },
      extensions = {
        -- Your extension configuration goes here:
        -- extension_name = {
        --   extension_config_key = value,
        -- }
        -- ADD: Just like builtin pickers, extension can take the same opts as the
        -- ADD: the defaults table in order to override them.
        -- please take a look at the readme of the extension you want to configure
      }
    }

    Valid keys for {opts.defaults}
    ADD: (Those keys are also valid for builtin pickers and extensions)

    -- And then the list of opts possible keys
```

I have to note that the builtin part of the builtin help kinda implies this but it doesn't state it clearly. What I struggled to understand is that the builtin pickers options **extend** the default ones. Maybe this section could be clarified?
:h telescope.builtin
```md
To use any of Telescope's default options or any picker-specific options, call
your desired picker by passing a lua table to the picker with all of the
options you want to use. Here's an example with the live_grep picker:

>
  :lua require('telescope.builtin').live_grep({
    prompt_title = 'find string in open buffers...',
    grep_open_files = true
  })
  -- or with dropdown theme
  :lua require('telescope.builtin').find_files(require('telescope.themes').get_dropdown{
    previewer = false
  })
```

### Neovim version

```markdown
NVIM v0.10.1
Build type: Release
LuaJIT 2.1.1713484068
```

Debian 12

0.1.X

### checkhealth telescope

```markdown
telescope: health#telescope#check

- OK plenary installed.

- OK rg: found ripgrep 14.1.0 (rev 47797d5476)
- WARNING fd: not found. Install [sharkdp/fd](https://github.com/sharkdp/fd) for extended capabilities

Telescope Extension: `conflicts` ~
- No healthcheck provided

Telescope Extension: `dap` ~
- No healthcheck provided

Telescope Extension: `luasnip` ~
- No healthcheck provided

- No healthcheck provided
```

### Steps to reproduce

1. :h telescope

### Expected behavior

_No response_

### Actual behavior

See desc

### Minimal config

```Lua

-- set stdpaths to use .repro
for _, name in ipairs { "config", "data", "state", "cache" } do
end

-- bootstrap lazy
  vim.fn.system {
    "git",
    "clone",
    "--filter=blob:none",
    lazypath,
  }
end

-- install plugins
local plugins = {
  {
    dependencies = {
      "nvim-lua/plenary.nvim",
    },
    config = function()
      require("telescope").setup {}
    end,
  },
}

  root = root .. "/plugins",
})
```

clorl:
# Description

Fixes #3238

The documentation isn't very clear about what keys are valid for pickers config options. There is a subset of of defaults options that pickers can override and it's not clear which ones are.

## Type of change

# How Has This Been Tested?

- [ ] Hot swapped telescope with this commit in my config, nothing broke
- [ ] `make test`
- [ ] `make lint`

**Configuration**:
* Neovim version (nvim --version): NVIM v0.10.1

# Checklist:

- [ N/A ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/lua/telescope/builtin/init.lua
+++ b/lua/telescope/builtin/init.lua
@@ -11,13 +11,17 @@
 ---
 --- :lua require('telescope.builtin').$NAME_OF_PICKER()
 ---
---- To use any of Telescope's default options or any picker-specific options, call your desired picker by passing a lua
---- table to the picker with all of the options you want to use. Here's an example with the live_grep picker:
+--- To configure your desired picker, call it by passing a lua table with all
+--- of the options you want to use. Pickers have their own specific set of
+--- options, but also accept a subset of the defaults options. The default
+--- options which are valid to configure a picker are marked as
+--- "Picker-Overridable" in telescope.setup.defaults
+--- Here's an example with the live_grep picker:
 ---
 --- <code>
 ---   :lua require('telescope.builtin').live_grep({
----     prompt_title = 'find string in open buffers...',
----     grep_open_files = true
+---     prompt_title = 'find string in open buffers...',   -- Default option overriden for this specific picker
+---     grep_open_files = true                             -- Option specific to live_grep picker
 ---   })
 ---   -- or with dropdown theme
 ---   :lua require('telescope.builtin').find_files(require('telescope.themes').get_dropdown{
--- a/lua/telescope/init.lua
+++ b/lua/telescope/init.lua
@@ -106,18 +106,26 @@ local telescope = {}
 ---   pickers = {
 ---     -- Default configuration for builtin pickers goes here:
 ---     -- picker_name = {
----     --   picker_config_key = value,
+---     --   default_config_key = value_one
+---     --   picker_config_key = value_two,
 ---     --   ...
 ---     -- }
----     -- Now the picker_config_key will be applied every time you call this
----     -- builtin picker
+---     -- Options set here will be applied everytime you call this builtin
+---     -- picker. Each entry accepts a subset of the "defaults" configuration
+---     -- options, as well as the picker's own specific configuration options.
+---     -- Defaults options that are also valid for pickers are marked as
+---     -- "Picker-Overridable" in telescope.setup.defaults. Learn about one
+---     -- picker's specific set of options at telescope.builtin.$PICKER_NAME
 ---   },
 ---   extensions = {
 ---     -- Your extension configuration goes here:
 ---     -- extension_name = {
 ---     --   extension_config_key = value,
 ---     -- }
----     -- please take a look at the readme of the extension you want to configure
+---     -- Extensions should generally accept the same subset of defaults config
+---     -- options as builtin pickers, although that may differ depending on the
+---     -- extension. Make sure to read the extension's README or builtin
+---     -- documentation to know its specific configuration options.
 ---   }
 --- }
 --- </code>
--- a/lua/telescope/config.lua
+++ b/lua/telescope/config.lua
@@ -123,6 +123,7 @@ local layout_config_description = string.format(
     the screen width for all strategies except 'center', which has width
     of 50%% of the screen width.
 
+    Picker-Overridable
     Default: %s
 ]],
   vim.inspect(layout_config_defaults, { newline = "\n    ", indent = "  " })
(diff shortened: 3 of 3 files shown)


lua/telescope/init.lua
```diff
@@ -106,18 +106,26 @@ local telescope = {}
 ---   pickers = {
 ---     -- Default configuration for builtin pickers goes here:
 ---     -- picker_name = {
----     --   picker_config_key = value,
+---     --   default_config_key = value_one
```
jamestrew, line 109:
```suggestion
---     --   default_config_key = value_one, -- optionally override any default options
```

lua/telescope/builtin/init.lua
```diff
@@ -11,13 +11,17 @@
 ---
 --- :lua require('telescope.builtin').$NAME_OF_PICKER()
 ---
---- To use any of Telescope's default options or any picker-specific options, call your desired picker by passing a lua
---- table to the picker with all of the options you want to use. Here's an example with the live_grep picker:
+--- To configure your desired picker, call it by passing a lua table with all
+--- of the options you want to use. Pickers have their own specific set of
+--- options, but also accept a subset of the defaults options. The default
+--- options which are valid to configure a picker are marked as
+--- "Picker-Overridable" in telescope.setup.defaults
```
jamestrew, line 18: It'll probably be easier to mark options that aren't "picker-overriable". 
In theory, all but a small handful can at least in technically be overriden by the each individual picker.

I would reword this bit slightly and let's leave out the marking of individual options an "Picker-Overridable" for now.

lua/telescope/init.lua
```diff
@@ -106,18 +106,26 @@ local telescope = {}
...
 ---     --   ...
 ---     -- }
----     -- Now the picker_config_key will be applied every time you call this
----     -- builtin picker
+---     -- Options set here will be applied everytime you call this builtin
+---     -- picker. Each entry accepts a subset of the "defaults" configuration
+---     -- options, as well as the picker's own specific configuration options.
+---     -- Defaults options that are also valid for pickers are marked as
+---     -- "Picker-Overridable" in telescope.setup.defaults. Learn about one
+---     -- picker's specific set of options at telescope.builtin.$PICKER_NAME
```
jamestrew, line 118:
```suggestion
---     -- Options set here will be applied everytime you call this builtin
---     -- picker. The pickers will also inherit the "default" options unless 
---     -- they are overwritten.
---     -- See `:h telescope.builtin` for options unique to each picker.
```

jamestrew (changes requested):
Apologies for the very late response.

I have a few suggestions but otherwise looks good.

PR #3258, opened 2024-08-15
fix(builtin.vim_options): fix getting current option values

sid-6581:
# Description

The current builtin picker `vim_options` did not correctly get the current value of all options. This PR fixes that.

## Type of change

# How Has This Been Tested?

1. `nvim test.lua`
2. `:Telescope vim_options`
3. Search for `'filetype`

Result before change: 

image

Result after change:

image

**Configuration**:
* Neovim version (nvim --version): NVIM v0.11.0-dev-611+g615f7bbff
* Operating system and version: Windows 11

# Checklist:

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1003,19 +1003,14 @@ function make_entry.gen_from_vimoptions(opts)
       display = make_display,
       value = {
         name = o.name,
-        value = o.default,
+        value = o.value,
         type = o.type,
         scope = o.scope,
       },
-      ordinal = string.format("%s %s %s", o.name, o.type, o.scope),
+      ordinal = string.format("%s %s %s %s", o.name, o.type, o.scope, utils.display_termcodes(tostring(o.value))),
     }
 
-    local ok, value = pcall(vim.api.nvim_get_option_value, o.name, {})
-    if ok then
-      entry.value.value = value
-      entry.ordinal = entry.ordinal .. " " .. utils.display_termcodes(tostring(value))
-      return make_entry.set_default_entry_mt(entry, opts)
-    end
+    return make_entry.set_default_entry_mt(entry, opts)
   end
 end
 
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1010,7 +1010,7 @@ function make_entry.gen_from_vimoptions(opts)
       ordinal = string.format("%s %s %s", o.name, o.type, o.scope),
     }
 
-    local ok, value = pcall(vim.api.nvim_get_option, o.name)
+    local ok, value = pcall(vim.api.nvim_get_option_value, o.name, { buf = opts.bufnr })
     if ok then
       entry.value.value = value
       entry.ordinal = entry.ordinal .. " " .. utils.display_termcodes(tostring(value))
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1010,15 +1010,12 @@ function make_entry.gen_from_vimoptions(opts)
       ordinal = string.format("%s %s %s", o.name, o.type, o.scope),
     }
 
-    local ok, value = pcall(vim.api.nvim_get_option_value, o.name, { buf = opts.bufnr })
+    local ok, value = pcall(vim.api.nvim_get_option_value, o.name, {})
     if ok then
       entry.value.value = value
       entry.ordinal = entry.ordinal .. " " .. utils.display_termcodes(tostring(value))
-    else
-      entry.ordinal = entry.ordinal .. " " .. utils.display_termcodes(tostring(o.default))
+      return make_entry.set_default_entry_mt(entry, opts)
     end
-
-    return make_entry.set_default_entry_mt(entry, opts)
   end
 end
 
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -647,7 +647,11 @@ end
 internal.vim_options = function(opts)
   local res = {}
   for _, v in pairs(vim.api.nvim_get_all_options_info()) do
-    table.insert(res, v)
+    local ok, value = pcall(vim.api.nvim_get_option_value, v.name, {})
+    if ok then
+      v.value = value
+      table.insert(res, v)
+    end
   end
   table.sort(res, function(left, right)
     return left.name < right.name
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -679,7 +679,9 @@ internal.vim_options = function(opts)
           end
 
           vim.api.nvim_feedkeys(
-            string.format("%s:set %s=%s", esc, selection.value.name, selection.value.value),
+            selection.value.type == "boolean"
+            and string.format("%s:set %s!", esc, selection.value.name)
+            or string.format("%s:set %s=%s", esc, selection.value.name, selection.value.value),
             "m",
             true
           )
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -679,9 +679,8 @@ internal.vim_options = function(opts)
           end
 
           vim.api.nvim_feedkeys(
-            selection.value.type == "boolean"
-            and string.format("%s:set %s!", esc, selection.value.name)
-            or string.format("%s:set %s=%s", esc, selection.value.name, selection.value.value),
+            selection.value.type == "boolean" and string.format("%s:set %s!", esc, selection.value.name)
+              or string.format("%s:set %s=%s", esc, selection.value.name, selection.value.value),
             "m",
             true
           )


lua/telescope/make_entry.lua
```diff
@@ -1010,7 +1010,7 @@ function make_entry.gen_from_vimoptions(opts)
       ordinal = string.format("%s %s %s", o.name, o.type, o.scope),
     }
 
-    local ok, value = pcall(vim.api.nvim_get_option, o.name)
+    local ok, value = pcall(vim.api.nvim_get_option_value, o.name, { buf = opts.bufnr })
```
jamestrew, line 1013: This will always look for buffer local option values even for global and window local options and throw an error (which will get caught by the pcall and not give you the true value of the option).

```
E5108: Error executing lua [string ":lua"]:1: 'buf' cannot be passed for global option 'breakat'
stack traceback:
        [C]: in function 'nvim_get_option_value'
```
```
E5108: Error executing lua [string ":lua"]:1: 'buf' cannot be passed for window-local option 'colorcolumn'
stack traceback:
        [C]: in function 'nvim_get_option_value'
```

Maybe something like this to account for this 
```diff
diff --git a/lua/telescope/make_entry.lua b/lua/telescope/make_entry.lua
index a1806f3..41462d4 100644
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -998,6 +998,16 @@ function make_entry.gen_from_vimoptions(opts)
     }
   end

+  local get_option = function(name, scope)
+    if scope == "global" then
+      return vim.api.nvim_get_option_value(name, {})
+    elseif scope == "win" then
+      return vim.api.nvim_get_option_value(name, { win = opts.winnr })
+    elseif scope == "buf" then
+      return vim.api.nvim_get_option_value(name, { buf = opts.bufnr })
+    end
+  end
+
   return function(o)
     local entry = {
       display = make_display,
@@ -1010,7 +1020,7 @@ function make_entry.gen_from_vimoptions(opts)
       ordinal = string.format("%s %s %s", o.name, o.type, o.scope),
     }

-    local ok, value = pcall(vim.api.nvim_get_option, o.name)
+    local ok, value = pcall(get_option, o.name, o.scope)
     if ok then
       entry.value.value = value
       entry.ordinal = entry.ordinal .. " " .. utils.display_termcodes(tostring(value))
```

jamestrew (changes requested):

lua/telescope/make_entry.lua
sid-6581: It doesn't throw an error for me on nightly, but I haven't tried on earlier versions. Which version are you on where it's throwing the error? I tried to see when `nvim_get_option_value` was introduced, and apparently it was in 0.9, but maybe the behavior changed. How does `{}` work for the opts for you?

Also, `opts.scope` doesn't seem to be documented anywhere, did I miss it? The scopes you listed don't align with the scopes `nvim_get_option_value` expects, so if that's not documented anywhere that is a bit confusing.

jamestrew: I'm on nightly as well.
I'm just doing `:lua =vim.api.nvim_get_option_value('colorcolumn', {buf=1})`. 
I didn't mean telescope throws an error. The `pcall` catches it so.

`o.scope` is coming from `vim.api.nvim_get_all_options_info()` and it is indeed different from the `global`/`local` scope scopes of `nvim_get_option_value`

sid-6581: Thanks, I see what you mean. I think just `{}` instead of `{ buf = opts.bufnr }` is more appropriate, since it will get the local value if there is one, and the global value otherwise.

I'm not quite sure what to think about `nvim_get_all_options_info()`. Here's what I'm getting:

```
:0Verbose =vim.api.nvim_get_all_options_info().aleph
{
  allows_duplicates = true,
  commalist = false,
  default = 224,
  flaglist = false,
  global_local = false,
  last_set_chan = 0,
  last_set_linenr = 0,
  last_set_sid = 0,
  name = "aleph",
  scope = "global",
  shortname = "al",
  type = "number",
  was_set = false
}
```

However, if I try to get that value:

`vim.api.nvim_get_option_value("aleph", {})`

image

`vim.api.nvim_get_option_value("aleph", { scope = "global" })`

image

Looking at the documentation, it looks like this was a vim option that was removed in neovim, but it's still being returned  by `nvim_get_all_options_info()`. I'm not quite sure what the best way is to filter those out, maybe that was the reason for the pcall in the first place? I'd appreciate your input here, you probably have more experience here than I do. To me it looks like the pcall with `{}` for opts may work, but as always there are a lot of edge cases with everything.

jamestrew: > Thanks, I see what you mean. I think just {} instead of { buf = opts.bufnr } is more appropriate, since it will get the local value if there is one, and the global value otherwise.

Well the problem is, if you use `{}` and there's a local option, it will current the current bufnr/winid, which is going to be the one associated with the telescope picker. So if you looked up the `filetype` option with `:Telescope vim_options` when you have a lua open, it won't say 'lua' but it will be blank. So your initial instinct for creating this PR is correct. We just need to make sure to pass bufnr/winid explicitly *ONLY* when it's appropriate to do so.

> Looking at the documentation, it looks like this was a vim option that was removed in neovim, but it's still being returned by nvim_get_all_options_info(). I'm not quite sure what the best way is to filter those out, maybe that was the reason for the pcall in the first place? I'd appreciate your input here, you probably have more experience here than I do. 

Yeah I guess filtering them out is better than displaying the default value.
We can do an early return (ie return nil) to filter out entries.
```diff
diff --git a/lua/telescope/make_entry.lua b/lua/telescope/make_entry.lua
index a1806f3..1613029 100644
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1015,7 +1015,7 @@ function make_entry.gen_from_vimoptions(opts)
       entry.value.value = value
       entry.ordinal = entry.ordinal .. " " .. utils.display_termcodes(tostring(value))
     else
-      entry.ordinal = entry.ordinal .. " " .. utils.display_termcodes(tostring(o.default))
+      return
     end

     return make_entry.set_default_entry_mt(entry, opts)
```

sid-6581: On my machine using `{}` returns the correct value for the current buffer, not the telescope buffer, but I did fear that might be a race condition. If you are seeing the telescope buffer values, that is definitely not what we want.

I'll do some more digging tomorrow and see if I can find a clean solution. I definitely didn't anticipate this behavior from these nvim APIs.

jamestrew: > So if you looked up the filetype option with :Telescope vim_options when you have a lua open, it won't say 'lua' but it will be blank.

You described this behavior yourself in your test instructions.

You can just implement the changes in the diffs. Pretty sure they work and are inline with the behavior expected from the APIs based on the docs.

sid-6581: > You described this behavior yourself in your test instructions.

No, that's not actually what's happening. With the `nvim_get_option_value` and `{}` for opts, it correctly gets the values for the buffer I'm in, not the Telescope buffer.

With the original `nvim_get_option`, it behaves very oddly. If I open nvim with a blank file with `nvim test.lua`, Telescope shows the filetype as blank. However, if I just do `nvim` and then `:set filetype=lua`, Telescope shows filetype as lua. This seems like a bug in `nvim_get_option`.
 
> You can just implement the changes in the diffs. Pretty sure they work and are inline with the behavior expected from the APIs based on the docs.

Unfortunately that is getting the wrong values. I tested it like this:

image

And the result is:

image

The values to the right of `|!=|` are from `nvim_get_option_value(o.name, {})` and are correct. The values on the left are from `get_option` and are incorrect.

jamestrew: Ok yeah it takes the global option value if buffer local option value is not set or set to some value that indicates the global value is to be used (undolevels' -123456).

I think using `nvim_get_option_value(o.name, {})` is the way to go like you say. My only concern is that it's relying on some lua event loop timings to work.

If you try these print statements one at a time, you'll see that what the "current" buffer is, is different depending on when the `nvim_get_option_value` call is made on the event loop.

```diff
diff --git a/lua/telescope/make_entry.lua b/lua/telescope/make_entry.lua
index 8ad849a..679a15c 100644
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1010,6 +1010,15 @@ function make_entry.gen_from_vimoptions(opts)
       ordinal = string.format("%s %s %s", o.name, o.type, o.scope),
     }
 
+    -- should be original buffer number and telescope prompt buffer bufnr
+    -- but it's not, they match
+    print("unscheduled", opts.bufnr, vim.api.nvim_get_current_buf())
+
+    vim.schedule(function()
+      -- now they are different
+      print("scheduled", opts.bufnr, vim.api.nvim_get_current_buf())
+    end)
+
     local ok, value = pcall(vim.api.nvim_get_option_value, o.name, {})
     if ok then
       entry.value.value = value
```

I think we should move the option value getting logic out of the entry maker and into the picker here https://github.com/nvim-telescope/telescope.nvim/blob/5972437de807c3bc101565175da66a1aa4f8707a/lua/telescope/builtin/__internal.lua#L649-L651

sid-6581: I agree, I don't trust the timing here, it seems prone to race conditions. Let me try to move things around like you suggested.

sid-6581: @jamestrew I moved the code over to where it more naturally belongs, and I also fixed the default handling of booleans. When selecting an option, it would put `:set option=value` in the command line, which doesn't work for booleans, so I changed it to `:set option!` for boolean options, which will toggle it. The other option would be to either use `:set option`/`:set nooption`, but I thought toggling would be the most useful.

sid-6581: I forgot to change my setup to fallback to stylua when a configuration file is present after I switched to formatting my own code with EmmyLua. I'll fix that tomorrow and take care of the formatting issues.

sid-6581: @jamestrew I think it should be good to go now if you agree with the changes I made

jamestrew:
@sid-6581 sorry I'm just getting around to this PR again. Is it alright I reopen?

sid-6581:
> @sid-6581 sorry I'm just getting around to this PR again. Is it alright I reopen?

Sure! Thanks.

jamestrew (approved):

jamestrew:
Thanks!

Issue #3032, opened 2024-04-07
Freezes on single line files, ignoring preview timeout


pejrich:
### Description

If I'm searching for a file, my Telescope will frequently lock up. I have quite a number of 5-20MB JSON files in the project. If the search query ever causes one of these files to be the top sorted result, my nvim would freeze for about 10-20 seconds. I tried adding preview timeout, no luck. I've just made the connection that it's specifically on _single line_ files that it locks up. If I format the file that was giving me issues, then trying the same keystrokes again in Telescope find_files will correctly trigger the "previewer timed out". But for unformatted, single-line files, it freezes the whole nvim considerably. 

### Neovim version

```markdown
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1703358377
```

### Operating system and version

macOS 14.0

### Telescope version / branch / rev

latest

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()

Checking for required plugins ~
- OK plenary installed.
- OK nvim-treesitter installed.

Checking external dependencies ~
- OK rg: found ripgrep 14.1.0
- OK fd: found fd 9.0.0

===== Installed extensions ===== ~

Telescope Extension: `file_browser` ~
- No healthcheck provided

Telescope Extension: `fzf` ~
- OK lib working as expected
- OK file_sorter correctly configured
- OK generic_sorter correctly configured

Telescope Extension: `ui-select` ~
- No healthcheck provided
```

### Steps to reproduce

1. download this directory of 2 3mb dummy json files(plus some empty files). Open Telescope find_files, move the selection up and down. Once you hit the unformatted file, Telescope will freeze for a few seconds, and even when you're no longer selecting that file, Telescope still lags.
2. Format the unformatted file, or duplicate the formatted file. Now we have roughly the same total amount of data, but this time Telescope is buttery smooth the whole way through.
[telescope_bug.zip](https://github.com/nvim-telescope/telescope.nvim/files/14895783/telescope_bug.zip)

In this demo the delay is shorter as I didn't think you'd want some huge files, so these are only 3MB, but it should be enough to demonstrate the issue. When the files get up above 10-20MB, it's unresponsive for a good 10+ seconds.

### Expected behavior

The previewer should timeout out without freezing nvim.

### Actual behavior

The previewer takes a few seconds to timeout.

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
[...]

PR #3261, opened 2024-08-18
fix(previewer): improve `file_maker` line splitting and timeouts

jamestrew:
Previous method of splitting files into lines and incorporating a timer is sub-optimal in two ways
- not that fast
- since it's only checking for timeouts between each line split, huge lines can and does lead to exceeding the timeout potentially significantly (https://github.com/nvim-telescope/telescope.nvim/issues/3032)

By avoiding using `vim.gsplit`, we can avoid some overheads it needs to account for patterns and options, and we can check for timeouts more incrementally (every 200 bytes) to avoid getting caught up in huge singular lines.

Based on quick and dirty benchmarks against large-ish files in the issue above, I'm seeing about a 4-5x speed increase.

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -60,25 +60,6 @@ local function defaulter(f, default_opts)
   }
 end
 
--- modified vim.split to incorporate a timer
-local function split(s, sep, plain, opts)
-  opts = opts or {}
-  local t = {}
-  for c in vim.gsplit(s, sep, plain) do
-    local line = opts.file_encoding and vim.iconv(c, opts.file_encoding, "utf8") or c
-    table.insert(t, line)
-    if opts.preview.timeout then
-      local diff_time = (vim.uv.hrtime() - opts.start_time) / 1e6
-      if diff_time > opts.preview.timeout then
-        return
-      end
-    end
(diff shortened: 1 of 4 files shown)


clason:
@jamestrew rebase?

Conni2461:
thanks :)

clason:
Sorry for not testing this earlier, but this triggers an error with file previews (on Nvim nightly):
```
vim.schedule callback: ...m/lazy/telescope.nvim/lua/telescope/previewers/utils.lua:267: attempt to compare boolean with number
stack traceback:
        ...m/lazy/telescope.nvim/lua/telescope/previewers/utils.lua:267: in function 'timed_split_lines'
        ...scope.nvim/lua/telescope/previewers/buffer_previewer.lua:189: in function 'fn'
        [string "vim/_core/editor"]:404: in function <[string "vim/_core/editor"]:403>
```

This is with `opts.preview.timeout = false`, which is a valid option to not time out.

**Workaround:** Set `opts.preview.timeout = math.huge` to disable the timeout.

Conni2461:
does this help? https://github.com/nvim-telescope/telescope.nvim/pull/3662

clason:
Yes! (I would also have considered adapting the documentation, but this is less breaking.)

PR #3266, opened 2024-08-22
fix: #3265 use pairs in Pickers:delete_selection

### LINKED ISSUE #3265: Picker:delete_selection doesn't work if entry_maker returns nil
### Description

`Picker:delete_selection` uses `ipairs` to iterate over the `finder.results`:

https://github.com/nvim-telescope/telescope.nvim/blob/5972437de807c3bc101565175da66a1aa4f8707a/lua/telescope/pickers.lua#L836-L840

If the `entry_maker` for a one shot job returns nil for one of the lines, it makes the indices discontinuous:

https://github.com/nvim-telescope/telescope.nvim/blob/5972437de807c3bc101565175da66a1aa4f8707a/lua/telescope/finders/async_oneshot_finder.lua#L66-L79

Which means the `ipairs` iteration won't see any of the results after the nil.

Having the `entry_maker` return `{}` instead isn't ideal because then the results count is off (i.e. the empty rows aren't displayed but they still show up in the count of results).

I think the correct behavior is for `Picker:delete_selection` to use `pairs` instead.

### Neovim version

```markdown
NVIM v0.10.1
Build type: Release
LuaJIT 2.1.1720049189
```

macos 14.6

telescope 0.1.8

### checkhealth telescope

```markdown

- OK plenary installed.

- OK rg: found ripgrep 14.1.0
- OK fd: found fd 10.1.0

Telescope Extension: `advanced_git_search` ~
- No healthcheck provided

Telescope Extension: `fzf` ~
- OK lib working as expected

Telescope Extension: `neoclip` ~
- No healthcheck provided

Telescope Extension: `session-lens` ~
- No healthcheck provided

- No healthcheck provided
```

### Steps to reproduce

1. Use minimal config
2. Run `touch .repro/data/nvim/sessions/ax.vim` to create an empty session file (do this in the directory used for the repro). This will cause the entry_maker to return `nil` for this file
3. run `nvim -u repro.lua`
5. Save a session via `:SessionSave`
6. Open session picker via `:SessionSearch`
7. Try to delete the session you just created (the only one in the picker) with <C-d>

### Expected behavior

The session is deleted and the picker is updated

### Actual behavior

Nothing happens because `Picker:delete_selection` doesn't find the result so it doesn't call the callback to perform the delete

### Minimal config
```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    lazypath,
  })
end

-- install plugins
local plugins = {
  {
    dependencies = {
      "nvim-lua/plenary.nvim",
    },
    config = function()
      require("telescope").setup({})
    end,
  },
  {
    "rmagatti/auto-session",
    lazy = false,
    opts = {},
  },
}

  root = root .. "/plugins",
})
```

cameronr:
ipairs won't find find all results if entry_maker happens to return nil for a line.

# Description

Fixing Pickers:delete_selection so it finds all results even if entry_maker returns nil

Fixes #3265

## Type of change

# How Has This Been Tested?

- [x] make test
- [x] fixes repro case described in the issue

**Configuration**:
NVIM v0.10.1
Build type: Release
LuaJIT 2.1.1720049189

macOS 14.6

# Checklist:

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -833,7 +833,7 @@ function Picker:delete_selection(delete_cb)
   end
 
   local selection_index = {}
-  for result_index, result_entry in ipairs(self.finder.results) do
+  for result_index, result_entry in pairs(self.finder.results) do
     if vim.tbl_contains(delete_selections, result_entry) then
       table.insert(selection_index, result_index)
     end


jamestrew:
Apologies for the delay and thanks!

PR #3271, opened 2024-08-30
fix(builtin.oldfiles): no output when cwd is root

ronandalton:
# Description

When oldfiles is configured to only show files in the current working directory and the current working directory is root, no results will be displayed because a slash is appended to a path string that already ends with a slash. The line that does this is not needed because the `buf_in_cwd` function already does this (and only adds it if needed).

## Type of change

# How Has This Been Tested?

- [ ] Test A:
      1. Launch nvim from the root directory
      2. Start the oldfiles picker
      3. Observe that nothing will be shown

**Configuration**:
* Neovim version (nvim --version): v0.10.1
* Operating system and version: Linux 6.9.3

# Checklist:


jamestrew:
Thanks!

Issue #2544, opened 2023-05-28
Option to show source in diagnostic picker


rafamadriz:
**Is your feature request related to a problem? Please describe.**
When there's multiple servers attached, and opening diagnostics picker, I don't know which diagnostic is coming from which server. 

**Describe the solution you'd like**
Add option to diagnostic picker to show source. 

**Describe alternatives you've considered**

**Additional context**
There's issue #2110 that proposes to directly append the source to diagnostic message, but I think it's better to add it as an option

jamestrew:
Should be fairly simple. I can take a look at this over the next week or so.

shashanktomar:
Hey Folks, any update on this?

PR #3281, opened 2024-09-07
feat: extend builtin diagnostic to allow displayed text customization

dansdantas:
This extends `builtin.diagnostic` opts, allowing to use a custom function. Solves #2544

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1831,6 +1831,8 @@ builtin.diagnostics({opts})                  *telescope.builtin.diagnostics()*
         {sort_by}             (string)          sort order of the diagnostics
                                                 results; see above notes
                                                 (default: "buffer")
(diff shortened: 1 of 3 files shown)


jamestrew:
How does this resolve #2544?

dansdantas:
Allow to when defining diagnostic picker to pass a function in the text option that formats how text is displayed on the picker.  First argument for that function is the diagnostic, allowing to add source into the displayed text. Then you can filter like normally you would do with telescope.
```lua
require('telescope.builtin').diagnostics({
    text = function(diagnostic)
        return vim.trim(diagnostic.message:gsub("[\n]", "") .. " (" .. diagnostic.source .. ")")
    end,
})
```

jamestrew:
I think I would prefer we just implement that as oppose to adding a mechanism that allows people the option to implement it.

Issue #3262, opened 2024-08-18
sort_lastused selection behavior


nowaylifer:
### Description

When using `buffers` picker with `sort_lastused` option enabled,  entry at index 2 stays selected until only one entry left, even though there's better match at index 1. With additional option 'ignore_current_buffer` enabled selection works as expected.

### Neovim version

```markdown
NVIM v0.10.1
Build type: Release
LuaJIT 2.1.1720049189
```

### Operating system and version

arch linux

### Telescope version / branch / rev

master branch commit 5972437

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()

Checking for required plugins ~
- OK plenary installed.
- OK nvim-treesitter installed.

Checking external dependencies ~
- OK rg: found ripgrep 14.1.0
- OK fd: found fd 10.1.0

===== Installed extensions ===== ~

Telescope Extension: `fzf` ~
- OK lib working as expected
- OK file_sorter correctly configured
- OK generic_sorter correctly configured
```

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. open some buffers
3. run `Telescope buffers sort_lastused=true`
4. start typing

### Expected behavior

After start typing in `buffers` picker with `sort_lastused` option enabled, selection goes to entry at index 1.

### Actual behavior

https://github.com/user-attachments/assets/675b79b5-dc3f-4c24-bdce-db10dde68364

With `ignore_current_buffer` enabled:

https://github.com/user-attachments/assets/6f4569ee-abd5-40cf-8cee-6d1610cd63c6

### Minimal config

```Lua
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs { "config", "data", "state", "cache" } do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.uv.fs_stat(lazypath) then
  vim.fn.system {
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  }
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  {
    "nvim-telescope/telescope.nvim",
    dependencies = {
      "nvim-lua/plenary.nvim",
    },
    config = function()
[...]

PR #3289, opened 2024-09-12
fix(buffers): bad `sort_lastused` result selection

jamestrew:
Bad condition led to the selection being stuck at idx = 2 until only one result is left, despite idx = 1 result being a better match.

Neither `sort_lastused` or `ignore_current_buffer` should affect the position of the selection.
Former is used simply to sort the buffer list. The latter is used to filter out the current buffer.

closes https://github.com/nvim-telescope/telescope.nvim/issues/3262

### DIFF
--- a/lua/telescope/builtin/__internal.lua
+++ b/lua/telescope/builtin/__internal.lua
@@ -938,10 +938,6 @@ internal.buffers = function(opts)
   for i, bufnr in ipairs(bufnrs) do
     local flag = bufnr == vim.fn.bufnr "" and "%" or (bufnr == vim.fn.bufnr "#" and "#" or " ")
 
-    if opts.sort_lastused and not opts.ignore_current_buffer and flag == "#" then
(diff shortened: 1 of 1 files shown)


Conni2461:
This is actually a breaking change and was here for a reason. Sort last used matches the behavoir of fzf where we always select the last used buffer Index 2. Index 1 is always the current buffer.

jamestrew:
Huh... That behavior doesn't really make sense to me 😅 but I can revert this and make a note of that in the docs instead of you prefer. I don't have a strong opinion on this.

Conni2461:
we just need other behavior in the core picker. selection idx should be reset if default selection idx was set and you start typing. the idea behind having the previous buffer selected is that you can switch between current and previous buffer using telescope without navigating (you can do this with vim only ik, but ppl are ppl), this now is no longer possible.

see also https://github.com/nvim-telescope/telescope.nvim/issues/179

Issue #3284, opened 2024-09-10
builtin.git_status doesn't handle quoted file names


Lattay:
### Description

If `git status` produces a quoted file name (if the file name contain a space for example) the builtin picker `git_status` will not parse the result properly and include literal quotes in the file name.

### Neovim version

```markdown
NVIM v0.11.0-dev-720b309c7-dirty
Build type: Release
LuaJIT 2.1.1720049189
```

### Operating system and version

Arch

### Telescope version / branch / rev

master, current commit 5972437

### checkhealth telescope

```markdown
telescope: require("telescope.health").check()

Checking for required plugins ~
- OK plenary installed.
- OK nvim-treesitter installed.

Checking external dependencies ~
- OK rg: found ripgrep 14.1.0
- WARNING fd: not found. Install [sharkdp/fd](https://github.com/sharkdp/fd) for extended capabilities

===== Installed extensions ===== ~

Telescope Extension: `file_browser` ~
- No healthcheck provided

Telescope Extension: `fzf` ~
- OK lib working as expected
- OK file_sorter correctly configured
- OK generic_sorter correctly configured

Telescope Extension: `ui-select` ~
- No healthcheck provided
```

### Steps to reproduce

With `minimal.lua` in home directory.

```
mkdir repro
cd repro
git init
echo "test file" > 'file with space.txt'
nvim -nu ~/minimal.lua
```
Then do `:lua require("telescope.builtin").git_status()`, type `file` and `<CR>`.

### Expected behavior

Navigate to the file `file with space.txt`.

### Actual behavior

A new buffer named `"file with space.txt"` is open.

### Minimal config

```Lua
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs { "config", "data", "state", "cache" } do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.uv.fs_stat(lazypath) then
  vim.fn.system {
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  }
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  {
    "nvim-telescope/telescope.nvim",
    dependencies = {
      "nvim-lua/plenary.nvim",
    },
    config = function()
[...]

PR #3292, opened 2024-09-12
fix: git status quotes path with spaces

Lattay:
# Description

First try to match the path with quotes, then try without.

Fixes #3284

## Type of change

Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list relevant details about your configuration

```
mkdir repro
cd repro
git init
echo "test file" > 'file with space.txt'
nvim -nu ~/minimal.lua
```

Then do `:lua require("telescope.builtin").git_status()`, type `file` and `<CR>`.

**Configuration**:
* Neovim version (nvim --version): 
* Operating system and version: Arch

# Checklist:

- [x] My code follows the style guidelines of this project (stylua)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (lua annotations)

PR #3301, opened 2024-09-22
fix(git_status): use `-z` over `--porcelain`

jamestrew:
`-z` is more machine parse-able than `--porcelain`. But this requires splitting NUL delimited entries. Adds support for this in `async_oneshot_finder`/`LinesPipe`


LinoWhy:
Track to this commit with the same problem: path with spaces or special char in git_status.
@jamestrew can you help to merge this commit to fix the problem?
thx!

clason:
@jamestrew can you explain (in the commit message) what specific benefit `-z` has _for us_ (I like "Problem:... Solution:..." messages ;)) Just looking at the diff, it looks like `-z` is _harder_ to parse (for us, i.e., the code looks slightly more complicated)?

(I'm not arguing against the change, mind you, just asking for an explanation for posterity.)

jamestrew:
Updated the commit message.

```
fix(git_status): parse NUL-delimited git status output

Problem: `git status --porcelain` will wrap file paths with "special
characters" (eg `\n`, spaces) in double quotes which complicates
downstream parsing of results.

Furthermore, if a file path HAS a newline char, the existing approach
for getting results by splitting by `\n` is completely broken.

Solution: Switch to `git status -z`.
`-z` is more machine parse-able than plain `--porcelain` as git will no
longer treat paths with special characters specially, delimiting results
by NUL (the only non-valid character in unix paths) rather than LF while
keeping the benefit of `--porcelain` (stable across git versions and
user config).
Adds support for this in `async_oneshot_finder`/`LinesPipe` via a
new `split_char` option.
```

> it looks like -z is harder to parse

The parsing difficulty is really not any higher. Basically just doing `string.find(text, '\0', ...)` rather than `string.find(text, '\n', ...)`.
Most of the diff is just adding an option to the `new_oneshot_job` finder and `LinesPipe` (the thing that actually does the splitting) to support this. Using `LinesPipe` to do NUL-delimited splitting is maybe a bit weird though. It is hardcoded into finders though.

closes https://github.com/nvim-telescope/telescope.nvim/issues/3284 https://github.com/nvim-telescope/telescope.nvim/pull/3292

Issue #3310, opened 2024-10-01
Does live_grep work with number instead of string?


zigotica:
### Description

I am getting an error when trying to grep for number, nit sure if this is expected or not.

### Neovim version

```markdown
NVIM v0.10.0-dev-2593+ga6b6d036b-Homebrew
Build type: Release
LuaJIT 2.1.1725453128
```

### Operating system and version

macOS 15.0

### Telescope version / branch / rev

latest master

### checkhealth telescope

```markdown
Checking for required plugins ~
- OK plenary installed.
- OK nvim-treesitter installed.

Checking external dependencies ~
- OK rg: found ripgrep 14.1.0
- OK fd: found fd 9.0.0

===== Installed extensions ===== ~

Telescope Extension: `glyph` ~
- No healthcheck provided

Telescope Extension: `ui-select` ~
- No healthcheck provided

Telescope Extension: `undo` ~
- No healthcheck provided
```

### Steps to reproduce

1. open a project
2. :Telescope grep_string search=map
3. This opens  a modal with fzf-able files including that string (map)
4. :Telescope grep_string search=3
5. Should open a similar modal but throws error seen below

### Expected behavior

Opens a modal with fzf-able files including that number (3)

### Actual behavior

Error executing Lua callback: ...im/lazy/telescope.nvim/lua/telescope/builtin/__files.lua:21: attempt to index local 'string' (a number value)
stack traceback:
        ...im/lazy/telescope.nvim/lua/telescope/builtin/__files.lua:21: in function 'escape_chars'
        ...im/lazy/telescope.nvim/lua/telescope/builtin/__files.lua:205: in function 'v'
        ...im/lazy/telescope.nvim/lua/telescope/builtin/__files.lua:644: in function 'v'
        .../nvim/lazy/telescope.nvim/lua/telescope/builtin/init.lua:587: in function <.../nvim/lazy/telescope.nvim/lua/telescope/builtin/init.lua:546>
        ...share/nvim/lazy/telescope.nvim/lua/telescope/command.lua:188: in function 'run_command'
        ...share/nvim/lazy/telescope.nvim/lua/telescope/command.lua:259: in function 'load_command'
        ...ocal/share/nvim/lazy/telescope.nvim/plugin/telescope.lua:108: in function <...ocal/share/nvim/lazy/telescope.nvim/plugin/telescope.lua:107>

### Minimal config

```Lua
local root = vim.fn.fnamemodify("./.repro", ":p")

[...]

PR #3319, opened 2024-10-06
fix(grep_string): cast `search` value to string

jamestrew:
When using the vim command API and passing a number to the `search` option of `grep_string` (eg. `:Telescope grep_string search=3`), the number is passed as a number. This eventually causes a type error at `string.gsub` as the `search` value is expected to be a string.

We'll just cast `search` value to a string.

closes https://github.com/nvim-telescope/telescope.nvim/issues/3310


zigotica:
thank you!

PR #3327, opened 2024-10-15
fix(previewer): `preview=false` option inheritance

### LINKED ISSUE #3325: preview flag does not get overriden (for find_files picker)
### Description

When using
```
opts = {
	defaults = { preview = true },
	pickers = { find_files = { preview = false } }
}
```
to setup telescope, the find_files picker still shows a preview. I also tried passing `preview = false` directly into the `find_files` function, but that didn't work either.

### Neovim version

```markdown
NVIM v0.10.2
Build type: RelWithDebInfo
LuaJIT 2.1.1725453128
```

Arch Linux x86_64

master

### checkhealth telescope

```markdown
telescope: health#telescope#check

- OK plenary installed.

- OK rg: found ripgrep 14.1.1
- OK fd: found fd 10.2.0

Telescope Extension: `fzf` ~
- OK lib working as expected

- No healthcheck provided
```

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. press `<Space>f`

### Expected behavior

I expected the find_files picker to not show a preview.

### Actual behavior

The find_files picker showed a preview.

### Minimal config

```Lua

-- set stdpaths to use .repro
end

-- bootstrap lazy
	vim.fn.system({
		"git",
		"clone",
		"--filter=blob:none",
		lazypath,
	})
end

-- install plugins
local plugins = {
	{
		keys = {
			{
				"<Space>f",
				function()
					require("telescope.builtin").find_files() -- also reproducible with { preview = true } as args
				end,
			},
		},
		dependencies = { "nvim-lua/plenary.nvim" },
		opts = {
			defaults = { preview = true },
			pickers = { find_files = { preview = false } },
		},
	},
}

	root = root .. "/plugins",
})
```

jamestrew:
According to `:h telescope.defaults.preview`, we should be able to do `preview=false` to disable the previewer for pickers. With how picker options resolve inheritance, something like this should work.

```lua
require('telescope').setup {
  defaults = { preview = true },
  pickers = { find_files = { preview = false } }
}
```
Here, find_files should not show a previewer but it previously would.

Corrects this by checking if `opts.preview` is explicitly set to `false`. In which case, we won't show the previewer.

Closes #3325

### DIFF
--- a/lua/telescope/command.lua
+++ b/lua/telescope/command.lua
@@ -112,6 +112,9 @@ command.convert_user_opts = function(user_opts)
             user_opts[key] = eval
           else
             -- otherwise return nil (allows split check later)
+            -- BUG: this doesn't allow things like `preview=false`, where an
+            -- option supports other types other than the default type
+            -- (`preview` has a default type of table)
             user_opts[key] = nil
           end
         end
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -36,6 +36,9 @@ local function defaulter(f, default_opts)
   default_opts = default_opts or {}
   return {
     new = function(opts)
+      if opts.preview == false then
+        return false
+      end
       if conf.preview == false and not opts.preview then
         return false
       end


weisbrja:
Can this be merged pls?

weisbrja:
Hello?

weisbrja:
Can someone assign a reviewer and pls just merge this? It's literally only a few lines of code for an obvious bug fix.

weisbrja:
Please move this forward...

PR #3366, opened 2024-11-19
refactor: expose path_truncate

sassanh:
# Description

Third parties and custom pickers would benefit from `path_truncate` being exposed in the API, specially considering it is using `calc_result_length` and it is too complicated and hacky for other developers to come to the conclusion they can achieve line length like that, I made it public for myself and now I can truncate paths in my custom picker like this:

```lua
utils.path_truncate(display_filename, 1, {})
```

I can't imagine how I could achieve this myself without investing so much time.

Fixes # (issue)

## Type of change

- This change requires a documentation update --- does it? I'm not sure if similar items in the api are documented, for example `is_path_hidden` is not documented.

# How Has This Been Tested?

N/A as it is just exposing something already used internally and has been tested.

**Configuration**:

# Checklist:

- [?] I have made corresponding changes to the documentation (lua annotations)

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -207,7 +207,7 @@ local calc_result_length = function(truncate_len)
   return type(truncate_len) == "number" and len - truncate_len or len
 end
 
-local path_truncate = function(path, truncate_len, opts)
+utils.path_truncate = function(path, truncate_len, opts)
   if opts.__length == nil then
     opts.__length = calc_result_length(truncate_len)
   end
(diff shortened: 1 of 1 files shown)


Conni2461:
We prefer if third party picker would use path_display instead

sassanh:
@Conni2461
I think I'm missing something here, `path_display` is a config variable taking values of `hidden`, `tail`, `absolute`, etc as far as I understand, `path_truncate` on the other hand is a utility function that actually truncates path based on the available width, I understand third party pickers should respect `path_display` config variable, but how can they "use" it to actually implement truncating paths respecting available width?

Conni2461:
extensions should use `transform_path` (see `:help telescope.utils.transform_path`) which will then consume `path_display` correctly.

maybe i didnt get your use case, why you want to have `path_truncate` available publicly. in the past i'd like to have a somewhat slim interface so less extensions depend on functions "being there with that interface". but that was also in Nov 2024 and now its april 2026 so idk if this is still something you want to have merged :sweat_smile: I am sorry

PR #3368, opened 2024-11-24
fix: offset_encoding is required now in `make_position_params`

rockyzhang24:
Adapt to the upstream change https://github.com/neovim/neovim/commit/629483e24eed3f2c07e55e0540c553361e0345a2 to get rid of the warning message "warning: offset_encoding is required, using the offset_encoding from the first client".

Since the commit https://github.com/neovim/neovim/commit/2dcbfe78fcec5f73ce061bb24b718187b9c6b134, `params` in `buf_request` can accept a function that takes the client as a parameter. So instead of sending the same params to all servers, the params with correct offset_encoding will send to each single server.

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -68,8 +68,29 @@ local function pick_call_hierarchy_item(call_hierarchy_items)
   return call_hierarchy_items[choice]
 end
 
+---@param win number? Window handler
+---@param extra table? Extra fields in params
+---@return table|(fun(client: vim.lsp.Client): table) parmas to send to the server
+local function client_position_params(win, extra)
+  win = win or vim.api.nvim_get_current_win()
+  if vim.fn.has "nvim-0.11" == 0 then
+    local params = vim.lsp.util.make_position_params(win)
+    if extra then
+      params = vim.tbl_extend("force", params, extra)
+    end
+    return params
+  end
+  return function(client)
+    local params = vim.lsp.util.make_position_params(win, client.offset_encoding)
+    if extra then
+      params = vim.tbl_extend("force", params, extra)
+    end
+    return params
+  end
+end
+
 local function calls(opts, direction)
-  local params = vim.lsp.util.make_position_params()
+  local params = client_position_params()
   vim.lsp.buf_request(opts.bufnr, "textDocument/prepareCallHierarchy", params, function(err, result)
     if err then
       vim.api.nvim_err_writeln("Error when preparing call hierarchy: " .. err)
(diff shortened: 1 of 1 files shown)


rockyzhang24:
I just realized that the `has` function for checking version numbers returns `0` or `1` instead of `true` or `false`. I’ve fixed it, and everything should be fine now. It’s ready to be merged.

lua/telescope/builtin/__lsp.lua
```diff
@@ -263,19 +281,19 @@ end
 
 lsp.references = function(opts)
   opts.include_current_line = vim.F.if_nil(opts.include_current_line, false)
-  ---@class lsp.TextDocumentPositionParams
-  local params = vim.lsp.util.make_position_params(opts.winnr)
-  params.context = { includeDeclaration = vim.F.if_nil(opts.include_declaration, true) }
+  local params = client_position_params(vim.api.nvim_get_current_win(), {
```
Conni2461, line 284: `opts.winnr` was here and is now missing

lua/telescope/builtin/__lsp.lua
```diff
@@ -68,8 +68,26 @@ local function pick_call_hierarchy_item(call_hierarchy_items)
...
+  win = win or vim.api.nvim_get_current_win()
+  if vim.fn.has "nvim-0.11" == 0 then
+    local params = vim.lsp.util.make_position_params(win)
+    if extra then
+      params = vim.tbl_extend("force", params, extra)
+    end
+    return params
+  end
+  return function(client)
+    local ret = vim.lsp.util.make_position_params(win, client.offset_encoding)
```
Conni2461, line 81: lets rename `ret` => `params` to keep it consistent

lua/telescope/builtin/__lsp.lua
```diff
@@ -68,8 +68,26 @@ local function pick_call_hierarchy_item(call_hierarchy_items)
...
+    if extra then
+      ret = vim.tbl_extend("force", ret, extra)
+    end
+    return ret
+  end
+end
+
 local function calls(opts, direction)
-  local params = vim.lsp.util.make_position_params()
+  local params = client_position_params(vim.api.nvim_get_current_win())
```
Conni2461, line 90: lets pass in nothing here as this is the default of the func

lua/telescope/builtin/__lsp.lua
```diff
@@ -68,8 +68,26 @@ local function pick_call_hierarchy_item(call_hierarchy_items)
   return call_hierarchy_items[choice]
 end
 
+local function client_position_params(win, extra)
```
Conni2461, line 71: can you add typing to that function

Conni2461:
thanks for the contribution :) i left a couple of nits

PR #3392, opened 2024-12-29
feat(live_grep): attach mapping is configurable through options

richardmarbach:
# Description

I started writing a new picker to improve the refinement of the `live_grep` picker search results and noticed that all I really wanted to do was add custom `attachment_mappings` to the existing `live_grep` picker. It would be nice if the `attachment_mappings` were configurable instead of needing to create a copy of an existing picker.

## Type of change

# How Has This Been Tested?

I changed my mapping for live grep to use a custom `attach_mapping`:

```lua
local refine_word_action = function(prompt_bufnr)
  require("telescope.actions.generate").refine(prompt_bufnr, {
    prompt_title = "Refine word results",
  })
end

local attach_mappings = function(_, map)
  map("i", "<c-space>", refine_word_action)
  return true
end

vim.keymap.set("<leader>sg", function() require("telescope.builtin").live_grep({ attach_mappings = attach_mappings }) end)
```

To check:

1. Execute the mapping, e.g. `<leader>sg`
2. search some text to populate the search buffer (e.g. `telescope`)
3. press the refinement shortcut (e.g. `<c-space>`)

The prompt title should now read "Refine word results"

**Configuration**:
* Neovim version (nvim --version): NVIM v0.10.3
* Operating system and version: macOS Sequonia Version 15.1.1

# Checklist:


alerque (commented):
Should this option availability be documented somewhere?

Kaiser-Yang:
I think this is the correct behaviour. Now, we can not delete the default <c-space> mappings by configuration.

To do so, we must use filetype autocmd to delete one mapping users may not want.

BTW, maybe you should update the pickers.new instead of calls of the function.

https://github.com/nvim-telescope/telescope.nvim/blob/5255aa27c422de944791318024167ad5d40aad20/lua/telescope/pickers.lua#L1543-L1551

Kaiser-Yang:
I find a way do this without this PR easily, just delete the default key mappings in attach_mappings:

```lua
      lsp_dynamic_workspace_symbols = {
        attach_mappings = function(buffer)
          vim.keymap.del('i', '<c-space>', { buffer = buffer })
          return true
        end,
      },
```

PR #3410, opened 2025-02-10
 Check for nil where nil was giving me problems

jam1015:
# Description

When picking buffers, when filtering, if there is one left, I have my telescope configured to pick the last remaining entry automatically:

```lua
  pickers = {
    -- picker_name = {
    --   ...
    -- }
    -- Now the picker_config_key will be applied every time you call this
    -- builtin picker
    find_files = {
      --theme = "ivy"
      hidden = true,
    },
    buffers = {
      --theme = 'ivy',
      on_complete = {
        function(picker)
          local index = 0
          for _ in picker.manager:iter() do
            index = index + 1
            if index > 1 then
              break
            end
          end
          if index == 1 then
            local bufno = tonumber(picker.prompt_bufnr)
            local to_call = require('telescope.actions').select_default
            local _ = pcall(to_call, bufno)
          end
        end
      }
    },
  },

```

see full setup [here](https://github.com/jam1015/dotfiles/blob/master/.config/nvim/lua/plugin_configs/telescope.lua)

the problem is that I get error messages if this automatic selection gets triggered for a terminal buffer.  This small check for nil skips the code that was triggering the error.

## Type of change

checks for nil and skips line centering if the buffer number is nil

# How Has This Been Tested?

Me just running `:Telescope  buffers` and narrowing the list down to a single terminal buffer so that it is all that is left; there are no error messages anymore.

**Configuration**:
NVIM v0.10.4
Build type: RelWithDebInfo
LuaJIT 2.1.1736781742
# Checklist:

- [ check] My code follows the style guidelines of this project (stylua)
- [ check] I have performed a self-review of my own code
- [check] I have commented my code, particularly in hard-to-understand areas
- [check] I have made corresponding changes to the documentation (lua annotations)

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -545,9 +545,11 @@ previewers.vimgrep = defaulter(function(opts)
 
       local middle_ln = math.floor(lnum + (lnend - lnum) / 2)
       pcall(vim.api.nvim_win_set_cursor, self.state.winid, { middle_ln + 1, 0 })
-      vim.api.nvim_buf_call(bufnr, function()
-        vim.cmd "norm! zz"
-      end)
+      if bufnr ~= nil then
+        vim.api.nvim_buf_call(bufnr, function()
+          vim.cmd "norm! zz"
+        end)
+      end
     end
   end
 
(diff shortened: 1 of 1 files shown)


Conni2461:
We can do this. Thanks :)

Issue #3414, opened 2025-02-14
builtin.git_status auto close the picker when working dir is clean, make screen flicker.


xieyonn:
### Description

When open telescope with `builtin.git_status` and  the working directory is complete clean, the behavior  is as follows:

1. The picker panel open and then closes very quickly, make screen flicker.
2. Sends an `ERROR` message `[telescope.builtin.git_status]: No changes found` via `vim.notify`.

If the working directory is complete clean, there is no need to open then close the picker, it would be better to remain empty panel. Alternatively, we can check the working directory before opening the picker. 

Working directory is clean is definitely **NOT** an `ERROR`,  sending an `INFO` message is fine enough, just like how `builtin.diagnostic` works.

### Neovim version

```markdown
NVIM v0.10.4
Build type: Release
LuaJIT 2.1.1713484068
```

### Operating system and version

Linux Debian 12

### Telescope version / branch / rev

master (78857db)

### checkhealth telescope

```markdown
==============================================================================
telescope: health#telescope#check

Checking for required plugins ~
- OK plenary installed.
- OK nvim-treesitter installed.

Checking external dependencies ~
- OK rg: found ripgrep 13.0.0
- OK fd: found fdfind 8.6.0

===== Installed extensions ===== ~

Telescope Extension: `fidget` ~
- No healthcheck provided

Telescope Extension: `fzf` ~
- OK lib working as expected
- OK file_sorter correctly configured
- OK generic_sorter correctly configured

Telescope Extension: `workspaces` ~
- No healthcheck provided
```

### Steps to reproduce

1. create git repo, commit all, make pwd clean.
```sh
git init && git add --all && git commit --message=add
```
2. `nvim -nu minimal.lua`
3. use command to open git_status picker.
```vim
Telescope git_status
```

### Expected behavior

1. remain `git_status` picker open if working dir is clean.
2. send `INFO` message, something like `The working directory is clean.`

### Actual behavior

1. `git_status` picker open and then close very quickly, make screen flicker.
2. send `ERROR` message `[telescope.builtin.git_status]: No changes found`

### Minimal config

```Lua
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs { "config", "data", "state", "cache" } do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
[...]

PR #3415, opened 2025-02-15
fix(git_status): keep picker open if pwd is clean

xieyonn:
# Description

Fixes # (https://github.com/nvim-telescope/telescope.nvim/issues/3414)

## Type of change

# How Has This Been Tested?

1. create git repo, commit all, make pwd clean.
```sh
git init && git add --all && git commit --message=add
```
2. `nvim -nu minimal.lua`
3. use command to open git_status picker.
```vim
Telescope git_status
```

Expect behavier:
1. Send a `INFO` message.
2. Remain picker panel open.

**Configuration**:
```lua

-- set stdpaths to use .repro
for _, name in ipairs { "config", "data", "state", "cache" } do
end

-- bootstrap lazy
  vim.fn.system {
    "git",
    "clone",
    "--filter=blob:none",
    lazypath,
  }
end

-- install plugins
local plugins = {
  {
    dependencies = {
      "nvim-lua/plenary.nvim",
    },
    config = function()
      require("telescope").setup {}
    end,
  },
}

  root = root .. "/plugins",
})
```

Debian 12
# Checklist:

### DIFF
--- a/lua/telescope/builtin/__git.lua
+++ b/lua/telescope/builtin/__git.lua
@@ -407,9 +407,8 @@ git.status = function(opts)
           if count == 0 and prompt == "" then
             utils.notify("builtin.git_status", {
               msg = "No changes found",
-              level = "ERROR",
+              level = "INFO",
             })
-            actions.close(self.prompt_bufnr)
           end
         end,
       },


shawnyu5:
Is there anyway to revert to the old behavior? 

If the current directory does not contain any git changes, there is no reason I'd want an empty telescope pane. The old behavior of closing telescope if the directory is clean made a lot of sense. Reading through the issue, I dont fully understand the justification to change the behavior.

xieyonn:
@shawnyu5 Because it cause the screen to flicker.

Picker run commad `git status` in a async way, open the picker first, show the content while waiting for the command results. That make sense if we are in a big repo, since `git status` is an operation with a very high execution cost.

 I think a better way is to wait until the command has been executed completely before opening the picker panel. In this way, there won't be screen flickering, just like the picker related to the LSP. However, it seems that this would require relatively significant modifications.

shawnyu5:
The old behaviour of telescope closing when there are no git changes is the correct behavior imho. What’s the point of opening the telescope prompt if there are no git changes? The screen flicker caused by the opening and closing of the prompt was an acceptable thing for me.

Could we at least add a config option to still have the old behavior? Needing to press `<esc> <esc>` to close the prompt when there are no git changes is kinda annoying.

xieyonn:
Yes i agree, if no git changes, no any prompt should open.

Infact, you can add a `git status` check in a custom keybind
```lua
vim.keymap.set("n", "some key", function()
  -- check git status ahead
  if
    vim.fn.system(
      "git -C " .. vim.fn.getcwd(-1, 0) .. " status --porcelain"
    ) ~= ""
  then
    require("telescope.builtin").git_status()
  end
end, keyopts({ desc = "telescope git_status" }))
```

PR #3418, opened 2025-02-19
fix(lsp): add position encoding when calling `symbols_to_items()`'

dustypomerleau:
# Description

neovim/neovim#30860 introduced the `position_encoding` parameter in `symbols_to_items()`. This breaks `:Telescope lsp_document_symbols`, `lsp_workspace_symbols`, and `lsp_dynamic_workspace_symbols` for anyone running nightly Neovim.

## Type of change

EDIT: draft status removed. Should be a bug-fix only now.

# How Has This Been Tested?

Manual testing of the `Telescope` commands above.
NVIM v0.11.0-nightly+4a2a54f
macOS 15.3.1

# Checklist:

### DIFF
--- a/lua/telescope/builtin/__lsp.lua
+++ b/lua/telescope/builtin/__lsp.lua
@@ -345,7 +345,7 @@ end
 
 lsp.document_symbols = function(opts)
   local params = client_position_params(opts.winnr)
-  vim.lsp.buf_request(opts.bufnr, "textDocument/documentSymbol", params, function(err, result, _, _)
+  vim.lsp.buf_request(opts.bufnr, "textDocument/documentSymbol", params, function(err, result, ctx, _)
     if err then
       vim.api.nvim_err_writeln("Error when finding document symbols: " .. err.message)
       return
@@ -359,7 +359,14 @@ lsp.document_symbols = function(opts)
       return
     end
 
-    local locations = vim.lsp.util.symbols_to_items(result or {}, opts.bufnr) or {}
+    local locations
+    if vim.fn.has "nvim-0.11" == 1 then
+      local client = assert(vim.lsp.get_client_by_id(ctx.client_id))
+      locations = vim.lsp.util.symbols_to_items(result or {}, opts.bufnr, client.offset_encoding) or {}
+    else
+      locations = vim.lsp.util.symbols_to_items(result or {}, opts.bufnr) or {}
+    end
+
     locations = utils.filter_symbols(locations, opts, symbols_sorter)
     if vim.tbl_isempty(locations) then
       -- error message already printed in `utils.filter_symbols`
@@ -396,13 +403,20 @@ end
 
 lsp.workspace_symbols = function(opts)
   local params = { query = opts.query or "" }
-  vim.lsp.buf_request(opts.bufnr, "workspace/symbol", params, function(err, server_result, _, _)
+  vim.lsp.buf_request(opts.bufnr, "workspace/symbol", params, function(err, server_result, ctx, _)
     if err then
       vim.api.nvim_err_writeln("Error when finding workspace symbols: " .. err.message)
       return
     end
 
-    local locations = vim.lsp.util.symbols_to_items(server_result or {}, opts.bufnr) or {}
+    local locations
+    if vim.fn.has "nvim-0.11" == 1 then
+      local client = assert(vim.lsp.get_client_by_id(ctx.client_id))
+      locations = vim.lsp.util.symbols_to_items(server_result or {}, opts.bufnr, client.offset_encoding) or {}
+    else
+      locations = vim.lsp.util.symbols_to_items(server_result or {}, opts.bufnr) or {}
+    end
+
     locations = utils.filter_symbols(locations, opts, symbols_sorter)
     if vim.tbl_isempty(locations) then
       -- error message already printed in `utils.filter_symbols`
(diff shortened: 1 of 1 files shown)


sibouras:
> I'm open to suggestions on how to handle that.

u can add the new change inside  `vim.fn.has('nvim-0.11') == 1`

dustypomerleau:
> > I'm open to suggestions on how to handle that.
> 
> u can add the new change inside `vim.fn.has('nvim-0.11') == 1`

Thanks @sibouras! I've made the change. Should be ready for review now.

lua/telescope/builtin/__lsp.lua
```diff
@@ -359,7 +359,11 @@ lsp.document_symbols = function(opts)
       return
     end
 
-    local locations = vim.lsp.util.symbols_to_items(result or {}, opts.bufnr) or {}
+    if vim.fn.has "nvim-0.11" == 1 then
+      local locations = vim.lsp.util.symbols_to_items(result or {}, opts.bufnr, "utf-8") or {}
```
dhruvmanila, line 363: I don't think you can hard code the encoding value as different clients may use different encoding. You should get the encoding from the `vim.lsp.Client` object. The request handler function here is passed a third parameter `ctx` (https://neovim.io/doc/user/lsp.html#lsp-handler) which contains the `ctx.client_id`. This can be used to get the client object (`vim.lsp.get_client_by_id(ctx.client_id)`) and access the encoding as `client.offset_encoding`.

lua/telescope/builtin/__lsp.lua
```diff
@@ -359,7 +359,11 @@ lsp.document_symbols = function(opts)
...
+    else
+      local locations = vim.lsp.util.symbols_to_items(result or {}, opts.bufnr) or {}
+    end
```
dhruvmanila, line 366: Also, the `locations` variable is local to the `if` and `else` block, they're not visible outside them. You'll need to define the variable before the `if` statement:
```lua
local locations
if ... then
	locations = ...
else
	locations = ...
end
```

lua/telescope/builtin/__lsp.lua
```diff
@@ -452,7 +460,11 @@ local function get_workspace_symbols_requester(bufnr, opts)
       if client_res.error then
         vim.api.nvim_err_writeln("Error when executing workspace/symbol : " .. client_res.error.message)
       elseif client_res.result ~= nil then
-        vim.list_extend(locations, vim.lsp.util.symbols_to_items(client_res.result, bufnr))
+        if vim.fn.has "nvim-0.11" == 1 then
+          vim.list_extend(locations, vim.lsp.util.symbols_to_items(client_res.result, bufnr, "utf-8"))
+        else
+          vim.list_extend(locations, vim.lsp.util.symbols_to_items(client_res.result, bufnr))
+        end
```
dhruvmanila, line 467: For the dynamic workspace symbol requester, the client id is the `_` variable in this for loop. So,
```lua
for client_id, client_res in pairs(results) do
	...
end
```

dustypomerleau:
Thanks for the detailed review @dhruvmanila - I wasn't familiar with the API. Please have a look @jamestrew and see if you're happy.

lua/telescope/builtin/__lsp.lua
```diff
@@ -359,7 +359,14 @@ lsp.document_symbols = function(opts)
       return
     end
 
-    local locations = vim.lsp.util.symbols_to_items(result or {}, opts.bufnr) or {}
+    local locations = {}
+    if vim.fn.has "nvim-0.11" == 1 then
+      local offset_encoding = vim.lsp.get_clients({ opts.bufnr })[1].offset_encoding
```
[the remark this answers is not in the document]
dustypomerleau: You're correct. The reason it seems to work either way is because there was a [subsequent commit](https://github.com/neovim/neovim/commit/e8b5dd1e89bfa984e7b943443a291484cba23fac) providing a fallback `bufnr` value in `symbols_to_items()`.

lua/telescope/builtin/__lsp.lua
```diff
@@ -359,7 +359,14 @@ lsp.document_symbols = function(opts)
       return
     end
 
-    local locations = vim.lsp.util.symbols_to_items(result or {}, opts.bufnr) or {}
+    local locations = {}
```
jamestrew, line 362:
```suggestion
    local locations
```

There's some linting errors related to this and similarly down below.

lua/telescope/builtin/__lsp.lua
```diff
@@ -359,7 +359,14 @@ lsp.document_symbols = function(opts)
...
+    if vim.fn.has "nvim-0.11" == 1 then
+      local offset_encoding = vim.lsp.get_clients({ bufnr = opts.bufnr })[1].offset_encoding
+      locations = vim.lsp.util.symbols_to_items(result or {}, opts.bufnr, offset_encoding) or {}
```
jamestrew, line 365:
```suggestion
      local client = assert(vim.lsp.get_client_by_id(ctx.client_id))
      locations = vim.lsp.util.symbols_to_items(result or {}, opts.bufnr, client.offset_encoding) or {}
```
Actually, we don't have to blindly just take the first client.
We can get the `client_id` from the `ctx` value that's passed into this handler callback:
```lua
vim.lsp.buf_request(opts.bufnr, "textDocument/documentSymbol", params, function(err, result, ctx, _)
```

We can do the same thing below as well

dustypomerleau: Thanks for this - appears to be working now.

jamestrew:
Thanks!

PR #3420, opened 2025-02-21
fix: swap items of tags entry

Whenrow:
# Description
Tags entry reserves 16 char for the tag name and the remaining place is left for the kind symbols.
This commit orders tags name *then* kind symbol as tags name can be long function name while kind is only one word.

Here is an example of what it looks like 
### Before
image

### After 
image

## Type of change

# How Has This Been Tested?

- [ ] Generate tags file with tags name longer than 16 chars. Call command `:Telescope tags`. See the tags name are not truncated.

**Configuration**:
* Neovim version (nvim --version): 0.10
* Operating system and version: LinuxMint 22

# Checklist:

### DIFF
--- a/lua/telescope/make_entry.lua
+++ b/lua/telescope/make_entry.lua
@@ -1063,8 +1063,8 @@ function make_entry.gen_from_ctags(opts)
             return path_style
           end,
         },
-        entry.tag,
         entry.kind,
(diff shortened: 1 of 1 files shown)


Conni2461:
thanks :)

PR #3416, opened 2025-02-15
feat(builtin.treesitter) send user message if no symbols found.

xieyonn:
# Description

- Remove redundant empty check in `builtin.treesitter`.
- Send user a `INFO` message if found no symbols, just like how other builtin pickers works.

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [x]  Manual testing - open a empty file and open ":Telescope treesitter"
- [x]  Manual testing - open example file `a.lua` and open ":Telescope treesitter"
a.lua 
```
var a = {}
```

**Configuration**:
* Neovim version (nvim --version):
NVIM v0.10.4
Build type: Release
LuaJIT 2.1.171348406
* Operating system and version:
Debian 12 

# Checklist:

- [x] My code follows the style guidelines of this project (stylua)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (lua annotations)

PR #3449, opened 2025-04-17
fix(builtin.treesitter) Force eval treesitter

juliancoffee:
# Description
I'm not sure if that's the best way to go about it, but that's the solution I came up with.

Fixes two problems.
- `treesitter` locals aren't populated until it's executed for highlighting, indents, etc, so if you disable highlighting for a particular language (or all of them), it will return no symbols. I worked around this by forcing `parser:parse()`. If performance is unacceptable, we could make it an option, but for me, it worked well.
- Some files for some languages simply don't return any definitions at all, and Treesitter silently ignores them.
This PR displays them as errors. #3416 does something very similar, but because I needed both fixes for my fork, I include it as well 😅 

## Type of change

# How Has This Been Tested?

- [x] Test A
I've tested it Rust file while disabling treesitter highlighting.
Returns nothing before this PR, after this PR returns symbols as expected.
- [x] Test B
Tested on my own config files, which don't have any symbols that the Lua parser recognised as locals.
```lua
return {
  'example',
}
```

**Configuration**:
```
NVIM v0.11.0
Build type: Release
LuaJIT 2.1.1744318430
```

```
Darwin Kernel Version 24.3.0: Thu Jan  2 20:23:36 PST 2025
```

# Checklist:


juliancoffee:
https://github.com/nvim-treesitter/nvim-treesitter/issues/7827

Ok, so I was told in treesitter repo that this is basically what we should do, and that parsing for already parsed file would be a no-op.

clason:
superseded by #3566 (but thanks!)

Issue #2779, opened 2023-11-12
The default maximum result display count seems to have been mistakenly set to `250`


leonasdev:
### Description

In: https://github.com/nvim-telescope/telescope.nvim/blob/20bf20500c95208c3ac0ef07245065bf94dcab15/lua/telescope/pickers.lua#L561C1-L565C44 , the comment seems to indicate that the default value for `max_result` should be `10000`.

But in: https://github.com/nvim-telescope/telescope.nvim/blob/20bf20500c95208c3ac0ef07245065bf94dcab15/lua/telescope/pickers.lua#L319 , `__scrolling_limit` is set to `250` when `opt.temp__scrolling_limit` is not present which seems wrong.

When users search for keywords, they may scroll beyond `250` entries to find a specific item. If the actual number of entries exceeds `250`, users might mistakenly believe that they have thoroughly searched for the keywords.

And traversing `250` entries is more likely compared to navigating through `10000` entries.

So, I believe setting `max_result` to `10000` is more reasonable, or alternatively, allowing users to configure this value based on their performance preferences would be a better option.

### Neovim version

```markdown
NVIM v0.9.2
Build type: Release
LuaJIT 2.1.1692716794
```

### Operating system and version

Ubuntu 22.04

### Telescope version / branch / rev

master

### checkhealth telescope

```markdown
==============================================================================
telescope: require("telescope.health").check()

Checking for required plugins ~
- OK plenary installed.
- OK nvim-treesitter installed.

Checking external dependencies ~
- OK rg: found ripgrep 13.0.0
- OK fd: found fd 8.3.1
```

### Steps to reproduce

1. `nvim`
2. `:Telescope highlights`
3. The prompt shows the total of result may beyond `250` (like `500` or something)
4. When you scolling results list, you'll find that not all results is on the Results list.

### Expected behavior

The `max_result` value should set to something like `10000` not `250`, or alternatively, allowing users to configure this value based on their performance preferences would be a better option.

### Actual behavior

The `max_result` seems accidentally set to `250`, whcih users might mistakenly believe that they have thoroughly searched for the keywords.

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
[...]

PR #3471, opened 2025-05-17
feat(pickers): allow the scrolling limit to be configured

davvid:
Promote the "temp__scrolling_limit" option into an officially supported
configuration field that controls the number of results displayed by pickers.
    
Closes: #2779


davvid:
@alerque thanks for considering this enhancement. I rebased and added a commit that documents this option. Let me know if there's a better place to document it.

BTW, from re-reading the code, it could be argued that `max_results` would be a better name for this option. `max_results` is the same name as the actual `self.max_results` field that `opts.scrolling_limit` gets assigned to, so it might be clearer to name them consistently.

I can add another commit to rename the field if this is being considered for merging and you agree that it's worth changing that detail. The current name came from removing the `temp__` prefix from the original variable's name. I didn't think about it too much at the time. cheers

PR #3473, opened 2025-05-23
fix(health): Update deprecated vim.validate usage

omrisarig13:
# Description

The vim.validate function has been changed, keeping the old implementation available for now, but deprecated, and therefore generates warnings when running checkhealth.

Update the usage of vim.validate across the whole project, to use the new API.

## Type of change

# How Has This Been Tested?

Re-run `checkhealth vim.deprecated` after change, and ensured that `vim.validate` is no longer a problem.

**Configuration**:
* Neovim version (nvim --version): NVIM v0.11.0, Build type: RelWithDebInfo, LuaJIT 2.1.1741730670
* Operating system and version: TUXEDO OS 24.04 (linux 6.11.0)

# Checklist:

### DIFF
--- a/lua/telescope/actions/utils.lua
+++ b/lua/telescope/actions/utils.lua
@@ -35,9 +35,7 @@ local utils = {}
 ---@param prompt_bufnr number: The prompt bufnr
 ---@param f function: Function to map onto entries of picker that takes (entry, index, row) as viable arguments
 function utils.map_entries(prompt_bufnr, f)
-  vim.validate {
-    f = { f, "function" },
-  }
+  vim.validate("f", f, "function")
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   local index = 1
   -- indices are 1-indexed, rows are 0-indexed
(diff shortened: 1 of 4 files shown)


clason:
This won't work; you'll need to introduce a compatibility shim to support both signatures.

(Or just live with the deprecation warnings; they're not harmful to users.)

omrisarig13:
@clason Thank you for looking at my pull request.

I did not know that the plugin should support several nvim versions in the past, and assumed that the new vim.validate has been around long enough (which I can now see was wrong).

I can create a compatibility shim, but as you write, it might be better to just keep the code as it is and ignore the errors, for simplicity.

What do you think is best?

I'll be happy to update the code to also support older versions, but if that seems unnecessary to you, we might as well close this pull request.

clason:
Honestly, as this plugin is on maintenance life support currently, I'd rather wait until something breaks and then aggressively update the baseline Neovim requirement, moving to newer API and removing deprecated calls. Right now, I'd say just papering over these warnings is not worth it.

omrisarig13:
Makes perfect sense, feel free to close this merge request, or keep open for future reference.

PR #3534, opened 2025-09-18
feat(lsp_jump_type): pass jump_type as a function

Blackmorse:
# Description

When calling LSP-Telescope functions (`lsp_definitions`, `lsp_references`, etc.), users may want to customize how the target buffer opens.

Currently:

 - With multiple results, customization is possible using the get_selection_window option:
```lua
require('telescope.builtin').lsp_definitions({
  get_selection_window = function() ... end,
})
```
 - With a single result, Telescope bypasses the picker and relies on jump_type, which only accepts string values (`tab`, `tab drop`, `split`, `vsplit`, `never`).
 - This makes customization in the single-result case very limited.

This PR extends `jump_type` to also accept a function.
When a function is provided, it will be executed instead of the predefined commands, giving users the same level of flexibility as in the multi-result case.

## Example use case

I want to choose the exact window in which a reference is opened when navigating with `lsp_definitions`.

Today: I can do this with a window-picker when multiple definitions are found.
Limitation: When only one definition is found, I have no such option.

With this PR, I can achieve it like so:

```lua
require('telescope.builtin').lsp_definitions(
    {
      get_selection_window = function()
        return require('window-picker').pick_window({ include_current_win = true });
      end,
      jump_type = function ()
        local window = require('window-picker').pick_window({ include_current_win = true, reuse_win = false });
        vim.api.nvim_set_current_win(window)
      end
    })
```

## Type of change

# How Has This Been Tested?

- [x] Manually tested with the configuration shown above.
- [x] Verified that behavior with multiple references remains unchanged
- [x] Verified that with a single reference, the provided function is executed (e.g. window-picker is triggered) and the reference opens in the selected window

**Configuration**:
* Neovim version (nvim --version): 0.11.2
* Operating system and version: MacOS

# Checklist:

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -1558,7 +1558,9 @@ builtin.lsp_references({opts})            *telescope.builtin.lsp_references()*
                                           only one and the definition file is
                                           different from the current file,
                                           values: "tab", "tab drop", "split",
-                                          "vsplit", "never"
+                                          "vsplit", "never" or a function,
+					  which will be executed before the
+					  jump
         {show_line}            (boolean)  show results text (default: true)
         {trim_text}            (boolean)  trim results text (default: false)
         {file_encoding}        (string)   file encoding for the previewer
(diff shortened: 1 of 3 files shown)


Blackmorse:
Hi @jamestrew,

This is the second time I’m submitting this change, but I haven’t received any feedback yet. This feature is quite important for my workflow - I even maintain a personal fork of Telescope just to use it, and some of my own plugins depend on it.

However, I believe this feature is generic enough to benefit the wider community. I’d really appreciate any feedback or suggestions on this PR. I’m more than willing to make improvements or rework it as needed. If the decision is to reject it, I’d also appreciate understanding the reasoning behind that choice.

Thank you!

Issue #3499, opened 2025-07-13
vim.validate is deprecated. Feature will be removed in Nvim 1.0


ivan-volnov:
### Description

subj

### Neovim version

```markdown
NVIM v0.11.0
Build type: Release
LuaJIT 2.1.1741730670
```

### Operating system and version

Freebsd 14

### Telescope version / branch / rev

telescope 0.1.8

### checkhealth telescope

```markdown
==============================================================================
telescope:                                 require("telescope.health").check()

Checking for required plugins ~
- OK plenary installed.
- WARNING nvim-treesitter not found. 

Checking external dependencies ~
- OK rg: found ripgrep 14.1.1
- OK fd: found fd 10.2.0

===== Installed extensions ===== ~

Telescope Extension: `fzf` ~
- OK lib working as expected
- OK file_sorter correctly configured
- OK generic_sorter correctly configured

Telescope Extension: `live_grep_args` ~
- No healthcheck provided
```

### Steps to reproduce

-

### Actual behavior

-

### Minimal config

```Lua
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs { "config", "data", "state", "cache" } do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.uv.fs_stat(lazypath) then
  vim.fn.system {
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  }
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  {
    "nvim-telescope/telescope.nvim",
    dependencies = {
      "nvim-lua/plenary.nvim",
    },
    config = function()
      -- ADD INIT.LUA SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
      require("telescope").setup {}
    end,
  },
}

require("lazy").setup(plugins, {
  root = root .. "/plugins",
})
```

Trishthedish:
I was just about to create an issue for this.

I found 5 usages across 4 files of the soon to be outdated vim.validate({}) syntax in use.

I've created a PR for those happening in /actions/utils.lua

I'd be happy to tackle others but wish to get feedback on my first PR before proceeding to the others.

[...]

PR #3552, opened 2025-11-03
feat!: drop support for Nvim 0.9

clason:
It's time; even Debian now ships with Nvim 0.10.4, and keeping compatibility for 0.9 makes fixing the many deprecation warnings on 0.12 harder.

- **feat!: drop support for Nvim 0.9**
- **refactor: remove deprecated shims**
- **refactor(lsp): remove obsolete workaround for item_to_location**
- **refactor: rename vim.loop to vim.uv**
- **refactor: proper grammar in confirmation messages**
- **refactor: prefer vim.fs.find over plenary.scandir**
- **refactor: centralize check for Nvim 0.11**
- **refactor: remove obsolete health shims** (already removed in Nvim 0.9 o.O)

@nvim-telescope/core

PR #3555, opened 2025-11-03
fix: adjust to vim.validate deprecation in Nvim 0.12

clason:
## requires #3552 

Closes https://github.com/nvim-telescope/telescope.nvim/issues/3499

### DIFF
--- a/lua/telescope/utils.lua
+++ b/lua/telescope/utils.lua
@@ -18,33 +18,21 @@ local utils = {}
 utils.iswin = vim.uv.os_uname().sysname == "Windows_NT"
 utils.nvim011 = vim.fn.has "nvim-0.11" == 1
 
----@param s string
----@param i number
----@param encoding "utf-8" | "utf-16" | "utf-32"
----@return integer
-utils.str_byteindex = function(s, i, encoding)
-  if utils.nvim011 then
-    return vim.str_byteindex(s, encoding, i, false)
-  else
-    return vim.lsp.util._str_byteindex_enc(s, i, encoding)
-  end
-end
-
----@param t table
-utils.flatten = function(t)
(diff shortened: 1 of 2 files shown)


clason:
Note to self: dispatch outside of function definition, not inside.

DrKJeff16:
Shouldn't the title be for `Nvim 0.11`? since we're doing checks against that version.

clason:
No, because the deprecation message got added in 0.12.

DrKJeff16:
Weird stuff, but I get it.

lua/telescope/utils.lua
```diff
@@ -15,26 +15,25 @@ local get_status = require("telescope.state").get_status
...
 
-utils.iswin = vim.loop.os_uname().sysname == "Windows_NT"
-
----@param s string
----@param i number
----@param encoding "utf-8" | "utf-16" | "utf-32"
----@return integer
-utils.str_byteindex = function(s, i, encoding)
-  if vim.fn.has "nvim-0.11" == 1 then
-    return vim.str_byteindex(s, encoding, i, false)
```
clason: `nil` is falsy in Lua, so we can use the same signature for both branches.

clason:
Follow-up (not critical but forward-looking):
* [x] replace `nvim_buf_get_option()` with `nvim_get_option_value()` (deprecated in Nvim 0.10; could use `vim.bo` but that is less stable API)
* [X] replace `nvim_buf_add_highlight()` with `vim.hl.range()` or `nvim_buf_set_extmark()` (deprecated in Nvim 0.11)
* [x] replace `err_writeln` with `utils.notify`/`error` (deprecated in Nvim 0.11)
* [x] pet peeve: make consistent use of `vim.api` or `vim.lsp` locals

PR #3557, opened 2025-11-05
refactor: replace deprecated nvim_buf_add_highlight with vim.hl.range

clason:


clason:
The refactors will continue until morale improves.

PR #3562, opened 2025-11-06
fix(deprecated): remove module with deprecated API call

clason:
Not sure what that is even supposed to do; in any case, we're way
past the deprecation period for any breaking change this may have warned
about.


Conni2461:
Most deprecations where handled in the changelog, lets get rid of this module. thanks :)

PR #3563, opened 2025-11-06
ci: add emmyluals workflow

clason:
Optional for now; can be run locally with `make luacheck`. (I just want to see how bad this is going to be...)

Also fix the lint workflow:
* no need to run as root
* format ALL THE FILES (there are Lua files outside `/lua` these days)

### DIFF
--- a/.emmyrc.json
+++ b/.emmyrc.json
@@ -1,0 +1,25 @@
+{
+  "$schema": "https://raw.githubusercontent.com/EmmyLuaLs/emmylua-analyzer-rust/refs/heads/main/crates/emmylua_code_analysis/resources/schema.json",
+  "format": {
+    "externalTool": {
+      "program": "stylua",
+      "args": [
+        "-",
+        "--stdin-filepath",
+        "${file}"
+      ]
+    }
+  },
+  "diagnostics": {
+    "disable": [
+      "unnecessary-if"
+    ]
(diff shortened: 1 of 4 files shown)


clason:
> Diagnosis complete, 2349 problems found

...pretty bad.

Conni2461:
yeah, we dont have typing, luals wasnt that mature back then :sweat_smile:

clason:
I guess all these luacats annotations we have are for something different then ;)

clason:
> yeah, we dont have typing, luals wasnt that mature back then 😅

Since we're basically starting from scratch, we might as well switch to emmyluals already, which is quite a bit faster and has better support for generics. (It's also quite a bit stricter and still has some rough edges, and the Nvim runtime itself is not fully emmyluals compliant, which may be a problem.)

Conni2461:
> I guess all these luacats annotations we have are for something different then ;)

annotations where used for doc generation, not type hints: https://github.com/tjdevries/tree-sitter-lua/blob/master/HOWTO.md but i think that one broke 2 years ago

clason:
Yeah, I noticed on another PR (which is what started this whole rampage) ;)

clason:
emmyluals has a docgen CLI; that could be another argument for switching, if we have to replace tools anyway.

PR #3564, opened 2025-11-06
refactor: prefer local alias for vim.api, vim.lsp

clason:
* be consistent
* use longer name that isn't as likely to shadow
* rule of thumb: > 5 occurrences merits an alias

### DIFF
--- a/lua/telescope/actions/init.lua
+++ b/lua/telescope/actions/init.lua
@@ -52,7 +52,7 @@
 --- file.
 ---@brief ]]
 
-local a = vim.api
+local api = vim.api
 
 local conf = require("telescope.config").values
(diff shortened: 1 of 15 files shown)


just1602:
Ain't no maintainer, but I reviewed it and LGTM! 

I'm really happy you're doing some cleanup @clason I was really not in the mood of switching to another picker, especially since fzf-lua doesn't support the send to qflist feature that I use a lot.

clason:
Thanks, but don't get used to it; this is just a slap of paint and a bandaid to keep the ball rolling at least for the Nvim 0.12 release. The project itself is still on life support (but no longer critical), with its future under discussion.

PR #3566, opened 2025-11-07
fix(treesitter): standalone implementation

clason:
Drop hard nvim-treesitter requirement, but assumes parser and
`locals` query (using the captures documented here:
https://github.com/nvim-treesitter/nvim-treesitter/blob/main/CONTRIBUTING.md#locals)
is available on `runtimepath`.


clason:
clason >> claude 💪

Conni2461 (approved):
Love to see that this is now everything possible with neovim itself 👍

PR #3569, opened 2025-11-12
docs(readme): update plugin manager example snippets

mrcjkb:
# Description

- Updated tag to latest release in install snippets.
- Removed `branch = 0.1.x` as this branch no longer seems to be maintained.
- Premoved packer.nvim example, as it has been archived.

@clason it appears you have taken over maintainership of this project. What's your current recommendation? Do you plan on releasing more frequently going forward? Or should users fetch the latest `HEAD`, given that previous releases have become quite outdated?

Depending on your release strategy, it might make sense to either remove the version from the install snippets, or to add a workflow that auto-updates them.

## Type of change

- This change is a documentation update

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,26 +84,14 @@ Using [vim-plug](https://github.com/junegunn/vim-plug)
 
 ```viml
 Plug 'nvim-lua/plenary.nvim'
-Plug 'nvim-telescope/telescope.nvim', { 'tag': '0.1.8' }
-" or                                , { 'branch': '0.1.x' }
+Plug 'nvim-telescope/telescope.nvim', { 'tag': '0.1.9' }
 ```
 
 Using [dein](https://github.com/Shougo/dein.vim)
 
 ```viml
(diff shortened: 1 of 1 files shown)


Conni2461:
Thanks, i dont think we need automation. So far we only agreed to not let it die because both @clason and I dont want to migrate to the alternatives. Its still on livesupport, and we are trying to find solutions for this. everything is tbd.

We will do a 0.2.0 shortly with the refactorings that clason made recently

Conni2461:
Ill also lock into luarocks publish

Conni2461:
I've published a new revision for scm-1, no idea if that broke more than it helped. :shrug:

clason:
Publish should work, at least it didn't error when I pushed the 0.1.9 tag.

Conni2461:
Changes to the spec were not published to scm, i think nix uses that rather than latest rel.

The release publish worked

clason:
...it's always Nix 😞

PR #3614, opened 2026-01-23
fix(previewer): use new winid after toggle preview window

xieyonn:
# Description
Can not move selection cursor after toggle preview window.

1. `:Telescope find_files`
2. Press `toggle_preview` key, close and open preview window, keep preview window open.
3. move selection cursor.

image

All usages of buffer_preview are affected, for example `git_files`.
This was introduced after https://github.com/nvim-telescope/telescope.nvim/pull/3561 @clason
> Could it be a typo?

image

Toggle preview will create a new preview window, `preview_winid` is the new winid, `preview_window_id` was captured and used in `opts.teardown()`, it should be a bug because this variable is no longer updated.

So the change is to set `preview_window_id = preview_winid`

## Type of change

**Configuration**:
NVIM v0.11.5
Build type: Release
LuaJIT 2.1.1741730670
debian 12

# Checklist:

### DIFF
--- a/lua/telescope/previewers/buffer_previewer.lua
+++ b/lua/telescope/previewers/buffer_previewer.lua
@@ -436,11 +436,11 @@ previewers.new_buffer_previewer = function(opts)
         end
       end)
 
-      vim.wo[preview_window_id].winhl = "Normal:TelescopePreviewNormal"
-      vim.wo[preview_window_id].signcolumn = "no"
-      vim.wo[preview_window_id].foldlevel = 100
-      vim.wo[preview_window_id].wrap = false
-      vim.wo[preview_window_id].scrollbind = false
+      vim.wo[preview_winid].winhl = "Normal:TelescopePreviewNormal"
+      vim.wo[preview_winid].signcolumn = "no"
+      vim.wo[preview_winid].foldlevel = 100
+      vim.wo[preview_winid].wrap = false
+      vim.wo[preview_winid].scrollbind = false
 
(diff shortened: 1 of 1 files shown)


xieyonn:
There is also a hidden bug here: 
`preview_window_id = preview_id`should be placed outside. Otherwise, any other references to `preview_window_id` will get the old value. 

Currently, it only affects the `keep_last_buf = true` feature, which isn’t used within Telescope, so leave it here for now.

```lua
   function opts.preview_fn(self, entry, status)
     local preview_winid = status.layout.preview and status.layout.preview.winid
+    preview_window_id = preview_winid
     if get_bufnr(self) == nil then
       set_bufnr(self, api.nvim_win_get_buf(preview_winid))
-      preview_window_id = preview_winid
     end
 
     if opts.get_buffer_by_name and get_bufnr_by_bufname(self, opts.get_buffer_by_name(self, entry)) then
```

clason:
Yes, that was a typo/mistake; I relied on the tests to cover this.

(I remember being confused by `preview_window_id` vs. `preview_winid` -- is the former actually used?)

xieyonn:
`preview_window_id` is a captured value, used in `opts.teardown`, I guess it was used in resued cache buffer render (affected by `opts.keep_last_buf`). It can be a hidden bug, since I did not use it, so I just leave it.

image

clason:
Yeah, the unused part threw me, too...

(The codebase is in need of a massive refactor and cleanup.)

PR #3647, opened 2026-04-10
feat: vendor and clean up plenary

### LINKED ISSUE #2552: Remove plenary dependency

clason:
This started as an ~~procrastination~~ investigation into what we actually use from Plenary, and which part of it can easily be replaced with current Neovim API. This is obviously not ready to merge yet, but I'm putting this here in the open now that Plenary is going to be archived, in case people want to follow along (and pitch in) as I keep chipping at it.

I think I have picked all of the low-hanging fruit, but there's obviously much more to be done. Roughly in order of difficulty:
* [ ] the `class.lua` module is only used in the (now renamed)  `async_job` telescope module to define a bunch of pipes objects. These should just be created directly, but care is required to adapt the downstream usage.
* [x] the `scandir.lua` is only used in one place: to list files in a directory previewer. ~~As I don't use that myself, I have been hesitant to convert it (should be a straightforward replacement using a variant of `vim.system('ls')`, so that's another nice, self-contained, project).~~ As far as I can tell, the directory preview is only used in the [telescope-file-browser](https://github.com/nvim-telescope/telescope-file-browser.nvim) extension, which isn't actively maintained and arguably too far out of scope for a core extension.
* [ ] the `path.lua` module usages are straightforward to replace with `vim.fs` and `vim.fn` calls; there's just quite a few of them. But this can be done file by file, so it's a good "fifteen minute project" when someone is bored. A first step could also be is to just replace suitable functions with wrappers of existing (libuv or Nvim ) API and prune then-unused internal functions. **EDIT:** In particular, the OO wrapper should be removed and replaced by flat string manipulation functions (and moved, if necessary, to `utils.lua`).
* [ ] the `strings.lua` module now contains only a few useful utility functions; these could be moved to a stand-alone `telescope.strings`. (Except for `dedent`; that should be replaced.)
* [ ] the `border.lua` and `popup.lua` can probably stay, but should be refactored to be a single UI utility module -- possibly moved into the `telescope` module. (That requires a bit more knowledge about the architecture than I have.) In any case, we don't need a wrapper that is expressly designed to make Nvim floating windows look like Vim popups (just something that makes _our_ life easier).
* [ ] the `job.lua` module should be replaced by `vim.system`; again, the deep integration makes this trickier but there should not be much feature disparity that would block this. One could start with refactoring the top `utils.lua` module and see where to go from there.
* [ ] the `async.lua` (including `vararg.lua`, which is only used there) and `log.lua` are probably the biggest modules; these will have to stay for a bit until `vim.async` and `vim.log` are done, in which case migrating to those will be a rather big (but necessary) refactor.
* [ ] **before merge:** rename `neoplen` to `shared`.

So the **plan** for _this PR_ is:
1. Remove all usage of `class`, `path`, and `scandir`.
2. Clean up `border`/`popup`, `job`, `log`, `async` and expose them for extensions.

A follow-up PR should rewrite `job` as a thin wrapper over `vim.system` and deprecate it (`log` and `async` will get the same treatment once the corresponding upstream modules are stable).

A couple of notes:
* I am using this branch myself but can only test the features I use myself; it's quite likely I inadvertently broke some feature I don't use (and that isn't covered by tests). So I would appreciate outside testing -- you can just specify the `feat/neoplen` branch in your plugin spec.
* The `plenary.busted` and `test_harness` modules are no longer in-tree but moved to an external [`plentest.nvim`](https://github.com/nvim-treesitter/plentest.nvim) module that is a test-only dependency.
* I have renamed the modules to `neoplen` exactly so they won't conflict with `plenary` installed as a dependency for an extension, but they of course will need to adapt in some way if we want to completely remove the dependency. I expect that this will involve a mix of using upstream API (for things like `path`) combined with simply pulling in one of the remaining `neoplen` modules. This is a bit tricky since I cleaned out modules that are still here, so I have probably removed functions that extensions rely on -- these will have to be added back. @delphinus 
* No more FFI.


jamestrew:
> Not sure how to deal with docgen 

I did make an effort to switch out the custom tree-sitter-lua based docgen for one based off of neovim core several years back. https://github.com/nvim-telescope/telescope.nvim/pull/3227

I can revive this.

delphinus:
Thank you @clason . I almost understood and agreed the changes. I will use this branch and check extensions' behaviors.

lua/telescope/utils.lua
```diff
@@ -219,7 +258,7 @@ local path_abs = function(path, opts)
   else
     cwd = vim.uv.cwd()
   end
-  return Path:new(path):make_relative(cwd)
+  return vim.fs.relpath(cwd, path) or path
```
clason, line 260: So here we run into the big issue: Nvim is aggressively standardizing on forward slashes for path separators even on Windows (as they should, this being standard now for many years!) but both `path.lua` and -- more importantly -- our `utils.lua` want to normalize to backslashes on Windows. (Hence the test failure.)

I feel strongly that we should follow upstream's lead on this and -- if necessary -- normalize to forward slashed at the edges (`fd` or `grep`?) only.

clason: (That makes the `fzy_sorter` tests fail, though, for reasons I can't quite trace yet.)

lua/telescope/previewers/utils.lua
```diff
@@ -5,57 +5,11 @@ local strings = require "neoplen.strings"
...
-    if match and match ~= "" then
-      return match
-    end
-
-    match = detect_from_modeline(p)
-    if match and match ~= "" then
-      return match
-    end
-  end
+  return vim.filetype.match { filename = filepath }
```
clason, line 11: Dropping the modeline etc. fallbacks since they never worked for me (help files).

delphinus:
@clason I migrated telescope-frecency to `neoplen.*` (draft PR nvim-telescope/telescope-frecency.nvim#343).

One blocker for us: `async.uv` was removed in b9b203f. frecency uses `async.uv.fs_{stat,open,close,read,write,unlink,realpath}` for non-blocking file lock acquisition and DB I/O. Two paths I see:

1. Restore `lua/neoplen/async/uv_async.lua` upstream.
2. Each extension wraps the specific `vim.uv.fs_*` it needs with `async.wrap`.

I prefer (1), but either works. How do you think for this?

clason:
Thank you! I'm happy to restore async/uv (that's exactly why I asked). If I can keep it minimal, that helps (so just keep the functions you use).

delphinus:
Thanks again. I made #3669 to add the minimal functions I want to use in frecency. Please merge it into here.

clason:
@jamestrew If I understand this correctly, `strings.dedent` is only used for docgen -- in that case, can we move this (in a simplified, less general form) into your docgen module that is used now?

jamestrew:
> @jamestrew If I understand this correctly, `strings.dedent` is only used for docgen -- in that case, can we move this (in a simplified, less general form) into your docgen module that is used now?

removing the `---@eval` and the `strings.dedent` dependency here https://github.com/nvim-telescope/telescope.nvim/pull/3670

emmanueltouzery:
I decided to migrate to that branch to get rid of plenary in my config and stumbled on a bug in this branch:

in `lua/telescope/make_entry.lua`, line 281:
```diff
-        if fn.isabsolutepath(t.filename) then
+        if fn.isabsolutepath(t.filename) == 1 then
```

these lua APIs that return 0|1 already did bite me in the past, since 0 is truthy...

apart from that, it's looking pretty good so far!

clason:
Thanks, fixed! (And thank you for testing the branch, that goes a long way!)

sliwkr:
Taken the branch for a spin: As of `0a931dc` something's going sideways (that doesn't happen on master `427b576`) when calling `vim.keymap.set` without `{opts}` on telescope.builtin.keymaps:

```
vim.keymap.set('n', 'a', require("telescope.builtin").keymaps) -- not ok
vim.keymap.set('n', 'x', require("telescope.builtin").keymaps, { desc = 'foo'} ) -- ok
```

produces `E5108: Lua: .../core/opt/telescope.nvim/lua/telescope/actions/utils.lua:112: attempt to call field 'read_file' (a nil value)`

init.lua used for testing: https://bpa.st/DUBQ

clason:
Thanks, fixed!

emmanueltouzery:
i don't want to request anyone to do any work or anything like that.. just my observation, from my POV this branch is 100% stable, i'm daily-driving it for quite a long time and it's rock solid for me, and i think i'm a heavy user, i tuned a lot of the pickers, wrote my own...

i think it would make sense to merge it as-is and to handle possible improvements separately down the line. but of course devs know better, i'm just very thankful to have the branch :+1:

clason:
As you can see, there are a number of regressions (which you may not hit but others will), so this is not ready to be merged.

nathanzeng:
Took this branch for a spin. Only bug I found was that it breaks one of my extensions [telescope-ui-select.nvim](https://github.com/nvim-telescope/telescope-ui-select.nvim). There are no errors, it just uses the default `vim.ui.select` instead of the extension. I reproduced on a lsp code action.

init.lua repro:
```lua
  vim.pack.add({
    'https://github.com/nvim-telescope/telescope-ui-select.nvim',
    { src = 'https://github.com/nvim-telescope/telescope.nvim', version = 'feat/neoplen' },
  })

  require('telescope').setup({
    extensions = {
      ['ui-select'] = {
        require('telescope.themes').get_dropdown(),
      },
    },
  })

  pcall(require('telescope').load_extension, 'ui-select')
```

PR #3650, opened 2026-04-15
feat(diagnostic): add option for workspace/diagnostic

clason:
Problem: Some language servers (e.g., emmyluals) only publish
diagnostics for the opened buffer(s), requiring a `workspace/diagnostic`
request to see diagnostics for the whole project.

Solution: Add `opts.workspace` (default false) option to `diagnostics`;
if enabled, it sends the `workspace/diagnostic` request and waits for
the response (max. 1 second) before collecting the diagnostics.

### DIFF
--- a/lua/telescope/builtin/__diagnostics.lua
+++ b/lua/telescope/builtin/__diagnostics.lua
@@ -139,6 +139,24 @@ diagnostics.get = function(opts)
     opts.path_display = vim.F.if_nil(opts.path_display, "hidden")
   end
 
+  -- call `workspace/diagnostic` request and wait for response before proceeding
+  if opts.workspace and next(vim.lsp.get_clients { method = "workspace/diagnostic" }) then
+    local got_response = false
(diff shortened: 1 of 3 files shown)


lua/telescope/builtin/__diagnostics.lua
```diff
@@ -139,6 +139,24 @@ diagnostics.get = function(opts)
     opts.path_display = vim.F.if_nil(opts.path_display, "hidden")
   end
 
+  -- call `workspace/diagnostic` request and wait for response before proceeding
+  if opts.workspace and next(vim.lsp.get_clients { method = "workspace/diagnostic" }) then
```
clason, line 143: Arguably, this should be the default if `bufnr = nil` (all buffers), and no separate option needed.

clason: Actually, no, "all buffers" and "whole workspace" (including files without open buffers) are meaningfully different, so this should be a separate option.

PR #3653, opened 2026-04-21
fix(previewer): remove unexpected trailing blank line

Thaumy:
# Description

There will be a blank line in the previewer buffer if the file ends with
a line feed. Since Neovim treats `\n` (or `\r\n`) as the
"line terminator" instead of the "line separator", the previewer buffer
should be consistent with the standard behavior.

This removes the last empty string from the split result to fix that.

## Type of change

# Checklist:


Conni2461:
thanks

PR #3663, opened 2026-05-06
fix: vim.deprecation of vim.F by providing a stub for now

Conni2461:


lua/telescope/utils.lua
```diff
@@ -15,6 +15,15 @@ local utils = {}
 
 utils.iswin = vim.uv.os_uname().sysname == "Windows_NT"
 
+utils.if_nil = (function()
+  -- TODO: remove with nvim 0.12 drop
+  if vim.F ~= nil then
```
clason: That won't solve the deprecation warning, since `vim.F` is still there (it's where the deprecation warning is triggered). The shim needs to be 
```lua
utils.if_nil = vim.nonnil or vim.F.if_nil
```
(which is also much simpler ;))

Conni2461: fair enough thanks.

Conni2461:
do we need a ci/lint check? to ensure that vim.F is no longer used?

clason:
No, `vim.F` is still there... A check to prevent regressions would be useful but requires Neovim to provide deprecations in a structured way.

clason:
@delphinus frecency has (or will have) a similar issue with `vim.F.npcall`.

lua/telescope/utils.lua
clason: If I may make a small suggestion: call this `utils.nonnil` to align with updated naming? (May make it easier to remove the shims...)

Not a blocker though.

Conni2461:
okay then the deprecated functions, new features/fixes coming from PRs should not contain usages of `vim.F.if_nil` or `vim.F.npcall` and that could be overlooked. i hoped that luacheck has something for that, but it doesnt seem like that.

clason:
Luals does report `Deprecated` as a warning, but will do so even if used like this... (And we can't use that with the current annotations...)

(You won't see it on CI until the nightly is updated tomorrow with the merged patch.)

delphinus:
Oh, I haven't noticed the deprecation of vim.F. I will fix for that.

Conni2461:
i would still patch plenary tbh, we said that its still supported this quarter

clason:
related fzf-native fix: https://github.com/nvim-telescope/telescope-fzf-native.nvim/pull/159

clason:
> i would still patch plenary tbh, we said that its still supported this quarter

Yeah, but the current deprecation warning will be dropped for now (removal moved to Nvim 0.15), so it's not really necessary. (But of course can still be done; just trying to save you some effort!)

clason:
Fixup is merged, warning is gone (for 0.12, so this change is still valuable) :)

Issue #343, opened 2020-12-17
Adjustable debounce time for input


steelsojka:
### Description
When using live grep, inputing text becomes unbearably slow in medium to large codebases. Is there a way to debounce the input to only search once I've stopped typing for N milliseconds?

**Expected Behavior**
Should be able to type my query without lag

**Actual Behavior**
Typing the first character causes the UI to freeze until the results for a single letter come back.

### Details

<details><summary>Reproduce</summary>

1. nvim -nu test.vim
2. Live grep
3. Start typing
</details>

<details><summary>Environment</summary>

- nvim version:
- Operating system: 
- Telescope commit: 

</details>

<details><summary>Configuration</summary>
<p>
 

```viml
set nocompatible hidden laststatus=2

if !filereadable('/tmp/plug.vim')
  silent !curl --insecure -fLo /tmp/plug.vim
    \ https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
endif

source /tmp/plug.vim
call plug#begin('/tmp/plugged')
Plug 'nvim-lua/popup.nvim'
Plug 'nvim-lua/plenary.nvim'
Plug 'nvim-telescope/telescope.nvim'
call plug#end()

autocmd VimEnter * PlugClean! | PlugUpdate --sync | close
lua << EOF

require('telescope').setup{
  defaults = {
    vimgrep_arguments = {
      'rg',
      '--color=never',
      '--no-heading',
      '--with-filename',
      '--line-number',
      '--column',
      '--smart-case'
    },
    prompt_position = "bottom", 
    prompt_prefix = ">",
    selection_strategy = "reset",
    sorting_strategy = "descending",
    layout_strategy = "horizontal",
    layout_defaults = {},
    file_ignore_patterns = {},
    shorten_path = true,
    winblend = 0,
    width = 0.75,
    preview_cutoff = 120,
    results_height = 1,
    results_width = 0.8,
    border = {},
    borderchars = { '─', '│', '─', '│', '╭', '╮', '╯', '╰'},
    color_devicons = true,
    use_less = true,
  }
}
EOF
```

rockerBOO:
Currently no debouncing as the goal is to get it as fast as possible without debouncing tricks. The bigger plan is to async searching so it wont block input. 

Maybe someone can suggest a quick and dirty way to debounce it for now.

steelsojka:
What is the road map for moving the searching async and are there any road blockers for it?

[...]

PR #3669, opened 2026-05-21
feat(neoplen): restore minimal async.uv

delphinus:
Per #3647 ([comment](https://github.com/nvim-telescope/telescope.nvim/pull/3647#issuecomment-4452651201)) — restoring `neoplen.async.uv` with only the wrappers extensions actually depend on, per your "keep it minimal" preference.

Surveyed against [telescope-frecency draft PR #343](https://github.com/nvim-telescope/telescope-frecency.nvim/pull/343), which uses these 7:

- `fs_stat`
- `fs_open`
- `fs_close`
- `fs_read`
- `fs_write`
- `fs_unlink`
- `fs_realpath`

`init.lua` lookups gets a `uv = "neoplen.async.uv_async"` entry so `require("neoplen.async").uv.fs_*` keeps working unchanged. Adding more wrappers later is a 1-line `add(name, argc)` call when another extension shows up needing them.

## Test plan

- [x] `make test` green locally (Apple Silicon macOS / Neovim stable).
- [x] stylua clean.

Net diff: +20 / -0 across 2 files.

### DIFF
--- a/lua/neoplen/async/init.lua
+++ b/lua/neoplen/async/init.lua
@@ -6,6 +6,7 @@
 local lookups = {
   util = "neoplen.async.util",
   control = "neoplen.async.control",
+  uv = "neoplen.async.uv_async",
 }
 
 local M = setmetatable(require "neoplen.async.async", {
--- a/lua/neoplen/async/uv_async.lua
+++ b/lua/neoplen/async/uv_async.lua
@@ -1,0 +1,19 @@
+local a = require "neoplen.async.async"
+local uv = vim.uv
+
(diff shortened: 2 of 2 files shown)


clason:
Thank you! I wonder if it would not be simpler to just add the functions directly? (I admit that the extra wrapper was one of the reasons I was happy to get rid of the module...)

Not a blocker of course, just wondering. 

(There's a bunchof gratuitous wrappers there...)

clason:
Thanks, cherry-picked!

PR #3670, opened 2026-05-25
wip: new setup docs

jamestrew:
Replacing the existing `config.lua` dynamic approach for documenting `telescope.setup(opts)`'s `opts` values to the `---@class` approach.

There's a somewhat complicated coupling and chain of inheritance between many of the fields, the base picker class and the many builtin pickers. Like how each builtin picker (and extension pickers) inherit many but not all of the `default` options (eg `vimgrep_arguments` are for only a couple builtins) which share many but not all the base picker options. Many of the base picker options are yet to be documented. And then some options are "config-only", and not passable to pickers at invocation time (eg `theme` and `mappings`).

- [ ] don't really like `telescope.setup.pickers` in the current form. worth keeping?
- [ ] want to make clearer that each picker inherit from `telescope.builtin.base_opts` set by `pickers` > `defaults`
- [ ] clean up `config.lua`, dropping existing descriptions, the `---@eval` call based on it
- [ ] field descriptions are pretty much copy/pasted from the old descriptions with minor docgen/markdown support clean up but probably more cleanup opportunities present

### DIFF
--- a/doc/telescope.txt
+++ b/doc/telescope.txt
@@ -79,6 +79,665 @@ https://github.com/nvim-telescope/telescope.nvim
   :h telescope.actions.history
   :h telescope.previewers
 
+*telescope.setup.defaults*
+
+    Fields: ~
+      • {path_display}?            (`table|string|fun(opts: table, path: string): string, table?`, default: `{}`)
+                                   Determines how file paths are displayed.
+
+                                   path_display can be set to an array with a
+                                   combination of:
+                                   • `"hidden"` hide file names
+                                   • `"tail"` only display the file name, and
+                                     not the path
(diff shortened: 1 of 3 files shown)


clason:
Ouch, that looks like a lot more work than I anticipated...

And, yeah, the dynamic nature and complex dependencies is a general issue that makes telescope maintenance so difficult... I am honestly not sure how much of this is necessary/useful and how much was just "we do what we must because we can" showing off...

I do think that the concept of extensions is core to the telescope identity (moreso than slight design choice differences that make some people prefer it over newer alternatives), but I'm sure that can be simplified or streamlined (with enough effort, which I don't expect anybody to put in!)

PR #3679, opened 2026-06-13
fix(pickers): highlight match against rendered buffer line

charbelnicolas:
# Description

`Picker:highlight_one_row` was given the `display` string that `entry_adder`
built using `entry_prefix`, then passed that to the sorter's highlighter to
compute match positions. The sorter (e.g. telescope-fzf-native) returns **byte**
offsets, which are painted onto the results buffer line.

The problem: by the time the highlight runs, `update_prefix` may have replaced
the line's leading prefix with `selection_caret`, and a caret can have a
different *byte* width than `entry_prefix` even when both are the same display
width — e.g. `selection_caret = " █ "` is 5 bytes (the `█`, U+2588, is 3 bytes)
while `entry_prefix = "   "` is 3 bytes. The byte offsets were computed against
the 3-byte-prefixed string but applied to the 5-byte-prefixed buffer line, so
every match highlight is shifted left by the difference.

Only the selected row goes through `update_prefix` with the caret, so the bug
surfaces on the auto-selected first row — i.e. on the first search, before any
keystroke redraws the row through the (correct) selection path.

This fixes it by highlighting against the line as it actually exists in the
buffer rather than the caller's `display`, making the byte offsets line up with
the rendered prefix regardless of its width. The `display` argument is kept as a
fallback for the rare case the row can't be read.

## Type of change

# How Has This Been Tested?

Reproduced manually before and after the change with the minimal config below,
and confirmed the existing suite, stylua, and luacheck stay green.

Minimal repro config:

```lua
require("telescope").setup({
  defaults = {
    entry_prefix = "   ",     -- 3 bytes
    selection_caret = " █ ",  -- 5 bytes (U+2588 is 3 bytes), same display width
  },
})
require("telescope").load_extension("fzf")
```

Steps:
1. `:Telescope find_files` in a repo containing a file whose path matches a
   query mid-string (e.g. `src/schemas/order_schema.js`).
2. Type `order`.
3. Before the fix: the match highlight on the auto-selected first row lands left
   of `order` (on `s/`), because the offsets were shifted by the 2-byte prefix
   difference.
   
image_2026_06_12_19_26_02

4. After the fix: the match highlight lands exactly where it needs to.

image_2026_06_12_23_34_39

- [x] manual repro above — highlight lands on the query on the selected row
- [x] `make test`, `stylua --check .`, and `make lint` all pass

**Configuration**:
* Neovim version: NVIM v0.12.3
* Operating system and version: Arch Linux (Linux 7.0.11-arch1-1 x86_64 GNU/Linux)

# Checklist:

### DIFF
--- a/lua/telescope/pickers.lua
+++ b/lua/telescope/pickers.lua
@@ -457,6 +457,12 @@ function Picker:highlight_one_row(results_bufnr, prompt, display, row)
     return
   end
 
+  -- Highlight against the line as it exists in the buffer, not the caller's
+  -- `display`: update_prefix may have written a caret whose byte width differs
+  -- from the entry_prefix `display` was built with, which would shift the
+  -- sorter's byte offsets onto the wrong columns. Fall back to `display` only
+  -- if the row can't be read (e.g. an invalid/empty buffer).
+  display = api.nvim_buf_get_lines(results_bufnr, row, row + 1, false)[1] or display
   local highlights = self.sorter:highlighter(prompt, display)
 
   if highlights then


charbelnicolas:
Hello, @clason or @tjdevries any of you mind giving this PR a look please, it was done using AI but it is solving a bug nonetheless. Maybe it is not the correct fix but a simple comment or review would be better than letting a bug linger forever.

Sorry if it was too much using the @ to ping you. Thank you.

Issue #1171, opened 2021-08-26
Files with this kind of title [...title] not opening correctly on find_files


efierros:
### Description

Each time I trigger the command to find files with Telescope, I search for a file with this kind of title

`[...title].js`

And it doesn't open, it shows me a blank file, but the file es full of code

I believe is because it asumes the extension is: `..title].js`, Formatter also has trouble with this files

### Neovim version

0.5.0

### Operating system and version

macOS 11.5.2

### Steps to reproduce

1.- Create a file with the title [...title].js
2.- Try to open it with find_files command, I use this: `<Cmd>Telescope find_files find_command=rg,--ignore,--hidden,--files<CR>`

### Actual behavior

If I open the file from nvimTree, it opens fine
![Captura de Pantalla 2021-08-26 a la(s) 15 02 58](https://user-images.githubusercontent.com/5409160/131028570-66caf3d9-c904-4138-be78-294af89eb5ed.png)

If I search the file it shows me correctly the preview
![Captura de Pantalla 2021-08-26 a la(s) 15 04 04](https://user-images.githubusercontent.com/5409160/131028588-edaac4d7-e860-48f9-b7e9-b1d85cb0aa8a.png)

But once I opened the file is blank, it doesn't overwrite the file, just show it blank
![Captura de Pantalla 2021-08-26 a la(s) 15 04 16](https://user-images.githubusercontent.com/5409160/131028683-e291694a-25a7-4b05-b550-11b57e26402a.png)

### Minimal config

```Lua
This is my telescope config

require('telescope').setup {
    defaults = {
        vimgrep_arguments = {
            'rg',
            '--no-heading',
            '--with-filename',
            '--line-number',
            '--column',
            '--smart-case',
            '-u'
        },
        prompt_prefix = " ",
        selection_caret = " ",
        entry_prefix = "  ",
        initial_mode = "insert",
        selection_strategy = "reset",
        sorting_strategy = "ascending",
        layout_strategy = "vertical",
        layout_config = {
			prompt_position = "top",
			horizontal = {mirror = false}, vertical = {mirror = false},
			--width = 0.75,
			--preview_cutoff = 120,
		},
        file_sorter = require'telescope.sorters'.get_fzy_sorter,
        file_ignore_patterns = {},
        path_display = {
			--'shorten'
		},
        winblend = 0,
        border = {},
        borderchars = {'─', '│', '─', '│', '╭', '╮', '╯', '╰'},
        color_devicons = true,
[...]

PR #3682, opened 2026-06-22
Fix: replace usage of `fnameescape()` with `bufadd()` approach when instantiating a new buffer

### LINKED ISSUE #2446: Telescope unable to preview and open files with sveltekit advanced layouts

lmgraf:
# Description

PLEASE NOTE: I am no neovim expert, please review this carefully. I did my best to ensure this PR is high quality and to think of every area this affects.

This PR removes the usage of `fnameescape` and replaces it with a different approach that introduces `bufadd()` to instantiate a new buffer.

- fixes #2446
- fixes remaining issues in #1171

## What
- replace `vim.cmd` string approach with table approach, as it removes the need for `fnameescape`
- replace `vim.cmd` + `fnameescape()` with `vim.fs.normalize()` + `bufadd()` and `bufload()` approach when creating a new buffer
- use `vim.fs.normalize` instead of plenary's `normalize` method, when opening a file
- add warning comment to `fnameescape` usage here: https://github.com/nvim-telescope/telescope.nvim/blob/9377230aa5305d9e9aca4ed8dadf1070fb4aa9fc/lua/telescope/from_entry.lua#L40

## Why
- recommended fix in https://github.com/neovim/neovim/issues/23530#issuecomment-4767653795
- avoids flawed `fnameescape` escaping method which relies on `isfname`. Because `isfname` does not include the `(` and `)` chars by default, `fnameescape` is unable to escape parentheses, which causes #2446.
- `vim.fs.normalize` uses `/` instead of `\`, which is required for `bufadd()` to work correctly and avoids the issues when using `\` as the path separator

## Type of change

# How Has This Been Tested?

- [x] I ran `make test` and all tests succeeded
- [x] I tested opening a file at `src\app\(protected)\layout.tsx` with my own configuration. Without my fix this results in issue #2446, with my fix it works as expected.
- [ ] testing `drop` and `tab drop` commands (because they are affected by this PR too)
- [ ] testing all of the above on linux

**Configuration**:
* Neovim version (nvim --version): v0.11.6
* Operating system and version: Windows 11 Home (25H2, OS build 26200.8655)

# Checklist:

  **Does it need comments?**
  **Doesn't need changes to the documentation as it's not a breaking change.**

### DIFF
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -94,9 +94,13 @@ do
       )
     end
     if buf_command ~= "drop" and buf_command ~= "tab drop" then
-      vim.cmd(string.format("%s %d", buf_command, bufnr))
+      pcall(function()
+        vim.cmd(string.format("%s %d", buf_command, bufnr))
+      end)
     else
-      vim.cmd { cmd = buf_command, args = { api.nvim_buf_get_name(bufnr) } }
+      pcall(function()
+        vim.cmd { cmd = buf_command, args = { api.nvim_buf_get_name(bufnr) } }
+      end)
     end
   end
 end
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -193,7 +193,6 @@ action_set.edit = function(prompt_bufnr, command)
     if api.nvim_buf_get_name(0) ~= filename or command ~= "edit" then
       filename = vim.fs.normalize(filename)
       local bufnr = vim.fn.bufadd(filename)
-      vim.fn.bufload(bufnr)
       vim.bo[bufnr].buflisted = true
       edit_buffer(command, bufnr)
     end
--- a/lua/telescope/actions/set.lua
+++ b/lua/telescope/actions/set.lua
@@ -10,7 +10,6 @@
 local api = vim.api
 
 local log = require "telescope.log"
-local Path = require "plenary.path"
 local state = require "telescope.state"
 local utils = require "telescope.utils"
 
@@ -97,7 +96,7 @@ do
     if buf_command ~= "drop" and buf_command ~= "tab drop" then
       vim.cmd(string.format("%s %d", buf_command, bufnr))
     else
-      vim.cmd(string.format("%s %s", buf_command, vim.fn.fnameescape(api.nvim_buf_get_name(bufnr))))
+      vim.cmd { cmd = buf_command, args = { api.nvim_buf_get_name(bufnr) } }
     end
   end
 end
@@ -192,8 +191,11 @@ action_set.edit = function(prompt_bufnr, command)
     -- check if we didn't pick a different buffer
     -- prevents restarting lsp server
     if api.nvim_buf_get_name(0) ~= filename or command ~= "edit" then
-      filename = Path:new(filename):normalize(vim.uv.cwd())
-      pcall(vim.cmd, string.format("%s %s", command, vim.fn.fnameescape(filename)))
+      filename = vim.fs.normalize(filename)
+      local bufnr = vim.fn.bufadd(filename)
+      vim.fn.bufload(bufnr)
+      vim.bo[bufnr].buflisted = true
+      edit_buffer(command, bufnr)
     end
   end
 
--- a/lua/telescope/from_entry.lua
+++ b/lua/telescope/from_entry.lua
@@ -37,7 +37,7 @@ function from_entry.path(entry, validate, escape)
     end
   end
   if escape then
-    -- WARNING: relies on `isfname` and does not escape `(` and `)`
+    -- WARNING: relies on `isfname`, thus does not escape `(` and `)` on Windows
     return vim.fn.fnameescape(path)
   end
   return path
--- a/lua/telescope/from_entry.lua
+++ b/lua/telescope/from_entry.lua
@@ -37,6 +37,7 @@ function from_entry.path(entry, validate, escape)
     end
   end
   if escape then
+    -- WARNING: relies on `isfname` and does not escape `(` and `)`
     return vim.fn.fnameescape(path)
   end
   return path


lua/telescope/from_entry.lua
```diff
@@ -37,6 +37,7 @@ function from_entry.path(entry, validate, escape)
     end
   end
   if escape then
+    -- WARNING: relies on `isfname` and does not escape `(` and `)`
```
justinmk:
```suggestion
    -- WARNING: relies on `isfname`, thus does not escape `(` and `)` on Windows
```

lmgraf:
Is there anything else I can do to make the reviewers job easier?

Conni2461:
I am happy merging this PR. Its just a hot code path that wasnt touched in years. I want to make sure it works. So I would like to run it tomorrow at work and see if there are any regressions. Afterwards i am fine merging it. Thanks for looking into this and opening a PR

clason:
That'll trigger conflicts with my plenarypocalypse PR, but +1 on removing the `plenary.path`. 

But note that `vim.fs.normalize` is not a 100% drop-in replacement for `plenary.path.normalize`, so please make doubly sure that this change is sound! It's important to keep in mind that `plenary.path` (and hence `telescope`) internally uses `\` on Windows, so this change may trigger unwanted side-effects down the pipeline.

lua/telescope/actions/set.lua
```diff
@@ -192,8 +191,11 @@ action_set.edit = function(prompt_bufnr, command)
     -- check if we didn't pick a different buffer
     -- prevents restarting lsp server
     if api.nvim_buf_get_name(0) ~= filename or command ~= "edit" then
-      filename = Path:new(filename):normalize(vim.uv.cwd())
-      pcall(vim.cmd, string.format("%s %s", command, vim.fn.fnameescape(filename)))
+      filename = vim.fs.normalize(filename)
+      local bufnr = vim.fn.bufadd(filename)
+      vim.fn.bufload(bufnr)
```
lmgraf:
```suggestion
```
I noticed this `bufload()` call is actually not needed, since the buffer gets loaded anyways when the `vim.cmd` command runs inside the `edit_buffer()` function.

lua/telescope/actions/set.lua
```diff
@@ -97,7 +96,7 @@ do
     if buf_command ~= "drop" and buf_command ~= "tab drop" then
       vim.cmd(string.format("%s %d", buf_command, bufnr))
```
lmgraf:
```suggestion
      pcall(function()
        vim.cmd(string.format("%s %d", buf_command, bufnr))
      end)
```

This `vim.cmd` should be wrapped in a `pcall`, to avoid showing errors to the user, e.g. when encountering an old swap file:

**Without the pcall:**
- after `(A)bort` it will show `E5108: Error executing lua: vim/_editor.lua:445: Keyboard interrupt`
- after `(O)pen Read-Only` it will show `E5108: Error executing lua: vim/_editor.lua:445: nvim_exec2(), line 1: Vim(buffer):E325: ATTENTION`

**With the pcall:**
- No errors show

lua/telescope/actions/set.lua
```diff
@@ -97,7 +96,7 @@ do
...
     else
-      vim.cmd(string.format("%s %s", buf_command, vim.fn.fnameescape(api.nvim_buf_get_name(bufnr))))
+      vim.cmd { cmd = buf_command, args = { api.nvim_buf_get_name(bufnr) } }
```
lmgraf:
```suggestion
      pcall(function()
        vim.cmd { cmd = buf_command, args = { api.nvim_buf_get_name(bufnr) } }
      end)
```

Same here ...

lua/telescope/actions/set.lua
justinmk: +1. Also note that `:edit` (no args) is recommended instead of `bufload()` (see https://github.com/neovim/neovim/pull/40378#issuecomment-4787851545 ), but not relevant here since telescope seems to eventually do that anyway.

mikejmcguirk: If you run bufload here and the buffer is not yet loaded, it will fire the `FileType` autocmd in an autocmd temporary window, causing any ftplugin setlocals for window options to do nothing.

Whatever cmd you ultimately use to load the file, that cmd will perform the load in the proper window context.

lmgraf:
I just committed my suggested changes (remove `bufload()` call and wrap `edit_buffer`'s `vim.cmd` calls in `pcall`) and performed a rebase.

What's missing (did I miss anything?):
- [ ] testing how the `drop` and `tab drop` commands were affected (I don't know how to do that properly, unfortunately)
- [ ] confirm that using `/` instead of `\` does not cause any issues

lua/telescope/actions/set.lua
```diff
@@ -95,9 +94,13 @@ do
...
     end
     if buf_command ~= "drop" and buf_command ~= "tab drop" then
-      vim.cmd(string.format("%s %d", buf_command, bufnr))
+      pcall(function()
+        vim.cmd(string.format("%s %d", buf_command, bufnr))
+      end)
     else
-      vim.cmd(string.format("%s %s", buf_command, vim.fn.fnameescape(api.nvim_buf_get_name(bufnr))))
+      pcall(function()
+        vim.cmd { cmd = buf_command, args = { api.nvim_buf_get_name(bufnr) } }
```
justinmk, line 102: pcall will silently fail. is that what you want?

lua/telescope/actions/set.lua
```diff
@@ -192,8 +195,10 @@ action_set.edit = function(prompt_bufnr, command)
     -- check if we didn't pick a different buffer
     -- prevents restarting lsp server
     if api.nvim_buf_get_name(0) ~= filename or command ~= "edit" then
-      filename = Path:new(filename):normalize(vim.uv.cwd())
-      pcall(vim.cmd, string.format("%s %s", command, vim.fn.fnameescape(filename)))
+      filename = vim.fs.normalize(filename)
+      local bufnr = vim.fn.bufadd(filename)
+      vim.bo[bufnr].buflisted = true
+      edit_buffer(command, bufnr)
```
justinmk, line 201: just FYI, setting 'buflisted' shouldn't be necessary if the edit-like command actually edits the buffer.

though maybe `:drop` doesn't set 'buflisted'? that would be interesting..

Conni2461 (changes requested):
The new code has the same classic issue that the previous code didnt. it opens files with an absolute rather than a relative path (when the other would be more appropriate). that would be a breaking change and also not desired.

1. Open telescope repo 
2. do find_files
3. Open Makefile
4. do `:ls` and observe "~/plugins/telescope.nvim/Makefile" as buffer name

With current master the file as just "Makefile" as buffername. not ideal.

This is not a simple change we can "just do". This is some of the most critical code path we have in this repo and everytime we change this in the past we made some mistakes and messed up something else. We havent changed that part since 14 May 2023 so we should be really careful

clason:
This is due to the switch to `vim.fs.normalize`, which as I wrote above is not a drop-in replacement. Is this change strictly necessary to fix your issue? (Avoid making unrelated "improvements" in bugfix PRs!)

justinmk:
without `vim.fs.normalize`, `bufadd()` won't work for "~" paths because `bufadd()` treats the name extremely "literally". Same for e.g. env vars: `foo/$BAR.txt` will be literal.

It should work if you skip `normalize()` for paths that don't have "~" (or env vars, if anyone cares about that)

clason:
That is not my point. What I'm saying is that plenary's `normalize` can _also_ return a relative path (which here is important); the refactor drops that. This is completely orthogonal to expanding `~` etc.

justinmk:
> plenary's `normalize` can _also_ return a relative path (which here is important); the refactor drops that

sounds like using plenary's normalize() instead of vim.fs should work fine 👍

clason:
Or manually make the path relative (plenary delendam est). The point is to be careful here; this is not a code path that takes kindly to a drive-by refactor.

lua/telescope/actions/set.lua
lmgraf: I am setting `buflisted` here, because not doing so keeps the buffer hidden (to test: remove buflisted line, open a file using `find_files`, `:ls` to confirm that the buffer is not listed). I don't know the exact cause of this, but I am guessing it has something to do with this:

https://github.com/nvim-telescope/telescope.nvim/blob/427b576c16792edad01a92b89721d923c19ad60f/lua/telescope/actions/set.lua#L65-L80

The `edit_buffer` function inherits the actual command it executes from the `command` argument in `action_set.edit()`, and translates it using the `map` I referenced above. As you can see, the `edit` command gets translated to `buffer`, which does not set `buflisted`.

In an earlier comment you mentioned that `:edit` should be used (see https://github.com/neovim/neovim/pull/40378#issuecomment-4787851545), but `:edit` is never actually called in `edit_buffer`.

If we don't set `buflisted`, and add `vim.cmd.edit()` after `vim.cmd` inside `edit_buffer()`, then the buffer gets listed. However, Neovim docs don't recommend `:edit` when editing files programatically, see

https://github.com/neovim/neovim/blob/bc3757b7bf9326a5856f3ead844f71dc59dea206/runtime/doc/editing.txt#L597-L616

Not sure how to handle this...

lmgraf: And for the record, we're setting `buflisted` like 10 lines above my suggestion as well:

https://github.com/nvim-telescope/telescope.nvim/blob/427b576c16792edad01a92b89721d923c19ad60f/lua/telescope/actions/set.lua#L186-L191

lua/telescope/actions/set.lua
lmgraf: You're right to call me out on this, I added the `pcall` because it avoids showing errors when encountering an old swap file - this does not mean that the user isn't made aware of the old swap file. Let me make it a bit clearer...

**Reproduction**
1. Open an existing file: `nvim test.md`
2. Make some changes, **without saving**
3. `pkill nvim` or `Stop-Process -Name nvim` on windows
4. Open Neovim, perform a telescope search via a picker, e.g. `find_files`, and then you will see the following messages:

    i. The "found a swap file" prompt by Neovim (the `pcall` does NOT hide this)

    Image

    ii. An error. Depending on which option you picked in the previous prompt, this will vary

    Image

My mistake: I just assumed neovim always handles errors like this, but I don't actually know.

What are you suggesting here? On one hand we might be hiding unknown errors, on the other hand, if neovim already tells me/the user, that it failed to open the file, showing an error afterwards is kind of annoying.

lua/telescope/actions/set.lua
lmgraf, line 201: Thanks everyone for reviewing this so carefully, I really do appreciate it. I initially made the switch to `vim.fs.normalize()`, because it fixed issues I had with `~` in the path, like @justinmk mentioned. However, I understand now that telescope's standard is relative paths, and we absolutely shouldn't force absolute paths on everyone, just because of this issue (thanks @clason for highlighting this).

~While writing this response, the (I think) ideal solution just came to my mind:~

```suggestion
      filename = Path:new(filename):normalize(vim.uv.cwd())
      pcall(function()
        vim.cmd { cmd = command, args = { filename } }
      end)
```

Advantages of this approach
1. We are able to switch back to the use of plenary's `normalize()` method, because this completely avoids `bufadd()` with its extremely literal interpretation of file paths (talking about `~` handling).
Also note that:
    - `Path:new([[C:\Users\username\some\path]]):normalize()` -> `~\some\path` 
    - `Path:new([[C:/Users/username/some/path]]):normalize()` -> `C:/Users/username/some/path` 

    No idea if this is intentional. This **was** the reason why I switched to `vim.fs.normalize()`, because `bufadd()` couldn't handle the `~`. But as I mentioned, not needed anymore with this new approach.
2. It is very close to the original code: the only difference is calling `vim.cmd` with a table argument instead of a string argument.
3. ~Successfully avoids `fnameescape()`, which was the whole point of this PR.~

~So, unless calling `vim.cmd` with a table argument significantly changes the behaviour of the executed command, I think this is the way to go. Thoughts on this?~ **EDIT: Calling `vim.cmd` with a table argument will rely on `isfname`, which we're trying to avoid, because it does not escape parentheses.**

PS: Obviously I will need to re-import `Path` from plenary. I will also remove the two `pcall`'s I added since it is not relevant to this bugfix PR.

lua/telescope/actions/set.lua
```diff
@@ -95,9 +94,13 @@ do
       )
     end
     if buf_command ~= "drop" and buf_command ~= "tab drop" then
-      vim.cmd(string.format("%s %d", buf_command, bufnr))
+      pcall(function()
+        vim.cmd(string.format("%s %d", buf_command, bufnr))
+      end)
```
lmgraf, line 99:
```suggestion
      vim.cmd(string.format("%s %d", buf_command, bufnr))
```

Remove `pcall`, because it is not relevant to this bugfix PR and I want to avoid this drive-by refactor. Also, `edit_buffer` is not called anymore when creating a new buffer, which was why I added this in the first place.

lua/telescope/actions/set.lua
```diff
@@ -95,9 +94,13 @@ do
...
     else
-      vim.cmd(string.format("%s %s", buf_command, vim.fn.fnameescape(api.nvim_buf_get_name(bufnr))))
+      pcall(function()
+        vim.cmd { cmd = buf_command, args = { api.nvim_buf_get_name(bufnr) } }
+      end)
```
lmgraf, line 103:
```suggestion
      vim.cmd { cmd = buf_command, args = { api.nvim_buf_get_name(bufnr) } }
```

Same reason as above.

lmgraf (commented):
Does this resolve all the concerns?

lmgraf:
I think I will close this PR as the original idea has multiple issues. I came to realize this is a really sensitive part of the codebase, and I don't think I am the right person to do this, also because I did a handful of rookie mistakes here and there.

One last opinion, though... What I think is best to do here: We can avoid switching to `bufadd()` alltogether if we follow upstream to enforce `/`, which is a completely different approach and should probably be a new PR. Frankly, I am not experienced enough to make this happen, and I don't know enough about what's currently going on, what the future plans are for telescope (with #3647 for example) etc. etc.

I am happy to follow any guidance here...

PR #3686, opened 2026-08-07
fix(sorters): allow prefilter tags containing whitespace

### LINKED ISSUE #3675: `lsp_document_symbols` and `lsp_workspace_symbols`: Default keybind `<C-l>` erroneously including icon
### Description

`lsp_document_symbols` and `lsp_workspace_symbols` have the default keymap `<C-l>` to filter results by type (see: `:h telescope.builtin.lsp_document_symbols`).

But it erroneously includes the icons, which breaks the filter:

Image

Remove the icons to leave only `:variable:`, for example, fixes it.
Also changing the filter to ignore the icon would also fix it.

### Neovim version

```markdown
NVIM v0.12.2
Build type: RelWithDebInfo
LuaJIT 2.1.1774896198
```

Arch Linux

main

### checkhealth telescope

```markdown
telescope:                                                                  ✅

- ✅ OK plenary installed.

- ✅ OK rg: found ripgrep 15.1.0
- ✅ OK fd: found fd 10.4.2

Telescope Extension: `fzf` ~
- ✅ OK lib working as expected
- ✅ OK file_sorter correctly configured
- ✅ OK generic_sorter correctly configured
```

### Steps to reproduce

1. `:Telescope lsp_document_symbols` (or lsp_workspace_symbols)
2. Press `<C-l>` to get the filter popup

Options in popup will have icons included

### Expected behavior

The popup options would either have:

- No icons (see screenshot)
- Filter logic would ignore the icons

### Actual behavior

Icons are included and filter doesn't work until they are removed manually.

### Minimal config

```Lua
-- AUTOCOMMANDS

-- WARN: This is a build for installation (and also update) step, so it must
--       come before adding the plugins in order to get triggered.
vim.api.nvim_create_autocmd('PackChanged', {
	desc = ' telescope: build extensions and setup it up in order',
	callback = function(ev)
		local name, kind = ev.data.spec.name, ev.data.kind
		if name == 'telescope-fzf-native.nvim' and (kind == 'install' or kind == 'update') then
			vim.system({ 'make' }, { cwd = ev.data.path })
		end
	end,
})

-- SETUP

vim.pack.add {
	-- Order matters to respect dependency tree.
	'https://github.com/nvim-lua/plenary.nvim',
	'https://github.com/nvim-telescope/telescope.nvim',
	'https://github.com/nvim-telescope/telescope-fzf-native.nvim',
}

require('telescope').setup {
	extensions = {
		fzf = {
			fuzzy = true,
			override_generic_sorter = true,
			override_file_sorter = true,
			case_mode = 'smart_case',
		},
	},
}

require('telescope').load_extension 'fzf'
```

MsfPablo:
Fixes #3675

### Root cause

`sorters.prefilter` (used by `lsp_document_symbols`, `lsp_workspace_symbols`, `treesitter` and `diagnostics`) matched the tag typed in the prompt with:

```lua
"^(" .. delimiter .. "(%S+)" .. "[" .. delimiter .. "%s]" .. ")"
```

`%S+` means a tag can never contain whitespace. The tag values come from `entry.symbol_type`, which telescope parses out of the `[<kind>] <name>` text produced by `vim.lsp.util.symbols_to_items()`, i.e. straight from `vim.lsp.protocol.SymbolKind`.

Plugins that add icons to LSP kinds patch that table — e.g. `MiniIcons.tweak_lsp_kind()` rewrites each kind to `icon .. ' ' .. kind`. The tags then look like `:<icon> field:`, and because of the space the pattern only ever matched `:<icon> `, so `<C-l>` completion inserted a tag that filtered nothing out.

### Fix

Try a fully delimiter-enclosed tag first (which may contain spaces), and fall back to the existing pattern that permits omitting the closing delimiter. The pattern is anchored and cannot cross a delimiter, so `:variable: str:x` still resolves to the tag `variable` with `str:x` left for the fuzzy sorter; both patterns are now also built once per sorter instead of on every entry.

### Verification

No neovim + LSP + mini.icons environment was set up; verified by reviewing the data flow above and by checking the two patterns against representative prompts:

| prompt | matched tag | remaining prompt |
|---|---|---|
| `:<icon> field: foo` | `:<icon> field:` | ` foo` |
| `:variable: foo` | `:variable:` | ` foo` |
| `:variable foo` | `:variable ` | `foo` |
| `:variable: str:x` | `:variable:` | ` str:x` |
| `foo` | none | `foo` |

Behaviour for whitespace-free tags is unchanged.

### DIFF
--- a/lua/telescope/sorters.lua
+++ b/lua/telescope/sorters.lua
@@ -580,16 +580,22 @@ local filter_function = function(opts)
   local scoring_function = utils.if_nil(opts.filter_function, substr_matcher)
   local tag = utils.if_nil(opts.tag, "ordinal")
 
+  local d = opts.delimiter
+  -- tags may contain whitespace, e.g. when a plugin prepends icons to the LSP
+  -- symbol kinds, so a fully enclosed tag takes precedence over the lenient
+  -- pattern that allows omitting the closing delimiter
+  local enclosed_filter = "^(" .. d .. "[^" .. d .. "]+" .. d .. ")"
+  local filter = "^(" .. d .. "%S+" .. "[" .. d .. "%s]" .. ")"
+
   return function(_, prompt, entry)
-    local filter = "^(" .. opts.delimiter .. "(%S+)" .. "[" .. opts.delimiter .. "%s]" .. ")"
-    local matched = prompt:match(filter)
+    local matched = prompt:match(enclosed_filter) or prompt:match(filter)
 
     if matched == nil then
       return 0, prompt
     end
     -- clear prompt of tag
     prompt = prompt:sub(#matched + 1, -1)
-    local query = vim.trim(matched:gsub(opts.delimiter, ""))
+    local query = vim.trim(matched:gsub(d, ""))
     return scoring_function(_, query, entry[tag], _), prompt
   end
 end


lua/telescope/sorters.lua
```diff
@@ -580,16 +580,22 @@ local filter_function = function(opts)
   local scoring_function = utils.if_nil(opts.filter_function, substr_matcher)
   local tag = utils.if_nil(opts.tag, "ordinal")
 
+  local d = opts.delimiter
+  -- tags may contain whitespace, e.g. when a plugin prepends icons to the LSP
+  -- symbol kinds, so a fully enclosed tag takes precedence over the lenient
+  -- pattern that allows omitting the closing delimiter
+  local enclosed_filter = "^(" .. d .. "[^" .. d .. "]+" .. d .. ")"
+  local filter = "^(" .. d .. "%S+" .. "[" .. d .. "%s]" .. ")"
```
clason, line 588: This is somewhat of a hot path, so prefer `string.format` here.

MsfPablo:
Bumping this — small sorter fix for prefilter tags with whitespace in them. Tests pass; let me know if you want a spec added.

Issue #3342, opened 2024-10-27
Unrelated search result list briefly showing up while typing


opsidjflksdf:
### Description

This is with Telescope files, observing the behavior with both fzf-native installed or not (defaulting to fzy-lua). With the default fzy - its easier to observe as the intermediary results linger for longer.

Say I'll start typing in the prompt, and after the first couple of letters I have a result list. Then I'll type two more letters quickly one after the other, and once I'm done typing them, the list changes to some kind of an intermediary result list of items that looks completely unrelated to the results to that point (as opposed to, say, a refinement of the list up to that point). That only shows briefly, and after that the "correct"/final list of results as expected, are shown, without me typing anything else after the last two letters. 
If I type the last two letters with some pause in between, I don't observe this effect, results change without "intermediary" results showing in between.
So it looks like Telescope is programmed to retrieve a first set of results through what is a presumably a quicker operation, in order to have something to display initially, but after that it will still wait for a better but slower set of results and once it gets them it dumps them on you basically while you are looking at the screen not typing anything.
I don't see this with fzf-lua.nvim or with fzf.vim
I'm sure this must be known behavior, can't figure how to search for it though in the discussions, could someone point  me to any discussion on the history/background and possible workarounds/settings to alleviate it? Feel free to mark this as a duplicate of an existing bug report.
I'm trying to look at the source code in the meantime.
Thank you a bunch!

### Neovim version

```markdown
NVIM v0.10.2
Build type: Release
LuaJIT 2.1.1713484068
Run "nvim -V1 -v" for more info
```

### Operating system and version

Windows 10

### Telescope version / branch / rev

0.1.x

### checkhealth telescope

```markdown
telescope: health#telescope#check

Checking for required plugins ~
- OK plenary installed.
- OK nvim-treesitter installed.

Checking external dependencies ~
- OK rg: found ripgrep 14.1.1 (rev 4649aa9700)
- OK fd: found fd 10.1.0

===== Installed extensions ===== ~

Telescope Extension: `file_browser` ~
- No healthcheck provided

Telescope Extension: `fzf` ~
- OK lib working as expected
- OK file_sorter correctly configured
- OK generic_sorter correctly configured

Telescope Extension: `json_history` ~
- No healthcheck provide
```

### Steps to reproduce

Please see bug description.

### Expected behavior

Expecting not to see "intermediary" results at all

### Actual behavior

Seeing intermediary results flashing briefly

### Minimal config

```Lua
[...]

PR #3689, opened 2026-08-17
fix(finders): stop yielding per entry in oneshot replay loop

joybro:
`index % await_count` is always truthy in Lua (0 included), so the oneshot finder's replay loop yields once per entry instead of once per `await_count`. The per-entry yields let redraws expose the partially re-scored list as a brief flash of unrelated results when typing quickly in large directories (related: #3342).

The streaming loop's per-entry yield is left as is; there it also escapes the libuv fast-event context.

Developed with LLM's help; reviewed and tested by me (`make test`, manual repro on a ~5k-file directory).

### DIFF
--- a/lua/telescope/finders/async_oneshot_finder.lua
+++ b/lua/telescope/finders/async_oneshot_finder.lua
@@ -87,7 +87,7 @@ return function(opts)
       local current_count = num_results
       for index = 1, current_count do
         -- TODO: Figure out scheduling...
-        if index % await_count then
+        if index % await_count == 0 then
           async.util.scheduler()
         end
 


clason:
Thank you, but please trim down the comment -- we don't need a user manual here. (In fact, no need for a comment at all, this is a simple bug fix for a by now very well known foot gun.)

(This is a tell-tale sign of naked LLM output; we don't have an AI policy yet, but general courtesy in _every_ project is to 1. declare LLM use and 2. clean up output _before_ making a PR -- this applied to PR and commit descriptions as well, which are _way_ too verbose.)

joybro:
Thanks, all fair points. Removed the comment and trimmed the commit message and PR description.

For the record: this fix was developed with LLM's help. I reviewed the change and tested it myself (`make test`, plus manual repro on a ~5k-file directory).

